PR #6, opened 2017-10-30
First version

matthew-white:
Using [Vue.js](https://vuejs.org/) on the front end, as well as npm and [Browserify](http://browserify.org/).

I'm also making use of Vue [single file components](https://vuejs.org/v2/guide/single-file-components.html), which separate each component into its own file that includes the HTML, CSS, and JavaScript for the component.

This first version of the front end uses submission endpoints on Jubilant, so you'll have to have the Jubilant server running on the old Jubilant branch `add-submissions` from https://github.com/nafundi/jubilant-garbanzo/pull/20. I've also updated that branch to enable CORS. We might not really want to do that at this point, and at Clint's suggestion, I'm going to investigate using nginx to integrate the front and back ends without CORS. However, enabling CORS on Jubilant is a quick fix to get Super Adventure up and running.


lib/main.js
```diff
@@ -1,7 +1,23 @@
-import Vue from 'vue'
-import App from './components/app.vue'
+import Vue from 'vue';
+import axios from 'axios';
 
-new Vue({ // eslint-disable-line no-new
+import App from './components/app.vue';
+const globalComponents = [
+  require('./components/error-message.vue'),
```
ggalmazor: Could we not mix the two import module syntaxes? `import` and `require` do the same thing. I'm in on adopting the new `import` syntax and would replace those `require` expressions to imports.

matthew-white: I also generally prefer `import` to `require()`. However, as far as I can tell, `import` can't be used in an array like this. Separately importing the global components then adding them to the array is slightly more verbose than using `require()` in the array. I also don't actually need to give any of the `require()`-d components a name, which I think `import` would require. Do you think it's alright to use `require()` just in this array, or do you think there's a better approach?

matthew-white:
FYI @lognaturel, I skipped adding a copyright notice and license information to three files:

1. `package.json`, which does not support comments
2. `package-lock.json`, which is autogenerated
3. `.gitkeep`, which is an empty file

lib/main.js
ggalmazor: I understand what you say but I see three major problems with this approach. 

- ‘require’, as opposed to ‘import’, is not (and will not be) supported by EcmaScript. By using ‘require’ you’re forcing this code to be transpiled, even if browsers start to understand ‘import’ statements which is the standard mechanism for loading modules in EcmaScript.
- Dynamic module loading is considered harmful in a web app context because it makes bundling more difficult. Browserify/Webpack can’t know what modules you’re ‘require’-ing and this impacts negatively in your options for building your deliverables.
- As a design guideline, it’s easier to understand a rule without exceptions like “always use import, no matter what”. Also, this way, all modules are imported at the header of each file, which is easier to read, as opposed to having modules being loaded in unexpected places.

I understand that you need to write some more lines but I think it’s worth the effort.

lognaturel:
@matthew-white That seems standard to me!

lib/main.js
matthew-white: Sounds good, I have made that change!

PR #20, opened 2017-11-28
Linting and routing

### LINKED ISSUE #11: Add router
We want a router so that folks can go back in their browser history and so that they can add individual pages of Super Adventure to their bookmarks. Investigate the Vue.js router:

https://router.vuejs.org/en/

matthew-white:
Closes #11, #18.


lib/components/breadcrumbs.vue
```diff
@@ -23,19 +23,16 @@ except according to the terms contained in the LICENSE file.
...
     }
   },
   methods: {
     linkBreadcrumb(index) {
-      const breadcrumb = this.breadcrumbs[index];
-      return breadcrumb.view != null && index != this.breadcrumbs.length - 1;
-    },
-    view(breadcrumb) {
-      this.$emit('view', breadcrumb.view, breadcrumb.props);
+      return this.list[index].to != null && index !== this.list.length - 1;
```
clint-tseng: Is there not a way to get it to give you some kind of rich data object instead of mucking about with array indices?

lib/components/loading.vue
```diff
@@ -0,0 +1,39 @@
...
+  },
+  watch: {
+    state: 'toggleMessage'
+  },
+  mounted() {
+    this.toggleMessage();
+  },
+  methods: {
+    toggleMessage() {
+      if (this.state) {
```
clint-tseng: I'd prefer some kind of explicit comparison operator here.

matthew-white:
@clint-tseng, do you think we're OK to merge this PR? Let me know if you'd like to see any other changes!

PR #23, opened 2017-12-10
Add NGINX

### LINKED ISSUE #10: Set up nginx
We need something like this as a more permanent solution for CORS.

matthew-white:
Closes #10.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -1,0 +1,21 @@
+events {
+}
+http {
+  # This path may be different on your local computer depending on how you
+  # installed nginx.
+  include /usr/local/etc/nginx/mime.types;
+
+  server {
+    listen 8080;
+    server_name localhost;
+    # Specify the root of the repository to `nginx -p` so that . is relative to
+    # the root directory of the repository.
+    root .;
+
+    location /api {
+      rewrite /api(.*) $1 break;
+      proxy_pass http://localhost:8383;
+      proxy_redirect off;
+    }
+  }
+}
--- a/README.md
+++ b/README.md
@@ -9,28 +9,54 @@ https://www.apache.org/licenses/LICENSE-2.0. No part of Super Adventure,
 including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
-
 # Super Adventure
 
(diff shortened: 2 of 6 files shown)


lognaturel:
The config works well for me. It's expected that we'd get 403s for form creation now, right?

nginx.conf
```diff
@@ -0,0 +1,21 @@
+events {
+}
+http {
+  # This path may be different on your local computer depending on how you
+  # installed nginx.
```
lognaturel: And depending on your OS.

matthew-white:
Sounds good, I'll go ahead and merge! You're right that the 403s are expected; my login PR should resolve that.

nginx.conf
```diff
@@ -0,0 +1,21 @@
...
+  include /usr/local/etc/nginx/mime.types;
```
clint-tseng: the other solution is to include a mime.types and relpath it as well

nginx.conf
```diff
@@ -0,0 +1,21 @@
+events {
```
clint-tseng: i would just leave this out entirely.

nginx.conf
```diff
@@ -0,0 +1,21 @@
...
+}
+http {
+  # This path may be different on your local computer depending on how you
+  # installed nginx.
+  include /usr/local/etc/nginx/mime.types;
+
+  server {
+    listen 8080;
```
clint-tseng: would recommend a more esoteric port (even something like 8989) in case someone is using this already locally.

lognaturel: I'm in favor of this. Ran into a conflict earlier today.

nginx.conf
matthew-white: You mean committing `mime.types` to the repository? I like that idea: I could see that being easier for contributors so that they don't have to modify paths. But we'd also miss any NGINX updates to `mime.types`, right (or have to monitor for them)? That might not be much of a concern though, as I doubt `mime.types` updates in big ways that often...

nginx.conf
matthew-white: The empty block looks funny to me as well, but NGINX results in an error when I don't have it:

```
nginx: [emerg] no "events" section in configuration
```

nginx.conf
matthew-white: Sounds good. I'll switch it to 8989 in https://github.com/nafundi/super-adventure/pull/24.

PR #24, opened 2017-12-12
Add login

matthew-white:


clint-tseng:
Yes; this is covered in the criteria (the button should not appear).

matthew-white:
OK, I've updated the navbar so that clicking the link doesn't reload the app and hits the router instead. I definitely agree that ideally, the button should not appear if you don't have the permissions to use it. We'll need a new Jubilant endpoint to cover that, right? Like a permissions API? Or is that already part of the users API?

lib/components/session/login.vue
```diff
@@ -0,0 +1,79 @@
...
+  data() {
+    return {
+      email: '',
+      password: '',
+      error: null,
+      disabled: false
+    };
+  },
+  mounted() {
+    this.$refs.email.focus();
```
clint-tseng: is there a way to generalize this for our forms so the first `input` is `focus()`ed? then we'll never forget to do it.

lib/router.js
```diff
@@ -34,4 +41,65 @@ const routes = [
...
+let historySize = 0;
+
+router.beforeEach((to, from, next) => {
+  historySize += 1;
+  next();
+});
+
+// Goes back to the page last visited in the app, navigating to `fallback` if
+// the app was just loaded and is on its first page.
+const safeBack = (fallback) => {
```
clint-tseng: can i suggest naming this `backOrElse`? "safe" is a bit ambiguous.

lib/components/session/login.vue
```diff
@@ -0,0 +1,79 @@
...
+        .then(response => {
+          const success = this.$session.set(response.data);
+          const headers = axios.defaults.headers.common;
+          if (success) {
+            headers.Authorization = `Bearer ${this.$session.token}`;
+            // Using safeBack() rather than this.$router.go(-1) in case the user
+            // navigated directly to /login rather than being redirected there.
+            safeBack('/forms');
+          } else {
+            delete headers.Authorization;
```
clint-tseng: why?

lib/components/session/login.vue
```diff
@@ -0,0 +1,79 @@
...
+      error: null,
+      disabled: false
+    };
+  },
+  mounted() {
+    this.$refs.email.focus();
+  },
+  methods: {
+    logIn() {
+      this.disabled = true;
```
clint-tseng: similarly, i worry about this flag getting set/unset each time a form exists.

at the least, is there a `.complete` or something in axios so that you don't have to apply it separately to the success and fail branches?

but ideally, is there some way we can have the behaviour automatically apply?

if the answers are not obvious then we'll punt until they are.

lib/router.js
clint-tseng: and actually, i don't think we want to do this operation this way at all. can we put the return-to URL in the queryparam? i think it is unexpected that logging in pushes your browser history backward.

lib/router.js
```diff
@@ -34,4 +41,65 @@ const routes = [
...
+router.beforeEach((to, from, next) => {
+  if (router.app.$session.isLoggedOut()) {
+    if (to.path === '/login') {
+      // If you are logged out and are navigating to login, you are going to the
+      // right place.
+      next();
+    } else {
+      // If you are logged out and are navigating somewhere other than login,
+      // you need to log in first.
+      next('/login');
```
clint-tseng: but will this then push the user to where they wanted to go afterwards?

lib/session.js
```diff
@@ -0,0 +1,44 @@
...
+See the NOTICE file at the top-level directory of this distribution and at
+https://github.com/nafundi/super-adventure/blob/master/NOTICE.
+
+This file is part of Super Adventure. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of Super Adventure,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+*/
+import moment from 'moment';
```
clint-tseng: i have been a happy moment user for many many years now, but they have a new one http://moment.github.io/luxon/ that seems cleaner, lighter, and better (immutable, for one). take a look?

lib/session.js
```diff
@@ -0,0 +1,44 @@
...
+
+export default class Session {
+  get token() { return this._token; }
+  // TODO: Use `expires` in some way.
+  get expires() { return this._expires; }
```
clint-tseng: are you actually using this for something? or planning to?

nginx.conf
```diff
@@ -1,16 +1,17 @@
...
 
   server {
     listen 8080;
     server_name localhost;
-    # Specify the root of the repository to `nginx -p` so that . is relative to
-    # the root directory of the repository.
+    # Specify the root directory of the repository to `nginx -p` so that . is
+    # relative to the root of the repository.
     root .;
+    expires -1;
```
clint-tseng: perhaps leave a comment mentioning that this setting is only desirable in a development environment, and for production you'll want to remove it.

lib/session.js
matthew-white: I was thinking that before each request, Super Adventure could check whether the token is expired. If it is, it doesn't send the request, but instead puts up an alert telling the user that they need to log in again.

nginx.conf
matthew-white: 👍

lib/components/session/login.vue
matthew-white: Ha, this is probably slightly overzealous. I'll go ahead and keep the header.

lib/components/session/login.vue
matthew-white: I definitely think we should focus the first `input` when the user is creating a new resource. Do you think we should focus the first `input` when the user is editing an existing resource? Part of me feels like when editing, the user should always have to explicitly select a field before it becomes editable (rather than the first field starting out as editable). That way, if the user accidentally types something, it won't go in the first field. But maybe that concern's overblown, and it'd definitely make implementation trickier...

lib/components/session/login.vue
matthew-white: > at the least, is there a `.complete` or something in axios so that you don't have to apply it separately to the success and fail branches?

axios just returns a `Promise`. I do wish that standard `Promise`s had a `finally()` method. It looks like there's a stage-3 [candidate for that method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally). Right now our Babel configuration is set to support stage-2 candidates, but not stage-3. However, it'd be easy to change that setting so that we have access to `finally()`.

> but ideally, is there some way we can have the behaviour automatically apply?
>
> if the answers are not obvious then we'll punt until they are.

Maybe we could write a fetch function that looks for a `disabled` property on the component. One option is to include the function as a method on the component using [Vue mixins](https://vuejs.org/v2/guide/mixins.html). But I'm not totally sure yet what the right answer is. I would opt for waiting until we have more forms, then seeing whether a refactor becomes clearer.

lib/components/session/login.vue
clint-tseng: no; focusing the first field is important for accessibility reasons if nothing else. it's in the criteria document, too.

lib/components/session/login.vue
clint-tseng: sounds like we should punt this one then.

lib/session.js
clint-tseng: okay, fair enough.

clint-tseng: but you will also have to react if the token has been invalidated for any other reason.

lib/session.js
matthew-white: I took a quick look, and I like the looks of it. I'm making a few other changes to this PR and am hoping to merge it soon, but how about I take a closer look the next time I reach for moment? I created the issue #28 so we don't lose track of it.

lib/session.js
matthew-white: Sounds good. I'm still thinking through the best way to implement this, but I'm thinking we should implement it before MJP. What do you think?

lib/components/session/login.vue
```diff
@@ -45,24 +43,21 @@ export default {
...
           const success = this.$session.set(response.data);
-          const headers = axios.defaults.headers.common;
           if (success) {
+            const headers = axios.defaults.headers.common;
             headers.Authorization = `Bearer ${this.$session.token}`;
-            // Using safeBack() rather than this.$router.go(-1) in case the user
-            // navigated directly to /login rather than being redirected there.
-            safeBack('/forms');
+            const next = this.$route.query.next;
+            const nextPattern = /^(\/|(\/[\w-]+)+)$/;
```
clint-tseng: ?

are you trying to determine that the route is local?

matthew-white: Exactly, I'm pretty sure the router will check this anyway, but I thought it didn't hurt to add a second check of the query parameter.

clint-tseng: hm. i think the safer way is to use the browser's own facilities:

    const link = document.createElement('a');
    link.href = this.$route.query.next;
    link.host === window.location.host;

clint-tseng:
Looks good. My final nit that you should ignore for now would be to see if the `let` can be avoided. Having to track the mutable variable's state introduces cognitive friction in understanding the code. Try to avoid that in the future, and at some point we'll fix this one up too if we can.

Issue #37, opened 2018-01-23
Deployment weight too heavy


clint-tseng:
The final JS bundle is 3.0MB, which is absolutely gargantuan. The bootstrap CSS is 112KB, which is more palatable but still seems highly excessive for a bunch of CSS. We should aim to bring these down by orders of magnitude by release.

matthew-white:
I thought I'd comment with the latest numbers. After `npm run build`, `build.js` is 417KB. With #44 merged, `build.css` is now 104KB (90KB Bootstrap / 14KB other). Are those sizes low enough for now, or is there more we should try to do before release?

clint-tseng:
those are still quite large; is there anything you can do to bring the js file down especially?

matthew-white:
I took a look at our dependencies, which seem to account for most of the size of the `.js` file:

| Name | Minified size (KB) |
| :- | :- |
| `jquery` | 85.1 |
| `vue` | 60.1 |
| `vue-router` | 23.9 |
| `moment` | 50.4 |
| `pako` (to compress the QR code) | 27.3 |
| `qrcode-generator` | 18.5 |
| `axios` | 13.6 |
| **Total** | 278.9 |

Maybe we could move the QR code generation server-side, though I think that would also require bandwidth down the road when users download those images. Other than that, I don't see a lot of low-hanging fruit among the dependencies.

I also don't think we're gzipping the CSS or JS. Maybe that'd be worth doing as well?

matthew-white:
@clint-tseng, now that [Central is using gzip](https://github.com/opendatakit/central/pull/15), do you think we're able to close this issue? Once minified and zipped, the CSS is 19KB and the JS is 155KB. I think we should keep our eyes on the JS, as that'll only grow as we add more pages, but maybe these are reasonable sizes for the time being?

clint-tseng:
yeah, that's much closer to an acceptable range. both are still larger than they need to be but we can live with it for now.

matthew-white:
Sounds good. I'll go ahead and close for now, but we can reopen or create a new issue if the CSS or JS starts getting too large.

PR #44, opened 2018-02-13
Trim Bootstrap

matthew-white:
See #37.


clint-tseng:
Did you do this manually? How do we replicate for future versions?

matthew-white:
Yes, all the removal was manual (though didn't take long). The current version of Bootstrap is v4, but we're using the latest version of v3: I believe v4 was still alpha when we got started. There are some big changes from v3 to v4, so I'd guess that we'll probably want to hold off on moving to v4 for a while, and once we do, we'll probably have to redo some of these removals.

clint-tseng:
:/ they don't have like a builder of some kind? the old jquery ui stuffs had a builder you could go and check the things you wanted and it would build you just that file. and then it also would leave a comment in that file that the builder understood so you could just feed it back in and it would give you the equivalent of the latest version.

matthew-white:
That does sound nice. I don't see anything official along those lines. I could try to see whether third-party options exists. I do also think that v3 will last us for a good while. How about we revisit this question once we start thinking about updating to v4?

PR #75, opened 2018-04-24
Revoke a field key

matthew-white:


lib/components/field-key/list.vue
```diff
@@ -188,41 +224,56 @@ export default {
...
       this.popoverLink = null;
     },
-    popoverContainsElement(element) {
-      if (this.popoverLink == null) return false;
-      const popover = $('#field-key-list-popover-content').closest('.popover');
-      return element === popover[0] || $.contains(popover[0], element);
+    elementIsOutsidePopover(element) {
+      if (this.popoverLink == null) return true;
+      const popover = $('#field-key-list-popover-content').closest('.popover')[0];
+      return !(element === popover || $.contains(popover, element));
```
clint-tseng: de morgan please? :)

clint-tseng: (i don't always prefer it that way but in this case i think it reads easier)

lib/components/field-key/list.vue
```diff
@@ -188,41 +224,56 @@ export default {
...
         this.hidePopover();
       }
     },
+    actionsId(index) {
+      return `field-key-list-actions${index}`;
+    },
+    showRevoke(fieldKey) {
+      // Bootstrap does not actually disable dropdown menu items marked as
+      // disabled.
+      if (fieldKey.isRevoked()) return;
```
clint-tseng: i think in this case it's fine, but in more complex cases it'd be nice to have the logic for both conditions in the same place somewhere, so the ui and the business logic absolutely agree on when disabled should happen

lib/components/field-key/revoke.vue
```diff
@@ -0,0 +1,80 @@
...
+  data() {
+    return {
+      alert: alert.blank(),
+      requestId: null
+    };
+  },
+  methods: {
+    revoke() {
+      this
+        .delete(`/field-keys/${this.fieldKey.id}`)
```
clint-tseng: whoa, no. you want to issue a `DELETE` with the `token` to `/v1/sessions`; see https://jubilantgarbanzo.docs.apiary.io/#reference/authentication/field-key-authentication/revoking-a-field-key

clint-tseng: this must be why you were seeing them disappear

clint-tseng:
also, all these changes adding `$`; are you doing that piecemeal or..?

matthew-white:
Just pushed changes to the PR, and I think it's ready for a second look.

> also, all these changes adding `$`; are you doing that piecemeal or..?

We had the `$` prefix on  jQuery variables everywhere in Adventure other than `field-key/list.vue`, so I added them there partly for consistency. Coming back to that file after a little bit of time, I was getting confused about what was a jQuery variable and what wasn't, so it felt useful and more consistent to have that distinction. I think we've discussed that the `$` prefix might not be the best way to make that distinction, but it felt like a good move here at least for now.

lib/components/field-key/revoke.vue
```diff
@@ -66,8 +66,9 @@ export default {
   },
   methods: {
     revoke() {
+      const encodedToken = encodeURIComponent(this.fieldKey.token);
```
clint-tseng: you shouldn't have to do this anymore, but it's not a problem for now.

clint-tseng:
Yeah, I have no problem with yes $ or no $; I just wanted to check in on how consistent that was across the code.

PR #83, opened 2018-05-17
Add form settings page

matthew-white:


lib/components/form/edit.vue
```diff
@@ -0,0 +1,114 @@
...
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of Super Adventure,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+<template>
+  <div id="form-edit" class="panel panel-simple">
+    <div class="panel-heading"><h1 class="panel-title">Form Stage</h1></div>
+    <div class="panel-body">
+      <form>
```
clint-tseng: need a `<fieldset>` here too to actually group the radiobuttons together. the tabstops are really goofy right now, i think the fieldset will help.

clint-tseng: relatedly, i think once you put a fieldset in you can simply set disabled on the fieldset and all the radio buttons will disable

lib/components/form/settings.vue
```diff
@@ -10,17 +10,79 @@ including this file, may be copied, modified, propagated, or distributed
...
+      }
+    };
+  },
+  methods: {
+    setAlert(alertObj) {
+      this.$emit('alert');
+      this.alert = alertObj;
+    },
+    afterDelete() {
+      const name = this.form.name != null
```
clint-tseng: i'm guessing this is a common pattern? is it worth centralizing?

lib/components/form/edit.vue
matthew-white: Very nice, that makes the disabling much more straightforward.

matthew-white: I've added the `<fieldset>`, but it didn't seem to have an effect on the tabstops. It seems like the typical behavior with radios it that tabbing brings you to the set of radios, then the arrow keys allow you to move between the radios: https://stackoverflow.com/questions/14322564/can-you-tab-through-all-radio-buttons. However, that's a bit awkward given that we send a request with each radio change...

lib/components/form/settings.vue
matthew-white: It's definitely common, but I haven't quite figured out how to centralize it yet. In templates, where there are more space constraints, I've used `form.name || form.xmlFormId`, but while `||` seems safe in this case, I've tended to avoid it outside templates, since it relies on JavaScript's truthiness conditions. A utility function feels a little heavy-handed, and for better or for worse, I've been reluctant to change the data returned by the server. In the forms table, we render slightly different content based on whether there's a form name, so I'd be reluctant to change the `name` property itself. Maybe I could add a `nameOrId` property? Do any of these solutions seem reasonable to you? Do other ideas come to mind?

matthew-white:
I feel a little conflicted about the order of `beforeEachResponse()`, because chronologically, it comes before the responses, but most of the time, you have to know what the responses are before you understand the `beforeEachResponse()` method. I think I've tended to put it after the `respondWith*()` calls for that reason, first specifying the requests/responses and only then specifying the tests to run for them. But I can really see it both ways, and I'm happy to switch it if you prefer the other way.

matthew-white:
@clint-tseng, this should be ready for you to take another look whenever you have the chance.

PR #88, opened 2018-06-01
Assorted issues

matthew-white:


clint-tseng:
looks good. my only comment is that in general when i fix bugs i like to write tests that fail before the fix and succeed after, and include the test with the commit. but this applies somewhat less to UI since there's just so much

PR #98, opened 2018-06-15
Add site root landing page

matthew-white:
This does not include the "latest news" panel, which I'll tackle once we know whether we'll update the Discourse server or pursue another path.


lib/components/root/home.vue
```diff
@@ -0,0 +1,103 @@
...
+    <div v-show="fetchCounter === 3">
+      <div class="panel panel-simple">
+        <div class="panel-heading"><h1 class="panel-title">You Have:</h1></div>
+        <div class="panel-body">
+          <div class="row">
+            <div class="col-xs-6">
+              <root-entity :icons="['user-circle', 'cog']" name="Web User"
+                data-from="/users" route-to="/users"
+                @fetched="incrementCounter">
+                who are allowed to administer projects through this website.
```
clint-tseng: let's modify this to "who can administer" so we don't have to worry about pluralization.

lib/components/root/entity.vue
```diff
@@ -0,0 +1,131 @@
+<!--
```
clint-tseng: tbh way more elaborate than i'd have personally gone for but you made it work and it seems to work good

PR #104, opened 2018-06-30
Update alerts

matthew-white:
This PR is ready to be reviewed. I took a first pass at the CSS of the alert at the top of the page, but would definitely welcome any updates to it at Clint's next CSS pass.

I ended up using two entirely separate alerts, one in `App` and one in `Modal`. Overall, I think using two alerts instead of one actually resulted in the most simplicity, as well as some nice flexibility.

Outside of `App` and `Modal` (which should change only infrequently), components no longer each need their own `<alert>` element and `alert` data property. There's also no longer the same need for alert-related communication between parent and child components, and there's no distinction between the current page's alert and the alert to be shown on the next page after routing. Instead, components just need to call `this.$alert()`. If a component is a modal or contained in a modal, `this.$alert()` shows an alert in the modal; otherwise, it shows an alert at the top of the page.

I initially tried implementing these changes using a single alert. This required hiding the alert whenever a modal was closed, because otherwise any alert that the modal triggered would be shown at the top of the page once the modal was closed. However, automating the hiding of that alert resulted in some complexity in some cases.

For example, if the modal is used to complete an action successfully, we often hide the modal, then show a success alert on the main page. If hiding the modal also automatically hides any alert that the modal triggered, that means we want the following events to take place in order:

1. The modal is hidden.
2. If the modal triggered an alert, the alert is hidden.
3. A success alert is shown at the top of the page.

However, this actually turned out to be nontrivial to implement. I implemented (2) using a Vue watcher on the `state` property of `Alert` and (3) using `$emit()`. Yet because Vue watchers are asynchronous and only executed at each tick, while `$emit()` is synchronous, (3) ended up taking place before (2): the success alert was set, then immediately hidden.

I tried to hack around Vue's reactivity system, including trying to use `$emit()` and `$nextTick()` together, but ultimately found that the logic became clearer when I used two alerts instead of one, separating the alert on the main page from the one in the modal. The separation means that the alert in the modal does not need to be automatically cleared when the modal is closed, because the modal's alert will not be shown on the main page. (Though if we want, we also have the flexibility to automatically reset the modal's alert when the modal is closed.) The separation means that hiding the alert in the modal and setting the alert on the main page are two distinct actions that modify two separate objects.


lib/alert.js
```diff
@@ -0,0 +1,82 @@
...
+
+// Returns a data object for a new blank alert.
+export const blankAlert = (type = 'danger') => ({
+  state: false,
+  type,
+  message: null,
+  at: new Date()
+});
+
+// Convenience class used to make it easier to modify a component's alert
```
clint-tseng: in what way? this comment doesn't really make sense on its own without looking for usages.

lib/router.js
```diff
@@ -96,11 +96,31 @@ const guardAuth = (router) => router.beforeEach((to, from, next) => {
...
+// ALERTS
+
+// Hides the alert of the App component. Any modal should be destroyed upon a
+// route change, so we do not need to hide modal alerts, and there are no other
+// alerts. (We may need to modify this if we add other alerts.)
+const hideAppAlert = (router) => router.afterEach(() => {
+  const root = router.app;
+  // When using mockRoute() in testing, `root` is the App component. In
+  // production, `root` is the parent of the App component, though it might not
+  // have children components before the first render.
```
clint-tseng: is this discrepancy fixable at the root?

matthew-white: I think so — it'd entail changing `mockRoute()`. I think that'll definitely be needed if/when we use `router.app` in other places, and may be worth it now to clean up this code. What do you think?

matthew-white:
I definitely hear you on `$nextTick()`. I was running into some of that in an earlier approach I tried, when I was using `$emit()` and `$nextTick()` together and running more into delay wars. In this PR, I think I've only used `$nextTick()` in `FormList`, and it's actually needed not because of the alerts, but because creating a new form results in a route change:

```javascript
afterCreate(form) {
  // Wait for the modal to hide.
  this.$nextTick(() => this.$router.push(`/forms/${form.xmlFormId}`, () => {
    const name = form.name || form.xmlFormId;
    this.$alert().success(`The form “${name}” was created successfully.`);
  }));
}
```

If we weren't using jQuery, I think we wouldn't need `$nextTick()`. Currently, we need to give jQuery the chance to hide the modal, because otherwise it won't complete the modal clean-up process before `this.$router.push()` runs. Actually, part of me wants to replace the Bootstrap jQuery code with something in Vue in order to avoid the `$nextTick()`...

That said, the reason this code was moved from `FormNew` to `FormList` was because, like you said, in `FormNew`, `this.$alert()` returns the modal's alert, while in `FormList`, it returns the alert at the top of the page. I think that's a pattern we'll see in general: if a modal action triggers an alert after the modal hides, the mechanism by which that'll occur is the modal emitting an event to its parent component, which will set the alert. We'd mostly been using that pattern already — we didn't here only because the modal triggers a route change — but now that pattern will be more required.

lib/alert.js
matthew-white: Good point, I'll update this comment!

lib/router.js
matthew-white: I merged the PR, but definitely let me know if you think we should create an issue to try to change this sooner rather than later.

PR #107, opened 2018-07-05
Bump node-sass to latest now that they have resolved `npm audit` warning

matthew-white:
The latest version of node-sass now uses request 2.87, which apparently
resolves the `npm audit` warning.

I think we have to continue pinning node-gyp for now, but I am hoping
that node-gyp will similarly upgrade to request 2.87 soon. Once that
occurs, we should be able to remove it as a dev dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6165,9 +6165,9 @@
       }
     },
     "node-sass": {
-      "version": "4.6.1",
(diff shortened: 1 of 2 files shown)


matthew-white:
Since this is just a dependency change, I'm going to go ahead and merge.

PR #109, opened 2018-07-19
5.02b See response table as part of submissions list

matthew-white:


lognaturel:
Requires being on the `cxlt/temp/meta` backend branch.

lib/components/form/submission-row.vue
```diff
@@ -0,0 +1,155 @@
...
+      return formatDate(this.submission.__system.submissionDate);
+    }
+  },
+  methods: {
+    // Returns the submission's value for the specified question column,
+    // formatting it according to the question's type.
+    questionValue(column) {
+      const rawValue = column.path.reduce((acc, name) => acc[name], this.submission);
+      if (rawValue == null) return '';
+      switch (column.type) {
```
matthew-white: Note that `type` can be `undefined`.

lib/components/form/submissions.vue
```diff
@@ -131,14 +178,51 @@ export default {
...
+    border-right: $border-bottom-table-heading;
+  }
+
+  > tbody > tr > td:last-child {
+    border-right: $border-top-table-data;
+  }
+}
+
+#form-submissions-table2-container {
+  overflow-x: scroll;
```
matthew-white: Take the bottom margin off the table and put it on the container so that the margin is below the scrollbar, not above it.

lib/components/form/submissions.vue
```diff
@@ -95,23 +119,46 @@ export default {
...
+    questionColumns() {
+      const columns = [];
+      for (let i = 0; columns.length < 10 && i < this.schema.length; i += 1) {
+        const { type, path } = this.schema[i];
+        // We already display __id as the instance ID, so if there is also a
+        // meta.instanceID, we do not display it.
+        const isInstanceId = type === 'string' && path.length === 2 &&
+          path[0] === 'meta' && path[1] === 'instanceID';
+        if (!(type === 'repeat' || isInstanceId)) {
+          const header = path.join('.');
```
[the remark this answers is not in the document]
clint-tseng: did this get answered?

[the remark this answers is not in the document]
clint-tseng: oh github displays these comments weirdly; hold.

test/data.js
```diff
@@ -225,7 +196,7 @@ const testData = Object.assign(
     sort: sortByUpdatedAtOrCreatedAtDesc,
     views: {
       simpleForms: (extendedForm) => {
-        const form = omit(extendedForm, ['lastSubmission', 'submissions']);
+        const form = R.omit(['lastSubmission', 'submissions'], extendedForm);
```
clint-tseng: be very very careful with rambda's omit ([explained here](https://github.com/opendatakit/central-backend/blob/master/lib/util/util.js#L27))

lib/components/form/submission-row.vue
clint-tseng: or `"unknown"` https://github.com/opendatakit/central-backend/blob/master/lib/data/schema.js#L54

clint-tseng: but it's more or less impossible that you'll actually get a data value for a field whose type is unknown.

clint-tseng:
minor comments, nothing major.

lib/components/form/submission-row.vue
matthew-white: If we do encounter an `unknown` field, the current code will display nothing for a value that is `null` or `undefined`. Otherwise, it will display the value as a string without change.

PR #112, opened 2018-08-03
Replace Moment with Luxon

### LINKED ISSUE #28: Replace moment with Luxon?
@clint-tseng writes in #24:

> i have been a happy moment user for many many years now, but they have a new one http://moment.github.io/luxon/ that seems cleaner, lighter, and better (immutable, for one). take a look?

### LINKED ISSUE #42: Standard date formatting
As part of this, add tests of any table data that includes a date.

matthew-white:
Closes #28. Closes #42.

### DIFF
--- a/lib/components/form/list.vue
+++ b/lib/components/form/list.vue
@@ -60,7 +60,7 @@ except according to the terms contained in the LICENSE file.
               {{ form.createdBy != null ? form.createdBy.displayName : '' }}
             </td>
             <td>
-              {{ updatedAt(form) }}
+              {{ lastModified(form) }}
             </td>
             <td>
               {{ lastSubmission(form) }}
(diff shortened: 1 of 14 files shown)


lib/components/form/list.vue
```diff
@@ -114,13 +113,11 @@ export default {
       const s = form.submissions !== 1 ? 's' : '';
       return `${count} submission${s}`;
     },
-    updatedAt(form) {
-      const updatedAt = form.updatedAt != null ? form.updatedAt : form.createdAt;
-      return moment(updatedAt).fromNow();
+    lastModified(form) {
+      return formatDate(form.updatedAt != null ? form.updatedAt : form.createdAt);
```
clint-tseng: can you not use the new `updatedOrCreatedAt` here?

matthew-white: I haven't added the form presenter to the `FormList` component yet, but once I do, I'll start using `updatedOrCreatedAt` and can probably simplify the code in other places as well. I'm thinking I'll add the presenter the next time I touch the component.

PR #114, opened 2018-08-09
Form attachments

matthew-white:


lib/components/form/attachment/list.vue
```diff
@@ -117,9 +117,15 @@ export default {
       return this.uploadStatus.current != null;
     }
   },
+  beforeRouteLeave(to, from, next) {
```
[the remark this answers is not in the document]
matthew-white: With the change to `pointer-events`, the user is only able to navigate to a different `FormShow` page for the same form if they manually edit the URL. Even in that case, it looks like the `uploadFiles()` promise still runs to completion, and `FormAttachmentList` correctly emits the `attachment-change` events. To see that:

1. Start an upload in `FormAttachmentList`.
2. While the upload is in progress, navigate to a different `FormShow` page for the same form by manually editing the URL.
3. Wait enough time for the upload to finish.
4. Return to `FormAttachmentList`: it should be updated to show the latest `attachment.updatedAt` values.

lib/components/form/submissions.vue
```diff
@@ -170,22 +163,10 @@ export default {
...
-          // Revoke the previous URL.
-          if (this.downloadHref !== '#')
-            window.URL.revokeObjectURL(this.downloadHref);
-          this.downloadHref = window.URL.createObjectURL(data);
-          this.$nextTick(() => this.$refs.downloadLink.click());
-        })
-        .catch(() => {});
+    // The `disabled` class on the <a> element does not prevent keyboard
+    // navigation.
+    clickDownloadButton(event) {
```
clint-tseng: you could consider making this a global jQuery thing instead;
`$('body').on('click', 'a.disabled', (event) => { event.preventDefault(); });`

clint-tseng: (but only if that semantic always holds, of course)

matthew-white: Ah good idea, I'll do that!

test/setup.js
```diff
@@ -17,6 +17,12 @@ import { router } from '../lib/router';
 import { setHttp } from './http';
 import './assertions';
 
+// "iTrim" for "internally trim"
+// eslint-disable-next-line no-extend-native
+String.prototype.iTrim = function iTrim() {
```
matthew-white: This method is only added in testing.

test/util.js
```diff
@@ -19,14 +19,16 @@ export class MockLogger {
...
 
 // Triggers an event that does not bubble.
-export const trigger = (eventName, wrapper) => {
-  wrapper.trigger(eventName);
-  return Vue.nextTick();
+export const trigger = (eventName, wrapper, selector = undefined) => {
+  const target = selector == null ? wrapper : wrapper.first(selector);
+  target.trigger(eventName);
+  const nextTick = Vue.nextTick();
+  return selector == null ? nextTick : nextTick.then(() => wrapper);
```
clint-tseng: i don't really understand this line. why return the wrapper only if you scoped the wrapper to selector?

test/event.js
```diff
@@ -1,16 +1,91 @@
...
+
+
+////////////////////////////////////////////////////////////////////////////////
+// trigger
+
+export const trigger = {};
+
+const simpleEventNames = ['change', 'click', 'submit'];
+for (const eventName of simpleEventNames) {
+  // Triggers an event that does not bubble.
```
clint-tseng: i know this isn't new code but i kind of feel like you should _always_ bubble, rather than never, since that's what the browser will do..

test/components/form/submissions.spec.js
```diff
@@ -337,6 +338,21 @@ describe('FormSubmissions', () => {
...
+
+      it('disables the download button', () =>
+        loadSubmissions(1)
+          .afterResponses(component =>
+            component.first('#form-submissions-download-button')
+              .hasClass('disabled').should.be.false())
+          .request(component => trigger.click(component, '.btn-refresh'))
+          .beforeEachResponse(component =>
+            component.first('#form-submissions-download-button')
+              .hasClass('disabled').should.be.true())
```
clint-tseng: given you just added custom assertions for visible-or-not perhaps disabled deserves one too?

test/setup.js
clint-tseng: that doesn't mean i like it!

test/util.js
matthew-white: This is definitely confusing as-is... Ideally I'd like to return the wrapper all the time. The function hadn't previously been returning the wrapper, and I didn't have the chance to make sure that changing it wouldn't break any existing promise chain, so I kept the current behavior for the non-selector case and added the return-wrapper behavior for the new selector case. I've created an [issue](https://github.com/opendatakit/central-frontend/issues/133) to track this, so I should be able to make that change soon.

test/event.js
matthew-white: I added that comment because it was looking to me like the Vue testing library that we're using, avoriaz, doesn't ever bubble events. In the code, I just call the avoriaz `trigger()` method (the method doesn't have any option for bubbling vs. not), but I thought I was seeing no bubbling. However, looking back at it, the issue I was running into might have been due to Vue not knowing about jQuery's listeners. Maybe I should mention avoriaz specifically in the comment, or maybe I should just remove it if I'm no longer sure that it's right (or link to this thread?).

test/components/form/submissions.spec.js
matthew-white: Ah good idea. The assertion would have to be smart, looking at the `disabled` attribute for buttons and the `disabled` class for links, but I think it'd make these tests more readable. I think I'll add that the next time I test whether an element is disabled.

PR #150, opened 2018-09-17
Replace .table with .panel-simple on backups page

### LINKED ISSUE #86: BackupList: replace .table with .panel-simple
`.panel-simple` is better semantically, since the element is not actually a table.

matthew-white:
The page does not contain table data, so we are replacing the table with
an identically designed panel. .panel-simple was added after the backups
page was created.

Closes #86.

### DIFF
--- a/lib/components/app.vue
+++ b/lib/components/app.vue
@@ -364,6 +364,7 @@ h1 {
 
     .panel-title {
       font-size: $font-size-table-heading;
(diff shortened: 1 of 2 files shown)


matthew-white:
Note that nothing nested within `#backup-list-status-message` changed. That code is the exact same as before; it's just less indented.

PR #159, opened 2018-10-25
Submissions infinite scroll

matthew-white:


lognaturel:
Nice! Can you please share your quick script for generating and submitting filled forms so it’s easier to try this out?

lognaturel:
In case anyone else wants to check it out, [myform.xml.txt](https://github.com/opendatakit/central-frontend/files/2527902/myform.xml.txt) is a blank form that works with @matthew-white's quick script [here](https://github.com/opendatakit/central-frontend/pull/159#issuecomment-433738103). To get a token:
```
 curl --header "Content-Type: application/json" \
 --request POST \
 --data '{"email": "<your login email>", "password": "<your password>"}' \
 http://localhost:8383/v1/sessions
```

It's working great! At some point, it would be really good to have a visual indication that more records are being loaded (spinner?). Right now it's easy to miss that the scrollbar has jumped and that there are more records beyond the end of the screen.

It would also be great for the background not to show through when scrolling quickly:
scroll-fast

Both of these can be issues for later.

matthew-white:
I've now added the loading message. @clint-tseng, what's the best way to address the background showing during scrolling? Is the solution to remove the background color from the `html` element, or is there a better approach? If there's a quick fix, I'm happy to implement that now, otherwise I can create a GitHub issue?

lognaturel:
The loading message looks good! 👍

matthew-white:
I just created an issue: #165. @clint-tseng, once the changes to the PR are approved, I think it'll be ready to merge.

matthew-white:
Also happy to jump on a quick interactive review tomorrow!

lib/components/form/submission/list.vue
```diff
@@ -181,26 +184,35 @@ export default {
...
-    submissionsURL() {
-      return `/forms/${this.form.encodedId()}.svc/Submissions`;
+    chunkURL({ top, skip = 0 }) {
+      const queryString = `%24top=${top}&%24skip=${skip}`;
+      return `/forms/${this.form.encodedId()}.svc/Submissions?${queryString}`;
     },
-    processSubmissions({ data }) {
-      this.submissions = data.value != null ? data.value : [];
+    processChunk(chunkIndex, { data }) {
+      if (chunkIndex === 0)
```
clint-tseng: nit: if one block in if/else is braced i prefer both to be braced. nbd.

lib/components/form/submission/list.vue
```diff
@@ -209,8 +208,15 @@ export default {
...
         if (data.value != null) {
+          const lastSubmission = this.submissions[this.submissions.length - 1];
+          const lastSubmissionDate = lastSubmission.__system.submissionDate;
           for (const submission of data.value) {
-            if (!this.instanceIds.has(submission.__id)) {
+            // If a submission has been created since the initial fetch or last
+            // refresh, the latest chunk of submissions may include a newly
+            // created submission or a submission that is already shown in the
+            // table.
+            if (submission.__system.submissionDate <= lastSubmissionDate &&
```
clint-tseng: hm. is this overzealous? does the `.has` check not do it?

lib/components/form/submission/list.vue
```diff
@@ -225,6 +231,37 @@ export default {
         }
         this.chunks += 1;
       }
+      const remaining = this.originalCount - this.submissions.length;
+      // A negative value should be rare or impossible.
+      if (remaining <= 0)
```
clint-tseng: same nit

clint-tseng:
just nits. looks good!

lib/components/form/submission/list.vue
matthew-white: I'm not sure how sensible my approach really is, but I usually add braces to all `if`/`else` blocks below the first block with a brace. So I'll sometimes have starting `if` and `else if` blocks without braces that are followed by blocks with braces. In general, I like avoiding braces where I can, but I also like the consistency of what you're suggesting. Do you find that consistency more readable or easier to edit?

On a separate note, I also wanted to make sure that you're looking at the latest code, since this particular line is outdated. I think you are, since you comment on newer lines below, but I just wanted to check.

clint-tseng: yeah i started a bit too early but did all the commits

clint-tseng: if it were up to me i'd leave every brace possible out. but i think for readability for others i try to stay consistent within each block. if one case needs multiple lines, they all get braces. sometimes this leads to me trying to reorganize the code so that the multiple lines aren't necessary.

lib/components/form/submission/list.vue
matthew-white: I think the `.has` check might not catch all submissions created since the submission shown in the first row was created, since such submissions wouldn't have been seen before, so wouldn't be caught by `.has`. I think that would be pretty rare though, because a whole chunk's worth of new submissions would have to be created in-between two chunks. But if that did happen, I think something like this would be possible:

* The user navigates to the submissions page and loads the first 250 submissions, then steps out for lunch.
* During lunch, 300 new submissions are created.
* After lunch, the user fetches the second chunk of submissions (`$top=250, $skip=250`). That returns the last 50 of the 300 newly created submissions, as well as 200 submissions that are already shown in the table.
* The 50 submissions would pass the `.has` test, but we wouldn't want to include them in the table.

That is definitely pretty unlikely. But the way I'm thinking about it, the `submissionDate` check helps ensure that there are no out-of-order submissions, then the `.has` check ensures that there are no duplicates.

lib/components/form/submission/list.vue
matthew-white: That sounds like a good approach to me. :+1: I might leave these lines as-is for now (and I suspect that there are quite a few like them throughout Frontend), but I'll use consistent braces in the future.

PR #169, opened 2018-11-07
Suggest expired link as possibility after AccountClaim request error

matthew-white:


lib/components/account/claim.vue
```diff
@@ -51,6 +51,11 @@ export default {
     };
   },
   methods: {
+    problemToAlert(problem) {
+      return problem.code === 401.2
+        ? `${problem.message} The password reset link may have expired, and the password may need to be reset again.`
```
clint-tseng: I think I would suggest: “The link in your email may have expired, and a new email may have to be sent.”

PR #171, opened 2018-11-08
Icon tweaks

matthew-white:


lib/components/backup/list.vue
```diff
@@ -24,11 +24,11 @@ except according to the terms contained in the LICENSE file.
           <button v-if="backups.status == 'notConfigured'"
             id="backup-list-new-button" type="button" class="btn btn-primary"
             @click="showModal('newBackup')">
-            <span class="icon-plus-circle"></span> Set up now
```
clint-tseng: so these changes are probably fine, but you may want to consider what happens if you want a newline after the closetag and before the text label. that will cause a different case, now, than what you have here where before they were isomorphic.

lib/components/field-key/row.vue
```diff
@@ -18,8 +18,8 @@ except according to the terms contained in the LICENSE file.
       <a v-if="fieldKey.token != null" ref="popoverLink" href="#"
         class="field-key-row-popover-link text-no-decoration" role="button"
         @click.prevent="showCode">
-        <span class="icon-qrcode"></span>
-        <span class="underline-within-link">See code</span>
+        <span class="icon-qrcode"></span><span class="underline-within-link">See
```
clint-tseng: and in fact i guess here’s one example already so you already know.

lib/components/field-key/list.vue
```diff
@@ -186,10 +186,6 @@ export default {
 #field-key-list-table {
   > tbody > tr > td {
     vertical-align: middle;
-
-    .dropdown {
```
clint-tseng: which dropdowns are these?

lib/components/backup/list.vue
matthew-white: For an icon within a button or link, going forward, I think I'm going to try to ensure that there is never a space or newline between the icon and the content that follows. How does that sound? Just to make sure I understand your comment, an icon followed by a space, then followed by content would look the same as an icon followed by a newline, then followed by content, right? Or are the space and newline treated differently?

lib/components/field-key/list.vue
matthew-white: These are the actions dropdowns in the rows of the web user table and app user table.

lib/components/backup/list.vue
clint-tseng: all whitespace is treated equivalently by browsers. so space and newline would be the same. if you’re committed to ensuring no whitespace anywhere and catching that in all code reviews then it’s fine. but it would seem to me that the other approach is more flexible.

lib/components/field-key/list.vue
clint-tseng: so what does this change do? they were locked to some width before and now..?

lib/components/backup/list.vue
matthew-white: I didn't have a big issue with the space between the icon and the content, but I think I needed to remove it in order to remove the `text-no-decoration` and `underline-within-link` classes. My aim with those classes was to remove a link's hover/focus underline from beneath the link's icon: since we vertically align the icon, the icon's underline was vertically positioned at a different level than the rest of the link's content. I was able to remove the icon underline without much trouble even without `.text-no-decoration` and `.underline-within-link`, but unless I removed the space between the icon and the rest of the content, the space between the icon and the content was underlined, which looked a little off. I'll try to catch any extra spaces going forward and will document the need for that. If that gets to be too much trouble, we might want to consider a different approach...

lib/components/field-key/list.vue
clint-tseng: oh i see. okay we can fix it later if we care.

lib/components/backup/list.vue
clint-tseng: oh, i see. uhh okay. that’s fine.

PR #202, opened 2019-02-28
Add project managers

matthew-white:


assets/scss/_variables.scss
```diff
@@ -18,6 +18,7 @@ $color-page-background: #f7f7f7;
 // Text
 $color-text: #333;
 $font-family-monospace: Menlo, Monaco, Consolas, "Courier New", monospace;
+$max-width-p: 600px;
```
matthew-white: Is it better to apply this to all `<p>` elements?

lib/components/project/user/row.vue
```diff
@@ -0,0 +1,119 @@
...
+      <form>
+        <div class="form-group">
+          <select ref="select" :value="selectedRole" :disabled="disabled"
+            :title="title" class="form-control" aria-label="Project Role"
+            @change="assignRole">
+            <option value="manager">Manager</option>
+            <option value="">None</option>
+          </select>
+          <span class="spinner-container">
+            <spinner :state="requesting"/>
```
matthew-white: Should we not wrap a `<spinner>` in a `<span>`? (We do it in `FormEdit` as well.)

lib/components/project/user/list.vue
```diff
@@ -0,0 +1,226 @@
...
+      The assigned Project Managers for this Project will be able to perform any
+      administrative or auditing task related to this Project. Sitewide
+      Administrators are automatically considered Managers of every Project. To
+      learn more about Projects and Managers, please see
+      <doc-link to="central-projects/#project-managers">this article</doc-link>.
+    </p>
+    <form id="project-user-list-search-form" @submit.prevent>
+      <!-- When search is disabled, we hide rather than disable this button,
+      because Bootstrap does not have CSS for .close[disabled]. -->
+      <button v-show="q != '' && !searchDisabled" type="button" class="close"
```
matthew-white: Should we make this a darker color? If so, what should the hover/etc. colors be?

lib/components/project/user/list.vue
```diff
@@ -0,0 +1,226 @@
...
+      </thead>
+      <tbody v-if="tableAssignments != null">
+        <project-user-row v-for="assignment of tableAssignments"
+          :key="assignment.actor.id" :project-id="projectId"
+          :assignment="assignment" @increment-count="incrementCount"
+          @decrement-count="decrementCount" @success="afterAssign"/>
+      </tbody>
+    </table>
+    <loading :state="$store.getters.initiallyLoading(['managers', 'users'])"/>
+    <p v-if="emptyMessageShown" class="empty-table-message">
```
matthew-white: Add a message for if the search returned no results?

PR #208, opened 2019-04-02
1.05 Grant/revoke administrative privileges

matthew-white:


lib/components/navbar.vue
```diff
@@ -26,25 +26,26 @@ except according to the terms contained in the LICENSE file.
       </div>
 
       <div id="navbar-collapse" class="collapse navbar-collapse">
-        <ul class="nav navbar-nav">
-          <li :class="{ active: $route.path === '/' || routePathStartsWith('/projects') }">
+        <ul v-if="currentUser != null" id="navbar-links" class="nav navbar-nav">
```
matthew-white: Hide the projects link if the user is not logged in?

lib/components/project/list.vue
```diff
@@ -89,33 +88,32 @@ except according to the terms contained in the LICENSE file.
...
-            <tbody>
-              <project-row v-for="project of projects" :key="project.id"
-                :project-count="projects.length" :project="project"
-                @show-introduction="showIntroduction"/>
-            </tbody>
-          </table>
-        </template>
+        <p v-if="projects != null && projects.length === 0"
+          class="empty-table-message">
+          You do not have access to any Projects.
```
matthew-white: Better empty message?

lib/components/user/list.vue
```diff
@@ -9,29 +9,48 @@ https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
...
+        from that Project's settings.
+      </p>
+      <p>
+        For more information,
+        <doc-link to="central-users/#managing-app-users">click here</doc-link>.
+      </p>
+    </div>
+    <div class="table-actions">
+      <refresh-button :configs="configsForRefresh"
+        :disabled="assignRequestCount !== 0"/>
```
matthew-white: Where should we position the refresh button? Near the create button?

lib/components/user/list.vue
```diff
@@ -9,29 +9,48 @@ https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
...
+either is an Administrator or has no role. -->
 <template>
   <div>
-    <float-row class="table-actions">
-      <refresh-button slot="left" :configs="refreshConfigs"/>
-      <button id="user-list-new-button" slot="right" type="button"
-        class="btn btn-primary" @click="showModal('newUser')">
+    <div class="heading-with-button">
+      <button id="user-list-new-button" type="button" class="btn btn-primary"
+        @click="showModal('newUser')">
```
matthew-white: Should this button be larger?

test/components/user/list.spec.js
```diff
@@ -1,43 +1,238 @@
...
+              })
+              .sorted())
+            .respondWithData(() =>
+              testData.toActor(testData.standardUsers.sorted().slice(0, 2)))
+            .complete()
+            .request(component => {
+              const select = component.find('table select')[rowIndex];
+              if (select.element.value === selectValue)
+                throw new Error('no change');
+              select.element.value = selectValue;
```
matthew-white: Instead of setting `value` on the `<select>`, is it better to set `selected` on the `<option>`? (We have a similar line to this in the `ProjectUserList` tests as well.)

matthew-white:
`beforeResponses()`

  - Rename to `beforeAnyResponse()`.
  - Remove multiple catches. (Also do this for `beforeEachResponse()`.)

lib/components/user/list.vue
matthew-white: It's OK where it is now.

PR #217, opened 2019-05-27
Switch from vueify to Vue CLI

matthew-white:


src/components/form/overview.vue
```diff
@@ -35,6 +35,7 @@ except according to the terms contained in the LICENSE file.
           Your Form design references files that we need in order to present
           your Form. You can upload these for distribution under the
           <router-link :to="formPath('media-files')">
+            <!-- eslint-disable-next-line vue/multiline-html-element-content-newline -->
```
issa-tseng: maybe just turn this one off.

PR #219, opened 2019-06-05
Make the project name a link on form pages

matthew-white:


src/components/page/head.vue
```diff
@@ -59,7 +59,12 @@ export default {
 
   > span:first-child {
     font-weight: bold;
-    margin-right: 5px;
+    margin-right: 10px;
```
issa-tseng: any particular reason this changed?

matthew-white: I removed a trailing space after the project name, which reduced the space between the two links, so I added that space back here. It should look the same as before!

issa-tseng: okay

PR #228, opened 2019-07-05
3.03 Filter audit log list

matthew-white:


src/assets/scss/app.scss
```diff
@@ -252,6 +253,16 @@ a {
   }
 }
 
+// The flatPickr input is readonly by default, but we do not want to style it as
```
issa-tseng: does it work by keyboard input if i tab to it though?

src/components/audit/list.vue
```diff
@@ -15,26 +15,53 @@ except according to the terms contained in the LICENSE file.
...
-      key: 'audits',
-      url: '/audits?action=nonverbose',
-      extended: true
-    }]).catch(noop);
+    this.fetchData(this.initialFilters);
+  },
+  methods: {
+    dateParams(dateRange) {
+      if (dateRange.length === 0) return '';
+      const start = encodeURIComponent(dateRange[0].toISO());
```
issa-tseng: do these generate the `z`?

issa-tseng: oh it generates the local tz offset?

issa-tseng: what if someone happens to be in GMT, will it still do `+0`?

test/components/audit/filters.spec.js
```diff
@@ -0,0 +1,116 @@
...
+  it('sends a request after a range of multiple dates is selected', () => {
+    const restoreLuxon = setLuxon({ defaultZoneName: 'UTC' });
+    return mockHttp()
+      .mount(AuditList)
+      .respondWithData(() => testData.extendedAudits.sorted())
+      .complete()
+      .request(component => {
+        const start = DateTime.fromISO('1970-01-02').toJSDate();
+        const end = DateTime.fromISO('1970-01-03').toJSDate();
+        // Ideally, we would actually open the flatPickr calendar and select the
```
issa-tseng: then we should put this on the list (which we should write down) of things the human testers should check with every release.

test/components/audit/filters.spec.js
```diff
@@ -0,0 +1,116 @@
...
+        config.url.should.containEql('start=1970-01-02T00%3A00%3A00.000Z');
+        config.url.should.containEql('end=1970-01-02T23%3A59%3A59.999Z');
+      })
+      .respondWithData(() => testData.extendedAudits.sorted())
+      .finally(restoreLuxon);
+  });
+
+  // The date range can be cleared if, for example, the user presses backspace
+  // after opening the flatPickr calendar. There doesn't seem to be an easy way
+  // to turn that behavior off, so here we test that Frontend reacts reasonably
```
issa-tseng: for definitions of "reasonably" that include "downloads every audit log ever maybe over a really awful connection"

src/assets/scss/app.scss
matthew-white: Right now it doesn't, which is the flatPickr default, and I think the user would have to type something goofy like "2019/01/01 to 2019/01/31". But flatPickr does seem to support direct keyboard input. I'm happy to configure that if that'd be useful!

src/components/audit/list.vue
matthew-white: Exactly, it generates the local TZ offset. It appends Z if the user happens to be in GMT. I could also make it always UTC (with the appended Z) if that'd be better than the local TZ offset?

test/components/audit/filters.spec.js
matthew-white: Haha now that you mention it, that sounds less reasonable than I thought. I'll go ahead and make it so that if the user clears the date range, it'll default to the current date (if that sounds good). There may be a brief flicker though between when the date range is cleared and when it's repopulated with the current date.

test/components/audit/filters.spec.js
matthew-white: Good idea! Shall I create that document, or would you like to? If I remember correctly, backups are the main other item that needs to be tested manually, right?

src/assets/scss/app.scss
issa-tseng: yes, please. keyboard input is important.

src/components/audit/list.vue
issa-tseng: okay

test/components/audit/filters.spec.js
issa-tseng: please do that

test/components/audit/filters.spec.js
issa-tseng: yes, sounds great

test/components/audit/filters.spec.js
matthew-white: The document is [here](https://docs.google.com/document/d/1VJmTS9akseiie8zFmUNe-lnCNv9A7sBt2w_ZVV2GVJ8/edit?usp=sharing)!

src/assets/scss/app.scss
matthew-white: I looked more into this, and it turns out that the default configuration actually does support keyboard navigation through the calendar: if you tab to the input, the calendar will open, then you can select a range using the arrow keys and Enter.

matthew-white: So I think it should work as-is.

PR #242, opened 2019-08-08
Remove Tableau from SubmissionAnalyze

matthew-white:


matthew-white:
@issa-tseng, I'm realizing that with this change, the OData URL is the same for all three tabs. I'm not sure that that's an issue or whether we need to solve it now even if it is, but I thought I'd mention it.

issa-tseng:
yeah i caught that too. we will just leave it alone for now. possibly tableau will be restored in the future.

PR #246, opened 2019-08-09
Do not show encryption message if fieldColumns is empty

matthew-white:


src/components/submission/row.vue
```diff
@@ -210,7 +210,7 @@ $icon-lock-margin-right: 12px;
   }
 
   ~ .encrypted-submission {
-    td:first-child {
+    td[colspan] {
```
issa-tseng: pretty slow. i'd advise using a class here.

PR #259, opened 2019-11-19
Import an XLSForm

matthew-white:


src/components/form/new.vue
```diff
@@ -99,64 +126,92 @@ export default {
...
   methods: {
-    readFile(files) {
-      if (files.length === 0) return;
+    processXLSForm(file) {
+      this.$alert().blank();
+      this.file = file;
+      this.xml = null;
+      this.warnings = null;
+    },
+    processXForm(file) {
```
matthew-white: Can we just send the `File` object in both cases?

PR #260, opened 2019-11-21
Make changes to Form Access tab

matthew-white:


src/components/project/form-access.vue
```diff
@@ -23,21 +23,22 @@ except according to the terms contained in the LICENSE file.
         Here you can set Form States, which control whether Forms are available
         for download and open for submission. You can also separately control
         which App Users may see each Form at all. For more information,
-        <doc-link to="central-projects/#managing-form-workflow">click here</doc-link>.
+        <doc-link to="central-projects/#central-projects-form-access">click here</doc-link>.
```
issa-tseng: i don't know what is happening here. when i build the docs on my local computer, i end up with the internal name as the hashlink. when i look on the actual site, it's the friendly name. i think therefore probably the correct url here is actually going to be #managing-form-access

matthew-white: Ah interesting. I've changed it to `central-projects/#managing-form-access`.

PR #266, opened 2019-11-28
Determine content type of form based on file extension

matthew-white:


issa-tseng:
you can't use `file.type` if it exists and do the fallback if it doesn't?

matthew-white:
Ah good point; I didn't think of that. Apparently browsers also just use the file extension to determine the `type` property of a `File` object, so I think the behavior should be pretty similar either way. I could see that potentially simplifying the code though. I'll go ahead and merge for now, but this could be something to circle back to.

PR #289, opened 2020-03-27
Update audits table for v0.8

matthew-white:


src/components/audit/filters.vue
```diff
@@ -68,6 +68,9 @@ export default {
         categoryOption('Form Actions', 'form'),
         actionOption('Create', 'form.create'),
         actionOption('Update Details', 'form.update'),
+        actionOption('Set Draft', 'form.update.draft.set'),
```
lognaturel: I mentioned to @matthew-white that I wouldn't really know what "set" means here. He described that when a draft is first created there would be two of these logged: creating the draft and then setting the draft form definition. I think it's pretty confusing that there are two of the same event. I don't think this is hugely important and I recognize that this is what the backend does but wanted to flag it. CC @issa-tseng

issa-tseng: what do you propose?

matthew-white: It's a little verbose, but would "Create or Update Draft" be clearer than "Set Draft"?

issa-tseng: sure

lognaturel: Yes, that sounds great to me.

PR #311, opened 2020-04-21
Update doc links

matthew-white:


issa-tseng:
but maybe also a getodk change here?

matthew-white:
I might run out of time for the opendatakit→getodk change, but I did create a PR for the "Open Data Kit"→ODK change.

PR #314, opened 2020-04-24
Use the same version of node for CircleCI as we use in production

matthew-white:


matthew-white:
@issa-tseng, do you think we can remove these lines from the config?

```
machine:
  node:
    version: stable
```

I'm not sure, but I think those lines might have been from our CircleCI 1.0 config. I'm not sure how they interact with how the version of node is specified elsewhere in the config (maybe these lines have no effect in the CircleCI 2.0 config?).

PR #320, opened 2020-05-09
Add internationalization

### LINKED ISSUE #230: Add autocomplete attribute to inputs
When a user navigates to the login page in Chrome, the following message is displayed in the console:

> [DOM] Input elements should have autocomplete attributes (suggested: "current-password"): (More info: https://goo.gl/9p2vKq)
>
> `<input type="password" placeholder="Password *" required="required" class="form-control">`

We specify `autocomplete="off"` for user email inputs (see #229). We should specify an autocomplete attribute for other inputs, as well. For example, for the login page, specifying current-password for the password input seems sensible.

### LINKED ISSUE #244: Show full date in title attribute
We format dates that we show in Frontend. @issa-tseng writes:

> within a week it shows the day name, and beyond that it shows the exact date. we should make it so that when you hover over a date you always get the exact full date/time though.

I'm thinking we could create a date component that does this.

### LINKED ISSUE #279: text: can we suggest that the encrypted submissions download modal can be closed?
right now it never says when you can dismiss the modal. we should do that if we can, but maybe there's a reason we can't.

### LINKED ISSUE #323: Audit action options are not indented
The audit action select element in /system/audits renders a series of options. Some options represent categories of actions, while others represent specific actions. The specific actions are supposed to be indented, but aren't right now:

Screen Shot 2020-06-06 at 5 10 24 PM

I'm guessing this is related to a change in how Vue handles white space and [this bug](https://github.com/vuejs/vue/issues/11059) in particular. If we don't want to wait for that bug to be resolved, I think we could also fix it by moving the non-breaking spaces from the template to `actionOptions` (specifically having `actionOption()` prefix the text with the non-breaking spaces).

matthew-white:
This PR adds internationalization to Frontend. It also:

Closes #230.
Closes #244.
Closes #279.
Closes #323.


matthew-white:
Things to review in the first code review:

- Structure of `en.json`
  - Root properties
  - `localeN`
  - `<i18n>` components
  - Modal buttons
  - `ProjectList` might be a good example.
- How we load locales
  - `src/i18n.js`
  - `src/util/i18n.js`
  - `src/router.js`
  - Maybe: associated tests.
- Problems
  - `src/util/request.js`
  - `AccountLogin` vs. `FormNew`
- Components
  - `TimeAndUser`
  - `FormGroup`

matthew-white:
Things we may review in code review:

- `restructure.js`: https://github.com/getodk/central-frontend/pull/320/commits/7635852225bd573b8a651f0f12b9e2ae138da477
- `NavbarLocaleDropdown`: https://github.com/getodk/central-frontend/pull/320/commits/38a8accc79473f1d6f4d768c56486e49cadf0455
- There have been enough changes to the router since the first code review that that may be worth another look: https://github.com/getodk/central-frontend/pull/320/commits/b625137af830cf08a8c2242145ed6c7a49315986
- I started using `$n()`: https://github.com/getodk/central-frontend/pull/320/commits/fbc1acc387b9235c6095537f82bcb96fef5d4d1b and https://github.com/getodk/central-frontend/pull/320/commits/7a3d080cbb28d2415b1949c39696585bb38ebf55
- Within testing, I updated the `mount()` function in `/test/util/lifecycle.js` to account for single file component `i18n` custom blocks: https://github.com/getodk/central-frontend/pull/320/commits/fbc1acc387b9235c6095537f82bcb96fef5d4d1b

lognaturel:
I think we don't ever want ODK or Central or ODK Central to be translated. The `appName` resource seems like it should not be a translatable string. Perhaps developer comments are sufficient to address ODK and Central standing alone.

src/components/field-key/list.vue
```diff
@@ -227,3 +223,29 @@ export default {
...
+
+<i18n lang="json5">
+{
+  "en": {
+    "action": {
+      "create": "Create App User"
+    },
+    "heading": {
+      "full": "App Users in this Project only will be able to download and use Forms within this Project. When you create a new App User, it will not have access to any Forms at first. To set the Forms each App User may access, use the {formAccess} tab. Multiple devices can use the same App User profile without problem. For more information, {clickHere}.",
+      "formAccess": "Form Access",
```
lognaturel, line 235: It seems this should always be the same as the tab name. Is it impractical to use the same resource?

matthew-white: I was worried about a change due to the accusative case, but maybe "tab" would receive the accusative marker, but "Form Access" wouldn't? Also, since "the Form Access tab" is a definite noun phrase, I think construct state might be used in Arabic or Hebrew that would change the spelling of "Form Access": https://en.wikipedia.org/wiki/I%E1%B8%8D%C4%81fah. But I don't know enough to know how likely that is here.

transifex/strings_en.json
```diff
@@ -0,0 +1,1800 @@
...
+    "close": {
+      "string": "Close",
+      "developer_comment": "This is the text for an action, for example, the text of a button."
+    },
+    "create": {
+      "string": "Create",
+      "developer_comment": "This is the text for an action, for example, the text of a button."
+    },
+    "createAnother": {
+      "string": "Create another",
```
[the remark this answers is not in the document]
matthew-white: Sounds good! I've made this change here: https://github.com/getodk/central-frontend/pull/320/commits/933ca922366830813d0186467b5bf2a6582c86db.

matthew-white:
> I think we don't ever want ODK or Central or ODK Central to be translated. The `appName` resource seems like it should not be a translatable string. Perhaps developer comments are sufficient to address ODK and Central standing alone.

I've removed `appName` here: https://github.com/getodk/central-frontend/pull/320/commits/23c96312e900fb685b5a987be8cdbdfd132dd5e8. Let me know if I should add comments anywhere!

src/components/user/list.vue
```diff
@@ -208,3 +206,29 @@ export default {
...
+      }
+    ],
+    "header": {
+      "sitewideRole": "Sitewide Role"
+    },
+    "alert": {
+      "create": "A user was created successfully for “{displayName}”.",
+      "assignRole": "Success! “{displayName}” has been given a Sitewide Role of “{roleName}”.",
+      "resetPassword": "The password for “{displayName}” has been invalidated. An email has been sent to {email} with instructions on how to proceed.",
+      "retire": "The user “{displayName}” has been retired."
```
matthew-white, line 230: @lognaturel, will `assignRole` and `retire` here require gendered conjugation in Spanish? Would it help to write something like "The account for {displayName} has&hellip;" rather than "The user {displayName} has&hellip;"?

lognaturel: It's not an issue in either language because of the passive construction.

src/components/project/list.vue
```diff
@@ -224,3 +222,51 @@ export default {
...
+    "gettingStarted": {
+      // This is a title shown above a section of the page.
+      "title": "Getting Started",
+      "body": [
+        {
+          "full": "If you’re not sure where to begin, we have a getting started guide and user documentation available on the {docsWebsite}.",
+          "docsWebsite": "ODK Docs website"
+        },
+        {
+          "full": "In addition, you can always get help from others on the {forum}, where you can search for previous answers or ask one of your own.",
```
lognaturel, line 242: This string doesn't really make sense. Maybe "search previous questions or ask one of your own"? Or "search through existing questions or ask a new one"? I think this would be nice to fix while we're here since it's on the landing page.

src/components/backup/list.vue
```diff
@@ -17,7 +17,7 @@ except according to the terms contained in the LICENSE file.
         @terminate="showModal('terminate')"/>
       <page-section v-if="audits.length !== 0">
         <template #heading>
-          <span>Latest Backups</span>
+          <span>{{ $t('auditsTitle') }}</span>
```
lognaturel, line 20: Do you maybe mean `backupsTitle`?

matthew-white: Ah interesting. This is the section at the bottom of the page that lists audit log entries, which is why I chose that key. But of course, they're also all backup-related audit log entries. `BackupStatus` is what shows the main title for the page, so I think I'd lean toward keeping `auditsTitle` to try to make that distinction clearer.

lognaturel: That makes sense!

src/locales/en.json
```diff
@@ -5,11 +5,81 @@
...
+    // Form States control the lifecycle state of each Form. "Closing" is one
+    // Form State.
+    "closing": "Closing",
+    // Form States control the lifecycle state of each Form. "Closed" is one
+    // Form State.
+    "closed": "Closed"
+  },
+  "audit": {
+    // This is a category of actions performed on the server. It is shown when
+    // filtering the log of actions by type. Note that title case is used.
```
lognaturel, line 29: This applies to the other comments below as well -- I don't know how helpful "Note that title case is used" is. Title case doesn't exist in most languages. It's unusual to have parentheses, though, and that might be worth calling out.

lognaturel: Is this part of a drop down? That context would help pick the right words, I think.

src/components/date-time.vue
```diff
@@ -16,22 +16,33 @@ except according to the terms contained in the LICENSE file.
...
+    }
   },
   computed: {
+    fromISO() {
+      return DateTime.fromISO(this.iso);
+    },
+    withLocale() {
+      return this.fromISO.locale === this.$i18n.locale
+        ? this.fromISO
+        : this.fromISO.setLocale(this.$i18n.locale);
```
lognaturel, line 37: Does `setLocale` not already do this check and do nothing if the specified locale is the same? Not a big deal but I was surprised to see this logic here.

src/components/date-time.vue
```diff
@@ -16,22 +16,33 @@ except according to the terms contained in the LICENSE file.
...
+    },
     title() {
-      return this.iso != null
-        ? DateTime.fromISO(this.iso).toFormat('y/MM/dd HH:mm:ss')
-        : null;
+      return this.iso != null ? formatDateTime(this.fromISO) : null;
```
lognaturel, line 40: Why no locale here?

CONTRIBUTING.md
```diff
@@ -170,6 +170,7 @@ To add a new locale to ODK Central Frontend:
 
 1. Add the locale to Transifex.
 2. Add the locale to `locales` in [`/src/i18n.js`](/src/i18n.js).
+3. Check that there is a flatpickr config for the locale.
```
lognaturel, line 173: And if there isn't one, chat in Slack to figure out what to do?

src/components/date-time.vue
matthew-white: `setLocale()` seems to return a new `DateTime` object even if the locale is the same, so I added the ternary so that `this.fromISO` can be reused if the locale is the same.

lognaturel: Fair enough!

src/components/date-time.vue
matthew-white: The title shows an absolute time stamp, not a relative date, so no localization takes place. And because we just use `this.fromISO` here instead of `this.withLocale`, if the locale changes, Vue won't have to re-evaluate the title. (I don't think it's a big optimization though, so I think it would have been fine to use `this.withLocale` here.)

lognaturel: I see that makes sense. For some reason I thought it could be relative there as well.

Issue #232, opened 2019-07-12
Lazy load routes


matthew-white:
Now that we're using Vue CLI / webpack, we should be able to [lazy load routes](https://router.vuejs.org/guide/advanced/lazy-loading.html) in order to reduce our bundle size.

PR #321, opened 2020-05-25
Added support for Password validation

### LINKED ISSUE #45: Use something like zxcvbn to indicate password strength
Currently, when a user sets or resets their password, we do not restrict their password in any way. We also do not indicate the strength of their password.

At the least, we wish to indicate to the user the strength of their password. We're thinking we could use zxcvbn or a library like it to do so.

We may also wish to add password restrictions rather than just warnings, such that the user cannot choose a password that fails to satisfy certain restrictions. I think we would consider at least a restriction on password length ("the password must be at least X characters"). If we add such a restriction, we may need to update Backend as well.

lakshyagupta21:
Closes #45 
* Minimum 8 characters required for password field
* Password will only be set when it meets certain criteria of zxcvbn and when it return scrore > 2 [https://github.com/dropbox/zxcvbn]

Screenshot 2020-05-25 at 7 28 55 PM


matthew-white:
Thanks, @lakshyagupta21! I'm excited to see you in the Frontend codebase!

This is reminding me that we need to discuss some of the details of how password validation should work. For example, we haven't figured out yet whether we want to require a strong password or just indicate password strength. We also have passphrases in `BackupNew` (where the passphrase is optional) and in `ProjectEnableEncryption`, so this behavior could be different in different places. I'll follow up with @lognaturel and @issa-tseng and see what their thoughts are.

Another thing to think about is the bundle size of zxcvbn, which is about 400kB gzipped. That's pretty big, especially on a slower connection, so I think we'll want to load it asynchronously somehow. Maybe the best approach would be to import it synchronously in `AccountClaim`, but [lazy load](https://router.vuejs.org/guide/advanced/lazy-loading.html) the `AccountClaim` route. We've been wanting to add lazy loading anyway (#232), so I'll think more about this. Let me know if you think of anything!

matthew-white:
Hi @lakshyagupta21! I chatted with @lognaturel and @issa-tseng and wanted to throw out some thoughts from that conversation:

- We want to make sure that there aren't too many restrictions around the password, which can impact the user experience.
- That said, requiring a minimum length for the password seems like a good idea. Let's go with a minimum of 10 characters rather than 8.
- I don't think we want to require a minimum password score, so I think we can remove `@score` and `@feedback`. However, the strength meter is a great addition, and the screenshot looks great.

The user can set their password in `AccountClaim`, and they can also set it in `UserEditPassword`. It'd be great for those two components to implement the same password validation. Note that `UserEditPassword` does some form validation already.

As I mentioned above, there are also passphrases in `BackupNew` and `ProjectEnableEncryption`. We don't want to have any restrictions on those passphrases, especially since the `BackupNew` passphrase is optional. However, we still want to show the strength meter.

We're thinking that we don't want to bundle zxcvbn, even with `AccountClaim` where the password field is shown immediately, because the package is so large. Instead, any component that wants password validation should load zxcvbn asynchronously. This means that there is the possibility that a password field will be shown and even filled out before zxcvbn has finished loading. That's OK: we'll have Frontend implement this password validation if it can if/when zxcvbn has loaded, but skip it while zxcvbn is still loading or if it fails to load.

There seems to be some related discussion in the vue-password-strength-meter repo [here](https://github.com/apertureless/vue-password-strength-meter/issues/20). Vue [async components](https://vuejs.org/v2/guide/components-dynamic-async.html#Async-Components) may come in handy here: we can import `Password` asynchronously, which I think will then import zxcvbn. As background, Vue School has a nice article on lazy loading [here](https://vueschool.io/articles/vuejs-tutorials/lazy-loading-and-code-splitting-in-vue-js/). One technique that it highlights that is also used with async components is [webpack dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports).

To be clear, we should still show the password field while the async load is in progress, and the user should still be able to submit the form. This makes me think that we should keep the existing `<input>` element, then use vue-password-strength-meter just for the strength meter (we can use the `strengthMeterOnly` prop). We'd lose the password count badge, but I think that's OK, especially since you've added an alert about the password length.

OK, I'll stop here! Let me know if you have any questions or thoughts about the above!

lakshyagupta21:
Thanks @matthew-white for the feedback, I'm new to frontend and backend so I might take time to get things implemented as expected and may need more feedback on the changes that I do.


Following are the screenshots of the changes in the UI I did.


Screenshot 2020-05-31 at 1 04 14 PM
Screenshot 2020-05-31 at 1 03 56 PM
Screenshot 2020-05-31 at 1 03 13 PM
Screenshot 2020-05-31 at 1 03 33 PM

lakshyagupta21:
Thanks @matthew-white for the feedback, Attached the UI screenshots as well with the new changes.
Not sure why lint didn't throw the error last time.

lognaturel:
Heya, @lakshyagupta21! The screenshots are looking great to me. You probably saw that we've been heads down making a pretty invasive and exciting change to pull out all strings for localization. Since I tagged you in https://github.com/GetODK/central-frontend/pull/335 I wanted to mention that @matthew-white will follow up shortly with more context. Basically he has to restructure some code for translations to be possible and he'll help you with the merge so we can get this out to users!!

matthew-white:
I suggested defining a computed property named `invalidPassword`, but thinking about it now, that won't work, because we only want to add the `has-error` class after the user has submitted the form. Instead, one option would be to make `invalidPassword` a data property, similar to `this.mismatch`. Like `this.mismatch`, `invalidPassword` could be set in `submit()`.

lakshyagupta21:
Hi @matthew-white  @lognaturel, 
Thanks for all your comments, sorry for keeping it open for a very long time. Now, I've resolved the merge conflicts as well also addressed the review comments.
I've also added the unit test cases as you suggested. Not sure if I wrote them in a correct way, let me know if there is anything which could've been done differently in that.

lakshyagupta21:
Hi @matthew-white @lognaturel ,

It seems again it got the resolved conflicts and now its fixed and there were some lint error which I think I somehow missed last time also got fixed in my last commit. Will look forward for your next round of review.

lognaturel:
Thank you, @lakshyagupta21!!

PR #336, opened 2020-07-06
Disable form preview button if form is not open

matthew-white:


src/components/form-version/standard-buttons.vue
```diff
@@ -113,7 +120,8 @@ export default {
       "xForm": "As XForm",
       "xlsForm": "As XLSForm"
     },
-    "previewDisabled": "Preview has not finished processing for this Form. Please refresh later and try again."
+    "previewDisabled": "Preview has not finished processing for this Form. Please refresh later and try again.",
+    "previewDisabledUnlessOpen": "In this version of ODK Central, preview is only available for Open Forms."
```
lognaturel, line 124: Ahh, interesting. Right, I hadn't considered that since we're using the same form list as for clients, there's some incompatibility between features. 🤔

matthew-white, line 124: "Forms in the Open state"

PR #349, opened 2020-07-22
Add a button to fill out a new submission using Enketo

matthew-white:


src/components/enketo/preview.vue
```diff
@@ -0,0 +1,64 @@
...
+    formVersion: {
+      type: Form,
+      required: true
+    }
+  },
+  computed: {
+    disabledTitle() {
+      if (this.formVersion.enketoId == null)
+        return this.$t('disabled.processing');
+      if (this.formVersion.publishedAt != null &&
```
lognaturel, line 38: Nit: reverse the test order? It seems that if the form is not open then it doesn't matter whether it's been processed or not and that's what provides the most info.

src/locales/en.json
```diff
@@ -87,6 +87,8 @@
     "cancel": "Cancel",
     "close": "Close",
     "create": "Create",
+    // This is the text of a button that is used to fill out a new Submission.
```
lognaturel, line 90: Might be helpful to say that it appears next to a headline that says "Submissions". Otherwise folks might be tempted to be helpful and add more text.

test/components/form-draft/testing.spec.js
```diff
@@ -7,9 +8,27 @@ import { load } from '../../util/http';
...
+  describe('new submission button', () => {
+    it('shows the button to an administrator', async () => {
+      mockLogin({ role: 'admin' });
+      testData.extendedForms.createPast(1, { draft: true });
+      const path = '/projects/1/forms/f/draft/testing';
+      const component = await load(path, { component: true }, {});
+      component.first(EnketoFill).should.be.visible();
+    });
+
+    it('does not render the button for a project viewer', async () => {
```
lognaturel, line 20: I'm kind of surprised that a project viewer can even see drafts? I guess they could be verifying test submissions that come in, is that the idea?

src/components/enketo/preview.vue
matthew-white: Ah good point, I've made that change for both `EnketoPreview` and `EnketoFill`.

test/components/form-draft/testing.spec.js
matthew-white: Yes, I think so. In addition to the Submissions and Testing tabs, project viewers can also see the Versions tab. It's not clear that project viewers will use the Versions and Testing tabs, but they have all the verbs necessary to view that data, and it's ultimately less frontend logic to let them see them. In the past, we initially locked down archived projects, then ended up reversing that, and since then, @issa-tseng and I have tended toward keeping things accessible unless there's a reason not to. But it also isn't hard to hide the tabs if that'd be better.

lognaturel: Pragmatic approach. 👍 Thanks for the explanation. I’m on board.

PR #358, opened 2020-07-30
Add public links

matthew-white:


src/components/public-link/create.vue
```diff
@@ -0,0 +1,109 @@
...
+
+<i18n lang="json5">
+{
+  "en": {
+    // This is the title at the top of a pop-up.
+    "title": "Create Public Access Link",
+    "field": {
+      "multiple": "Multiple responses"
+    },
+    "multipleHelp": "Allow this Form to be submitted more than once from the same device."
```
lognaturel, line 106: Maybe this should make it a little clearer that this has to do with the link rather than with the Form? "Allow this link to be used to submit the Form more than once from the same device."? Or is that just too pedantic since the context makes it clear?

src/components/public-link/row.vue
```diff
@@ -0,0 +1,106 @@
...
+    "action": {
+      "revoke": "Revoke"
+    },
+    // This text is shown for a Public Access Link that has been revoked. The
+    // text appears on its own and is not part of a longer sentence.
+    "revoked": "Revoked",
+    "unavailable": {
+      // This text is shown for a Public Access Link that is not available yet.
+      "text": "Not available yet",
+      "title": "Public Link is not available yet. It has not finished being processed. Please refresh later and try again."
```
lognaturel, line 102: Public Link -> Public Access Link for consistency?

src/presenters/form.js
```diff
@@ -22,7 +22,10 @@ const props = [
   'version',
   'keyId',
   'sha256',
+  // Form and form draft, not form version
   'enketoId',
+  // Form only, not form draft or form version
+  'enketoSingleId',
```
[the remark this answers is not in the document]
matthew-white: Frontend uses the same property as the Backend response, which is `enketoSingleId`, so I think changing this would require a Backend change. But I like the sound of `enketoOnceId`: it feels like a nice parallel with the `once` property of public links.

src/components/public-link/create.vue
matthew-white: I like having a shorter sentence there, but we could probably make the text clearer. What do you think of "Allow a User to submit more than once from the same device.", which doesn't mention the form? Or "Allow a User with this Link to submit more than once from the same device.", which mentions the link?

lognaturel: Of course! "Allow a User with this Link to submit more than once from the same device." sounds great to me.

src/components/project/submission-options.vue
```diff
@@ -0,0 +1,118 @@
...
+except according to the terms contained in the LICENSE file.
+-->
+<template>
+  <modal :state="state" hideable backdrop @hide="$emit('hide')">
+    <template #title>{{ $t('title') }}</template>
+    <template #body>
+      <div class="modal-introduction">
+        <p>{{ $t('introduction[0]') }}</p>
+        <ul>
+          <i18n tag="li" path="introduction[1].full">
```
lognaturel, line 19: Nit: the vertical spacing between list items looks very small to me.

PR #360, opened 2020-08-11
Add new locales

matthew-white:


bin/util/transifex.js
```diff
@@ -0,0 +1,574 @@
...
+    if (this.translated.isEmpty()) return undefined;
+    return key === 'full' && this.translated.length !== 1
+      ? Array.from(this.translated)
+      : this.translated.toVueI18n();
+  }
+}
+
+// `Translations` stores an object of source messages with the corresponding
+// translations. It provides methods to modify the translations.
+class Translations {
```
lognaturel, line 341: Comment: mirrors the Transifex json structure.

bin/util/transifex.js
```diff
@@ -0,0 +1,574 @@
...
+
+  get root() { return this.parent == null ? this : this.parent.root; }
+
+  // Returns a translation or another set of translations. In either case, the
+  // result is "live": a change to this node will be reflected in its children.
+  get(key) {
+    const sourceValue = this._source[key];
+    if (sourceValue == null) return undefined;
+    if (sourceValue instanceof PluralForms) return new Translation(this, key);
+    if (this._translated[key] == null) this._translated[key] = {};
```
lognaturel, line 362: I am terrified by a getter that potentially mutates!

bin/util/transifex.js
```diff
@@ -0,0 +1,574 @@
...
+};
+
+const verifyDestructure = (_, { source, translated }) => {
+  if (translated == null || !equals(Array.from(source), Array.from(translated)))
+    logThenThrow({ source, translated }, 'mismatch for source locale');
+};
+
+// If a component interpolation is only partially translated, we remove the
+// partial translation so that the resulting text is not a mix of locales. We
+// also remove an array that is missing a translation, because JSON does not
```
lognaturel, line 467: What does "an array that is missing a translation" concretely refer to? I'm not seeing where there are arrays involved.

bin/util/transifex.js
matthew-white: Ah interesting, I hadn't thought of that. I'm going to throw out a few different thoughts — let me know what you think.

- Here we're mutating internal data within the `Translations` object, but we're not mutating anything externally facing: any descendant `Translation` object will have a `translated` property that is `null`. `this._translated` was passed to the constructor, but `Translations` assumes control of `source` and `translated` once it's passed them.
- Setting `this._translated` here means that `Translations` can always assume that `this._translated` is an object, avoiding a `null` check in a few places.
- Setting `this._translated` is one way to ensure that parent and child `Translations` stay in sync. If `this._source` is an object, and `this._translated` is `null`, then that means that `this._source` is an object that only contains linked locale messages (or other objects that only contain linked locale messages). Eventually we add those linked locale messages to `this._translated` using `set()`, but when doing so, we want to ensure that the parent `Translations` is aware of each addition. We could do so in `set()` instead, but that would require mutating `this.parent._translations`:

```diff
 get(key) {
   const sourceValue = this._source[key];
   if (sourceValue == null) return undefined;
   if (sourceValue instanceof PluralForms) return new Translation(this, key);
-  if (this._translated[key] == null) this._translated[key] = {};
-  return new Translations(this, key, sourceValue, this._translated[key]);
+  return new Translations(
+    this,
+    key,
+    sourceValue,
+    this._translated != null ? this._translated[key] : null
+  );
 }
 
 set(key, translated) {
   if (!(this._source[key] instanceof PluralForms))
     logThenThrow(key, 'invalid key');
   if (!(translated instanceof PluralForms))
     logThenThrow(translated, 'invalid translated');
+
+  const nodes = [];
+  for (let node = this; node != null && node._translated == null; node = node.parent) {
+    nodes.push(node);
+  while (nodes.length !== 0) {
+    const node = nodes.pop();
+    node._translated = {};
+    if (node.parent != null)
+      node.parent._translated[node._key] = node._translated;
+  }
+
   this._translated[key] = translated;
   return this;
 }
 ```

Happy to add a comment about all this or to update `set()`. Or let me know if there's a better way to do it!

File: bin/util/transifex.js (hunk 3 above)
Comment (matthew-white) [reply bin/util/transifex.js:467]: Let's say the source JSON looks like this:

```
{
  "introduction": [
    "one",
    "two",
    "three"
  ]
}
```

and "one" and "three" are translated, but "two" is not. In the translation JSON, it's not clear what the second element of the array should be. (It has to be something, because JSON does not support sparse arrays.)

```
{
  "introduction": [
    "ūnus",
    ?,
    "trēs"
  ]
}
```

It doesn't work to use `""` (https://github.com/kazupon/vue-i18n/issues/563). `null` may work, but Vue I18n logs a warning to the console, and I think this workaround might not be available in the next major version (which has some related changes). For now, we simply discard the entire array.

We're able to work around this if the empty array element is an object or another array. (If the second element were a missing component interpolation, we could use `{}`.) However, if an array directly contains an untranslated string, we discard the entire array.

lognaturel: I see. What I find confusing is that this is part of `destructure` and the source Transifex file for that doesn't have any arrays. I see now that you consider the indexed members the arrays on the structured JSON side. A tiny bit of context around that would really help someone who comes in without deep experience with the different JSON formats involved. Something like "In vue-i18n, strings can be grouped in an array. Those are displayed separately in Transifex and may not all be translated. If any string in an array is missing in a translation, we discard the whole array."

bin/util/transifex.js
lognaturel: If this were code we were planning to build on, I'd like to spend more time thinking through alternatives. Given that it's probably (hopefully?) not going to need much touching in the future, it's probably ok. All of this work is pretty mutation-heavy.

bin/util/transifex.js
lognaturel: Ooh, right, it mirrors the source Vue I18n json and reads the Transifex translations into it. This is a very helpful comment.

PR #361, opened 2020-08-12
Add developer comments to strings that need context

lognaturel:
From comments and observations in Transifex.

### DIFF
--- a/src/components/project/list.vue
+++ b/src/components/project/list.vue
@@ -244,6 +244,7 @@ export default {
   "en": {
     "heading": [
       "Welcome to Central.",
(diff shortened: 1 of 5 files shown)


src/components/project/list.vue
```diff
@@ -244,6 +244,7 @@ export default {
   "en": {
     "heading": [
       "Welcome to Central.",
+      // This is a tagline displayed next to "Welcome to Central". The text adds visual balance to the page and does not need to be translated literally to your language. The text should be action-oriented, collaborative, and a little bit fun.
```
matthew-white, line 247: No need if you prefer a single line, but just FYI, `restructure.js` does lets you split a comment over multiple lines.

lognaturel: Good to know, thanks!

PR #375, opened 2020-09-08
Rename en.json to en.json5

matthew-white:


lognaturel (approved):
Weird things happened.

lognaturel (commented):
I see you're not renaming files for locales other than `en`. I think the correct files will be generated on the next language import. But won't the stale `.json` files also remain? Seems easier to rename them here.

matthew-white:
One of the main reasons I made this change was to improve the display of `en.json5` in GitHub: after this PR, comments will no longer be highlighted as syntax errors. But I actually like that `en.json5` ends up with a different file extension from the other locale files, and my thinking is that the other files will continue to have a `.json` extension. `en.json5` is the only locale file that is actually JSON5: it's the source locale, so it's the only locale file with comments. Because of that, it's also loaded differently: `en.json5` is loaded in a different place from the other locale files and using a different loader (the `json5-loader`). I think that with this change, the locale files will also nicely parallel the i18n custom blocks: in each component, the first i18n block is the source locale and JSON5, and the second i18n block is the translated locales and JSON.

matthew-white:
I just merged #376, and it looks like there are no merge conflicts. 🎊

bin/util/transifex.js
```diff
@@ -346,7 +346,7 @@ const readSourceMessages = (localesDir, filenamesByComponent) => {
 
   // Read the root messages.
   const messages = parse(
-    fs.readFileSync(`${localesDir}/${sourceLocale}.json`).toString(),
+    fs.readFileSync(`${localesDir}/${sourceLocale}.json5`).toString(),
```
lognaturel, line 349: This is what made me think all files were expected to have a `json5` extension.

lognaturel:
Thanks for the extra information. Always helpful to get a small PR description when we do async review! It can also helpful for future archaeology to capture why things are being done.

All that makes sense, but I'm still uncertain about one change I've marked inline.

lognaturel (changes requested):
We're still getting me up to speed and I can't seem to dismiss my review.

matthew-white:
> Thanks for the extra information. Always helpful to get a small PR description when we do async review!

I tried to add some context in the commit message, but I can see now that that was rather succinct. 😅

> All that makes sense, but I'm still uncertain about one change I've marked inline.

Is that the comment above about the .json5 extension in `readSourceMessages()`? Is that resolved now? You’re exactly right that it’s .json5 there because it’s reading the source locale: `writeTranslations()` outputs .json files for the translated locales.

bin/util/transifex.js
lognaturel: Sourcelocale is a constant set to ‘en’

lognaturel (approved):

PR #376, opened 2020-09-14
Add Indonesian

matthew-white:


lognaturel (approved):
I double checked the developer comment changes. For the translations themselves, I scrolled through really quickly looking for any strange formatting or anything else that would jump out and found none.

PR #379, opened 2020-09-29
4.21 View Form XML in browser

### LINKED ISSUE #370: Standard form buttons wrap in French
The "standard form buttons" include the Preview button and the Download button. In English, they are shown on a single line. However, in French (and possibly other locales), they can wrap onto a second line, because the text of the buttons is longer.

On the form overview page:

overview_fr

On the draft status page:

status_fr

Given the length of the button text, it might be hard to keep the buttons on a single line along with the version string. I think it might work best to move the buttons below the version string. For example:

On the form overview page:

Screen Shot 2020-08-20 at 5 59 29 AM

On the draft status page:

Screen Shot 2020-08-20 at 5 57 40 AM

On the draft status page, we could also consider moving the "Upload new definition" button down to the Actions section, joining the "Publish Draft" and "Abandon Draft" buttons.

matthew-white:
Closes getodk/central#102. Also closes #370.


test/components/form-version/view-xml.spec.js
```diff
@@ -0,0 +1,12 @@
+import FormVersionViewXml from '../../../src/components/form-version/view-xml.vue';
+import { mount } from '../../util/lifecycle';
+
+describe('FormVersionViewXml', () => {
+  it('formats the XML', () => {
+    const modal = mount(FormVersionViewXml, {
+      propsData: { state: true },
+      requestData: { formVersionXml: '<x><y/></x>' }
+    });
+    modal.first('code').text().should.equal('<x>\r\n    <y/>\r\n</x>');
```
lognaturel, line 10: `\r\n`? That's what the library does?

lognaturel (commented):
Looks good! I'm not convinced that both "Download XML" and "Download as XForm (.xml)" are needed. It seems confusing to me and I'd rather just have "Download as XForm (.xml)".

test/components/form-version/view-xml.spec.js
matthew-white: Yes! It was also surprising to me. It's possible to specify something else, but `\r\n` is the default and seems to render OK in the browser.

src/components/form-version/def-dropdown.vue
```diff
@@ -0,0 +1,101 @@
...
+      <li>
+        <a href="#" @click.prevent="viewXml">{{ $t('action.viewXml') }}</a>
+      </li>
+      <li>
+        <a :href="defPath('xml')" :download="`${version.xmlFormId}.xml`">
+          <template v-if="version.excelContentType == null">
+            {{ $t('action.downloadXml') }}
+          </template>
+          <template v-else>
+            {{ $t('action.downloadXForm') }} (.xml)
```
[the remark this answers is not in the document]
matthew-white: Ah good point! I've left a comment so that we can come back to it for when we first add an RTL language.

lognaturel (approved):

PR #381, opened 2020-10-02
set the id

banushanv:


matthew-white:
Hi @banushanv! I don't think we need to set the `id` attribute of these elements. We specify the `key` attribute for `v-for`, and use the `form-row` class to style. I'm going to close this PR for now, but let me know if I've missed anything!

PR #383, opened 2020-11-06
Do not strip 'uuid:' prefix from submission instanceID

matthew-white:


lognaturel (approved):

lognaturel:
For posterity, this came out of the discussion at https://forum.getodk.org/t/incorrect-documentation-api-mock-for-instanceid/31027/6. The intent is for instanceID to simply be passed through and that's the case except for frontend display. This can be misleading/confusing when comparing with an export.

PR #384, opened 2020-11-12
Add submission download options

### LINKED ISSUE #133: Consistent return from methods of `trigger`
In testing, all methods of `trigger` return a promise. However, the promise resolves inconsistently: if a selector was specified to the method, the promise resolves with the wrapper; otherwise, it resolves with `undefined`. We should have these promises resolve consistently, always resolving with the wrapper. As we do so, we should check whether the change breaks any existing promise chain.

matthew-white:
This PR contains the frontend work for getodk/central#150. It also closes #133.

Most of the work is done in a new component, `SubmissionDownloadDropdown`. We use event bubbling in that component. We rarely use event bubbling, and I needed to make a change to `trigger` in order to test the event bubbling. I also made a couple of other small changes to `trigger`.

### DIFF
--- a/src/components/submission/download-dropdown.vue
+++ b/src/components/submission/download-dropdown.vue
@@ -20,12 +20,12 @@ except according to the terms contained in the LICENSE file.
     </button>
     <ul class="dropdown-menu"
       aria-labelledby="submission-download-dropdown-toggle" @click="download">
-      <li>
+      <li :class="{ disabled: disablesDownloadWithMedia }">
         <a :href="`${baseUrl}/submissions.csv.zip`" :target="target">
           {{ $t('action.download.withMedia') }}
         </a>
       </li>
-      <li :class="{ disabled: disablesDownloadWithoutMedia }">
+      <li>
         <a :href="`${baseUrl}/submissions.csv.zip?media=false`"
           :target="target">
           {{ $t('action.download.withoutMedia') }}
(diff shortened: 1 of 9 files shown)


src/components/submission/download-dropdown.vue
```diff
@@ -0,0 +1,112 @@
...
+    formVersion: {
+      type: Form,
+      required: true
+    }
+  },
+  computed: {
+    ...requestData(['fields']),
+    ...mapGetters(['managedKey']),
+    buttonText() {
+      if (this.formVersion.submissions <= 1)
```
lognaturel, line 65: It kind of feels like this either should be a special case (just display 0 or 1 record) or the button should be disabled for 0.

lognaturel (commented):
I get a 403.1 when trying to request a download with https://github.com/getodk/central-backend/pull/304. Any ideas what that might be about?

I find the user experience for a form without media or repeats confusing:
- All data and media files (.zip) active. This makes me review the form definition to see whether there is media I'm forgetting about.
- All data without media files (.zip) inactive. From a user perspective, I have no idea why this is inactive. If I don't have media files, why isn't the first option inactive?
- Primary data table (.csv) active. This makes sense but since I only have one data table it feels a little redundant.

Could the last option be `All data without media files (.csv)` if there are no repeats and `Primary data table (.csv)` if there are repeats?

lognaturel (approved):
Yes, this all feels a lot more expected to me, thank you!

lognaturel:
> Cookie auth / HTTPS is required for download — could that be it?

Ugh, how many more times/different ways am I going to ask this. It's unintuitive to me what does and doesn't work locally.

PR #385, opened 2020-11-13
Add forward to default Enketo config

lognaturel:
As we add more functionality that depends on Enketo, it would be good to make it easier to use in development. This addition should have no effect for folks who don't have Enketo running.

It likely won't be reasonable to get working with submission (non-draft, non public) links because those require cookie auth. Maybe that makes it not worth doing, I'm not sure.


lognaturel:
This doesn't really help since most of the interesting functionality is cookie-based.

PR #386, opened 2020-11-24
Synchronize all clients to server Form list

### LINKED ISSUE #61: Hide field key QR code popover upon window resize
Otherwise, I think the popover would end up positioned incorrectly? I think this is mentioned in the Bootstrap documentation.

### LINKED ISSUE #64: Removing jQuery listeners and data
I'm not sure we're removing all jQuery listeners and data as part of the Vue component destroy process. (Or at other times when we remove jQuery-managed elements, for example, removing elements in a data table before fetching fresh data.) One case: I think we might not be removing listeners and data from the popovers on the field key list page.

### LINKED ISSUE #118: Show modal within updated hook
Right now we show the modal (by calling the Bootstrap `modal()` method) within the `state` watcher, but I think that could have the effect of displaying the modal before it is fully re-rendered. Probably a better place to call `modal()` is the `updated` hook.

matthew-white:
This PR adds the ability to switch between a Managed and a Legacy Code.

The new app user popover is more complex than what Bootstrap's popover plugin naturally supports, and there are limitations to the approach I took. However, I tried to document those limitations.

The new `Popover` component implements the basic popover functionality. `Popover` involves a little asynchronous work (though less so now than in an earlier version of the component). I was looking at the asynchronous work in the `callWait` mixin for inspiration and noticed a small issue there: the `catch()` callback does not look at the `canceled` flag, which means that a previously canceled call might itself cancel the current call. I committed a fix for that issue along with some small clarifying changes.

`Popover` uses `$nextTick()`, which we generally try to avoid, and you'll also see `$nextTick()` in tests. The `.popover` element is appended to the body where avoriaz can't access it, so some tests access the document directly and don't use avoriaz wrapper methods.

The new `FieldKeyQrPanel` component uses nested component interpolation. I have updated `restructure.js` to support that case, autogenerating Transifex developer comments. It'd be helpful to get feedback on whether the resulting comments are clear enough. (It's a little hard to describe the structure of a nested component interpolation in comments.)

For more notes, see the commit message.

This PR closes #61, closes #64, and closes #118.


lognaturel:
Looks really slick. I haven't looked at the implementation but the behavior and text look good to me. One minor suggestion is to say "Managed QR Code" and "Legacy QR Code".

I'll hopefully come back to the Transifex changes soon but wanted to get initial impressions out.

lognaturel:
Regarding the Transifex-related change. I probably would have manually crafted a comment for that specific case but the general approach seems fine now that you've written it!

issa-tseng (approved):

matthew-white (commented):

PR #398, opened 2020-12-16
Add help dropdown to navbar

### LINKED ISSUE #350: Add navbar links to get help
@lognaturel, @issa-tseng, what do you think about adding link(s) to the navbar dropdown that link to the docs and/or the forum? One possibility is that these links could go between "Edit profile" and "Log out".

Also CC @florianm, who made a related [comment](https://forum.getodk.org/t/odk-central-is-ready-for-translating/28491/7) on the forum.

matthew-white:
Closes #350.

### DIFF
--- a/src/components/navbar/help-dropdown.vue
+++ b/src/components/navbar/help-dropdown.vue
@@ -1,0 +1,53 @@
+<!--
+Copyright 2020 ODK Central Developers
+See the NOTICE file at the top-level directory of this distribution and at
+https://github.com/getodk/central-frontend/blob/master/NOTICE.
+
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+<template>
+  <li id="navbar-help-dropdown" class="dropdown">
+    <a class="dropdown-toggle" href="#" data-toggle="dropdown" role="button"
+      aria-haspopup="true" aria-expanded="false">
+      <span class="icon-question-circle"></span><span class="caret"></span>
+    </a>
+    <ul class="dropdown-menu">
+      <li>
+        <doc-link to="central-intro/">{{ $t('docs') }}</doc-link>
+      </li>
+      <li>
+        <a href="https://forum.getodk.org/" target="_blank">{{ $t('forum') }}</a>
+      </li>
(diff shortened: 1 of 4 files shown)


src/components/navbar/help-dropdown.vue
```diff
@@ -0,0 +1,53 @@
...
+};
+</script>
+
+<i18n lang="json5">
+{
+  "en": {
+    // This is the text of a link that links to the documentation.
+    "docs": "Docs",
+    // This is the text of a link.
+    "forum": "Community forum",
```
lognaturel, line 47: I think this should match the website and just say "Forum." That also harmonizes better with the length of "Docs".

src/components/navbar/help-dropdown.vue
```diff
@@ -0,0 +1,53 @@
...
+    </a>
+    <ul class="dropdown-menu">
+      <li>
+        <doc-link to="central-intro/">{{ $t('docs') }}</doc-link>
+      </li>
+      <li>
+        <a href="https://forum.getodk.org/" target="_blank">{{ $t('forum') }}</a>
+      </li>
+      <li>
+        <a href="/version.txt" target="_blank">{{ $t('version') }}</a>
```
lognaturel, line 26: This will only work if using our Docker setup, right? It feels like Frontend shouldn't have that requirement. I wonder whether there maybe should be a pre-emptive query and the link is only shown if it's available?

lognaturel (changes requested):

src/components/navbar/help-dropdown.vue
matthew-white: That's possible, but it's definitely more work. Do you have a use case in mind that doesn't use our Docker setup?

lognaturel: Right. I guess we can make a note in the user documentation. It's true that we don't know of folks deploying frontend without the Docker setup.

lognaturel (approved):

lognaturel:
It feels like the dropdown should expand towards the right rather than the left but I don't feel strongly about it.

matthew-white:
> It feels like the dropdown should expand towards the right rather than the left but I don't feel strongly about it.

My instinct is that if the help dropdown expands towards the right, then the locale dropdown should as well. However, the locale dropdown is too wide to do so, so I think I'd lean toward keeping the help dropdown as-is. What do you think? If you want to take a look at what it would look like as it expands towards the right, try adding the `dropdown-menu-left` class to the `ul`!

matthew-white:
Oh just saw your comment update!

PR #404, opened 2020-12-20
Rename ?media query parameter to ?attachments

matthew-white:


lognaturel:
I see from https://github.com/getodk/central-backend/pull/304/files#diff-b48115a435dee55dae6048783669f3de6e30e561e4ed20195d0be48fad879c50R240 that the actual query param name is indeed `attachments`. And I see that it's documented as `media?` at https://github.com/getodk/central-backend/pull/304/files#diff-9eddf4dc51bf6a0125ca7fb094468ad284112270385c41cebce4f8f0a29620abR43. I think `attachments` makes sense given we use that word in other places. But I Can't remember, was there discussion around which was intended?

matthew-white:
@issa-tseng and I discussed some [here](https://docs.google.com/document/d/17gt5oJkCpLwW9p44jiPrfu-nZFQhxcSnGBWjpJP66YI/edit?disco=AAAAK_PokAY) and landed on `?attachments` as the preferred name. We should also update the API docs!

lognaturel (approved):

PR #407, opened 2020-12-28
Add favicon

### LINKED ISSUE #332: Add favicon
How about we use the ODK logo as our icon? Right now Frontend doesn't use an icon.

matthew-white:
Closes #332.


lognaturel (approved):
Forum, docs and Central will now all have the same favicon. I think that's fine for now but wanted to make sure everyone realized that.

PR #415, opened 2021-02-09
Improve login/logout

### LINKED ISSUE #122: Coordinate logout across tabs
The work so far to coordinate login across multiple tabs (#116, #120) doesn’t do as much to coordinate logout. For example, if you log out in one tab, you’re redirected to login in that tab, but you can continue navigating throughout the app in the other tabs (though outside of cached data, you should mostly just see alerts about expired credentials).

Perhaps we could implement something like, if the user tries to navigate to a different page and their session has expired, they are automatically redirected to login.

Alternatively, if we're worried about the user losing their work, we could implement it so that the navigation from the current page is simply aborted, and an alert is shown indicating that the user's session has expired. That may be quite a bit trickier though if it requires the router to access `App`'s `alert` data property.

### LINKED ISSUE #362: Handle case where user is logged into one tab and attempts to log in to another
Currently a user sees `Incorrect email address and/or password.` even if their credentials are correct. This is because the request includes the session cookie but not the CSRF token.

@matthew-white has suggested sending a second request to `/sessions/restore` and in case of a different user attempting to log in, display a message like `“You are already logged in as X. Please refresh the page to continue.”`

matthew-white:
Closes #122. Closes #362.

### DIFF
--- a/src/components/account/login.vue
+++ b/src/components/account/login.vue
@@ -43,8 +43,11 @@ except according to the terms contained in the LICENSE file.
 <script>
 import FormGroup from '../form-group.vue';
 import Spinner from '../spinner.vue';
+
 import request from '../../mixins/request';
 import { enketoBasePath } from '../../util/util';
+import { localStore } from '../../util/storage';
+import { logIn } from '../../util/session';
 
 export default {
   name: 'AccountLogin',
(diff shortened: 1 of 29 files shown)


issa-tseng (approved):
testNoRequest is confusing af that's all

PR #422, opened 2021-03-18
Indicate state in Project and Form tables with icons

ktuite:
* See project encryption status in project list
* See form submission state (open/closed/closing) in form list
* See form draft state (if it's unpublished and only a draft exists) in form list
* Each new icon has a tooltip

Screen Shot 2021-03-18 at 11 19 17 AM


ktuite:
The failing test has to do with what is shown in the "submissions" column for a form with no published version, only a draft. The release docs say the column should be empty, and the test expects a link to draft/testing. Which behavior do we want?

src/components/form/row.vue
```diff
@@ -13,7 +13,13 @@ except according to the terms contained in the LICENSE file.
   <tr class="form-row">
     <td class="name">
       <link-if-can :to="primaryFormPath(form)">
-        <span>{{ form.nameOrId() }}</span><span class="icon-angle-right"></span>
+        <span v-if="form.state == 'closed'" class="icon-times-circle text-danger form-icon"
```
matthew-white, line 16: We tend not to use Boostrap helper classes like `text-danger`. Instead, we specify those styles in the CSS, often selecting on some other class. (Here, that other class would probably be `icon-times-circle`.) We often use Sass variables, and if the styles are complex, we use a Sass mixin. That can lead to more complex CSS, but it's nice that it simplifies the template and arguably makes things more semantic.

src/components/form/row.vue
```diff
@@ -13,7 +13,13 @@ except according to the terms contained in the LICENSE file.
...
+          :title="$t('formClosedTip')"></span>
+        <span v-if="form.state == 'closing'" class="icon-clock-o text-warning form-icon"
+          :title="$t('formClosingTip')"></span>
+        <span v-if="form.publishedAt == null" class="icon-edit text-muted form-icon"
+          :title="$t('formUnpublishedTip')"></span>
```
matthew-white, line 21: I think we can use `v-else-if` for these cases.

src/components/form/row.vue
```diff
@@ -13,7 +13,13 @@ except according to the terms contained in the LICENSE file.
...
+        <span v-bind:class="[form.state == 'closed' ? 'form-name-closed' : '']">{{ form.nameOrId() }}</span><span class="icon-angle-right"></span>
```
matthew-white, line 22: `:class` is a nice shorthand for `v-bind:class`.

src/components/form/row.vue
```diff
@@ -146,7 +162,10 @@ export default {
     "lastSubmission": "(last {dateTime})",
     "action": {
       "fill": "Fill Form"
-    }
+    },
+    "formClosedTip": "This Form is closed. It is no longer accepting submissions.",
+    "formClosingTip": "This Form is closing. It is not visible on new devices, but it is still accepting submissions.",
```
matthew-white, line 167: We often capitalize Central-specific terms, like "Submissions", "Closed", and "Closing". I'm also wondering whether we should try to match the language in `FormOverviewRightNow` (that might be an question for @issa-tseng):

https://github.com/getodk/central-frontend/blob/fb714d1a7643815a78af1fb8d662bceec668710f/src/components/form/overview/right-now.vue#L118-L120

src/components/form/row.vue
```diff
@@ -112,6 +118,16 @@ export default {
         vertical-align: 2px;
       }
     }
+
+    .form-icon {
```
matthew-white, line 122: This might need some `margin-right` to match the mock-up.

matthew-white:
> The failing test has to do with what is shown in the "submissions" column for a form with no published version, only a draft. The release docs say the column should be empty, and the test expects a link to draft/testing. Which behavior do we want?

I think we can make the change that the release criteria suggests and then update the failing test.

ktuite:
The icon for a closed form on the form list is still not *ideal*.

I see it should be the ban/not allowed 🚫 icon, but I'm going to wait for Matt's other icon-related PRs to get merged in first. Whatever it is changed to, it should be more consistent with form state icon in the "right now" section of the form overview. I can change that icon to 🚫, too?

ktuite:
Updated appearances:

Screen Shot 2021-03-18 at 3 02 01 PM
Screen Shot 2021-03-18 at 3 01 51 PM

src/components/project/row.vue
```diff
@@ -14,6 +14,8 @@ except according to the terms contained in the LICENSE file.
     <td>
       <div class="name">
         <router-link :to="projectPath(project.id)">
+          <span v-if="project.keyId" class="icon-lock text-muted project-icon"
```
matthew-white, line 17: I think this Bootstrap helper class can also be replaced with CSS.

matthew-white (commented):
This is looking good! The updated appearances look great. 🚀

I think the main thing left to do is tests of the new functionality. In general, we tend not to test the styles, but rather test that the correct elements are rendered and that those have the correct classes and attributes. So here for example, you can test that if a form is closing, a `.form-icon` element is rendered, that element also has the classes `form-icon-closing` and `icon-clock-o`, and it has the correct `title` attribute; but you don't need to test that the icon is yellow.

> I can change that icon to 🚫, too?

That sounds great!

ktuite:
Ban icon now in place!!

Forms table:
Screen Shot 2021-03-19 at 4 12 32 PM

Form overview "right now" page also shows the same icon: 
Screen Shot 2021-03-19 at 4 12 46 PM

src/assets/css/icomoon.css
```diff
@@ -1,15 +1,15 @@
...
+    url('/fonts/icomoon.ttf?kps16g') format('truetype'),
+    url('/fonts/icomoon.woff?kps16g') format('woff'),
+    url('/fonts/icomoon.svg?kps16g#icomoon') format('svg');
   font-weight: normal;
   font-style: normal;
   font-display: block;
 }
 
-[class^="icon-"] {
+[class^="icon-"], [class*=" icon-"] {
```
matthew-white, line 12: We actually just removed this second selector (in #416), in part for performance reasons and in part because it simplifies other icon-related CSS in `app.scss`.

test/components/form/row.spec.js
```diff
@@ -21,6 +21,30 @@ describe('FormRow', () => {
...
       app.first('.form-row .name').text().trim().should.equal('f');
     });
+
+    it('shows ban icon in form name link for closed form', async () => {
+      testData.extendedForms.createPast(1, { xmlFormId: 'f', state: 'closed' });
+      const app = await load('/projects/1');
+      const item = app.first('.form-row .name a');
+      item.find('.icon-ban').length.should.equal(1);
+      item.first('.form-icon').getAttribute('title').should.equal('This Form is Closed. It is not downloadable and does not accept Submissions.');
+    });
```
matthew-white, line 31: Maybe also check that there is a `.form-name-closed` element?

test/components/form/row.spec.js
```diff
@@ -21,6 +21,30 @@ describe('FormRow', () => {
...
+    it('shows clock icon in form name link for closing form', async () => {
+      testData.extendedForms.createPast(1, { xmlFormId: 'f', state: 'closing' });
+      const app = await load('/projects/1');
+      const item = app.first('.form-row .name a');
+      item.find('.icon-clock-o').length.should.equal(1);
+      item.first('.form-icon').getAttribute('title').should.equal('This Form is Closing. It is not downloadable but still accepts Submissions.');
+    });
+
+    it('shows edit/pencil icon for unpublished draft form', async () => {
+      testData.extendedForms.createNew(1, { draft: true, publishedAt: null });
```
matthew-white, line 42: You can actually use `createPast()` here too. `createPast()` is for setting up data that exists before the page is loaded, whereas `createNew()` is for data created after the page is loaded. `createNew()` is relatively uncommon, but you'll see it in tests of components that create a new resource.

matthew-white (commented):
The new tests are looking great! Also, super exciting to have 🚫 in there. 😄

matthew-white (approved):
Looks great! 🚀

PR #425, opened 2021-03-23
Update documentation about testing

matthew-white:
Updating documentation after conversations with @ktuite. Some of the language is the same as the Backend contribution guide. @ktuite, would be great to get your review as well!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -320,7 +320,7 @@ Our tests use a number of external packages:
 
 We extend Should.js assertions in [`/test/assertions.js`](/test/assertions.js).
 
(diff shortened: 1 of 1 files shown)


lognaturel (approved):

ktuite:
Looks great!

One thing I'd suggest adding (not about testing) about translations is how to run `restructure.js` *with* the full command `node bin/transifex/restructure.js`. Also, that there is a git hook that runs it to make sure.

Also, I found the avoriaz documentation helpful for understanding the kinds of test assertions: https://eddyerburgh.gitbooks.io/avoriaz/content/ 
In our meeting, you said something about how avoriaz uses mount() to create the elements, and then those elements are traversable and inspectable through the avoriaz commands linked above.

matthew-white:
Thanks, @ktuite! I added notes about `restructure.js`. I also added a short description of avoriaz and a link to their docs. I think I'm going to go ahead and merge, but let me know if there are ways to improve upon these changes, or if you have other suggestions for the documentation!

ktuite:
Looks good to me!

PR #435, opened 2021-04-16
Update audit log for new actions

matthew-white:
Related to getodk/central-backend#274.


issa-tseng (approved):

matthew-white (commented):
Notes from interactive review

PR #438, opened 2021-04-22
Reduce width of Time column of audit log

### LINKED ISSUE #432: Give time columns (e.g. in backups or logs) a fixed or maximum width
Time columns have a known maximum width. Currently they are responsive just like other columns in the tables meaning they become unnecessarily wide even on not-so-wide screens. It would help with readability if it were fixed or had a maximum width.

Context from comment:
> This came about because I was looking at other systems' logs and those felt a lot tighter to me. I think it's because they either have more columns (e.g. [Zoho](https://www.zoho.com/analytics/help/account-audit-logs.html)) or have a fixed-width table (e.g. [WPEngine](https://wpengine.com/blog/user-portal-activity-log/)). All that had this tighter feeling seemed to have some columns with fixed width (and/or match the content width). Not saying we should do any of those things but that's how I ended up thinking in this space.

matthew-white:
Closes #432. See the commit message for additional notes.

### DIFF
--- a/src/components/audit/table.vue
+++ b/src/components/audit/table.vue
@@ -13,11 +13,11 @@ except according to the terms contained in the LICENSE file.
   <table id="audit-table" class="table">
     <thead>
       <tr>
-        <th>{{ $t('header.time') }}</th>
+        <th class="logged-at">{{ $t('header.time') }}</th>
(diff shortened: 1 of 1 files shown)


lognaturel (approved):
Looks good to me!

PR #439, opened 2021-04-22
Add warning about Enketo-based links to "Set up Backups" modal

matthew-white:
Using our existing `modal-warnings` class. Screenshot:

Screen Shot 2021-04-22 at 5 45 34 PM


lognaturel (approved):
Ugh, not sure users will know what "externally link directly to Web Forms for making new Submissions" means but I can't think of a better way to phrase it.

PR #441, opened 2021-04-27
Add issue template config

matthew-white:
Using the same config as the `central` repository. See getodk/central#216.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,11 @@
+blank_issues_enabled: true
+contact_links: 
+  - name: "Report an issue"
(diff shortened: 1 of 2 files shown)


lognaturel (changes requested):
We should update this based on https://github.com/getodk/central/pull/218

lognaturel (changes requested):
No seriously, I want some changes.

lognaturel (changes requested):
I EXTRA request changes.

(See @getodk/testers repeating review states can be helpful. 😉)

matthew-white:
Just updated `config.yml` to reflect the latest. @lognaturel, this should be ready for another look!

lognaturel (approved):

PR #443, opened 2021-04-28
v1.2 style cleanup

issa-tseng:
see commits

### DIFF
--- a/src/components/form-version/def-dropdown.vue
+++ b/src/components/form-version/def-dropdown.vue
@@ -11,7 +11,7 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <div class="form-version-def-dropdown btn-group">
-    <button :id="toggleId" type="button" class="btn btn-primary dropdown-toggle"
+    <button :id="toggleId" type="button" class="btn btn-default dropdown-toggle"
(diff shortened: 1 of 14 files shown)


issa-tseng:
four broken tests, all because of button class changes. mind if i leave you to decide what to do with them? happy to take a crack otherwise.

src/components/form-version/def-dropdown.vue
```diff
@@ -11,7 +11,7 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <div class="form-version-def-dropdown btn-group">
-    <button :id="toggleId" type="button" class="btn btn-primary dropdown-toggle"
+    <button :id="toggleId" type="button" class="btn btn-default dropdown-toggle"
```
matthew-white, line 14: It looks like this fades into the background in the Your Current Draft section of the form overview:

Screen Shot 2021-04-28 at 11 27 35 AM

matthew-white (approved):
Awesome! Very excited to get these styles in there. ✨

I've updated the tests, and those are passing now. I noticed one issue, but I'm thinking that can be addressed in a follow-up PR: I think I'm going to go ahead and merge, since this PR touches some of the same components that @ktuite and I are working on.

PR #445, opened 2021-04-29
Add ability to filter submissions table by review state

matthew-white:
This PR adds the review state filter to `SubmissionList`. The presence of the additional filter means that the filters and buttons might not fit on a single row. I have set it up so that if possible, everything will remain on a single row. However, if there is not enough space, the download dropdown will go onto its own row above the filters. We may revisit this row as part of block 13.

### DIFF
--- a/src/components/submission/list.vue
+++ b/src/components/submission/list.vue
@@ -13,21 +13,23 @@ except according to the terms contained in the LICENSE file.
   <div id="submission-list">
     <loading :state="$store.getters.initiallyLoading(['fields'])"/>
     <div v-show="fields != null">
-      <form class="form-inline" @submit.prevent>
-        <submission-filters v-if="!draft" v-bind.sync="filters"/>
-        <submission-field-dropdown
-          v-if="fields != null && selectableFields.length > 11"
-          v-model="selectedFields"/>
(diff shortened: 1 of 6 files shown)


src/components/submission/list.vue
```diff
@@ -305,14 +311,25 @@ export default {
...
+}
+#submission-list-refresh-button {
+  margin-left: 10px;
+  margin-right: 5px;
 }
 #submission-download-dropdown {
-  float: right;
-  top: 3px;
+  margin-bottom: 10px;
+  margin-left: auto;
```
matthew-white, line 329: This is what floats the download dropdown to the right: see https://css-tricks.com/how-auto-margins-work-in-flexbox/

src/components/submission/list.vue
```diff
@@ -305,14 +311,25 @@ export default {
...
+
+  // Needed to work with align-items above.
+  .btn { float: none; }
```
matthew-white, line 332: Bootstrap floats `.btn-group > .btn` to the left, but I think it's not needed for this dropdown (or maybe for any of our uses of `.btn-group`?).

ktuite (approved):

PR #448, opened 2021-05-02
Check for session expiration on an interval

matthew-white:
This PR fixes an issue related to automatic logout before session expiration. Previously, we used `setTimeout()` to schedule logout for some point in the future. However, it seems that `setTimeout()` does not count time while the computer is asleep (!). Instead, I've switched to using `setInterval()` to check for session expiration once a second. I've also changed it so that a logout request is not sent if the session has already expired. Right now, if your computer was asleep, and you are later (belatedly) automatically logged out, an error message is shown, because the logout request fails.

Things have been shuffled around, but the core logic is mostly the same. I'll also add some comments to the PR now about specific lines.

### DIFF
--- a/src/store/modules/request.js
+++ b/src/store/modules/request.js
@@ -308,9 +308,10 @@ export default {
         const { cancelId } = requestsForKey;
 
         const baseConfig = { method: 'GET', url };
-        baseConfig.headers = extended
-          ? { ...headers, 'X-Extended-Metadata': 'true' }
-          : headers;
+        if (extended)
+          baseConfig.headers = { ...headers, 'X-Extended-Metadata': 'true' };
+        else if (headers != null)
+          baseConfig.headers = headers;
         const { session } = data;
         const token = session != null ? session.token : null;
         const axiosConfig = configForPossibleBackendRequest(baseConfig, token);
--- a/test/unit/session.spec.js
+++ b/test/unit/session.spec.js
@@ -323,7 +323,7 @@ describe('util/session', () => {
       const cleanup = useSessions(router, store);
       setData({
         session: testData.sessions.createNew({
-          expiresAt: '1970-01-01T00:05:01Z'
+          expiresAt: '1970-01-01T00:05:00Z'
         })
       });
       return mockHttp()
@@ -331,7 +331,7 @@ describe('util/session', () => {
         .respondWithData(() => testData.extendedUsers.first())
         .complete()
         .testNoRequest(() => {
-          clock.tick(240000);
+          clock.tick(239000);
         })
         .request(() => {
           clock.tick(1000);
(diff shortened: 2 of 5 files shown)


test/unit/session.spec.js
```diff
@@ -71,7 +71,6 @@ describe('util/session', () => {
     it('sends the correct request', () => {
       testData.extendedUsers.createPast(1);
       setData({ session: testData.sessions.createNew({ token: 'foo' }) });
-      store.commit('setSendInitialRequests', false);
```
matthew-white, line 74: I realized while writing the new tests that this line doesn't seem to be needed. Maybe it used to be, but I changed something in #420 such that it's no longer needed? I think I used to use it in any test that ran `logOut()` after calling `mockHttp()` (rather than `load()`). But there's no navigation in that case, so I don't think it's needed. (If there were navigation, the router would send the initial requests, and `mockHttp()` would throw an error about a request without a response.) Right now the tests use `load()` when there's navigation.

test/unit/session.spec.js
```diff
@@ -345,39 +339,21 @@ describe('util/session', () => {
...
       return load('/users', {}, false)
-        .restoreSession(true)
+        .respondWithData(() =>
+          testData.sessions.createNew({ expiresAt: '1970-01-01T00:05:00Z' }))
+        .respondWithData(() => testData.extendedUsers.first())
         .respondFor('/users')
         .complete()
         .request(() => {
-          clock.runAll();
+          clock.tick(240000);
```
matthew-white, line 356: `clock.runAll()` doesn't do what's needed anymore, since we now use `setInterval()` instead of `setTimeout()`. Instead, I specify `expiresAt` explicitly and use `clock.tick()`.

test/unit/session.spec.js
```diff
@@ -345,39 +339,21 @@ describe('util/session', () => {
         .respondWithSuccess()
         .afterResponse(() => {
           should.not.exist(store.state.request.data.session);
-        });
-    });
-
-    it('logs out immediately for a login within a minute of expiration', () => {
```
matthew-white, line 351: This is no longer a special case.

test/unit/session.spec.js
```diff
@@ -477,17 +528,21 @@ describe('util/session', () => {
...
-
-    it('logs out after sessionExpires changes', () =>
-      load('/users')
+    it('logs out after sessionExpires changes', () => {
+      testData.extendedUsers.createPast(1);
+      const cleanup = useSessions(router, store);
+      setData({ session: testData.sessions.createNew() });
+      return mockHttp()
+        .request(() => logIn(router, store, true))
+        .respondWithData(() => testData.extendedUsers.first())
```
matthew-white, line 545: Now that `App` isn't setting up the event listener, it's possible to test it using `mockHttp()` (similar to the tests related to session expiration).

src/store/modules/request.js
```diff
@@ -308,9 +308,10 @@ export default {
         const { cancelId } = requestsForKey;
 
         const baseConfig = { method: 'GET', url };
-        baseConfig.headers = extended
-          ? { ...headers, 'X-Extended-Metadata': 'true' }
-          : headers;
+        if (extended)
+          baseConfig.headers = { ...headers, 'X-Extended-Metadata': 'true' };
+        else if (headers != null)
+          baseConfig.headers = headers;
```
matthew-white, line 314: This makes it so that `baseConfig.headers` is not set at all if the `headers` option was not specified. I noticed that it was being set while working on the new tests, because (I believe) the only place it comes up right now is the request to restore the session.

issa-tseng (approved):

PR #449, opened 2021-05-04
Update dependencies for v1.2

matthew-white:


ktuite (approved):
Looked through this interactively with @matthew-white

Issue #333, opened 2020-07-02
Update page title after navigation


matthew-white:
Right now we display "ODK Central" as the page title. I think it'd be nice to update the title as the user navigates within Frontend. I think the main benefit of this would be improved history. We push a new entry into the history stack with each navigation, which allows the user to use the browser back button. However, without updating the title, it's hard to differentiate entries:

<img width="162" alt="Screen Shot 2020-07-01 at 4 33 34 PM" src="https://user-images.githubusercontent.com/5970131/86358840-2bf9b200-bc3e-11ea-8ca8-125be46a2af2.png">

One approach would be to define a route meta field for the title, then register a navigation guard that sets the document title to the value of the meta field after the navigation is confirmed.

However, this wouldn't be enough to show something like the project name in the title, because the response for the project will be received after the navigation is confirmed and the page is initially rendered. For pages like that, a watcher on the data might be the best place to set the title.

@lognaturel, @issa-tseng, do you think updating the page title would be useful? Do you think the title should include things like the project/form/user name?

lognaturel:
Huh, I would have expected to notice this while using Central but I hadn't. Thinking about it, I think it's because there are usually direct ways to get to what I need so I don't use the browser stack much for Central though I do on other sites. This suggests it's probably not very high priority.

I do think it'd be nice to address, though. I'm not exactly sure what should go first. Something static like "project summary", "test submissions", etc might be most helpful because users are likely to be in the context of one project at a time. So I'm thinking something like "Project Roles - My Cool Project - ODK Central", "Submissions - My Cool Form - My Cool Project - ODK Central". Maybe the static part is enough and that would be simpler.

matthew-white:
I feel like something like "Project Roles - My Cool Project - ODK Central" would be ideal, but if we expect that users will usually be in the context of a single project/form/etc. at a time, maybe just "Project Roles - ODK Central" would be enough to differentiate the most recent history entries. Showing a static title per page would definitely be much simpler in terms of code complexity: that way, the implementation wouldn't have to interact with the Vuex store or account for changing data.

Also, I'm thinking that this could be a good issue to label as "help wanted".

lognaturel:
It can also be staged. I think a contributor could very well use the localized page title to get something like "Project Roles - ODK Central" and that would be hugely helpful already. Then possibly in the future the project/form name could be added.

matthew-white:
[...]

PR #450, opened 2021-05-04
Update translations for v1.2

matthew-white:
Almost all of this is automated changes: I pulled the latest Transifex JSON, then ran `destructure.js` on it. However, I had to change one line in the destructure logic. I'll add a comment about that below.


bin/util/transifex.js
```diff
@@ -697,6 +697,7 @@ const deletePartialTranslation = ({ translated, parent }) => {
 };
 
 const validateTranslation = (locale) => ({ source, translated, path }) => {
+  if (translated == null) return;
```
matthew-white, line 700: `translated == null` means that it is a linked locale message that has not been translated. If `translated == null`, then there's nothing to validate, and we immediately return: the function could throw an error otherwise. I'm actually a bit surprised that this hasn't come up before, because there are a few places in this file that deal with untranslated linked locale messages. Before v1.2, the only place we used linked locale messages was the audit log actions. Those tend to be simple strings near the top of the list, so maybe actually they were all translated? 🤔 In any case, I've glanced over some of the logic related to untranslated linked locale messages, and I think things should work (other than this line). There are untranslated linked locale messages in German, but `de.json` looks OK and is correctly missing linked locale messages that are untranslated. `cs.json` has more linked locale messages than `de.json`, because all of them have been translated.

transifex/strings_es.json
```diff
@@ -1,4 +1,78 @@
...
+      "backups": {
+        "string": "Copias de seguridad",
+        "developer_comment": "This is the text of a navigation tab, which may also be shown as the page title (browser tab)."
+      },
+      "audits": {
+        "string": "Registro y auditoría de la actividad del servidor",
+        "developer_comment": "This is the text of a navigation tab, which may also be shown as the page title (browser tab)."
+      }
+    }
+  },
```
ktuite, line 75: I'm glad these title strings that got moved out of their components got fully translated for some languages! But not all languages :(

transifex/strings_id.json
```diff
@@ -1,4 +1,78 @@
 {
+  "title": {
+    "resetPassword": {
+      "string": "",
+      "developer_comment": "This is the text shown as the title of a page including in the browser tab."
+    },
+    "setPassword": {
+      "string": "",
+      "developer_comment": "This is the text shown as the title of a page including in the browser tab."
+    },
```
ktuite, line 10: Missing title strings that weren't re-translated :(

ktuite (commented):
I'm not totally sure what to be on the lookout for, but I did glance at all these files! I see new strings and their translations and I see some translated strings that changed a little bit.

I also see the effects of the page title translation strings that I moved around for issue #333. I noticed some of those strings got re-translated (Czech, Spanish) but not some of the other languages. Since these strings are going to show up (untranslated, presumably) in page titles, is there any sort of programatic hack to move the old strings for these phrases into the new slots where they are missing? Would that even be worth doing?

src/components/project/form-access/states.vue
```diff
@@ -134,7 +134,7 @@ export default {
       },
       {
         "full": "{closing} Formulare akzeptieren neue Übermittlungen, aber können {not} auf mobile Geräte heruntergeladen werden.",
-        "closing": "Wird geschlossen",
+        "closing": "Schließen",
```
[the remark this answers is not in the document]
matthew-white: Interesting. This seems to be a translation of the "Closing" form state. That's translated elsewhere as "Schließen", so maybe this was an attempt by a translator to make things consistent?

src/components/submission/decrypt.vue
```diff
@@ -225,7 +225,7 @@ export default {
   "es": {
     "title": "Descifrar y descargar",
     "introduction": [
-      "Para descargar la información, usted necesitará ingresar su contraseña. Su contraseña se usará para desencriptar la información descargada, al finalizar el proceso el servidor la olvidará."
+      "Para descargar la información, usted necesitará ingresar su contraseña. Su contraseña se usará para desencriptar la información descargada , al finalizar el proceso el servidor la olvidará."
```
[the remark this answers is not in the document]
matthew-white: Ah good catch, I'll fix that in Transifex and will pull again before release.

ktuite (approved):

matthew-white:
Thank you for taking a look, @ktuite! 🙏

> Since these strings are going to show up (untranslated, presumably) in page titles, is there any sort of programatic hack to move the old strings for these phrases into the new slots where they are missing? Would that even be worth doing?

Part of me would love to implement something like that. We could build upon the existing comment system by adding a way to indicate that a message used to be located elsewhere. It would then stay in that old location in Transifex. For example:

`en.json5`
```json5
{
  "title": {
    // @transifexKey component.PageNotFound.title
    "pageNotFound": "Page Not Found"
  }
}
```

On the flip side, our Transifex scripts are already a little complicated, and this would be an additional piece of complexity. And even though a lot of messages moved in your PR, it's actually pretty rare that that happens.

PR #453, opened 2021-05-05
Last minute block 12 updates

ktuite:
* Added "Detail: " as a prefix to submission detail page title (in response to QA feedback)
* Changed markdown/sanitization to be stricter and only allow certain tags

### DIFF
--- a/test/components/submission/feed-entry.spec.js
+++ b/test/components/submission/feed-entry.spec.js
@@ -181,8 +181,8 @@ describe('SubmissionFeedEntry', () => {
       mountComponent().first('.body').html().should.equal('<div class="body"></div>');
     });
 
-    it('removes style and class attributes', () => {
-      testData.extendedComments.createPast(1, { body: '<b style="color: red;" class="c">foo</b>' });
+    it('removes unwanted attributes', () => {
+      testData.extendedComments.createPast(1, { body: '<b style="color: red;" class="c" data-foo="bar">foo</b>' });
       mountComponent().first('.body').html().should.equal('<div class="body"><p><b>foo</b></p>\n</div>');
     });
   });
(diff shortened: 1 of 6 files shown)


matthew-white (commented):
This looks great! ✨

I wonder whether we'll want to forbid additional attributes in the future (there are so many interesting attributes out there). However, it'd be easy to do so if need be, and I think we're forbidding the most important ones. :+1:

test/components/submission/feed-entry.spec.js
```diff
@@ -176,9 +176,14 @@ describe('SubmissionFeedEntry', () => {
...
 
-    it('removes script tags and sanitizes html', () => {
-      testData.extendedComments.createPast(1, { body: '<script>foo</script>' });
+    it('removes script and svg tags and sanitizes html', () => {
+      testData.extendedComments.createPast(1, { body: '<script>foo</script><svg>bar</svg>' });
       mountComponent().first('.body').html().should.equal('<div class="body"></div>');
     });
+
+    it('removes style and class attributes', () => {
+      testData.extendedComments.createPast(1, { body: '<b style="color: red;" class="c">foo</b>' });
```
matthew-white, line 185: Ah somehow this comment wasn't attached to my previous review. 😅

Could we also test a data attribute here? (Those are actually the attributes I worry the most about.) Could be a fake data attribute like `data-foo="bar"`.

ktuite: Done!

matthew-white (approved):

matthew-white:
Not critical, but I just noticed that `parts()` should maybe return `[]` or `[null]` while `submission` is `null`:

https://user-images.githubusercontent.com/5970131/117239197-4f603100-adfc-11eb-89bd-6d197c935e8b.mov

PR #459, opened 2021-05-12
Use darker background color for .btn-default in #form-overview-draft

matthew-white:
Right now the background color of `.btn-default` is the same as the background color of `#form-overview-draft`. #ccc is a little darker than the current color, #ddd.


matthew-white:
This is what the new color looks like:

Your Current Draft

issa-tseng (approved):
beautiful

PR #460, opened 2021-05-12
Do not show alert for 401.2 response to logout request

matthew-white:
Two different tabs may send a logout request at the same time. Depending on the timing of the requests, it's possible for one of the requests to succeed and for that to cause the other request to fail (because the session has already been logged out). That case used to be quite common when we scheduled logout for a particular moment (before #448). It's maybe less common now, but it will happen if the user clears local storage (immediately logging out all tabs) and can probably also happen if there are many tabs open. To account for this case, Frontend would previously ignore a 403.1 response to a logout request: other error responses would be shown to the user, but a 403.1 would not.

However, that intended behavior is not what's being seen right now. Instead, logging out one tab can result in an alert being shown in another tab. I think the issue is that the correct code for Frontend to account for is 401.2, not 403.1. I thought that if you try to log out a session that has already been logged out, that will result in a 403.1. I do think that's how the session endpoint works, but actually, I'm pretty sure that a request that will fail here won't even reach the session endpoint: it'll fail in the `sessionHandler()` preprocessor. That's because not only is the request attempting to delete a session that has already been deleted, but that deleted session's token is also what's used to auth, which will result in a 401.2.

There is likely more that could be done here in order to send a duplicate logout request in fewer cases. However, for now I'm just patching the existing functionality.

This issue is a little intermittent, which is maybe why I missed it before. I think it's possible for two different DELETE requests of the same session to both be successful, if the transaction for the second request begins before the transaction for the first request ends.

### DIFF
--- a/src/util/session.js
+++ b/src/util/session.js
@@ -47,7 +47,7 @@ expiration date is also stored in local storage. This approach is designed to:
 
   - Support cookie auth
   - Ensure the user knows when they are logged out; prevent the user from seeing
-    403 messages after their session has been deleted
+    401 messages after their session has been deleted
   - Prevent one user from using another user's cookie
   - Ensure that the cookie is removed when the user logs out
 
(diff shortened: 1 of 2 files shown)


matthew-white:
Hmm, I just deployed this commit to the QA server, and I still sometimes see a 403.1, not a 401.2. 🤔 Is it possible that if one request deletes the session while a second request is in progress, the second request will pass `sessionHandler()`, but then fail in the session endpoint? Otherwise I'm not sure how this scenario could result in a 403.1 instead of a 401.2.

I'm going to push a commit that ignores both 401.2 and 403.1. Really the main errors that we care about and want to show to the user is a 500 or a network error in sending the request, so I think it's fine to ignore both 401.2 and 403.1.

matthew-white:
I've deployed this commit to the QA server, and things seem to be working now, so I think this is the right approach.

lognaturel (approved):
Took a brief look at this last night and it all sounds reasonable.

PR #461, opened 2021-05-12
Display geopoint, geotrace, and geoshape fields as WKT

matthew-white:


test/data/submissions.js
```diff
@@ -40,15 +39,8 @@ const odataValue = (field, instanceId) => {
...
-      // [longitude, latitude], not [latitude, longitude]
-      const coordinates = [
-        faker.random.number({ min: -180, max: 180, precision: 0.0000000001 }),
-        faker.random.number({ min: -85, max: 85, precision: 0.0000000001 })
-      ];
-      if (faker.random.boolean()) coordinates.push(faker.random.number());
-      return { type: 'Point', coordinates };
-    }
+    case 'geopoint':
+      return 'POINT (0 90)';
```
ktuite, line 43: Is there a test that checks the rendering of the value, beyond just checking that the `geopoint-field` class is there?

ktuite: Oh I see, it's just passing through exactly what it gets from the backend, so no need to check the rendering. And the geopoint is the only one that you want to make wider, because it could be wide enough to show the whole value, and the other geo types (GeoTrace, GeoShape) are too long to reasonably display.

ktuite (approved):

PR #463, opened 2021-05-13
Remove sessionExpires from local storage after session restore 404

matthew-white:


src/router.js
```diff
@@ -69,19 +69,11 @@ const initialLocale = () => {
...
-  // logout error.
-  if (sessionExpires != null && parseInt(sessionExpires, 10) <= Date.now())
-    return;
-  await store.dispatch('get', [{
-    key: 'session',
-    url: '/v1/sessions/restore',
-    alert: false
-  }]);
-  await logIn(router, store, false);
+const restoreSessionIfShould = async (to) => {
```
matthew-white, line 72: I don't love the name of this function, but I couldn't think of anything better. 🤔

src/util/session.js
```diff
@@ -72,6 +72,22 @@ import { forceReplace } from './router';
 import { localStore } from './storage';
 import { noop } from './util';
 
+const removeSessionFromStorage = () => {
```
matthew-white, line 75: This is copied from below. It's in its own function now that it is used in a couple of different places.

src/router.js
```diff
@@ -69,19 +69,11 @@ const initialLocale = () => {
 };
 
 // Implements the restoreSession meta field.
-const restoreSession = async (to) => {
```
matthew-white, line 72: Now that the logic here is slightly more complicated, I've moved most of this to `util/session.js`. I also think it's good to have most session-related logic consolidated in that file.

test/unit/session.spec.js
```diff
@@ -31,6 +98,16 @@ describe('util/session', () => {
           .respondWithData(() => testData.extendedUsers.first());
       });
 
+      it('saves currentUser', () => {
```
matthew-white, line 101: This test is a little unrelated, but I noticed it was missing while working on the tests above.

src/router.js
ktuite: `tryRestoreSession`? `restoreSessionIdAble`? I guess the "should" comes from it being in the router file, and the router saying "hey we are at the restoreSession route, we _should_ restore". So it's not really trying/checking if it should or should not do something based on some special condition, but it needs to do the whole restore session process because that's what the router says. `reallyRestoreSession` `doRestoreSession` `restoreSessionNow` `routerRestoreSession` `handleRestoreSession`... any of those inspire you?

ktuite (commented):
This is great! I feel like I've run into this issue on my dev server, too (when my session is about to expire, or I've turned on/off the backend in some weird combination, or I was restoring a database...) 

This explanation resonates:
> [This happened] because of a mismatch between what the frontend believes is true about your session and what is actually true on the server. The frontend believes that the session still exists and hasn't expired yet, but since the session wasn't part of the backup, the session will actually no longer exist after the backup is restored. This can be resolved by clearing cookies and site data for Central. It should also resolve itself in 24 hours or less, because the session will expire by then.

src/router.js
matthew-white: Haha thank you, `routerRestoreSession` got me thinking of a better name! The "should" from before was actually there because some routes won't attempt to restore the session. (It's a configurable route meta field.) With that in mind, I've renamed the function to `restoreSessionForRoute()`. I like that that captures that the function is used specifically in the router and takes the initial route as an argument. Maybe there's a better name out there, but I think I'm satisfied with this new one!

matthew-white:
> This is great! I feel like I've run into this issue on my dev server, too (when my session is about to expire, or I've turned on/off the backend in some weird combination, or I was restoring a database...)

I do feel like we're making progress on the behavior of sessions! Definitely restoring a database could cause this behavior before. Logging in on the latest commit, then checking out v1.1 and logging out and logging back in could also cause this before. I haven't personally encountered an issue related to session restore when the session is about to expire — let me know if you see that again or if you encounter any other unexpected behavior around sessions!

Actually, thinking about it, now that #448 is in place, there could be an issue if your session is restored in the last 15 seconds of your session, because then there will be some seconds in which your session has expired, but `setInterval()` hasn't logged you out yet. However, that seems quite unlikely, and I think the worst-case scenario there is that the user sees some ugly 401 messages. We could detect that case and immediately log out after restoring the user's session, but my instinct is that that's not worth the additional logic. However, I've added a comment about that scenario.

I think this PR should be ready for another look!

matthew-white:
I also thought I'd mention that some of the scenarios discussed above will still apply to a tab that is already open. For example:

- The user logs in.
- A backup is restored.
- The user uses the existing tab to do something that sends a request.
- The user will see a 401 message.

In that case, things should work again if the user refreshes the page or logs out and logs back in.

For a future release, we could also consider automatically logging out the user as soon as they receive a 401. Though a 401 may be normal in some cases, for example, if the user tries to change their password and enters their old password incorrectly.

ktuite (approved):

PR #465, opened 2021-05-14
Update German translations for v1.2

matthew-white:


lognaturel:
Nooo, why must you torture me like that and not include all language updates including the fixes to egregious things in French?! Is there a benefit to selectively updating languages? I figured you'd pull everything from Transifex and run your script.

matthew-white:
Haha I already updated French in 88344049fe69e81af72672f80ce8851c84a38358! German is the only language that's changed since my last pull!

PR #467, opened 2021-05-19
Add instructions for running Enketo for development

ktuite:
Fixes getodk/central#219


README.md
```diff
@@ -59,7 +59,62 @@ ODK Central Frontend communicates with ODK Central Backend in part using a sessi
...
+```
+
+- Configure Nginx in ODK Central Frontend. Add the following to `nginx.conf` inside `server`, before or after `location /v1/`. This will create a reverse proxy for the Enketo server in the same way there is a reverse proxy to the ODK Central Backend server. You will need to restart Nginx after this.
+
+```
+    location /- {
+      proxy_pass http://localhost:8005/-;
+      proxy_redirect off;
+      proxy_set_header Host $host;
+    }
```
matthew-white, line 110: I'm thinking it might be easiest for you to go ahead and put this in `nginx.conf`: I don't think there's any harm in always having it there. If Enketo isn't running, there'd be an error message, but that's also the case with the current `nginx.conf`. The production `nginx.conf` puts the Enketo `location` block before the other `location` blocks, so I think it'd work well to do that here as well.

README.md
```diff
@@ -59,7 +59,62 @@ ODK Central Frontend communicates with ODK Central Backend in part using a sessi
...
+}
+```
+
+- Configure ODK Central Backend. Add the following to `config/default.json`.
+
+```
+    "enketo": {
+      "url": "http://localhost:8005/-",
+      "apiKey": "enketorules"
+    },
```
matthew-white, line 100: Similar to the above, I'm also wondering whether we should go ahead and put this in `config/default.json` now (though I feel less sure about that). Currently, `config/default.json` doesn't specify a configuration for Enketo, but it does specify a configuration for pyxform-http (localhost:5000). If pyxform-http isn't running, then the developer will see an error message, but if it is, the developer doesn't need to do anything extra to get Backend to start communicating with pyxform-http: it works automatically. Similarly, I think that if we add this configuration to `config/default.json`, then if Enketo isn't running, the developer will see an error message, but if it is, things will start working right away. I think right now, the developer will also see an error message, so I'm thinking there might not be much downside to making that change.

README.md
```diff
@@ -59,7 +59,62 @@ ODK Central Frontend communicates with ODK Central Backend in part using a sessi
 
 - To upload an XLSForm, run [pyxform-http](https://github.com/getodk/pyxform-http). ODK Central Frontend communicates with pyxform-http through ODK Central Backend.
 - You can use ODK Collect to scan an app user QR code, download a form, and submit data. One option to do so is to use [`ngrok`](https://ngrok.com/download). ODK Central Frontend is available on port 8989, so you can run `ngrok http 8989` to expose a temporary HTTPS URL that you can use. Within ODK Central Backend, you will also need to set the `default.env.domain` property in [`config/default.json`](https://github.com/getodk/central-backend/blob/master/config/default.json) to the HTTPS URL, then restart ODK Central Backend if it is already running.
-- Enketo is a web form engine used to show form previews and allow for web-based data entry. The development `nginx.conf` is not configured to work with Enketo.
+- Enketo is a web form engine used to show form previews and allow for web-based data entry. The following section describes how to set up an Enketo server for use by ODK Central Frontend and ODK Central Backend.
+
+## (Optional) Running Enketo for development
```
matthew-white, line 64: I think it'd be nice to keep the readme pretty short. Now that we have a `docs` directory, what do you think about putting this in a separate file there and linking to it?

matthew-white (commented):
Very excited about this PR! I feel like these instructions will really ease local development.

README.md
ktuite: I agree that it feels fine to put it here in `nginx.conf`! Less sure on putting it in the backend, but your reasoning below sounds good!

docs/enketo.md
```diff
@@ -0,0 +1,61 @@
...
+- Configure ODK Central Backend. Add the following to `config/default.json`.
+
+```
+    "enketo": {
+      "url": "http://localhost:8005/-",
+      "apiKey": "enketorules"
+    },
+```
+
+- Configure Nginx in ODK Central Frontend. The following lines have already been added to `nginx.conf`. They create a reverse proxy for the Enketo server in the same way as the reverse proxy to the ODK Central Backend server.
```
lognaturel, line 40: Nit: Now that the `nginx` config includes this, maybe this step could go last and the first sentence could go? Nice to have the explanation but this implies there's an action to take.

lognaturel (approved):
Very clear, thanks!

README.md
lognaturel: I think that's right but would be good to review how/if `config/default.json` is used to get to a prod setup. I'd merge this first and maybe explore next time we have to look at Enketo config things.

matthew-white:
@ktuite, when you have a chance, how about we update the `central` readme to mention that Central can be configured to work with Enketo in development? Probably that could link to the new .md file here?

PR #468, opened 2021-05-21
Add Japanese

### LINKED ISSUE #440: i18n: space between sentences
There are places in Frontend where we insert a space between translated sentences:

<details>
<summary>Examples</summary>

https://github.com/getodk/central-frontend/blob/49ae7b080103e3a03572ff57c2616d498c0b0220/src/components/backup/new.vue#L18-L22
https://github.com/getodk/central-frontend/blob/4f366fd674d83b7f2f156822e5a8d2ceca563b32/src/components/field-key/qr-panel.vue#L59-L62
https://github.com/getodk/central-frontend/blob/4f366fd674d83b7f2f156822e5a8d2ceca563b32/src/components/field-key/qr-panel.vue#L25-L47

There are also cases where we insert a `&nbsp;` before or after an `i18n` component.
<hr>
</details>

However, some locales do not generally have a space between sentences, for example, Japanese. In Japanese, characters tend to be fixed-width, so the additional space can cause the text to be misaligned.

matthew-white:
Closes #440. See the commit messages for additional notes.


lognaturel (approved):

matthew-white:
I looked into whether our styles would need to change to account for Japanese. I was expecting some things to need to change, but was happily surprised that things seem to work as-is. (Maybe it really is mostly just LTR vs. RTL that is challenging!)

I increased the width of a few elements, which we've needed to do for other locales in the past. The width of `DateRangePicker` increased quite a bit (by over 45px), so I conditioned that width increase on the selection of Japanese. Since `DateRangePicker` is used in `SubmissionFilters`, it's important for it to be as narrow as possible.

While researching styles, I encountered some recommendations that were straightforward to implement:

- Use a different default font for Japanese. This seems to be particularly important for sentences that contain a mix of Japanese and Latin characters.
- Limit text to 35 characters per line. 35 was at the upper end of the recommendations I saw. This ends up making text blocks 15–20% less wide in Japanese.
- Use bold rather than italics for `<em>`, as Japanese doesn't generally use italics. (Actually, I wonder whether other locales we support might not use italics to convey emphasis? Though right now I think the only place we use `<em>` is `ProjectFormAccessStates`.)

I'm sure there's more we can do here (I saw a recommendation to reduce font sizes, for example). However, I've tried to implement all the quick wins I saw.

Another issue to consider is #440, which seems noticeable in Japanese. However, it's not clear to me what the answer is there, so I think that can be addressed in a follow-up PR.

Resources:

- https://ux.shopify.com/html-localization-7eae6dcf3794
- https://aqworks.com/en/blog/2016/09/20/perfect-japanese-typography/
- https://gist.github.com/vanquang9387/12b7894223e73288414119c84027fb9b

matthew-white:
I think things should be all set with this PR. @lognaturel, it'd be great if you could take another look! I've made some changes to the first commit and added a commit to address #440; however, the last commit just involves running `destructure.js`.

Tomonori Hoshi says that they may continue updating translations through June, so I think we can go ahead and merge this PR and release Japanese as part of v1.3. This PR makes some slight changes to a few source messages, so it'd probably be a good idea to merge it before too long.

I looked into #440 a little more, and I think I've come to the conclusion that there's no elegant answer there. I also realized that removing instances of `&nbsp;` could be beneficial in other locales as well, since `&nbsp;` can affect word-wrapping. The solution I landed on is to use a component to insert spaces between sentences when the locale isn't Japanese. I like that this approach doesn't involve a global utility, a mixin, or logic in the component definition. I would probably consider a different approach in the future if:

- This issue comes up outside a template; or
- There ends up being another character that varies based on the locale. For example, maybe one day there will be a locale where it doesn't work to use `*` to indicate that a field is required?

That said, I'm happy with the current approach, and it also shouldn't be hard to change it in the future if need be.

src/assets/scss/app.scss
```diff
@@ -446,8 +453,6 @@ $form-group-padding-bottom: 15px;
 .empty-table-message {
   color: #555;
   font-size: 15px;
-  // Unless the message is quite long, we want it to fit on one line.
-  max-width: 650px;
```
matthew-white, line 450: The value of `max-width` in the `text-block` mixin will scale with the font size, so this line is no longer needed.

lognaturel (approved):

matthew-white:
I'm going to continue pushing updates to the `add-ja` branch, so I won't delete it for now.

PR #471, opened 2021-06-03
Improve submission actions

matthew-white:


src/components/submission/table.vue
```diff
@@ -90,34 +120,36 @@ export default {
       return this.project != null && this.project.permits('submission.update');
     }
   },
-  mounted() {
```
matthew-white, line 93: Previously, I manually added event listeners in the `mounted` hook instead of using Vue event handling. I thought I had to do that in order to use event delegation, but it turns out that that's not the case, so I've switched things over to Vue event handling. We have more event listeners now, so it's nice to be able to use the Vue event handling. Other than that change, most of the actual logic should be the same.

src/components/submission/table.vue
```diff
@@ -90,34 +120,36 @@ export default {
...
+      const dataRow = event.target.closest('tr');
+      const index = Number.parseInt(dataRow.dataset.index, 10);
+      if (index === this.actions.dataHover) return;
+      const { metadataBody } = this.$refs;
+      if (this.actions.dataHover != null)
+        metadataBody.querySelector('.data-hover').classList.remove('data-hover');
+      const metadataRow = metadataBody.querySelector(`tr:nth-child(${index + 1})`);
+      // The SubmissionMetadataRow element does not have a class binding, so I
+      // think we can add this class without Vue removing it.
+      metadataRow.classList.add('data-hover');
```
matthew-white, line 137: I've renamed this class from `actions-shown` to `data-hover`, since now hovering over a `SubmissionDataRow` doesn't necessarily mean that actions are shown for the row (because the user might be using the keyboard to navigate). The only purpose of `data-hover` is to indicate whether the cursor is over the `SubmissionDataRow`.

test/components/submission/update-review-state.spec.js
```diff
@@ -59,55 +44,45 @@ describe('SubmissionUpdateReviewState', () => {
...
-      await trigger.input(modal, 'textarea', 'Some notes');
-      modal.setProps({ state: false });
-      await modal.vm.$nextTick();
-      modal.setProps({ state: true });
-      await modal.vm.$nextTick();
-      modal.first('input[value="hasIssues"]').element.checked.should.be.true();
-      modal.first('textarea').element.value.should.equal('');
-    });
-
-    it('resets review state to approved if current state is null', async () => {
```
matthew-white, line 101: The review state selection is now set in only one place, not two. That means that this is no longer a special case and is covered by the tests above. I've removed this test.

src/components/submission/metadata-row.vue
```diff
@@ -167,7 +171,7 @@ export default {
   }
 
   .icon-angle-right {
-    bottom: #{$padding-right-table-data + 1px};
+    bottom: #{$padding-bottom-table-data + 1px};
```
matthew-white, line 174: I'm pretty sure this was a typo. `$padding-right-table-data` and `$padding-bottom-table-data` have the same value, so there's no functional change.

test/components/submission/activity.spec.js
```diff
@@ -70,9 +76,77 @@ describe('SubmissionActivity', () => {
       const component = mountComponent();
       component.find('#submission-activity-update-review-state-button').length.should.equal(0);
     });
+
+    describe('after a successful response', () => {
```
matthew-white, line 80: These tests are moved from elsewhere.

issa-tseng (approved):

PR #475, opened 2021-06-22
Add project settings to Collect QR codes

### LINKED ISSUE #474: Add project name to configuration QR code
Collect v2021.2 is in beta as of Jun 17 and adds support for configuring multiple projects.

Project name and icon can be included in configuration QR codes by adding:
```
  "project": {
    "name": <project name>
    "icon": <single character>
  }
```

https://github.com/getodk/collect/blob/master/collect_app/src/main/java/org/odk/collect/android/configure/qr/AppConfigurationKeys.kt

- [ ] Add project name to App User QR codes
- [ ] Add form title preceded by `[Draft]` and 📝 ([memo emoji](https://emojipedia.org/memo/)) to draft QR codes

(Note: project icon defaults to the first character of the project name and color is generated from a hash of project name. If there's no project name, it defaults to the domain name)

matthew-white:
Closes #474.

### DIFF
--- a/src/components/collect-qr.vue
+++ b/src/components/collect-qr.vue
@@ -21,7 +21,6 @@ import pako from 'pako/lib/deflate';
 export default {
   name: 'CollectQr',
   props: {
-    // General settings only (not admin settings)
     settings: {
       type: Object,
       required: true
@@ -36,19 +35,9 @@ export default {
     }
   },
   computed: {
-    fullSettings() {
-      return {
-        general: {
-          ...this.settings,
-          server_url: `${window.location.origin}${this.settings.server_url}`
-        },
-        // Collect requires the settings to have an `admin` property.
-        admin: {}
-      };
-    },
     imgHtml() {
       const code = qrcode(0, this.errorCorrectionLevel);
-      const json = JSON.stringify(this.fullSettings);
+      const json = JSON.stringify(this.settings);
       code.addData(btoa(pako.deflate(json, { to: 'string' })));
       code.make();
       return code.createImgTag(this.cellSize, 0);
--- a/src/components/field-key/qr-panel.vue
+++ b/src/components/field-key/qr-panel.vue
@@ -70,26 +70,31 @@ import DocLink from '../doc-link.vue';
 
 import FieldKey from '../../presenters/field-key';
 import { apiPaths } from '../../util/request';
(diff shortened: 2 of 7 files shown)


src/components/collect-qr.vue
```diff
@@ -36,19 +35,9 @@ export default {
...
-    fullSettings() {
-      return {
-        general: {
-          ...this.settings,
-          server_url: `${window.location.origin}${this.settings.server_url}`
-        },
-        // Collect requires the settings to have an `admin` property.
-        admin: {}
-      };
-    },
```
matthew-white, line 48: `FieldKeyQrPanel` and `FormDraftTesting` now do this work. I thought about leaving it here, but the code would have been surprisingly hairy, since `settings` now contains nested objects. I also think the new approach is a little clearer. Part of me doesn't love that `FieldKeyQrPanel` and `FormDraftTesting` now have some duplicate code, but it's just two components and not a lot of duplicate code. And probably we'll return to all this before too long once we start working on project settings.

src/components/field-key/qr-panel.vue
```diff
@@ -70,26 +70,31 @@ import DocLink from '../doc-link.vue';
...
 
 export default {
   name: 'FieldKeyQrPanel',
   components: { CollectQr, DocLink },
   props: {
-    fieldKey: FieldKey, // eslint-disable-line vue/require-default-prop
-    managed: {
-      type: Boolean,
-      default: false
-    }
```
matthew-white, line 82: The linter no longer flags `vue/require-default-prop`, so this can be simplified.

src/components/form-draft/testing.vue
```diff
@@ -90,16 +90,20 @@ export default {
...
+        general: { server_url: `${window.location.origin}${url}` },
+        project: {
+          name: this.$t('collectProjectName', this.formDraft),
+          icon: '📝'
+        },
+        // Collect requires the settings to have an `admin` property.
+        admin: {}
       };
-    },
-    baseUrl() {
```
matthew-white, line 101: I randomly noticed that this computed property isn't used anywhere anymore! Probably it should have been removed as part of 43ceaa0c7ec9b74d3833c3398c9d8e2f7c4ca711.

lognaturel (approved):
🎉

PR #476, opened 2021-06-23
Replace avoriaz with Vue Test Utils

### LINKED ISSUE #173: Moving from avoriaz
We currently use avoriaz for our testing, but [Vue Test Utils](https://vue-test-utils.vuejs.org/) seems to be [where the community is heading](https://github.com/eddyerburgh/avoriaz#deprecation). Vue Test Utils is still in beta, but it has been under development for a while, which makes me think that it should be fine to move to it.

matthew-white:
Closes #173. See the commit message for additional notes.


ktuite:
Met with Matt. Some notes from the meeting (many notes match what is described in the commit message): https://docs.google.com/document/d/1uEeOOikBnTZIyOkIsa1DG9hgWw6ORHH7ICqRWRJ1Qs4/edit?usp=sharing

matthew-white (commented):
@ktuite and I did an interactive review of this PR yesterday. We didn't cover everything, but we covered all the changes to the testing infrastructure and to `src/` files. We reviewed changes to the tests of the router and to unit tests. We also reviewed changes to the tests of a few components, including `SubmissionFeedEntry`, `FormDraftChecklist`, and `Download`.

One thing we talked about is the differences between `mount()`, `mockHttp()`, and `load()`. As part of that, we took a look at the new documentation in `test/util/http.js`. With the changes in this PR, I think all that is clearer than it's ever been. But I think some things are still confusing, so we're going to continue to think about ways to improve the documentation (probably in a follow-up PR).

We spent some time on the interactive review, and I don't think it's necessary to review the remaining component tests. @ktuite will maybe look at a few more on her own, and I'm also going to double-check some things. But I think the overall risk is low, so I'm planning to timebox that and am hoping to merge this PR by next Tuesday.

test/util/http.js
```diff
@@ -253,9 +270,48 @@ is equivalent to:
...
+Which of these approaches should I take?
+----------------------------------------
+
+Every test of a component will need to mount the component. There are a few ways
+to do so:
+
+  1. mount()
+  2. mockHttp()
+  3. load()
+  4. load(), specifying `root` as `false`
```
ktuite, line 282: I suggest reordering this list and having the most common/straightforward ones at the top. It also matches the explanations below. 

```
  1. load() - used when loading the entire App is necessary
  2. mount() - used for testing a component in isolation (fastest)
  3. load(), specifying `root` as `false`
  4. mockHttp()
```

ktuite (approved):
I've taken another look! Looks good!

A high level suggestion is possibly moving the large comment about writing tests and mount/load/mockhttp in `test/util/http.js` into a separate documentation file about testing. (Especially since there is a docs folder now for some doc images.)

ktuite:
Oh! You might want to update some of the wording in [CONTRIBUTING.md](https://github.com/getodk/central-frontend/blob/master/CONTRIBUTING.md) about using avoriaz.

test/util/http.js
matthew-white: I actually think that in some ways, the current order already goes from most straightforward to least straightforward. It starts with `mount()`, which comes from Vue Test Utils and only mounts a component. It then moves on to `mockHttp()`, which mocks `$http` and can also mount a component by calling `mount()`. After that is `load()`, which calls `mockHttp()`, then specifies responses.

I'm not sure that this section is successful at doing so, but what I want is to set up `mount()` and `mockHttp()` as the defaults while also presenting `load()` as the go-to function in certain circumstances. `mount()` and `mockHttp()` are more explicit and do less, and now that we can mock `$route`, we should be able to use `mount()` and `mockHttp()` in more tests. Yet if you need to mount `App`, or if you need to mount a route component that sends a request, then `load()` is definitely the most convenient option.

Let me know if you can think of changes along those lines that we could make to this section! Like maybe we should say here that `mockHttp()` calls `mount()`, and `load()` calls `mockHttp()` — maybe that would give a sense of increasing complexity? I also think your comment that `mount()` is fastest is helpful — I'll add that to this section now.

matthew-white:
> A high level suggestion is possibly moving the large comment about writing tests and mount/load/mockhttp in `test/util/http.js` into a separate documentation file about testing. (Especially since there is a docs folder now for some doc images.)

I think it's useful to have that section close to the other comments about `mockHttp()` and `load()`: I think a natural question after reading the previous comments will be when to use `mockHttp()` and when to use `load()`, which this section tries to answer. We could move all the comments in the file elsewhere, but I think I like having those comments close to the functions they refer to. CONTRIBUTING.md links to this file, but maybe it should say explicitly that the file contains a guide to `mockHttp()` and `load()`, including when to use them instead of `mount()`?

> Oh! You might want to update some of the wording in [CONTRIBUTING.md](https://github.com/getodk/central-frontend/blob/master/CONTRIBUTING.md) about using avoriaz.

I think I replaced all the references to avoriaz there, but let me know if I missed any!

matthew-white:
I've made some small changes after code review. I think I'm going to go ahead and merge, since this PR touches so many files, and I want to start working on other stories. However, I haven't had the chance to double-check component tests. I'm hoping to do that as I also work on other stories, and if I see anything that should change, I'll push a follow-up PR. I'd also love to continue discussing with @ktuite how to improve comments/documentation. I think changes there could also come in a follow-up PR.

PR #477, opened 2021-06-23
Visualizing submission diffs

ktuite:
Edits to a submission are shown on the Submission activity feed. (Behind the scenes, the backend computes diffs between each version of a submission and the previous version. There is an audit event for each submission update, and these audit events are linked with the calculated diff information to produce a visualization for this page.)

Edits/diffs can consist of single fields changing value, or nested changes such as an entire instance being added or removed (e.g. when a new element is added inside a repeat group such as a person and their name and age being added to a list of people in a family.)
  
Screen Shot 2021-07-22 at 9 52 19 PM


src/components/submission/diff-item.vue
```diff
@@ -0,0 +1,226 @@
...
+        } else if (typeOfChange === 'added') {
+          changes.push({ path, old: null, new: node[k] });
+        } else {
+          changes.push({ path, old: node[k], new: null });
+        }
+      }
+      return changes;
+    },
+    binaryHref(value) {
+      return `/v1${this.$route.fullPath}/attachments/${value}`;
```
matthew-white, line 150: You can access `projectId`, `xmlFormId`, and `instanceId` in order to use the `apiPaths` helper:

**Option 1:** `$this.route.params` (not preferred)

```js
this.$route.params.projectId
this.$route.params.xmlFormId
this.$route.params.instanceId
```

**Option 2:** the store

```js
computed: {
  ...requestData(['form', 'submission']),
  binaryHref(value) {
    return apiPaths.submissionAttachment(
      this.form.projectId,
      this.form.xmlFormId,
      false,
      this.submission.instanceId,
      value
    );
  }
}
```

src/components/submission/diff-item.vue
```diff
@@ -0,0 +1,226 @@
...
+
+.submission-diff-item {
+  border-bottom: 1px solid #ccc;
+}
+
+.submission-diff-item:last-child {
+  border-bottom: 0;
+}
+
+.full-path, .field-name, .nested-change-type {
```
matthew-white, line 167: You can nest class selectors within `.submission-diff-item` so that they're scoped to the component.

src/components/submission/diff-item.vue
```diff
@@ -0,0 +1,226 @@
...
+</script>
+
+<style lang="scss">
+@import '../../assets/scss/mixins';
+
+.submission-diff-item {
+  border-bottom: 1px solid #ccc;
+}
+
+.submission-diff-item:last-child {
```
matthew-white, line 163: You can also use `&` in Sass:

```scss
.submission-diff-item {
  &:last-child { border-bottom: 0; }
}
```

src/components/submission/diff-item.vue
```diff
@@ -0,0 +1,226 @@
...
+
+.field-name, .nested-change-type {
+  width: 150px;
+  margin: 10px;
+}
+
+.field-name {
+  color: #000;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
matthew-white, line 194:
```scss
@include text-overflow-ellipsis;
```

src/components/submission/diff-item.vue
```diff
@@ -0,0 +1,226 @@
...
+    </div>
+    <div>
+    <div class="diff-details">
+      <template v-if="isAtomicChange">
+        <div class="field-name" :title="fieldName">
+          {{ fieldName }}
+        </div>
+        <div class="old-to-new">
+          <span v-if="entry.old" class="data-old">
+            <span v-if="isBinary">
```
matthew-white, line 38: Edge case:

- A form is published, and a submission is created.
- A binary field is removed from the form, and a new form version is published.
- The submission is edited.
- The diff will show **filename -> empty**, but filename won't be linked.

This seems very unlikely, and there isn't an easy way to address it, but just wanted to note that we thought about this case.

matthew-white (commented):
This is looking really good so far! ✨ Adding notes from our interactive review.

src/components/submission/diff-item.vue
ktuite: I'm trying this (**option 2** in particular) but `form` doesn't seem to be in the store! The `projectId` and `xmlFormId` are passed around as props to the main components, but they are not plumbed deep enough to get to the components I need them in. I'm temporarily falling back to **option 1** for the form fields.

matthew-white: Ah you're totally right. `SubmissionShow` doesn't send a request for the form, because it generally doesn't need it, but that means that `form` won't exist. Other components on the page that use `apiPaths` do so by receiving `projectId` and `xmlFormId` as props. You could definitely consider doing so as well. But I do see that that'd require a lot of plumbing, passing those props from `SubmissionActivity` to `SubmissionFeedEntry` to `SubmissionDiffContainer` to `SubmissionDiffItem`. Maybe that's a reason to consider removing `SubmissionDiffContainer`, now that it's doing less?

I think it's also fine to continue using `this.$route.params`. In general, Vue Router [encourages](https://router.vuejs.org/guide/essentials/passing-props.html) the use of props over `$route`. However, I think it'd be reasonable to use `$route` in this case if that seems easier! (And you could use `$route` for `instanceId` if that seems nicer.) The main thing is that when testing `SubmissionDiffItem` in isolation, you'll need to mock `$route`.

ktuite: I'm going to try removing `SubmissionDiffContainer` and getting the props through to `SubmissionDiffItem`!

src/components/submission/diff-item.vue
```diff
@@ -0,0 +1,267 @@
...
+        this.xmlFormId,
+        false,
+        this.instanceId,
+        value
+      );
+    },
+    formatData(content) {
+      const length = 240;
+      return content.length > length ? content.slice(0, length).concat('...') : content;
+    }
```
matthew-white, line 182: Rather than doing this in the JavaScript, I think it'd be better to do it in the CSS using the `text-overflow-ellipsis` mixin. Here are a couple of options that might work:

Option 1: use `inline-block`

```scss
.data-old, .data-new {
  @include text-overflow-ellipsis;
  display: inline-block;
  max-width: 250px; // or a different number
}
```

Option 2: use flexbox

```scss
// You might need some additional flexbox options?
.old-to-new { display: flex; }

.data-old, .data-new {
  @include text-overflow-ellipsis;
  max-width: 250px; // or a different number
}
```

src/components/submission/diff-item.vue
```diff
@@ -0,0 +1,267 @@
...
+
+<i18n lang="json5">
+{
+  "en": {
+    // The description of how a specific field in a form submission changed
+    "editCaption": {
+      "added": "(added)",
+      "deleted": "(deleted)",
+    },
+    "empty": "empty", // Text showing that a value in a submission edit is empty
```
matthew-white, line 264: The i18n comment mechanism is pretty particular: the comment needs to be before the property name in order to be picked up. (You can see that the comment doesn't currently appear in `strings_en.json`.)

src/components/submission/diff-item.vue
```diff
@@ -0,0 +1,267 @@
...
+        <div class="field-name" :title="fieldName">
+          {{ fieldName }}
+        </div>
+        <div class="old-to-new">
+          <span v-if="entry.old" class="data-old">
+            <template v-if="isBinary">
+              <a :href="binaryHref(entry.old)" :title="entry.old">{{ formatData(entry.old) }}</a>
+            </template>
+            <span v-else :title="entry.old">{{ formatData(entry.old) }}</span>
+          </span>
```
matthew-white, line 39: If you prefer, you could also put the `title` attribute on the `.data-old` and `.data-new` elements rather than their children. One nice result of that is that the `v-else` case could remain a `<template>` rather than rendering another `<span>`.

src/components/submission/diff-item.vue
ktuite: I'm trying out the flexbox route and it's ok except I wish I could get it to expand to fill the whole space while still doing the overflow ellipses trick.

Screen Shot 2021-07-27 at 4 24 02 PM

It also looks pretty weird when the page is too small and needs to scroll horizontally.

Screen Shot 2021-07-27 at 4 25 50 PM

matthew-white: > I'm trying out the flexbox route and it's ok except I wish I could get it to expand to fill the whole space while still doing the overflow ellipses trick.

I'm not totally sure, but I think that's possible using flexbox! Maybe remove the `width` property from `.data-old` and `.data-new` and instead specify the `flex` property (or maybe just `flex-grow`)? I think it should be possible to keep the icon a fixed width, then divvy up the remaining space to `.data-old` and `.data-new`.

This also feels like a nice-to-have, so if it ends up being challenging, I think we could come back to it.

> It also looks pretty weird when the page is too small and needs to scroll horizontally.

I think this comes from specifying a fixed `width`. If you're able to solve the above and get `.data-old` and `.data-new` to divvy up the full width, I think that'll also solve this issue. Alternatively, you could try replacing `width` with `max-width`.

That said, Frontend also isn't designed for smaller viewports, and there'll be other overflow issues if the viewport is too narrow. We generally assume that the viewport is at least 1000px wide. As long as things look OK at 1000px, I think that's probably fine!

test/components/submission/feed-entry.spec.js
```diff
@@ -195,4 +195,54 @@ describe('SubmissionFeedEntry', () => {
...
+        action: 'submission.update',
+        details: { instanceId: '5678' }
+      });
+
+      // Mount the component with diff-specific data
+      const component = await mount(SubmissionFeedEntry, {
+        propsData: {
+          projectId: '1',
+          xmlFormId: testData.extendedForms.last().xmlFormId,
+          instanceId: '1234',
```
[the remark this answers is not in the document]
ktuite: Also, mount is synchronous, don't need async/await.

test/components/submission/feed-entry.spec.js
```diff
@@ -195,4 +195,54 @@ describe('SubmissionFeedEntry', () => {
...
+          fields: testData.extendedForms.last()._fields
+        },
+        stubs: { RouterLink: RouterLinkStub },
+        mocks: { $route: '/projects/1/submissions/s' }
+      });
+
+      // Two SubmissionDiffItem components should be present
+      component.findAll('.submission-diff-item.outer-item').length.should.equal(2);
+    });
+  });
```
ktuite, line 247: Test filtering out deprecated ID and instance ID

src/components/submission/diff-item.vue
```diff
@@ -0,0 +1,275 @@
...
+
+
+  .old-to-new {
+    display: flex;
+    align-items: center;
+  }
+
+  .data-old, .data-new {
+    @include text-overflow-ellipsis;
+    width: 250px;
```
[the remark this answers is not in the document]
ktuite: we've also been experimenting with `flex: 1 0 49%;` but it has some issues with shorter boxes

ktuite (commented):
Doing an interactive code review with @matthew-white

test/components/submission/feed-entry.spec.js
```diff
@@ -195,4 +195,54 @@ describe('SubmissionFeedEntry', () => {
...
+      });
+
+      // Diffs attach to audits with the same details.instanceId
+      testData.extendedAudits.createPast(1, {
+        action: 'submission.update',
+        details: { instanceId: '5678' }
+      });
+
+      // Mount the component with diff-specific data
+      const component = await mount(SubmissionFeedEntry, {
```
matthew-white, line 214:
```js
const mountComponent = (options = {}) => mount(SubmissionFeedEntry, {
  propsData: {
    projectId: '1',
    xmlFormId: testData.extendedForms.last().xmlFormId,
    instanceId: 'foo', // or anything else
    entry: testData.extendedAudits.size !== 0
      ? new Audit(testData.extendedAudits.last())
      : testData.extendedComments.last()
  },
  requestData: {
    diffs: {},
    fields: testData.extendedForms.last()._fields,
    ...options.requestData
  },
  stubs: { RouterLink: RouterLinkStub },
  mocks: { $route: '/projects/1/submissions/s' }
});

it('shows diffs joined with a submission.update audit event', () => {
  // Create the audit test data.

  // The beforeEach() hook above will create a submission (and thereby a form),
  // so you might not need to create a second form.

  const component = mountComponent({
    requestData: {
      diffs: {
        5678: {
          // ...
        }
      }
    }
  });
});
```

src/components/submission/feed-entry.vue
```diff
@@ -129,12 +130,27 @@ export default {
...
         last(entry.path) !== 'deprecatedID');
+    },
+    newVersionId() {
+      const { details } = this.entry;
+      if (details == null) return null;
+      const { instanceId } = details;
+      return instanceId;
+    },
+    oldVersionId() {
+      const deprecatedIdDiff = this.allDiffs.filter((entry) => last(entry.path) === 'deprecatedID')[0];
```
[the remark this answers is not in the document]
ktuite: i changed this to `find()` but left the comparison as just a string comparison on the last element since array comparison with `['meta', 'deprecatedID']` seems more complicated.

src/components/submission/diff-item.vue
ktuite: Going with the condensed 250px wide & ellipsis overflow solution for now since getting something that consistently worked and looked ok was challenging.

matthew-white (approved):
I thought of one additional case to handle, but otherwise this is looking 💯. I'll add a ✅ so that you can merge after handling that case. I think this PR would be a good candidate for squashing. If you use the "squash and merge" option below, that'll preserve the commits/history here while adding just a single commit to `master`.

PR #478, opened 2021-06-28
Update translations for v1.2.2

matthew-white:


lognaturel (approved):
Scanned filenames and diff to try to catch anything other than translations that might have accidentally changed. Didn't see anything. I didn't make any effort to make sure the update was the latest or anything like that.

PR #479, opened 2021-06-30
Add basic user access logging to audit log

matthew-white:


test/components/audit/filters/action.spec.js
```diff
@@ -10,26 +10,21 @@ const mountComponent = ({ value = 'nonverbose' } = {}) =>
...
       option.element.textContent.should.containEql('\u00a0\u00a0\u00a0Create');
-      const attributes = option.attributes();
-      attributes.value.should.equal('user.create');
-      should.not.exist(attributes.class);
+      should.not.exist(option.attributes().class);
     });
 
     it('renders option correctly for an action with multiple periods', () => {
-      const option = mountComponent().findAll('option').at(14);
+      const option = mountComponent().get('option[value="form.update.draft.set"]');
```
matthew-white, line 25: Usually I don't have to update the tests of `AuditFiltersAction` when I add a new action to the filter. However, the index of this option changed, so this test started failing. I updated the test so that it selects the option using its value rather than its index. I also made similar changes to the tests above.

src/components/audit/row.vue
```diff
@@ -106,6 +106,7 @@ export default {
       const species = acteeSpeciesByCategory[this.category];
       if (species == null) return null;
       const { actee } = this.audit;
+      if (actee == null) return null;
```
matthew-white, line 109: Unlike other user actions, `user.session.create` isn't associated with an actee. The user who logged in is the actor. I could also/instead make the user the actee, but that seemed potentially confusing.

src/locales/en.json5
```diff
@@ -115,6 +115,7 @@
         "update_publish": "Publish Draft",
         "update_draft_delete": "Abandon Draft",
         "attachment_update": "Update Attachments",
+        "submission_export": "Download Submissions",
```
matthew-white, line 118: The release criteria specified "Submissions Export", but I thought a verb would nicely match the other actions. I used "Download" instead of "Export", since that's the word used for that button.

src/components/audit/row.vue
issa-tseng: i think we usually in this case set the actor as both the actor and actee. then we don't have the null.

issa-tseng (approved):
yeah i think i'd set the actee if it were me. guess that's on the backend review. but it looks good.

PR #480, opened 2021-07-02
Add Italian

matthew-white:


matthew-white:
I've split this PR into two commits. The second just involved pulling the translations and running `destructure.js`. The first sets up i18n for Italian, which was just a few lines. I'll continue updating translations as they're modified and will force-push changes to the second commit. However, this PR should be ready for review.

src/components/submission/field-dropdown.vue
```diff
@@ -302,6 +302,8 @@ export default {
     },
     "disabled": "Cannot select more than 100 columns.",
     "action": {
+      // This text is shown in a dropdown that allows the user to select which
+      // columns to display in a table.
```
ktuite, line 306: this doesn't seem related to the Italian translation but i'm assuming you saw other translation things (comments) you wanted to fix up.

ktuite (approved):

PR #496, opened 2021-09-17
Improve experience after Frontend is rebuilt

matthew-white:


test/data/forms.js
```diff
@@ -55,9 +56,11 @@ const forms = dataStore({
       enketoId: !draft ? enketoId : null,
       enketoOnceId,
       state,
-      createdAt: inPast
-        ? fakePastDate([lastCreatedAt, project.createdAt, createdBy.createdAt])
-        : new Date().toISOString(),
+      createdAt: publishedAt != null
```
matthew-white, line 59: Comment that we're setting publishedAt to createdAt in order to ensure that createdAt is not after publishedAt.

matthew-white (commented):
Notes from interactive review

ktuite (approved):

PR #502, opened 2021-09-24
Adding separate table of submission states and changing project title

ktuite:
Update to the analytics metrics preview on frontend to make the title/copy a little smaller and better explained, and to move the stats about submission states to their own table.

Screen Shot 2021-09-24 at 10 00 24 AM

### DIFF
--- a/src/components/analytics/metrics-table.vue
+++ b/src/components/analytics/metrics-table.vue
@@ -56,10 +56,6 @@ export default {
   .metric-value {
     text-align: right;
   }
-
-  tbody {
-    border-bottom: 1px solid #ddd;
-  }
 }
 </style>
 
@@ -82,7 +78,7 @@ export default {
       "num_public_access_links": "Number of Public Access Links",
       "num_forms": "Number of Forms",
       "num_forms_with_repeats": "Number of Forms with repeats",
-      "num_forms_with_geospatial": "Number of Forms with geosdata",
+      "num_forms_with_geospatial": "Number of Forms with geodata",
       "num_forms_with_encryption": "Number of Forms with encryption",
       "num_forms_with_audits": "Number of Forms with audits",
       "num_submissions_received": "Number of Submissions - Received",
--- a/src/components/analytics/metrics-table.vue
+++ b/src/components/analytics/metrics-table.vue
@@ -56,6 +56,10 @@ export default {
   .metric-value {
     text-align: right;
   }
+
+  tbody {
+    border-bottom: 1px solid #ddd;
+  }
 }
 </style>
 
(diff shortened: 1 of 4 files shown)


src/components/analytics/preview.vue
```diff
@@ -146,8 +165,11 @@ export default {
     "projects": {
       // This is the title shown above a series of metrics about Project usage.
       "title": "Project Summaries",
-      "subtitle": "Showing 1 Project of {numProjects}"
-    }
+      "subtitle": "(Showing the most active Project of {numProjects} Projects)"
```
matthew-white, line 168: With this rewording, I think this does need pluralization now. Even in English, there's the possibility of there being just one project, in which case I think we should show "Project" at the end instead of "Projects". Vue I18n has good docs around pluralization, but the upshot is that this message will change to:

```
"subtitle": "(Showing the most active Project of {count} Project) | (Showing the most active Project of {count} Projects)"
```

Notice that `{numProjects}` becomes `{count}`. That's because Vue I18n passes in the count as a variable named `{count}`.

Other changes:

```diff
-<span class="explanation">{{ $t('projects.subtitle', { numProjects }) }}</span>
+<span class="explanation">{{ $tcn('projects.subtitle', numProjects) }}</span>
```

```diff
 numProjects() {
-  return this.$n(this.analyticsPreview.projects.length);
+  return this.analyticsPreview.projects.length;
 }
```

We no longer call `this.$n()` in `numProjects()`, because Vue I18n needs a number, not a localized string, in order to pluralize. However, `$tcn()` (a little utility I wrote) will call `$n()` on the count before passing it to the message via `$tc()`.

By the way, I don't think the test will need to change at all.

src/components/analytics/metrics-table.vue
```diff
@@ -56,6 +56,10 @@ export default {
   .metric-value {
     text-align: right;
   }
+
+  tbody {
+    border-bottom: 1px solid #ddd;
```
matthew-white, line 61: We don't usually put a border below a table, so I feel unsure about this change. But if you feel like it's an improvement, I'd say feel free to run with it. But in that case, I think it'd be better to use the Sass variable `$border-top-table-data` in order to ensure that this border is the same as the one between rows.

src/components/analytics/metrics-table.vue
```diff
@@ -78,14 +82,14 @@ export default {
       "num_public_access_links": "Number of Public Access Links",
       "num_forms": "Number of Forms",
       "num_forms_with_repeats": "Number of Forms with repeats",
-      "num_forms_with_geospatial": "Number of Forms with geospatial",
+      "num_forms_with_geospatial": "Number of Forms with geosdata",
```
matthew-white, line 85: No "s" in "geodata"

test/components/analytics/preview.spec.js
```diff
@@ -10,12 +10,18 @@ const analyticsPreview = {
...
+        num_submissions_from_web_users: { recent: 0, total: 0 }
+      }
     },
     {
       users: { num_managers: { recent: 1, total: 1 } },
       forms: { num_forms: { recent: 1, total: 1 } },
-      submissions: { num_submissions_received: { recent: 1, total: 1 } }
+      submissions: {
+        num_submissions_received: { recent: 1, total: 1 },
+        num_submissions_from_web_users: { recent: 0, total: 0 }
```
matthew-white, line 23: I made all the numbers here 1 in order to differentiate this project from the other project. That way you can be sure that the numbers are coming from the second project, not the first.

matthew-white (approved):
Looking good! ✨

src/components/analytics/preview.vue
ktuite: I added a single project test as well!

src/components/analytics/metrics-table.vue
ktuite: It looked like an odd amount of whitespace at first, but now when I try it without the line, I think it looks good! I guess I'll leave it out.

PR #503, opened 2021-09-24
Made comment in update review state textarea not required

ktuite:
Fixes an issue that @matthew-white noticed where the comment in the Submission Update Review State modal was required when it said it wasn't. This was done by adding a `required` prop (which defaults to `false`) in the `MarkdownTextarea` component that gets applied to the inner `<textarea>`.

`SubmissionComment` on the Submission show page now explicitly sets this to `true`. `SubmissionUpdateReviewState`'s code does not change, but the comment is now optional.

### DIFF
--- a/test/components/submission/comment.spec.js
+++ b/test/components/submission/comment.spec.js
@@ -154,6 +154,13 @@ describe('SubmissionComment', () => {
     });
   });
 
+  it('should make the textarea required', async () => {
+    testData.extendedSubmissions.createPast(1);
+    const component = mountComponent();
+    const required = await component.get('textarea');
+    should.exist(required);
+  });
+
   describe('after a successful response', () => {
     const submit = () => {
       testData.extendedForms.createPast(1, { xmlFormId: 'a b' });
--- a/test/components/submission/comment.spec.js
+++ b/test/components/submission/comment.spec.js
@@ -157,8 +157,7 @@ describe('SubmissionComment', () => {
   it('should make the textarea required', async () => {
     testData.extendedSubmissions.createPast(1);
     const component = mountComponent();
-    const required = await component.get('textarea');
-    should.exist(required);
+    component.getComponent(MarkdownTextarea).props().required.should.equal(true);
   });
 
   describe('after a successful response', () => {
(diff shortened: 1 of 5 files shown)


test/components/submission/update-review-state.spec.js
```diff
@@ -73,6 +73,15 @@ describe('SubmissionUpdateReviewState', () => {
     modal.get('input[value="hasIssues"]').should.be.focused();
   });
 
+  it('does not require a comment in the text area', async () => {
+    testData.extendedSubmissions.createPast(1, { reviewState: 'hasIssues' });
+    const modal = mountComponent();
+    await modal.setProps({ state: true });
+    modal.getComponent(MarkdownTextarea).props().required.should.equal(false);
+    const required = await modal.get('textarea').attributes('required');
```
matthew-white, line 81: I don't think you need `await` here, since there's no promise. And actually, I feel like as long as there's a test of the `required` prop among the `MarkdownTextarea` tests, there's no need to test the attribute in addition to the prop. (And like I said above, I actually think this test isn't strictly necessary as long as there's that `MarkdownTextarea` test.)

test/components/submission/comment.spec.js
```diff
@@ -154,6 +154,13 @@ describe('SubmissionComment', () => {
     });
   });
 
+  it('should make the textarea required', async () => {
+    testData.extendedSubmissions.createPast(1);
+    const component = mountComponent();
+    const required = await component.get('textarea');
+    should.exist(required);
```
matthew-white, line 161: I think this is just testing that the `<textarea>` exists.

matthew-white (approved):
Looks great!

For tests, I think the main test that's needed is a test of `MarkdownTextarea`, testing that if the `required` prop is `true`, the `required` attribute is set. I think you could update the existing `mountComponent()` function in that file to accept a `required` prop?

After that, I personally actually wouldn't test the uses of the `required` prop in `SubmissionUpdateReviewState` or `SubmissionComment`, since there's nothing dynamic or complicated about how the `required` prop is used in those components. However, feel free to do so if that feels right!

test/components/submission/update-review-state.spec.js
ktuite: i'm going to keep these tests of the `MarkdownTextarea` `required` prop in these two components (as a kind of... reminder that the behavior is that one is required and one is optional) but not have it actually check the textarea directly, just the prop.

PR #506, opened 2021-09-27
Update translations for v1.3

matthew-white:


ktuite (approved):
Translations look good!

Since most of them are on the system > usage reporting page, I checked how the analytics preview modal looks with translations. Many phrases are longer, which makes the tables expand and then stack vertically, which I think is better than if they crammed together awkwardly. I can see some minor CSS tweaks to make, though. Including an image for context.

Screen Shot 2021-09-30 at 8 37 46 AM

matthew-white:
It's neat how they automatically stack like that! Feel free to make those CSS tweaks. There probably isn't time to get them into v1.3, but we could ship them with v1.4.

PR #509, opened 2021-09-30
Metrics preview CSS adjustments to even out columns

ktuite:
Evens out some padding that had been put in between the two columns in the English metrics modal. When translated to a different language, the strings are generally longer and the tables widen and stack vertically. The padding meant to separate columns was making the single column layout look uneven. That's been fixed by adding more even margins to both columns (and negative margins on the outer element) so the tables look good in both 1-column and 2-column modes.

Before and After:
before after

English still okay:
Screen Shot 2021-09-30 at 9 54 07 AM

Screen Shot 2021-09-30 at 9 54 21 AM

### DIFF
--- a/src/components/analytics/metrics-table.vue
+++ b/src/components/analytics/metrics-table.vue
@@ -15,7 +15,7 @@ except according to the terms contained in the LICENSE file.
       <tr>
         <th>{{ title }}</th>
         <th class="metric-value recent-col">{{ $t('recent') }}</th>
-        <th class="metric-value">{{ $t('common.total') }}</th>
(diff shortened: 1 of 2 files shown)


src/components/analytics/metrics-table.vue
```diff
@@ -56,6 +56,10 @@ export default {
   .metric-value {
     text-align: right;
   }
+
+  .total-col {
+    width: 50px;
```
ktuite, line 61: Flexbox doesn't totally seem to respect this width, but making it this small does seem to get it to choose a consistent width.

matthew-white (approved):

PR #510, opened 2021-10-06
Restore full margin between navbar links

matthew-white:
See the commit message for additional notes.


matthew-white:
@ktuite, I think this one can be reviewed async.

PR #515, opened 2021-10-22
Improve appearance of <select> elements in Safari

matthew-white:
Right now `select` elements look different in Safari than other browsers. Even within Safari, `select` elements look different from `input` elements. Right now in Safari, a `select` element looks like this:

select_safari_default

In Chrome, it looks like:

select_chrome

Firefox:

select_firefox

After this change, Chrome and Firefox will look like the same as now, but Safari will look like this:

select_safari_none


issa-tseng (approved):
lol

PR #517, opened 2021-11-05
Update translations before upgrading to Vue I18n v9

matthew-white:
We have to upgrade to Vue I18n v9 as part of moving to Vue 3. As part of that, I need to make a few changes to `bin/util/transifex.js`. I want to make sure that `restructure.js` and `destructure.js` still work as before after those changes. However, the structure of the Transifex translation files no longer match Frontend, because a message has been removed since the last time we updated translations. `destructure.js` won't work under those circumstances, so I'm doing a quick update of translations before upgrading to Vue I18n v9.

### DIFF
--- a/src/components/analytics/form.vue
+++ b/src/components/analytics/form.vue
@@ -332,8 +332,25 @@ export default {
       "null": [
         "後で通知する。",
         "管理者には引き続きスクリーン上部にメッセージが表示されます。"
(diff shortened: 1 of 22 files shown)


matthew-white:
@ktuite, I think this can be reviewed async!

ktuite (approved):

PR #534, opened 2022-02-11
Add "Download Submissions" modal

### LINKED ISSUE #236: Inform user if iframe is loading
The "Decrypt and Download" modal contains an iframe. The iframe is not visible to the user, but it must load before the user can download their data. Right now if the user clicks the Download button, and the iframe has not loaded, nothing happens. Instead, we should inform the user if the iframe is loading. For example, we could disable the button so that the user doesn't try to click it until the iframe has finished loading. Alternatively, we could keep it enabled, but display an alert about the iframe if the user clicks the button before the iframe has finished loading.

matthew-white:
Closes #236.


test/components/submission/decrypt.spec.js
```diff
@@ -1,183 +1,420 @@
...
+  describe('passphrase input', () => {
+    it('shows the input if there is a managed key', () => {
+      testData.extendedForms.createPast(1, {
+        key: testData.standardKeys.createPast(1, { managed: true }).last()
+      });
+      const modal = mountComponent();
+      modal.find('input[type="password"]').exists().should.be.true();
+    });
+
+    it('does not show the input if there is not a managed key', () => {
```
matthew-white, line 144: Reword test description.

PR #536, opened 2022-02-13
Add form version and user agent to SubmissionBasicDetails

matthew-white:
Related to getodk/central-backend#398 and getodk/central#254.


ktuite (approved):

matthew-white (commented):
Notes from interactive review with @ktuite

PR #537, opened 2022-02-14
Publish draft modal shows version input on duplicate version problem

ktuite:
In the scenario where a Form Draft version conflicts with another Form in the system, but that duplication conflict is not known ahead of time (e.g. a Form with the same ID and version that is currently in the trash), when the user tries to publish the Draft Form, backend returns a Problem. That problem shows up as an alert in the Form Draft publish modal.

This PR changes the modal behavior to let the user specify a new Form version after they encounter that specific problem. The version input field was already part of the modal for situations where the version duplication issue was known, so this was a simple change to unhide it when appropriate. 

Screen Shot 2022-02-14 at 3 17 45 PM

### DIFF
--- a/src/components/form-draft/publish.vue
+++ b/src/components/form-draft/publish.vue
@@ -36,7 +36,7 @@ except according to the terms contained in the LICENSE file.
         <p>{{ $t('introduction[1]') }}</p>
         <p v-if="draftVersionStringIsDuplicate">{{ $t('introduction[2]') }}</p>
       </div>
-      <form v-if="draftVersionStringIsDuplicate" @submit.prevent="publish">
+      <form v-if="draftVersionStringIsDuplicate || versionConflict" @submit.prevent="publish">
         <form-group ref="versionString" v-model.trim="versionString"
           :placeholder="$t('field.version')" required autocomplete="off"/>
         <!-- We specify two nearly identical .modal-actions, because here we
@@ -92,7 +92,8 @@ export default {
   data() {
     return {
       awaitingResponse: false,
-      versionString: ''
+      versionString: '',
+      versionConflict: false
     };
   },
   computed: {
@@ -134,7 +135,11 @@ export default {
           this.versionString !== this.formDraft.version
             ? { version: this.versionString }
             : null
-        )
+        ),
+        fulfillProblem: (problem) => {
+          if (problem.code === 409.6)
+            this.versionConflict = true;
+        }
       })
         .then(() => {
           this.$emit('success');
@@ -173,7 +178,7 @@ export default {
       "version": "Version"
     },
     "problem": {
-      "409_6": "The version name you’ve specified conflicts with a past version of this Form. Please change it to something new and try again."
+      "409_6": "The version name of this Draft conflicts with a past version of this Form or a deleted Form. Please use the field below to change it to something new or upload a new Form definition."
     }
   }
 }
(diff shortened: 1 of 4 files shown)


src/components/form-draft/publish.vue
```diff
@@ -134,7 +135,11 @@ export default {
           this.versionString !== this.formDraft.version
             ? { version: this.versionString }
             : null
-        )
+        ),
+        fulfillProblem: (problem) => {
+          if (problem.code === 409.6)
+            this.versionConflict = true;
+        }
```
matthew-white, line 142: The `fulfillProblem` function actually isn't meant to have side effects: rather than being an "on error" callback, it's a way to identify error responses that shouldn't be treated as errors (the function should return a boolean). Such responses will then be handled in the `then()` callback. I think it's best to use promise methods to process the response, but `fulfillProblem()` helps determine which promise callback will be used.

With that in mind, one option is to use `fulfillProblem` + `then()` along these lines:

```js
this.request({
  method: 'POST',
  url: '...',
  fulfillProblem: (problem) => problem.code === 409.6
})
  .then(({ data }) => {
    if (!isProblem(data)) {
      this.$emit('success');
    } else {
      this.$alert().danger(this.$t('problem.409_6'));
      this.versionConflict = true;
    }
  })
  .catch(noop);
```

In this case, `fulfillProblem` identifies a 409.6 Problem as not an error, so it doesn't do the normal error handling, including showing an alert. That's why the `then()` callback here calls `this.$alert()`.

To see another example of `fulfillProblem` + `then()`, check out the `FormNew` component.

I know I suggested that `fulfillProblem` might be the way to approach this issue, but maybe `catch()` would be clearer after all. That way the normal error handling would take place. That would look something like this:

```js
this.request({
  method: 'POST',
  url: '...',
  fulfillProblem: (problem) => problem.code === 409.6
})
  .then(() => {
    this.$emit('success');
  })
  .catch(error => {
    if(error.response != null && isProblem(error.response.data))
      this.versionConflict = true;
  });
```

Also, by the way, if there's a way to improve the comments about `fulfillProblem` in the `request` mixin to make all this clearer, please feel free to do so (either now or later)!

test/components/form-draft/publish.spec.js
```diff
@@ -269,11 +269,51 @@ describe('FormDraftPublish', () => {
...
+          'The version name of this Draft conflicts with a past version of this Form or a deleted Form. Please use the field below to change it to something new or upload a new Form definition.'
         );
       });
   });
 
+  it('shows the version input field after request returns duplicate version problem', () => {
+    // The scenario here is a user trying to publish a form that conflicts with
+    // a form/version combo probably found in the trash. This component doesn't
+    // have access to trashed forms so it doesn't know about the problem until the
+    // request from the backend returns the problem.
```
matthew-white, line 281: I think this comment is helpful, and I wonder whether it makes sense to move it to the component itself so that the motivation behind the approach there is clear. Just my two cents, it also works well here!

matthew-white (approved):
I left some thoughts about `fulfillProblem`, but otherwise this is looking good! I think the tests are quite clear.

src/components/form-draft/publish.vue
ktuite: I went with the first approach and added an example to the request mixin comment!

test/components/form-draft/publish.spec.js
ktuite: Now included in the component, too!

src/components/form-draft/publish.vue
matthew-white: Awesome, comment is great!

PR #539, opened 2022-02-16
Make changes related to password inputs

matthew-white:


ktuite (approved):

src/components/user/edit/password.vue
```diff
@@ -23,10 +23,10 @@ except according to the terms contained in the LICENSE file.
           autocomplete="current-password"/>
         <form-group id="user-edit-password-new-password" v-model="newPassword"
           type="password" :placeholder="$t('field.newPassword')" required
-          :has-error="mismatch || invalidPassword" autocomplete="new-password" strengthmeter/>
+          :has-error="tooShort || mismatch" autocomplete="new-password"/>
         <form-group id="user-edit-password-confirm" v-model="confirm"
```
matthew-white, line 27: The id attribute is being added to the input, but should it be added to the `.form-group` instead?

matthew-white (commented):
Notes from interactive review

src/components/user/edit/password.vue
matthew-white: I took a quick look, and I'm actually not seeing examples outside this component of `id` being specified for a `FormGroup`. I think I'm going to leave this as-is for now, but I think it's something to keep in mind going forward: this pattern seems a little confusing. If this pattern does come up again, we might want to add a different prop.

PR #542, opened 2022-02-17
Update translations for v1.4

matthew-white:


matthew-white:
@ktuite, just needs a very quick skim I think.

transifex/strings_id.json
```diff
@@ -3233,9 +3313,15 @@
       "reviewState": {
         "string": "Status ulasan"
       },
+      "formVersion": {
+        "string": ""
```
ktuite, line 3317: I noticed some of these new blank strings and I don't know if this is an intermediate transifex thing that is supposed to be like this or not. When the translation string is empty like this, does it show nothing or does it fall back to the original english??

ktuite (approved):
Just one question about how transifex works and what happens with untranslated strings. Other than that, it looks fine!

transifex/strings_id.json
matthew-white: Yes, that's normal! This is one of the ways in which the Transifex JSON is different from the Vue I18n JSON. In the Transifex JSON, an untranslated string appears as an empty string, whereas in the Vue I18n JSON, it doesn't appear at all. Vue I18n then falls back to the fallback locale (English).

PR #543, opened 2022-02-18
Fix left margin of submission actions in FormDraftTesting

matthew-white:
In `FormDraftTesting`, the leftmost control (either the refresh button or the field dropdown) is not aligned with the edge of the page:

Screen Shot 2022-02-17 at 7 46 25 PM

The margin is there because in `FormSubmissions`, those elements always have something to the left, because that page has the filters. This PR tweaks the styles so that if the element is the first child, it won't have left margin.

### DIFF
--- a/src/components/submission/list.vue
+++ b/src/components/submission/list.vue
@@ -349,6 +349,8 @@ export default {
   align-items: baseline;
   display: flex;
   flex-wrap: wrap-reverse;
+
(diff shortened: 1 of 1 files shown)


matthew-white:
@ktuite, I think this can be reviewed async!

ktuite (approved):

PR #544, opened 2022-02-19
Add Swahili

matthew-white:
This PR adds Swahili to Frontend. This required a couple of changes in addition to the usual tasks.

First, there was an error parsing the Transifex JSON, because the translator used a `\n`. That shouldn't have been a problem, because we convert `\n` in Transifex strings to space. Actually, the Transifex JSON for Japanese also seems to contain strings with `\n` without issue. However, the presence of `\n` specifically within a pluralized string resulted in an error here. Luckily, a small change to a regular expression was all that was needed.

Second, Swahili is the first locale for us for which there isn't a flatpickr localization. I think we should consider submitting a PR to flatpickr for that, since I think it'd be pretty easy. However, for now, I modified `DateRangePicker` so that it works even if there isn't a flatpickr localization (it falls back to English). I didn't want to add more flatpickr-related logic to `src/util/i18n.js` as part of that and instead decided to move the existing flatpickr logic in that file into `DateRangePicker`. The result is more consistent with how we approach internationalization in other components. I had also already made that change as part of moving to Vue 3 (in #526). That PR makes a few changes to `src/util/i18n.js`, in large part because the i18n object will no longer be a global object. More about flatpickr localization: https://flatpickr.js.org/localization/

I may update the last commit as translations are updated, but I don't plan to make changes to the other commits. Because of that, I think we can go ahead and review this PR. An interactive review would probably be easiest, but I think async would also work. Because this PR touches the same file as the Vue 3 PR, it'd be good to merge it soon. (Swahili will hopefully be ready by the time we want to release, but even if we decide to release it in a patch, we'd be able to temporarily hide these translations.)

### DIFF
--- a/bin/util/transifex.js
+++ b/bin/util/transifex.js
@@ -18,7 +18,8 @@ const locales = {
   fr: { pluralCategories: ['one', 'other'] },
   id: {},
   it: {},
-  ja: { warnVariableSeparator: false }
+  ja: { warnVariableSeparator: false },
+  sw: {}
 };
 
 const sourceLocale = 'en';
--- a/bin/util/transifex.js
+++ b/bin/util/transifex.js
@@ -87,7 +87,7 @@ class PluralForms {
   // Transifex uses ICU plurals.
   static fromTransifex(string, locale) {
     const forms = [];
-    const icuMatch = string.match(/^({count, plural,).+}$/);
+    const icuMatch = string.match(/^({count, plural,).+}$/s);
     if (icuMatch == null) {
       forms.push(string);
     } else {
--- a/src/components/date-range-picker.vue
+++ b/src/components/date-range-picker.vue
@@ -30,11 +30,16 @@ except according to the terms contained in the LICENSE file.
 
 <script>
 import flatpickr from 'vue-flatpickr-component';
-
 import { DateTime } from 'luxon';
 import 'flatpickr/dist/flatpickr.css';
 
-import { flatpickrLocales } from '../util/i18n';
+import 'flatpickr/dist/l10n/cs';
+import 'flatpickr/dist/l10n/de';
+import 'flatpickr/dist/l10n/es';
+import 'flatpickr/dist/l10n/fr';
+import 'flatpickr/dist/l10n/id';
+import 'flatpickr/dist/l10n/it';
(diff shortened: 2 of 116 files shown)


bin/util/transifex.js
```diff
@@ -87,7 +87,7 @@ class PluralForms {
   // Transifex uses ICU plurals.
   static fromTransifex(string, locale) {
     const forms = [];
-    const icuMatch = string.match(/^({count, plural,).+}$/);
+    const icuMatch = string.match(/^({count, plural,).+}$/s);
```
matthew-white, line 90: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll

src/components/date-range-picker.vue
```diff
@@ -63,28 +68,13 @@ export default {
...
     config() {
       const config = {
         mode: 'range',
         // See https://github.com/flatpickr/flatpickr/issues/1549
         dateFormat: 'Y/m/d'
       };
-      // If this.$i18n.locale changes, this.config will change, but flatpickr
-      // itself won't change:
-      // https://github.com/flatpickr/flatpickr/issues/1882
-      // https://github.com/flatpickr/flatpickr/issues/2019
```
matthew-white, line 86: This seems to be working now! Because of that, I removed these comments.

test/components/date-range-picker.spec.js
```diff
@@ -278,16 +278,21 @@ describe('DateRangePicker', () => {
...
-          locale.weekdays.longhand[0].should.equal('Domingo');
-        })
-        .finally(() => loadLocale('en')));
+    it('renders correctly for es', async () => {
+      await loadLocale('es');
+      const component = mountComponent({ attachTo: document.body });
+      await component.get('input').trigger('click');
+      const text = document.querySelector('.flatpickr-weekday').textContent.trim();
+      text.should.equal('Lun');
+    });
```
matthew-white, line 296: Instead of inspecting the `config` computed property, we actually cause the flatpickr element to be rendered, then inspect the element. That feels like a surer test that things are working and actually wasn't that much more complicated to write.

ktuite (approved):

PR #547, opened 2022-02-22
Call submit() rather than requestSubmit() to circumvent Firefox bug

matthew-white:
Fixing a bug that the QA team found:

> I’m not able to download submissions from encrypted project on Firefox &hellip; Everything is fine in Chrome and Safari


matthew-white:
@ktuite, this is short and could be reviewed async, but would probably be easier to review interactively.

ktuite (approved):
🙈

PR #549, opened 2022-02-22
Hide alert about passphrase length after successful response

matthew-white:
While working on #547, I noticed the following sequence:

- In the Enable Encryption modal, submit a passphrase that is too short. An alert is shown.
- Now submit a passphrase that is long enough. The modal, which is multistep, renders a new screen. However, the alert is still shown.

Elsewhere, we hide any alert when moving from one step of a multistep modal to the next:

https://github.com/getodk/central-frontend/blob/a273f313e5977593750e400f6b3e042e8d21fb81/src/components/field-key/new.vue#L119-L134

The goal of this PR is to do the same thing for the Enable Encryption modal.

@ktuite, I think this PR can be reviewed async!

### DIFF
--- a/test/components/project/enable-encryption.spec.js
+++ b/test/components/project/enable-encryption.spec.js
@@ -129,9 +129,42 @@ describe('ProjectEnableEncryption', () => {
   });
 
   describe('after a successful response', () => {
-    const submit = () => {
+    beforeEach(() => {
       testData.extendedProjects.createPast(1);
-      return load('/projects/1/settings')
+    });
+
+    it('shows a success icon', () =>
+      mockHttp()
+        .mount(ProjectEnableEncryption, mountOptions())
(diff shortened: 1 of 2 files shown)


test/components/project/enable-encryption.spec.js
```diff
@@ -129,9 +129,42 @@ describe('ProjectEnableEncryption', () => {
   });
 
   describe('after a successful response', () => {
-    const submit = () => {
```
matthew-white, line 132: It's a little hard to see with the diff, but there used to be a helper function named `submit()` that called `load()`. However, the new test couldn't use `submit()`, because it needs to set the input twice. And actually, only one test needed to be using `load()`. I changed the other existing test (about the icon) so that it calls `mockHttp()` instead.

ktuite (approved):

PR #566, opened 2022-03-19
Add the ability to format a date/time as an amount of time in the past

matthew-white:


src/components/date-time.vue
```diff
@@ -35,7 +39,7 @@ export default {
   },
   methods: {
     format(relative) {
```
matthew-white, line 41: Rename to `useRelative`.

matthew-white (commented):
Notes from interactive review

ktuite (approved):

PR #573, opened 2022-03-28
Editable project descriptions!! 

ktuite:
Some screenshots of the new Project description feature:

Screen Shot 2022-04-01 at 5 09 51 PM

Screen Shot 2022-04-01 at 5 09 45 PM

Screen Shot 2022-04-01 at 5 09 40 PM

### DIFF
--- a/src/components/markdown/textarea.vue
+++ b/src/components/markdown/textarea.vue
@@ -14,7 +14,7 @@ except according to the terms contained in the LICENSE file.
     <div class="form-group">
       <textarea :value="value" class="form-control"
         :placeholder="defaultText" :aria-label="defaultText"
-        :required="required" rows="2" @input="$emit('input', $event.target.value)">
+        :required="required" :rows="rows" @input="$emit('input', $event.target.value)">
       </textarea>
     </div>
     <div v-if="value !== '' && previewMode" class="preview-container">
@@ -44,16 +44,18 @@ export default {
       required: true
     },
     showFooter: {
-      type: Boolean,
-      default: false
+      type: Boolean
     },
     defaultText: {
       type: String,
       default: ''
     },
     required: {
-      type: Boolean,
-      default: false
+      type: Boolean
+    },
+    rows: {
+      type: String,
+      default: '2'
     }
   },
   data() {
--- a/src/components/markdown/textarea.vue
+++ b/src/components/markdown/textarea.vue
@@ -112,7 +112,7 @@ export default {
   }
 
   .md-preview-btn {
-    background-color: #ccc;
+    background-color: #bbb;
     color: #333;
     &:hover, &:focus, &:active:focus { background-color: #888; }
   }
--- a/src/components/markdown/textarea.vue
+++ b/src/components/markdown/textarea.vue
@@ -11,13 +11,11 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <div class="markdown-textarea" :class="activeBorderClass">
-    <div class="form-group">
-      <textarea :value="value" class="form-control"
-        :placeholder="defaultText" :aria-label="defaultText"
-        :required="required" :rows="rows" @input="$emit('input', $event.target.value)">
-      </textarea>
-    </div>
-    <div v-if="value !== '' && previewMode" class="preview-container">
+    <textarea :value="value" class="form-control"
+      :placeholder="defaultText" :aria-label="defaultText"
+      :required="required" :rows="rows" @input="$emit('input', $event.target.value)">
+    </textarea>
+    <div v-if="!emptyValue && previewMode" class="preview-container">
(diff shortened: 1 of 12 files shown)


src/components/project/edit.vue
```diff
@@ -16,8 +16,12 @@ except according to the terms contained in the LICENSE file.
     </div>
     <div class="panel-body">
       <form @submit.prevent="submit">
-        <form-group v-model.trim="name" :placeholder="$t('field.name')" required
```
ktuite, line 19: Might have taken this out by accident

ktuite (commented):
notes from an interactive review with @matthew-white 
... to be continued!

src/components/markdown/textarea.vue
```diff
@@ -14,14 +14,14 @@ except according to the terms contained in the LICENSE file.
     <div class="form-group">
       <textarea :value="value" class="form-control"
         :placeholder="defaultText" :aria-label="defaultText"
-        :required="required" rows="2" @input="$emit('input', $event.target.value)">
+        :required="required" :rows="rows" @input="$emit('input', $event.target.value)">
       </textarea>
     </div>
     <div v-if="value !== '' && previewMode" class="preview-container">
```
matthew-white, line 20: Is this another place where `emptyValue` is needed? Or is it impossible for `value` to be `null` if `previewMode` is `true`? (In that case, maybe a comment would be useful?)

src/components/project/overview/description.vue
```diff
@@ -0,0 +1,111 @@
+<!--
+Copyright 2022 ODK Central Developers
+See the NOTICE file at the top-level directory of this distribution and at
+https://github.com/getodk/central-frontend/blob/master/NOTICE.
+This file is part of ODK Central. It is subject to the license terms in
```
matthew-white, line 5: Would you mind adding a newline between these two just for consistency?

src/components/markdown/textarea.vue
```diff
@@ -44,16 +44,18 @@ export default {
       required: true
```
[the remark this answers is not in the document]
ktuite: It was triggering a warning so I made it not required.

src/components/project/edit.vue
```diff
@@ -18,6 +18,10 @@ except according to the terms contained in the LICENSE file.
       <form @submit.prevent="submit">
         <form-group v-model.trim="name" :placeholder="$t('field.name')" required
           autocomplete="off"/>
+        <label class="form-group">
+          <markdown-textarea v-model="description" rows="10"/>
```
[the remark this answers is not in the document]
ktuite: I took `form-group` completely out of markdown textarea so it can be added here at the outer level. This meant I also had to add it to the submission comment component in the div that wraps the markdown textarea there.

I also made the button a lighter gray but still darker than the surrounding background, but with dark text so it doesn't look disabled. 

Didn't want to add a slot for a label, just wanted to think of it as a textarea that you can use wherever. Though it kind of expects to be nested within a `.form-group` now to get the background color of the text area to be consistent with the rest of the markdown stuff.

src/components/markdown/textarea.vue
ktuite: emptyValue is needed here, yes!

src/components/project/edit.vue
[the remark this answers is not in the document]
ktuite: two things a form label does (that are hard to get it to do in this context with the way the textarea is nested inside another div):
1. hide itself it the input doesn't have any value
2. becomes blue color when input is active

src/components/submission/comment.vue
```diff
@@ -14,13 +14,15 @@ except according to the terms contained in the LICENSE file.
...
       <span class="icon-pencil"></span>{{ $t('editWithoutComment') }}
     </div>
-    <markdown-textarea v-model="body" :default-text="$t('field.comment')"
-      :show-footer="editWithoutComment || awaitingResponse" required>
-      <button type="submit" class="btn btn-primary"
-        :disabled="awaitingResponse">
-        {{ $t('action.comment') }} <spinner :state="awaitingResponse"/>
-      </button>
-    </markdown-textarea>
+    <div class="form-group">
```
ktuite, line 17: update review state modal will probably also need a `form-group`

src/components/project/overview/description.vue
```diff
@@ -0,0 +1,107 @@
...
+
+</style>
+
+<i18n lang="json5">
+{
+  "en": {
+    // This is shown in place of an empty project description and serves as a
+    // guide for where and how to change the description.
+    "instructions": {
+      "full": "Add project notes, links, instructions and other resources to this space in {projectSettings}.",
```
ktuite, line 99: Capitalize "P" in "project"

matthew-white (approved):

PR #574, opened 2022-04-01
Home Page Project List

ktuite:
Displays a list of projects and forms nested within those projects to give ODK users quicker access to important parts of their workflows.

Screen Shot 2022-05-12 at 2 12 24 PM


src/components/home.vue
```diff
@@ -48,7 +48,7 @@ export default {
     fetchData() {
       this.$store.dispatch('get', [{
         key: 'projects',
-        url: '/v1/projects',
+        url: '/v1/projects-forms', // todo: should this be using apiPaths?
```
matthew-white, line 51: Just thought I'd comment on this todo, I usually don't use `apiPaths` if the URL is constant.

src/components/project/form-row.vue
```diff
@@ -0,0 +1,102 @@
+<!--
+Copyright 2017 ODK Central Developers
```
matthew-white, line 2: 2017 → 2022

src/components/project/form-row.vue
```diff
@@ -0,0 +1,102 @@
...
+    // The component assumes that this data will exist when the component is
+    // created.
+    ...requestData(['project']),
+    submissionsPath() {
+      return this.formPath(
+        this.form.projectId,
+        this.form.xmlFormId,
+        this.form.publishedAt != null ? 'submissions' : 'draft/testing'
+      );
+    }
```
matthew-white, line 84: The links for the review states will follow a similar pattern. For the review state of `null` (received), link to `submissions?reviewState=null`. For `edited`, `submissions?reviewState='edited'` (with the single quotes around `edited`). For `hasIssues`, `submissions?reviewState='hasIssues'`.

Not sure what to do if the form doesn't have a published version. The draft Testing page doesn't have filters, so I'm not sure it makes sense to link to a filtered view of that table. The counts will also be zero in that case, right?

matthew-white: Actually, I think those single quotes should be encoded, so (for example) `submissions?reviewState=%27edited%27`.

src/components/project/home-block.vue
```diff
@@ -0,0 +1,68 @@
...
+    },
+    maxForms: {
+      type: Number,
+      default: 3
+    }
+    // todo: default # of proj to show
+  },
+  computed: {
+    visibleForms() {
+      return this.project.nestedForms.slice(0, this.maxForms).map((f) => new Form(f));
```
matthew-white, line 47: Instead of wrapping the forms in a presenter here, I think it'd be better to do it when the response is received (in the store). That would involve changing the `projects` transform function from:

```js
({ data }) => data.map(project => new Project(project))
```

to something like:

```js
({ data }) => data.map(project => new Project({
  ...project,
  forms: forms.map(form => new Form(form))
}))
```

src/components/project/form-row.vue
```diff
@@ -0,0 +1,100 @@
...
+import routes from '../../mixins/routes';
+import { requestData } from '../../store/modules/request';
+
+export default {
+  name: 'ProjectFormRow',
+  components: { DateTime, LinkIfCan },
+  mixins: [routes()],
+  props: {
+    form: {
+      type: Form,
```
matthew-white, line 64: In anticipation of a change I'll be making soon, let's go ahead and make this `Object` rather than `Form`! Soon it's not going to be possible to import presenter classes or use them as types.

src/components/project/home-block.vue
```diff
@@ -0,0 +1,68 @@
...
+
+import ProjectFormRow from './form-row.vue';
+
+export default {
+  name: 'ProjectHomeBlock',
+  components: { ProjectFormRow },
+  mixins: [routes()],
+  props: {
+    project: {
+      type: Project,
```
matthew-white, line 36: Similarly, let's go ahead and change `Project` to `Object`.

matthew-white (commented):
I've been thinking about i18n and presenters and just thought I'd leave a few related comments!

src/components/project/form-row.vue
```diff
@@ -0,0 +1,111 @@
...
+<template>
+  <tr class="project-form-row">
+    <!--todo: come back and figure out all the permissions-->
+    <td class="name">
+      <link-if-can :to="primaryFormPath(form)">
+        {{ form.nameOrId() }}
+      </link-if-can>
+    </td>
+    <td class="review-state">
+      {{ $n(form.reviewState.received) }}<span class="icon-dot-circle-o"></span>
```
matthew-white, line 21: I usually specify the optional second argument of `$n()` (which would be `'default'` here), since that will use the cached `Intl.NumberFormat`.

src/components/project/form-row.vue
```diff
@@ -0,0 +1,111 @@
...
+    <td class="review-state">
+      {{ $n(form.reviewState.hasIssues) }}<span class="icon-comments"></span>
+    </td>
+    <td class="review-state">
+      {{ $n(form.reviewState.edited) }}<span class="icon-pencil"></span>
+    </td>
+    <td class="last-submission">
+      <div v-if="form.lastSubmission != null">
+        <router-link :to="submissionsPath">
+          <date-time :iso="form.lastSubmission" relative="past"/>
```
matthew-white, line 32: I think the `blank` prop could be specified here. (You can grab that from `ProjectRow`. I think that message will now be used in multiple components, so should probably be moved to `en.json5`.)

src/components/project/home-block.vue
```diff
@@ -0,0 +1,92 @@
...
+
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+<template>
+  <div class="project-home-block">
+    <h2 class="title">
```
matthew-white, line 14: We don't use a ton of `<h*>` elements (and it can be hard to do so in a semantic way correctly). Bootstrap has equivalent classes (for example, `.h2`), but we also don't use those a lot. It might end up being easiest to style these elements separately.

matthew-white (commented):
I know this is still a draft PR, but I thought I'd leave a few small comments before heading out.

matthew-white:
I'm thinking that I probably won't have the chance to merge #576 before I head out, so you might want to go ahead and rebase this PR on that branch in order to avoid merge conflicts.

matthew-white:
I think I just noticed a mistake in the existing `ProjectRow` component, which is that it should no longer have an `introduction` prop (as of [`bd2b8ae`](https://github.com/getodk/central-frontend/pull/561/commits/bd2b8ae3dfb259854c94ffbc04639c743b0567cf)). I don't think we need to change that, since presumably this PR will delete that component, but I thought I'd make a note of it.

src/components/project/home-block.vue
```diff
@@ -0,0 +1,140 @@
...
+
+import ProjectFormRow from './form-row.vue';
+
+export default {
+  name: 'ProjectHomeBlock',
+  components: { ProjectFormRow },
+  mixins: [routes()],
+  props: {
+    project: {
+      type: Object,
```
matthew-white, line 49: I know I just said to specify `Object` as the prop type instead of `Project`, but I actually figured out a way for us to continue importing presenter classes (in 359a8c650292b437c3095c2825a672a0ecfd6463). That means that it actually does work to specify `Project` as the prop type here.

src/components/project/list.vue
```diff
@@ -50,24 +56,71 @@ except according to the terms contained in the LICENSE file.
...
+    ProjectSort
+  },
   mixins: [modal(), routes()],
   data() {
     return {
       newProject: {
         state: false
-      }
+      },
+      sortMode: 'latest'
```
matthew-white, line 81: I think at one point we talked about making alphabetical the default so that the default matches what's shown today. (Just randomly noticed that while taking a look at the QA server.)

src/components/project/list.vue
```diff
@@ -80,12 +133,23 @@ export default {
...
+    font-weight: 500;
+  }
+}
+
 </style>
 
 <i18n lang="json5">
 {
   "en": {
+    "archived": "Archived Projects",
```
matthew-white, line 152: I usually add a comment when a message is used as a title, in case the translator needs to use title case. For example, "This is a title shown above a section of the page."

src/components/project/sort.vue
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+}
+
+</style>
+
+<i18n lang="json5">
+{
+  "en": {
+    "sort": "Sort",
```
matthew-white, line 60: I'm guessing this should go in `en.json5`, since it'll be reused in `FormList`.

src/components/project/sort.vue
```diff
@@ -0,0 +1,68 @@
...
+    "sortOptions": {
+      "alphabetical": "Alphabetical",
+      "latest": "Latest Submission",
+      "newest": "Newest"
```
matthew-white, line 64: It might be fine to reuse `header.lastSubmission` from `en.json5` for `latest`. Either way, I'm thinking this message should live in `en.json5`, since it'll be reused in `FormList`. For `alphabetical` and `newest` though, since those are adjectives that depending on the language may need to grammatically match the noun (project or form), I think they probably can't be reused.

matthew-white (commented):
i18n thoughts that crossed my mind

src/components/project/sort.vue
matthew-white: For that reason, `alphabetical` and `newest` would probably benefit from comments indicating what the corresponding noun is.

src/components/project/form-row.vue
```diff
@@ -0,0 +1,149 @@
...
+<template>
+  <tr class="project-form-row">
+    <!--todo: come back and figure out all the permissions-->
+    <td class="name">
+      <link-if-can :to="primaryFormPath(form)">
+        {{ form.nameOrId() }}
+      </link-if-can>
+    </td>
+    <template v-if="form.publishedAt != null">
+      <td class="review-state">
```
matthew-white, line 21:
```vue
<td v-for="reviewState of reviewState" :key="reviewState" class="review-state">
  <router-link :to="submissionsPath[reviewState]">
    {{ $n(form.reviewStates[reviewState], 'default') }}<span :class="reviewStateIcon(reviewState)"></span>
  </router-link>
</td>
```

You'll need to define a computed property to return `reviewStates` ([received, hasIssues, edited]).

src/components/project/home-block.vue
```diff
@@ -0,0 +1,141 @@
...
+</style>
+
+
+<i18n lang="json5">
+{
+  "en": {
+    "encrypted": "Encrypted",
+    "encryptionTip": "This Project uses managed encryption.",
+    "showMore": "Show {count} total",
+    "showFewer": "Show fewer of {count} total"
```
matthew-white, line 138: I thought of one other i18n thing. I think it'd be helpful to comment here what the object being counted is, noting that these messages have to do with projects specifically. I might be wrong, but I think the Japanese translator will need to choose a [Japanese counter word](https://en.wikipedia.org/wiki/Japanese_counter_word) to go with the number. (I see counter words being used in similar Japanese translations.)

src/components/project/home-block.vue
```diff
@@ -0,0 +1,141 @@
...
+
+  .project-form-table tr:nth-child(3n + 2) {background: #eee;}
+}
+</style>
+
+
+<i18n lang="json5">
+{
+  "en": {
+    "encrypted": "Encrypted",
```
matthew-white, line 135: I think a comment would be useful here as well, noting that this refers to projects.

src/components/project/list.vue
ktuite: After the tuesday meeting, we're going to try leaving it as `latest` and get feedback from issa.

src/components/project/home-block.vue
```diff
@@ -0,0 +1,146 @@
...
+        :title="$t('encryptionTip')">
+        <span class="icon-lock project-icon"></span>
+        {{ $t('encrypted') }}
+        </span>
+    </div>
+    <table v-if="visibleForms != null" class="project-form-table table">
+      <project-form-row v-for="form of visibleForms" :key="form.xmlFormId" :form="form"/>
+    </table>
+    <div v-if="showExpander">
+      <span class="expand-button" @click.prevent="toggleExpanded">
```
matthew-white, line 26: I think that using a `<span>` as a button here might not work well from an accessibility standpoint. It might help to specify `role="button"` (the [related page](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) on MDN has useful guidance in general), but even then, I don't know that a keyboard user would be able to navigate to the element. One strategy that I've used (and that comes up sometimes in Bootstrap) is to render the element as a link (specify `href="#" role="button"` or maybe even just `role="button"`). You could then use the `text-link` mixin. (I'm sure there are also other approaches.)

src/components/project/list.vue
```diff
@@ -81,12 +134,24 @@ export default {
 </script>
 
 <style lang="scss">
-#project-list table { table-layout: fixed; }
+@import '../../assets/scss/mixins';
+
+#archived-projects {
+  .project-title {
+    a { @include text-link; }
+    color: $color-action-foreground;
```
matthew-white, line 142: Similar question as above, do we need to specify `$color-action-foreground` here?

src/components/project/list.vue
```diff
@@ -19,28 +19,34 @@ except according to the terms contained in the LICENSE file.
...
           {{ $t('emptyTable') }}
         </p>
       </template>
     </page-section>
+    <page-section v-if="archivedProjects.length > 0">
+      <template #heading>
+        <span>{{ $t('archived') }}</span>
+      </template>
+      <template #body>
+        <div id="archived-projects">
```
matthew-white, line 41: Since it's an id, it'd be great to prefix it with the name of the component: `project-list-archived` maybe.

src/components/project/sort.vue
```diff
@@ -0,0 +1,66 @@
...
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+<template>
+  <div id="project-sort">
+    <form class="form-inline" @submit.prevent>
+      <span id="project-sort-label">{{ $t('action.sort') }}</span>
+      <label id="project-sort-modes" class="form-group">
```
matthew-white, line 16: It looks like `#project-sort-modes` isn't used anywhere?

test/util/http.js
```diff
@@ -51,7 +51,7 @@ component once the responses have been processed:
     .respondWithData(() => testData.extendedProjects.sorted())
     .respondWithData(() => testData.standardUsers.sorted())
     .afterResponses(component => {
-      component.findAllComponents(ProjectRow).length.should.equal(3);
+      component.findAllComponents(ProjectHomeBlock).length.should.equal(3);
```
matthew-white, line 54: Yay for keeping examples up-to-date!

test/data/forms.js
```diff
@@ -51,6 +51,7 @@ const forms = dataStore({
     createdBy = extendedUsers.size !== 0
       ? extendedUsers.first()
       : extendedUsers.createPast(1).last(),
+    reviewStates = undefined,
```
matthew-white, line 54: Just a thought, maybe should default to `{}` instead?

matthew-white (commented):
This is coming together really nicely! I'm leaving some comments async, but I know we'll take a closer look during our interactive review. (We can also chat about any of these questions during our interactive review.)

src/components/project/form-row.vue
```diff
@@ -12,31 +12,56 @@ except according to the terms contained in the LICENSE file.
 <template>
   <tr class="project-form-row">
     <!--todo: come back and figure out all the permissions-->
-    <td class="name">
-      <link-if-can :to="primaryFormPath(form)">
+    <td class="form-name">
+      <template v-if="canLinkToFormOverview">
+        <router-link :to="primaryFormPath(form)">{{ form.nameOrId() }}</router-link>
```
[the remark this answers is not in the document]
matthew-white: Oh wait, never mind, I see that you need this to decide whether to navigate to `FormOverview` or `FormDraftStatus`!

test/components/project/form-row.spec.js
```diff
@@ -0,0 +1,287 @@
...
+        link.props().to.should.equal('/projects/1/forms/a%20b/submissions');
+      });
+
+      it('submission count links to the right place', () => {
+        const link = mountComponent().get('.total-submissions a');
+        link.props().to.should.equal('/projects/1/forms/a%20b/submissions');
+      });
+    });
+
+    describe('Project Viewer', () => {
```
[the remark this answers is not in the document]
ktuite: ah yes, i meant for it to be about project managers

test/components/project/form-row.spec.js
```diff
@@ -0,0 +1,287 @@
...
+  container: { router: mockRouter('/projects/1') }
+});
+
+describe('ProjectFormRow', () => {
+  describe('form name', () => {
+    beforeEach(mockLogin);
+
+    it('renders the form name correctly', () => {
+      testData.extendedForms.createPast(1, { name: 'My Form', xmlFormId: 'f' });
+      const link = mountComponent().find('.form-name a');
```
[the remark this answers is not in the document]
ktuite: I'm struggling to use these to get a specific instance (using a CSS selector) of a specific kind of component. In this and similar tests, I'm looking for a `RouterLinkStub` but there are several in the component now. Will ask for more advice in interactive review!

test/components/project/home-block.spec.js
```diff
@@ -0,0 +1,109 @@
...
+
+  it('shows nothing when there are zero forms', () => {
+    testData.extendedProjects.createPast(1);
+    const table = mountComponent().get('.project-form-table');
+    table.isEmpty().should.be.true();
+  });
+
+  it('shows the correct number of forms if there are only a few', () => {
+    const project = testData.extendedProjects.createPast(1).last();
+    testData.extendedForms.createPast(3, { reviewStates: {} });
```
ktuite, line 48: dont need reviewStates thign anymore

test/components/project/home-block.spec.js
```diff
@@ -0,0 +1,109 @@
...
+    const rows = mountComponent().findAllComponents(FormRow);
+    // Test component's sort function defined above will sort by xmlFormId
+    rows.map((row) => row.props().form.name).should.eql(['Bravo', 'Charlie', 'Alpha']);
+  });
+
+  it('shows correctly sorted forms before and after cutoff forms', async () => {
+    const project = testData.extendedProjects.createPast(1).last();
+    testData.extendedForms.createPast(1, { name: 'aaa_z', xmlFormId: 'z' });
+    testData.extendedForms.createPast(1, { name: 'bbb_y', xmlFormId: 'y' });
+    testData.extendedForms.createPast(1, { name: 'ccc_w', xmlFormId: 'w', reviewStates: {} });
```
ktuite, line 97: why is there a lone reviewStates

test/components/project/list.spec.js
```diff
@@ -13,15 +16,123 @@ const mountComponent = () => mount(ProjectList, {
...
+
 describe('ProjectList', () => {
   beforeEach(mockLogin);
 
+  it('shows a message if there are no projects', () => {
+    // TODO: show something fancier when there are no projects and update test
+    mountComponent().get('.empty-table-message').should.be.visible();
+  });
+
   it('renders a row for each project', () => {
```
ktuite, line 41: we dont need this one anymore

matthew-white (approved):

ktuite (commented):
Interactive review with @matthew-white

PR #575, opened 2022-04-11
Account for changes to translation component

matthew-white:


ktuite (approved):
Looks good!

matthew-white:
FYI @ktuite, I did some refactoring of the `<i18n-t>` component in 3ae00e8386c7b0264daecbd4cc65a3d1750c7553:

- I think I was confused about what `parent` refers to in a functional component. It seems  to be different from `$parent` and (conveniently) actually seems to be the ancestor that we were searching for. So the ancestor search can be removed.
- Previously, the component wouldn't generate pluralized messages from the root Vue I18n instance after the locale was changed. (I don't think this actually comes up for us anywhere, but that case is fixed now.)

The component is a little simpler now, and I feel confident about these changes. Since this is a temporary component, I'm not sure you really need to review these changes, but of course feel free to take a look at that commit (or we can it review together later today).

PR #579, opened 2022-04-19
Adding emits options on the way to Vue 3

ktuite:
Cherry-pick of this commit from the massive Vue3 PR https://github.com/getodk/central-frontend/pull/526/commits/330c708a277498f501782eedf2842813988fe7e8

### DIFF
--- a/src/components/audit/filters/action.vue
+++ b/src/components/audit/filters/action.vue
@@ -34,7 +34,7 @@ export default {
       required: true
     }
   },
-  emits: ['input'],
+  emits: ['update:modelValue'],
   computed: {
     options() {
       return [
--- a/src/components/audit/filters/action.vue
+++ b/src/components/audit/filters/action.vue
@@ -34,6 +34,7 @@ export default {
       required: true
     }
   },
+  emits: ['input'],
   computed: {
     options() {
       return [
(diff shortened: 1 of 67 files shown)


src/components/project/form-access.vue
```diff
@@ -34,7 +34,7 @@ except according to the terms contained in the LICENSE file.
     <template v-if="dataExists">
       <project-form-access-table :changes-by-form="changesByForm"
         @update:state="updateState"
-        @update:field-key-access="updateFieldKeyAccess"
+        @update:fieldKeyAccess="updateFieldKeyAccess"
```
matthew-white, line 37: Actually, would you mind renaming this back to `@update:field-key-access` here? Just here in the template and in the template in `ProjectFormAccessTable`, but not anywhere else in the JS (the change works well in the JS). I think making this camel-case here was a mistake on my part: I ended up reverting it after the Vue 3 linter yelled at me.

src/components/audit/filters/action.vue
```diff
@@ -34,6 +34,7 @@ export default {
       required: true
     }
   },
+  emits: ['input'],
```
matthew-white, line 37: Could you change all instances of `'input'` in the `emits` option to the (somewhat verbose&hellip;) `'update:modelValue'` (which `v-model` uses in Vue 3)?

ktuite: done!

src/components/project/form-access.vue
ktuite: done!

matthew-white:
The link in the commit message doesn't seem to work anymore, but here's the latest link to the relevant section of the migration guide: https://v3-migration.vuejs.org/breaking-changes/emits-option.html

matthew-white (approved):

PR #580, opened 2022-04-21
Update link to Vue Test Utils

matthew-white:


matthew-white:
@ktuite and I reviewed this change in #526, so I think I'm going to go ahead and merge.

PR #599, opened 2022-05-20
Make changes to router as part of move to Vue 3

matthew-white:


matthew-white:
One helpful page about asynchronous navigation: https://www.vuemastery.com/blog/vue-router-4-route-params-not-available-on-created-setup/

PR #609, opened 2022-06-01
Dynamic number of forms and filtering out closed forms

ktuite:
Filters closed forms from project overview list.

Also dynamically changes the number of forms to show: e.g. if there are only a few forms total, show them all!

### DIFF
--- a/test/components/project/list.spec.js
+++ b/test/components/project/list.spec.js
@@ -210,7 +210,7 @@ describe('ProjectList', () => {
       blocks[2].findAllComponents(FormRow).length.should.equal(5);
     });
 
-    it('uses the samee form limit across home blocks', () => {
+    it('uses the same form limit across home blocks', () => {
       createProjectsWithForms(
         [
           { name: 'Project 1' },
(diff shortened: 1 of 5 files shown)


test/components/project/list.spec.js
```diff
@@ -158,8 +158,121 @@ describe('ProjectList', () => {
...
+          { name: 'Project 2' }
+        ],
+        [
+          [{}, {}, {}, { state: 'closed' }],
+          [{}, {}]
+        ]
+      );
+      const component = mountComponent();
+      const blocks = component.findAllComponents(ProjectHomeBlock);
+      blocks[0].props().maxForms.should.equal(4);
```
matthew-white, line 272: I think that even if the component didn't account for closed forms, `maxForms` would equal 4, because `shownForms` would still equal `totalForms`. Maybe a way to test this would be to add one more closed form to the first project.

test/components/project/list.spec.js
```diff
@@ -152,14 +152,144 @@ describe('ProjectList', () => {
...
+      );
+      const component = mountComponent();
+      component.findAllComponents(FormRow).length.should.equal(6);
+      const blocks = component.findAllComponents(ProjectHomeBlock);
+      blocks[0].props().maxForms.should.equal(4);
+      blocks[0].findAllComponents(FormRow).length.should.equal(4);
+      blocks[1].findAllComponents(FormRow).length.should.equal(2);
+    });
+
+    it('it renders correctly with a small number of forms but where one project is above the limit', () => {
```
ktuite, line 257: description fits better with test above?

matthew-white (approved):

ktuite (commented):
Interactive review with @matthew-white.
In the home block, consider adding a test that uses/tests the expander and filtering out closed forms together

PR #613, opened 2022-06-01
Changing styles based on feedback

ktuite:
1. Bolder project title on project list
2. More padding of form rows on project list
3. More discoverable review state count links on form list
4. Encrypted project badge changed slightly

### DIFF
--- a/src/components/project/home-block.vue
+++ b/src/components/project/home-block.vue
@@ -95,7 +95,7 @@ export default {
 
   .title {
     font-size: 24px;
-    font-weight: 600;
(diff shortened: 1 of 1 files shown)


src/components/project/home-block.vue
```diff
@@ -95,14 +95,16 @@ export default {
 
   .title {
     font-size: 24px;
-    font-weight: 500;
+    font-weight: 600;
```
ktuite, line 98: 'bold'

matthew-white (approved):

PR #614, opened 2022-06-03
Update translations before v1.5 release

matthew-white:


lognaturel (approved):
Scrubbed through modified file list and diff quickly and confirmed all changes are localization-related.

PR #616, opened 2022-06-20
Showing form IDs next to duplicated form names

ktuite:
Added form ID next to form name for duplicate forms on both the Project List and the Form List.

Screen Shot 2022-06-20 at 2 40 24 PM

Screen Shot 2022-06-20 at 2 41 10 PM

### DIFF
--- a/src/components/project/form-row.vue
+++ b/src/components/project/form-row.vue
@@ -92,7 +92,6 @@ import Form from '../../presenters/form';
 import Project from '../../presenters/project';
 import routes from '../../mixins/routes';
 
-import { requestData } from '../../store/modules/request';
 import useReviewState from '../../composables/review-state';
 import { enketoBasePath } from '../../util/util';
 
--- a/src/components/project/form-row.vue
+++ b/src/components/project/form-row.vue
@@ -26,6 +26,7 @@ except according to the terms contained in the LICENSE file.
           {{ form.nameOrId() }}
         </template>
       </template>
+      <span v-if="showIdForDuplicateName" class="duplicate-form-id">({{ form.xmlFormId }})</span>
     </td>
     <template v-if="form.publishedAt != null">
       <td v-for="reviewState of visibleReviewStates" :key="reviewState" class="review-state">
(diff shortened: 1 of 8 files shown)


test/components/project/home-block.spec.js
```diff
@@ -125,4 +125,19 @@ describe('ProjectHomeBlock', () => {
     const rows = block.findAllComponents(FormRow);
     rows.map((row) => row.props().form.name).should.eql(['a', 'c', 'd', 'e']);
   });
+
+  it('shows form ID in parentheses next to duplicated form names', () => {
+    const project = testData.extendedProjects.createPast(1).last();
+    testData.extendedForms.createPast(1, { name: 'Same Name', xmlFormId: 'a' });
+    testData.extendedForms.createPast(1, { name: 'Same Name', xmlFormId: 'b' });
```
yanokwa, line 132: Is this check case sensitive? If so, the test is fine. If not, it might be good to compare "Same Name" and "same name".

src/store/modules/request/keys.js
```diff
@@ -197,6 +197,16 @@ const dataGetters = {
...
     }
     return result;
+  },
+
+  // Returns a set containing just the form names that appear more than once
+  // in a project. Used on project overview to show form ID next to form name
+  // when form names are duplicated.
+  duplicateFormNames: ({ data: { forms, deletedForms } }) => {
+    const allForms = [...forms || [], ...deletedForms || []];
+    const formNames = allForms.map((form) => (form.nameOrId().toLowerCase()));
```
matthew-white, line 207: How about using `toLocaleLowerCase()`? I like that we're using `localeCompare()` to sort form names in src/util/sort.js.

src/components/project/form-row.vue
```diff
@@ -137,6 +138,10 @@ export default {
     enketoPath() {
       const encodedId = encodeURIComponent(this.form.enketoId);
       return `${enketoBasePath}/${encodedId}`;
+    },
+    showIdForDuplicateName() {
+      const formNames = this.project.formList.map((form) => (form.xmlFormId !== this.form.xmlFormId ? form.nameOrId() : null));
+      return formNames.includes(this.form.nameOrId());
```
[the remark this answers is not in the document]
ktuite: I'm using some Sets now, but I haven't added the bit to skip computation entirely if too many forms. Maybe I should!

test/components/project/home-block.spec.js
ktuite: Ahh good idea. It's case-insensitive now!

src/components/project/form-row.vue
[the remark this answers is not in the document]
matthew-white: My intuition is that if you're using Sets, that should be enough!

PR #617, opened 2022-06-27
Update text in a couple of components for v1.5.1

matthew-white:


src/components/submission/decrypt.vue
```diff
@@ -350,9 +350,9 @@ $actions-padding-left: $label-icon-max-width + $margin-right-icon;
     "field": {
       "splitSelectMultiples": "Split “select multiple” choices into columns",
       "removeGroupNames": "Remove group names",
-      "deletedFields": "Include fields not in the published version of the Form"
+      "deletedFields": "Include fields not in the published Form"
     },
-    "deletedFieldsHelp": "Use this option if you need to see data collected with previous versions of this Form.",
+    "deletedFieldsHelp": "Use this option if you need to see fields referenced in previous Form versions.",
```
lognaturel, line 355: What do you think of referenced -> only? This is a bit challenging to express precisely!

lognaturel:
Are you thinking you'll update English but not pull translations for now? A bad outcome would be for all of those strings to revert to English.

src/components/submission/decrypt.vue
matthew-white: I agree that it's challenging to express this precisely! Here I'm using the text that @yanokwa suggested in Slack (on June 7). I don't have strong feelings myself, but I personally think that either way would be reasonably clear.

lognaturel: Let's leave it. With "only" there are two readings: "just the fields in older versions, not the ones in the current version" or "fields that only existed in previous versions and now don't exist"

I don't really know what "referenced" means but I think you're right it's clear enough.

matthew-white:
> Are you thinking you'll update English but not pull translations for now? A bad outcome would be for all of those strings to revert to English.

I was just wondering about that. I agree that that would be a bad outcome. I think the two main options are:

1. Update the English, but don't update the translations for these components (or at least don't update them for `SubmissionDownloadDropdown`, since that text is quite prominent).
2. Wait for v2022.3 to update the English (or at least wait for `SubmissionDownloadDropdown`).

I don't love the idea of making manual deviations in these components' autogenerated `<i18n>` blocks, which are based on the strings pulled from Transifex. That might be a reason to consider (2). But it's also just a few messages, so I'm also fine with (1).

lognaturel:
I would probably go with 1.5: update English and don't update any other translations but if 1 isn't too painful, that's fine by me. I do think the long text on the export options modal is worth changing asap.

matthew-white:
There have been a number of translations contributed since v1.5.0 (several languages are now at 100%), so I'm currently leaning toward (1).

I just pushed a new commit whose tests actually pass. 😅 Would you mind reviewing/approving if you have time?

lognaturel (approved):

PR #619, opened 2022-07-01
Some ideas around making dev setup easier 

alxndrsn:
1. use node-foreman to build vue / watch for changes / start nginx
2. make `nginx.conf` less dependent on os-specific details / require less manual configuration
3. update README setup instructions to reflect 1 & 2

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,13 +43,7 @@ Follow these instructions to run ODK Central Frontend in development. For deploy
 
 First, run ODK Central Backend.
 
-Next, build ODK Central Frontend files for development by running `npm run dev`. The files will be outputted to `dist/`. As you update the source code, the files will be automatically rebuilt.
-
-Finally, run NGINX by changing the working directory to the root directory of the repository, then typing the following:
-
-```bash
-nginx -c "$PWD/nginx.conf" -p "$PWD"
-```
+Next, start nginx and build ODK Central Frontend files for development by running `npm run dev`. The files will be outputted to `dist/`. As you update the source code, the files will be automatically rebuilt.
 
 NGINX effectively places ODK Central Frontend and ODK Central Backend at the same origin, avoiding cross-origin requests. We specify `-p "$PWD"` so that relative paths in [`nginx.conf`](/nginx.conf) are relative to the repository root.
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -5347,6 +5347,18 @@
       "integrity": "sha1-gQaNKVqBQuwKxybG4iAMMPttXoA=",
       "dev": true
     },
+    "foreman": {
+      "version": "3.0.1",
+      "resolved": "https://registry.npmjs.org/foreman/-/foreman-3.0.1.tgz",
+      "integrity": "sha512-ek/qoM0vVKpxzkBUQN9k4Fs7l0XsHv4bqxuEW6oqIS4s0ouYKsQ19YjBzUJKTFRumFiSpUv7jySkrI6lfbhjlw==",
+      "dev": true,
+      "requires": {
+        "commander": "^2.15.1",
+        "http-proxy": "^1.17.0",
+        "mustache": "^2.2.1",
+        "shell-quote": "^1.6.1"
+      }
+    },
     "forever-agent": {
       "version": "0.6.1",
       "resolved": "https://registry.npmjs.org/forever-agent/-/forever-agent-0.6.1.tgz",
@@ -7968,6 +7980,12 @@
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/multicast-dns-service-types/-/multicast-dns-service-types-1.1.0.tgz",
       "integrity": "sha1-iZ8R2WhuXgXLkbNdXw5jt3PPyQE=",
+      "dev": true
+    },
+    "mustache": {
+      "version": "2.3.2",
+      "resolved": "https://registry.npmjs.org/mustache/-/mustache-2.3.2.tgz",
+      "integrity": "sha512-KpMNwdQsYz3O/SBS1qJ/o3sqUJ5wSb8gb0pul8CO0S56b9Y2ALm8zCfsjPXsqGFfoNBkDwZuZIAjhsZI03gYVQ==",
       "dev": true
     },
     "mute-stream": {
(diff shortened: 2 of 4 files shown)


nginx.conf
```diff
@@ -13,13 +13,104 @@
...
+      video/mpeg                            mpeg mpg;
+      video/quicktime                       mov;
+      video/webm                            webm;
+      video/x-flv                           flv;
+      video/x-m4v                           m4v;
+      video/x-mng                           mng;
+      video/x-ms-asf                        asx asf;
+      video/x-ms-wmv                        wmv;
+      video/x-msvideo                       avi;
+  }
```
alxndrsn, line 113: I doubt this is all necessary - it's just a copy of my local `mime.types`

nginx.conf
```diff
@@ -62,9 +153,7 @@ http {
     }
 
     location / {
-      # Specify the dist/ directory of the repository to `nginx -p` so that . is
-      # relative to dist/.
-      root .;
+      root ../dist;
```
alxndrsn, line 156: Now running from `tmp/`.  Could also run from the project root, and then explicitly put logs in `tmp/`.

package-lock.json
```diff
@@ -757,70 +757,6 @@
...
-          "resolved": "https://registry.npmjs.org/loader-utils/-/loader-utils-2.0.2.tgz",
-          "integrity": "sha512-TM57VeHptv569d/GKh6TAYdzKblwDNiumOdkFnejjD0XwTH87K90w3O7AiJRqdQoXygvi1VQTJTLGhJl7WqA7A==",
-          "dev": true,
-          "optional": true,
-          "requires": {
-            "big.js": "^5.2.2",
-            "emojis-list": "^3.0.0",
-            "json5": "^2.1.2"
-          }
-        },
```
alxndrsn, line 823: :shrug:

nginx.conf
```diff
@@ -13,13 +13,104 @@
 #
 # One difference between this file and the configuration for production is that
 # this configuration does not support request body decompression.
+daemon off;
+worker_processes 1;
+error_log stderr;
+pid ./tmp/nginx.pid;
```
matthew-white, line 19: Are these changes needed? I think it's really helpful to minimize the number of differences between this file and the nginx configuration in [prod](https://github.com/getodk/central/blob/master/files/nginx/odk.conf.template).

nginx.conf
```diff
@@ -13,13 +13,104 @@
...
 
 events {
 }
 http {
-  # This path may be different on your local computer depending on your OS and
-  # how you installed nginx.
-  include /usr/local/etc/nginx/mime.types;
+  access_log ./tmp/nginx-access.log;
```
matthew-white, line 24: Similar as above, is this addition needed?

nginx.conf
matthew-white: Sounds good. :+1: I like the idea of specifying these inline rather than including a file whose path changes based on the OS.

nginx.conf
matthew-white: Is `tmp/` just for storing logs? For better or for worse, I'm actually not sure that I've ever used the logs from this local nginx.

package-lock.json
matthew-white: Sorry, I'm hoping that soon we won't have to deal with this! These unrelated changes should go away if you remove `node_modules`, then run `npm install` again.

README.md
```diff
@@ -33,7 +33,7 @@ First, install Node.js 14+.
 
 Next, install dependencies by running `npm install`.
 
-Install NGINX. Depending on your operating system and how you install NGINX, you may need to change the absolute paths in the development [`nginx.conf`](/nginx.conf).
+Install NGINX. Depending on your operating system and how you install NGINX.
```
matthew-white, line 36: Maybe we can now remove the second sentence entirely?

matthew-white (commented):
Thank you, @alxndrsn! I'm definitely excited about changes that could make Frontend development easier.

> 1. use node-foreman to build vue / watch for changes / start nginx
> 2. make `nginx.conf` less dependent on os-specific details / require less manual configuration

(2) feels like a big win to me!

The value of (1) feels slightly less clear. Is the benefit there that the user doesn't need to start nginx manually (thereby running one command instead of two)? I think I could get behind that, especially if it's not a huge code change. I think it would be nice to minimize the changes to nginx.conf.

matthew-white:
Oh, and it'd be awesome if commit messages were a little more descriptive! Especially the commit message of the first commit, since if the PR is squash-merged, I think GitHub will use that message in the commit message it suggests for the squash-merged commit. You could also consider doing some of that squashing / commit cleanup locally, before pushing. Or if you've already pushed, you could do that locally, then force-push.

README.md
alxndrsn: :grin:  yes, it makes no sense now!

nginx.conf
alxndrsn: It stores the access log, and the nginx pid file.  I think ideally the pidfile wouldn't be necessary, and the logs would go to stdout...

alxndrsn: Ref pidfile, there isn't much in the nginx docs: https://nginx.org/en/docs/ngx_core_module.html#pid

This thread looks relevant, but doesn't suggest there's a reliable & convenient alternative: https://github.com/docker-library/httpd/issues/166

nginx.conf
alxndrsn: `daemon off`: this means that nginx will run as a foreground process, which seems a lot less surprising for frontend devs not familiar with nginx

`worker_processes 1`: this is the default, so can be removed

`error_log stderr`: i think this is both necessary and helpful - it removes one point of OS-specificity, and means that devs will see nginx errors in their terminal rather than have to tail the error log file

`pid ./tmp/nginx.pid`: again this removes another place where nginx may default to an absolute path which the current user does not have access to

nginx.conf
alxndrsn: Yes, similarly to `error_log`, this may otherwise default to an absolute path which a normal user may not have permissions for, and if they are using nginx for other purposes may pollute their logs.

alxndrsn:
@matthew-white I've separated the nginx config changes into their own PR at https://github.com/getodk/central-frontend/pull/621.

If that looks like it'll be merged, I can update this PR to be foreman-specific.

alxndrsn:
Ideas addressed in separate PRs.

PR #621, opened 2022-07-02
nginx.conf: remove OS-specific paths

alxndrsn:


nginx.conf
```diff
@@ -13,13 +13,103 @@
...
 
 events {
 }
 http {
-  # This path may be different on your local computer depending on your OS and
-  # how you installed nginx.
-  include /usr/local/etc/nginx/mime.types;
+  access_log ./tmp/nginx-access.log;
+
+  types {
```
alxndrsn, line 25: The second column in this list is file extensions.  In the frontend `dist` directory, I can only find the following: `html ico js svg ttf woff`

We might get away with reducing the mime declarations to only these 6 extensions.

ref: https://nginx.org/en/docs/http/ngx_http_core_module.html#types

matthew-white: I'd support limiting the size of this list. I also think it'd be nice if we rarely had to modify this list. What do you think about making this list those 6 extensions, then also adding extensions for a limited number of very common media types (whichever ones you think make the most sense)?

.gitignore
```diff
@@ -27,3 +27,6 @@ npm-debug.log*
 *.njsproj
 *.sln
 *.sw?
+
+/tmp/
```
matthew-white, line 31: What do you think about using `/nginx/` instead of `/tmp/`? We haven't needed a `tmp` directory for other reasons, and if this directory is only going to hold `nginx` files, maybe `/nginx/` would be more immediately understandable as a name. 🤔

alxndrsn: How about `.nginx` - i think the dot prefix would make it clearer it's a temporary directory

README.md
```diff
@@ -48,10 +48,10 @@ Next, build ODK Central Frontend files for development by running `npm run dev`.
 Finally, run NGINX by changing the working directory to the root directory of the repository, then typing the following:
 
 ```bash
-nginx -c "$PWD/nginx.conf" -p "$PWD/dist/"
+nginx -c nginx.conf -p $PWD
```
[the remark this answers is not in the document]
alxndrsn: No real point messing with it if it might be required by some nginx versions - i'll just re-instate.

nginx.conf
alxndrsn: Updated.

N.B. the following known-used types were missing:

* `ttf`
* `csv`

It might be worth checking the mime types that are currently being served for these in production.

alxndrsn: `ttf`: does not have it's own type - correctly served as `application/octet-stream`

alxndrsn: `csv`: this seems to be handled manually in `central-backend`: https://github.com/getodk/central-backend/blob/master/lib/resources/submissions.js#L307

matthew-white: Thanks for checking those! :+1:

matthew-white (approved):
Very awesome to no longer have OS-specific paths! Thanks, @alxndrsn!

nginx.conf
```diff
@@ -13,13 +13,34 @@
...
 
 events {
 }
 http {
-  # This path may be different on your local computer depending on your OS and
-  # how you installed nginx.
-  include /usr/local/etc/nginx/mime.types;
+  access_log ./.nginx/nginx-access.log;
+
+  types {
```
lognaturel, line 25: What are the implications of this? Like if I have a jpg or mov or m4v or any other file type that's not explicitly in this list specified as a form or submission attachment, what happens?

alxndrsn: My understanding is that the content-type header for attachments is set in backend code.

alxndrsn: https://github.com/getodk/central-backend/blob/master/lib/resources/submissions.js#L397 -> https://github.com/getodk/central-backend/blob/master/lib/util/http.js#L100

tests at: https://github.com/getodk/central-backend/blob/master/test/integration/api/submissions.js#L3517 etc

PR #622, opened 2022-07-04
nginx.conf: document dev-specific hacks

alxndrsn:
Moves `X-Forwarded-Proto` to top of `location` block, as per https://github.com/getodk/central/blob/master/files/nginx/odk.conf.template#L36.

### DIFF
--- a/main.nginx.conf
+++ b/main.nginx.conf
@@ -73,14 +73,27 @@ http {
       proxy_pass http://localhost:8383;
       proxy_redirect off;
 
-      include ./common-headers.nginx.conf;
-      add_header Set-Cookie $session_cookie;
-      proxy_set_header X-Forwarded-Proto https;
-
       # buffer requests, but not responses, so streaming out works.
       proxy_request_buffering on;
       proxy_buffering off;
       proxy_read_timeout 2m;
+
+      # Dev-specific hacks:
+
+      # In conjunction with the map{} definition above, remap
+      # "Set-Cookie: __Host-session=..." to "Set-Cookie: session=..."
+      #
+      #   1. Cookies cannot use the "__Host-" prefix in non-HTTPs requests
+      #      see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#cookie_prefixes
+      #   2. central-backend cookie parsing is relaxed and will consider the
+      #      first cookie ending in "session" to be the session cookie
+      add_header Set-Cookie $session_cookie;
+      # re-add common headers after add_header call
+      include ./common-headers.nginx.conf;
+
+      # Trick central-backend from thinking connections are coming
+      # over HTTPS so that ExpressJS will set "secure" cookies.
+      proxy_set_header X-Forwarded-Proto https;
     }
 
     location / {


matthew-white (commented):
In general, I definitely want the development nginx.conf to be as close as possible to what we have in production, including the order of lines. Here though, I think this line is doing something a little different from what happens in production. In production, the nginx.conf just sets `X-Forwarded-Proto` to `$scheme`. But here, we specifically set `X-Forwarded-Proto` to `https` so that cookie auth works in local development even when you're not using HTTPS. That's needed because Backend checks for HTTPS before trying to use cookie auth. We made this change to the development nginx.conf in #418, and it's definitely made local development smoother.

The other thing we do in this file to make cookie auth work in local development is to strip the `__Host-` prefix from the `__Host-session` cookie used for cookie auth. That happens when the cookie is received (using the `map` above). We then add our own cookie named `session` (on L55).

I've grouped those two lines together because they're oriented around the same goal of making cookie auth work in local development. I think it makes sense to keep them together, but writing this explanation now, I also think that it'd be helpful to add comments around why things are set up the way they are. (If you have any interest in that, please feel free!)

alxndrsn:
Thanks, great explanation.  I'll try to capture this in a simple comment to add around the relevant dev-specific config and update the PR

alxndrsn:
@matthew-white I've updated the PR with what I've understood from your last comment, but I'm not clear how renaming the `Host-session` cookie as simply `session` is helpful/relevant in devlopment :\

nginx.conf
```diff
@@ -52,13 +52,20 @@ http {
...
       # buffer requests, but not responses, so streaming out works.
       proxy_request_buffering on;
       proxy_buffering off;
       proxy_read_timeout 2m;
+
+      # Dev-specific hacks:
+
+      # Remap "Set-Cookie: __Host-session=..." to "Set-Cookie: session=..."
+      # TODO but why is this helpful?
+      add_header Set-Cookie $session_cookie;
```
matthew-white, line 64: The browser will only use a cookie prefixed with `__Host-` if it meets certain conditions, including that the response is from an HTTPS origin. Backend doesn't use HTTPS in local development, but we still want to use the cookie that it returns, so we (1) strip `__Host-` from the cookie, saving the result in `$session_cookie`, then (2) use `$session_cookie` for the cookie header when sending the next request. `map` on L24 above implements (1), then `add_header` here implements (2). I think that `map` can't be in `server`, which is why (1) and (2) are so far apart. This strategy works because Backend looks for a cookie whose name ends with `session`, not specifically for `__Host-session`: https://github.com/getodk/central-backend/blob/5aa8ff4a4e91a65574b19774e655a54c36999209/lib/http/preprocessors.js#L103

nginx.conf
```diff
@@ -52,13 +52,20 @@ http {
...
+
+      # Trick central-backend from thinking connections are coming
+      # over HTTPS so that ExpressJS will set "secure" cookies.
```
matthew-white, line 67: I'm not sure whether we need Express to set Secure cookies in local development, but if we didn't specify `X-Forwarded-Proto`, any attempt to use cookie auth would fail this Backend check: https://github.com/getodk/central-backend/blob/5aa8ff4a4e91a65574b19774e655a54c36999209/lib/http/preprocessors.js#L99

matthew-white (commented):
Thanks for adding these comments! I like the idea of putting these lines into a "dev-specific hacks" section. This whole thing does feel pretty hack-y, but it is quite handy to be able to use cookie auth in local development.

nginx.conf
alxndrsn: Aha!  I suspect that check may technically be redundant due to a combination of:

1. https://github.com/getodk/central-backend/blob/master/lib/resources/sessions.js#L42:

```js
response.cookie('__Host-session', session.token, { path: '/', expires: session.expiresAt,
  httpOnly: true, secure: true, sameSite: 'strict' });
```

2. https://expressjs.com/en/4x/api.html#res.cookie: 



Property | Type | Description
-- | -- | --
`secure` | Boolean | Marks the cookie to be used with HTTPS only.

3. https://expressjs.com/en/4x/api.html#req.protocol

> ### `req.protocol`
>
> Contains the request protocol string: either http or (for TLS requests) https.
>
> When the [trust proxy setting](https://expressjs.com/en/4x/api.html#trust.proxy.options.table) does not evaluate to false, this property will use the value of the X-Forwarded-Proto header field if present. This header can be set by the client or by the proxy.

nginx.conf
alxndrsn: Interesting!  That cookie parsing looks risky...

alxndrsn: @matthew-white i've updated the cookie comments :slightly_smiling_face:

nginx.conf
matthew-white: Interesting. It seems like (3) means that the two conditions in the check could be identical. And (1) and (2) mean that the check as a whole shouldn't be needed. My instinct is to leave Backend as-is for now and just document how all this is working, but we could also consider removing the check from Backend.

matthew-white (commented):
It looks like an empty file named .nginx/asg was committed as well.

alxndrsn:
I've removed the accidental file!

nginx.conf
matthew-white: What do you think about "Trick central-backend from thinking connections are coming over HTTPS so that it will use the session cookie." ? I believe that Chrome and Firefox are actually both happy to accept a Secure cookie over http://localhost (maybe Safari too these days, not sure).

matthew-white:
Just one last comment I think! https://github.com/getodk/central-frontend/pull/622#discussion_r926804512

nginx.conf
alxndrsn: With less assumed knowledge about the project, I prefer the current reading.

matthew-white (approved):
I think this is a helpful change. ✅ Thanks again, @alxndrsn. Unfortunately, there's now a conflict with #630, which I just merged. Sorry for the delay! @alxndrsn, feel free to merge this PR after resolving the conflict!

alxndrsn:
Conflicts resolved.

PR #623, opened 2022-07-08
Move select getters out of store

matthew-white:


lognaturel (approved):
I can't comment on how well this structure sets things up for the future. However, I can follow where things come from and where they go and I can't think of a source of risk.

PR #625, opened 2022-07-12
Pin node version for volta users

alxndrsn:


matthew-white (approved):

matthew-white:
Would it be convenient in any way to make this change in Backend as well? Or do we already have enough checks of the Node version there?

PR #627, opened 2022-07-13
Use foreman to manage dev processes

alxndrsn:
One less thing to think about for devs.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,2 @@
+vue-builder: vue-cli-service build --mode development --watch
+nginx: nginx -c "$PWD/nginx.conf" -p "$PWD"
--- a/README.md
+++ b/README.md
@@ -43,13 +43,7 @@ Follow these instructions to run ODK Central Frontend in development. For deploy
 
 First, run ODK Central Backend.
 
-Next, build ODK Central Frontend files for development by running `npm run dev`. The files will be outputted to `dist/`. As you update the source code, the files will be automatically rebuilt.
-
-Finally, run NGINX by changing the working directory to the root directory of the repository, then typing the following:
-
-```bash
-nginx -c "$PWD/nginx.conf" -p "$PWD"
-```
+Next, start nginx and build ODK Central Frontend files for development by running `npm run dev`. The files will be outputted to `dist/`. As you update the source code, the files will be automatically rebuilt.
 
 NGINX effectively places ODK Central Frontend and ODK Central Backend at the same origin, avoiding cross-origin requests. We specify `-p "$PWD"` so that relative paths in [`nginx.conf`](/nginx.conf) are relative to the repository root.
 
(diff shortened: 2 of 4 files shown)


Procfile
```diff
@@ -0,0 +1,2 @@
+vue-builder: vue-cli-service build --mode development --watch
```
matthew-white, line 1: What do you think about using a shorter name like `build` or `vue-build`?

README.md
```diff
@@ -43,13 +43,7 @@ Follow these instructions to run ODK Central Frontend in development. For deploy
...
 First, run ODK Central Backend.
 
-Next, build ODK Central Frontend files for development by running `npm run dev`. The files will be outputted to `dist/`. As you update the source code, the files will be automatically rebuilt.
-
-Finally, run NGINX by changing the working directory to the root directory of the repository, then typing the following:
-
-```bash
-nginx -c "$PWD/nginx.conf" -p "$PWD"
-```
+Next, start nginx and build ODK Central Frontend files for development by running `npm run dev`. The files will be outputted to `dist/`. As you update the source code, the files will be automatically rebuilt.
```
matthew-white, line 46: With the thought that not all frontend devs will know nginx, I would maybe explain that part a little more. But not overly highlight it either, since part of the beauty of Foreman is that the dev doesn't really need to understand nginx. How about:

> Next, build ODK Central Frontend files for development by running `npm run dev`. The files will be outputted to `dist/`. As you update the source code, the files will be automatically rebuilt. `npm run dev` will also start NGINX, which will serve the files.

README.md
```diff
@@ -43,13 +43,7 @@ Follow these instructions to run ODK Central Frontend in development. For deploy
...
 
 NGINX effectively places ODK Central Frontend and ODK Central Backend at the same origin, avoiding cross-origin requests. We specify `-p "$PWD"` so that relative paths in [`nginx.conf`](/nginx.conf) are relative to the repository root.
```
matthew-white, line 48: I'm not quite sure what to do with this paragraph. The second sentence now feels like it doesn't belong in the readme. Maybe it could move to nginx.conf or Procfile or just go away? The first sentence also might be unnecessary at this point. I feel like it's nice to have some explanation of why nginx is useful, but maybe "nginx will serve the files" is enough.

matthew-white (commented):
I still feel a little uncertain about this change. It adds another component to our workflow, and I'm sure we'll need to update the dependency over time, even if just for security. But on the other hand, it's rare that we'll ever have to modify Procfile, and the end result of this change is that a new dev or contributor will have fewer things to think about, because they won't have to manually start nginx. I like the idea of making setup easier than it's been in the past, so ultimately I think I come down on the side of making this change.

Procfile
alxndrsn: Fine with any option :slightly_smiling_face:   Maybe even just `vue`?

README.md
alxndrsn: LGTM :+1:

README.md
alxndrsn: :+1:

PR #631, opened 2022-08-16
Add CSP headers

alxndrsn:
Follows up: https://github.com/getodk/central/pull/308
Related to: https://github.com/getodk/central/pull/287


matthew-white:
I'm thinking we should close this PR:

- The first commit was merged via #630.
- We don't add a Content-Security-Policy header in production (the second commit). We add a Content-Security-Policy-Report-Only header in production, but I don't think we need that in development.
- The third commit is just a small white space change.

PR #632, opened 2022-09-12
improve: upgraded node version to 16

sadiqkhoja:
Set node version to 16

Related to https://github.com/getodk/central/pull/318


matthew-white (approved):
Thanks, @sadiqkhoja! 🎉

PR #635, opened 2022-09-21
Update translations for v1.5.3

matthew-white:
As part of this PR, I've changed one part of how we import translations from Transifex. Previously, if any element of an array was an untranslated message, we would discard the entire array. Now, we only discard the elements after the untranslated message. (See the code for an explanation of why we have to do this at all.) In general, this isn't a case that comes up a lot: usually an array is either fully translated or fully untranslated. However, right now it comes up a lot in the `BackupStatus` component because we just modified some of the source arrays there. Without making this change, a lot of that page would no longer be translated.

I've also discarded the changes to `SubmissionDownloadDropdown` as was done in #618.

### DIFF
--- a/bin/util/transifex.js
+++ b/bin/util/transifex.js
@@ -394,7 +394,7 @@ const _restructure = (
           logThenThrow(value, 'cannot link to a linked locale message');
         }
         if (value.full != null || Array.isArray(value)) {
-          // Supporting this case would add complexity to
+          // Supporting these cases would add complexity to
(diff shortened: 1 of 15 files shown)


lognaturel (commented):
In today's episode of "Nothing is Every Trivial!"

I did not run this but I did verify both that the code makes sense and that the resulting translation files are as I would expect them to be.

lognaturel (approved):

PR #636, opened 2022-09-25
Add first version of Datasets tab

matthew-white:
This PR adds a first version of the Datasets tab. I think we should try to merge this PR, then rebase #629 onto it.

This PR is intended as a starting point. Once this PR is merged, follow-up work can branch off `master`. Before then, follow-up work can branch off this branch.

This PR makes a couple of assumptions about how Backend will work that I think aren't currently true. Partly because of that, I haven't yet tried out this PR locally. (There are tests though!) Here are the assumptions that this PR makes:

**1. The extended-metadata response for the project returns the count of datasets in a property named `datasets`.**

This parallels what happens for forms and app users (counts are currently returned for those). Frontend needs to know the count of datasets because it will use that count to determine whether to show the Datasets tab. (If there are no datasets, the tab is not shown.) A different way we could approach this would be to request the dataset list in order to get the count that way. However, it seems better to avoid an extra request if possible. There would also be complications with sending that request because of Data Collectors. I think that Data Collectors won't have access to all datasets, so if we were to immediately request the dataset list (before the project response, when we know what the current user can do on the project), there would have to be some logic around handling a 403 response. Alternatively, we could send the request for the dataset list only after the response for the project, but that would require two round trips to the server before the page is rendered, which I think we should try to avoid.

**2. The admin, project manager, and project viewer roles include a verb named `dataset.list`. The Data Collector role does not.**

As mentioned above, my understanding is that Data Collectors won't have access to all datasets. That means that even if a project has one or more datasets, we will still need to hide the Datasets tab from Data Collectors. To do so, I think we will need to check for one or more verbs. If I recall correctly, we were thinking of adding a new set of dataset-specific verbs, so this PR uses a verb named `dataset.list`.


src/components/dataset/list.vue
```diff
@@ -0,0 +1,48 @@
...
+  created() {
+    this.fetchData();
+  },
+  methods: {
+    fetchData() {
+      this.$store.dispatch('get', [{
+        key: 'datasets',
+        url: apiPaths.datasets(this.projectId),
+        resend: false
+      }]).catch(noop);
```
matthew-white, line 44: For now, this doesn't request extended metadata. If we add a "latest entity" timestamp as extended metadata at some point, we can change this to request extended metadata.

test/data/datasets.js
```diff
@@ -0,0 +1,22 @@
+import faker from 'faker';
+import { comparator } from 'ramda';
+
+import { dataStore } from './data-store';
+import { extendedProjects } from './projects';
+
+// eslint-disable-next-line import/prefer-default-export
+export const extendedDatasets = dataStore({
```
matthew-white, line 8: I named this `extendedDatasets` for now with the thought that we might end up requesting extended metadata. If we don't end up doing so, we can rename this to `standardDatasets`.

test/router.spec.js
```diff
@@ -408,45 +419,57 @@ describe('createCentralRouter()', () => {
     describe('project viewer', () => {
       beforeEach(() => {
         mockLogin({ role: 'none' });
+      });
 
-        testData.extendedProjects.createPast(1, { role: 'manager', forms: 1 });
+      describe('ProjectSettings', () => {
```
matthew-white, line 424: I noticed that most of the tests under `describe('project routes', ...)` weren't quite right: they were navigating to /projects/1/&hellip;, but then using the data from the second project. The tests of `ProjectSettings` are the only ones that need two projects, so I've split that off into its own `describe()` block. Each `describe()` block will create its own test data. That way, there won't be a need to change the tests of the other project routes.

matthew-white: The tests under `describe('form routes', ...)` had a similar issue. That `describe()` block will now create its own test data.

test/router.spec.js
```diff
@@ -539,10 +574,11 @@ describe('createCentralRouter()', () => {
...
       beforeEach(() => {
         mockLogin({ role: 'none' });
-        testData.extendedProjects.createPast(1, { role: 'formfill', forms: 1 });
+        testData.extendedProjects.createPast(1, {
+          role: 'formfill',
+          datasets: 1
+        });
         testData.extendedForms.createPast(1);
-        testData.extendedFormVersions.createPast(1, { draft: true });
-        testData.standardFormAttachments.createPast(1);
```
matthew-white, line 545: In these tests, the only data that is returned is the data for /projects/1. That means that we don't need to create this form-level data.

matthew-white (commented):

sadiqkhoja (approved):
Did synchronous code review, ran this PR locally, everything looks good and can be merged

PR #640, opened 2022-10-21
Get hint about testing Datasets in alpha

sadiqkhoja:
# Changes:
- Added dataset testing hint on testing page of draft form

This is how it looks:
image

### DIFF
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -722,6 +722,7 @@ becomes more complicated.
 
     .panel-title {
       font-size: 16px;
+      font-weight: bold;
       line-height: inherit;
       color: $color-accent-primary;
     }
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -709,6 +709,27 @@ becomes more complicated.
   }
 }
 
+.panel-dialog {  
+  border: 2px solid $color-accent-primary;
+  border-radius: 5px;
+  margin-bottom: 10px;
+  padding: 5px;
+  background-color: #fff;
+  max-width: 77ch;
+
+  .panel-heading {    
+    border: none;
+
+    .panel-title {
+      font-size: 16px;
+      line-height: inherit;
+      color: $color-accent-primary;
+    }
+  }
+  .panel-body {
+    padding: 0 15px;
+  }
+}
 
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/src/components/form-draft/testing.vue
+++ b/src/components/form-draft/testing.vue
@@ -21,6 +21,25 @@ except according to the terms contained in the LICENSE file.
             </enketo-fill>
           </template>
           <template #body>
+            <div v-if="form && form.entityRelated" class="panel-dialog">
+              <div class="panel-heading">
+                <span class="panel-title">
+                  <span class="icon-database"></span>
+                  {{ $t('datasetsPreview.title') }}
+                </span>
+              </div>
+              <div class="panel-body">
+                <p>                  
+                  {{ $t('datasetsPreview.body[0]') }}
+                </p>
+                <p>
+                  {{ $t('datasetsPreview.body[1]') }}
+                </p>
+                <p>
+                  {{ $t('datasetsPreview.body[2]') }}
+                </p>
+              </div>
+            </div>
             <p>{{ $t('body[0]') }}</p>
             <p>
               <span>{{ $t('body[1]') }}</span>
@@ -90,7 +109,8 @@ export default {
     const { resourceView, createResource } = useRequestData();
     const formDraft = resourceView('formDraft', (data) => data.get());
     const keys = createResource('keys');
-    return { formDraft, keys };
+    const form = createResource('form');
+    return { formDraft, keys, form };
   },
   computed: {
     qrSettings() {
@@ -147,7 +167,15 @@ export default {
     ],
     // This text will be shown in ODK Collect when testing a Draft Form. {name}
     // is the title of the Draft Form.
-    "collectProjectName": "[Draft] {name}"
+    "collectProjectName": "[Draft] {name}",
+    "datasetsPreview": {
+      "title": "Datasets Preview",
+      "body": [
+        "This form can add entities to a dataset.",
+        "In this preview release of Datasets, you must publish your Form to see the Dataset update. In a future release of Central, you will be able to test Dataset functionality in a draft state.",
+        "For now, we recommend publishing your Dataset-related Forms, and not approving any submissions (and therefore not create any unwanted entities) until you are sure the Form is ready."
+      ]
+    }
   }
 }
 </i18n>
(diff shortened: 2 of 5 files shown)


src/components/form-draft/testing.vue
```diff
@@ -108,7 +127,7 @@ export default {
   computed: {
     // The component does not assume that this data will exist when the
     // component is created.
-    ...requestData([{ key: 'formDraft', getOption: true }, 'keys']),
+    ...requestData([{ key: 'formDraft', getOption: true }, 'keys', 'form']),
```
matthew-white, line 130: Components now access this data by using the `useRequestData()` composable.

src/components/form-draft/testing.vue
```diff
@@ -21,6 +21,25 @@ except according to the terms contained in the LICENSE file.
             </enketo-fill>
           </template>
           <template #body>
+            <div v-if="form && form.entityRelated" class="panel-dialog">
```
matthew-white, line 24: `form` will now be a resource object and will never be `null`. You can check whether the resource has data using `form.dataExists`.

matthew-white, line 24: Just wondering, shouldn't we check for `entityRelated` on the draft rather than the form? Even if a form doesn't have a published version, I think we want to show this warning if the draft form definition would create entities.

src/presenters/form.js
```diff
@@ -39,6 +39,7 @@ const props = [
   'lastSubmission',
   'reviewStates',
   'createdBy',
+  'entityRelated',
```
matthew-white, line 42: Presenters have been removed, so this list of properties no longer exists, so you won't need to add `entityRelated` to it.

matthew-white (commented):
I haven't taken a look at the code, and I believe that @ktuite is going to do the code review for this PR once it's ready. However, I wanted to add some comments now that #629 is merged.

src/components/form-draft/testing.vue
```diff
@@ -147,7 +166,15 @@ export default {
...
     // This text will be shown in ODK Collect when testing a Draft Form. {name}
     // is the title of the Draft Form.
-    "collectProjectName": "[Draft] {name}"
+    "collectProjectName": "[Draft] {name}",
+    "datasetsPreview": {
+      "title": "Datasets Preview",
+      "body": [
+        "This form can add entities to a dataset.",
+        "In this preview release of Datasets, you must publish your Form to see the Dataset update. In a future release of Central, you will be able to test Dataset functionality in a draft state.",
+        "For now, we recommend publishing your Dataset-related Forms, and not approving any submissions (and therefore not create any unwanted entities) until you are sure the Form is ready."
```
matthew-white, line 175: I think this string has changed. We often iterate on text after the screenshot/mock-up is created: the body of the release criteria has the most up-to-date text. (That's true in general: where it differs from the screenshot, it's the body/text of the release criteria that reflects the latest truth.)

src/components/form-draft/testing.vue
```diff
@@ -147,7 +166,15 @@ export default {
     ],
     // This text will be shown in ODK Collect when testing a Draft Form. {name}
     // is the title of the Draft Form.
-    "collectProjectName": "[Draft] {name}"
+    "collectProjectName": "[Draft] {name}",
+    "datasetsPreview": {
+      "title": "Datasets Preview",
+      "body": [
+        "This form can add entities to a dataset.",
+        "In this preview release of Datasets, you must publish your Form to see the Dataset update. In a future release of Central, you will be able to test Dataset functionality in a draft state.",
```
matthew-white, line 174: These two strings differ from the body of the release criteria just in capitalization I think.

matthew-white (commented):
Just taking a look at text/i18n.

src/components/form-draft/testing.vue
```diff
@@ -147,7 +166,15 @@ export default {
     ],
     // This text will be shown in ODK Collect when testing a Draft Form. {name}
     // is the title of the Draft Form.
-    "collectProjectName": "[Draft] {name}"
+    "collectProjectName": "[Draft] {name}",
+    "datasetsPreview": {
+      "title": "Datasets Preview",
```
matthew-white, line 171: One other thought related to i18n: particularly for strings that aren't a full sentence, it can be helpful to add context for translators about how the string is used. Here, I would add a comment that explains why "Preview" is capitalized, something like:

```json5
// This is a title shown above a section of the page.
```

Our Transifex workflow will automatically add comments like that to Transifex for translators to see there.

src/assets/scss/app.scss
```diff
@@ -709,6 +709,27 @@ becomes more complicated.
...
+  margin-bottom: 10px;
+  padding: 5px;
+  background-color: #fff;
+  max-width: 77ch;
+
+  .panel-heading {    
+    border: none;
+
+    .panel-title {
+      font-size: 16px;
```
ktuite, line 724: I think this text should also be bold, e.g.  add `font-weight: bold;`

src/components/form-draft/testing.vue
```diff
@@ -147,7 +166,15 @@ export default {
...
+      "body": [
+        "This form can add entities to a dataset.",
```
ktuite, line 173: Capitalize Form, Entities, and Dataset.

src/components/form-draft/testing.vue
ktuite: Capitalization looks good except for final "Draft".

ktuite (approved):
All it needs is more comments for Transifex translators about what the strings mean/how they are used and double-checking the capitalization!

The release criteria shows the Testing tab as being called "Testing & Data". Is that a change we are making? The string itself for that tab is [here](https://github.com/getodk/central-frontend/blob/master/src/locales/en.json5#L30).

matthew-white:
> The release criteria shows the Testing tab as being called "Testing & Data". Is that a change we are making?

I don't believe that's a change we're planning to make: that change isn't mentioned in the body of the release criteria. There are sometimes small differences like this one between screenshots/mock-ups in release criteria and how Central currently looks.

src/components/form-draft/testing.vue
sadiqkhoja: Recent text in release criteria is bit confusing, I have left a question there

sadiqkhoja: As discussed in the weekly meeting, we are keeping this text as it is and putting document link on the manage attachments (media) page

src/components/form-draft/testing.vue
```diff
@@ -147,7 +166,16 @@ export default {
...
     // is the title of the Draft Form.
-    "collectProjectName": "[Draft] {name}"
+    "collectProjectName": "[Draft] {name}",
+    "datasetsPreview": {
+      // This is a title shown above a section of the page.
+      "title": "Datasets Preview",
+      "body": [
+        "This Form can add Entities to a Dataset.",
+        "In this preview release of Datasets, you must publish your Form to see the Dataset update. In a future release of Central, you will be able to test Dataset functionality in a Draft state.",
+        "For now, we recommend publishing your Dataset-related Forms, and not approving any submissions (and therefore not create any unwanted entities) until you are sure the Form is ready."
```
sadiqkhoja, line 176: Capitalize "Submissions" and "Entities"

PR #641, opened 2022-11-04
Dataset summary and update to publish dialog

sadiqkhoja:
#Changes:
- Updated publish dialog box to show changes to datasets
- Added dataset summary widget on form overview page
- Added dataset summary widget on form draft status page

PR #642, opened 2022-11-08
new: link datasets to the attachments

sadiqkhoja:
#Changes:
- Autolinks attachment to the match dataset
- User can override linked attachment with an uploaded file
- User can restore dataset link


test/components/form/show.spec.js
```diff
@@ -76,7 +76,9 @@ describe('FormShow', () => {
     testData.extendedForms.createPast(1, { xmlFormId: 'f', draft: true });
     testData.standardFormAttachments.createPast(1);
     return load('/projects/1/forms/f/draft/attachments')
-      .beforeEachResponse(app => {
+      .beforeEachResponse((app, { url }) => {
+        // TODO need to find a way to show loading in show.vue when dataset is loading
```
sadiqkhoja, line 80: @matthew-white I would need your help here

matthew-white: Many components show their own loading message when they request an additional resource. For example, like `FormAttachmentList`, `PublicLinkList` is a child component of `FormShow`. `PublicLinkList` sends a request for `publicLinks`, which `FormShow` doesn't request, so it renders its own loading message just for `publicLinks`:

- `FormShow` requests `project`, `form`, `formDraft`, and `attachments`. Until it receives them, it shows a loading message.
- Even while `FormShow` is waiting for those resources, it renders the `<router-view>`. As a result, `PublicLinkList` is rendered right away. It sends its request for `publicLinks` and renders its loading message.
- Even though `PublicLinkList` is rendered, it won't be shown/visible until `FormShow` has received its resources.
- Then `PublicLinkList` will show its loading message until it has received `publicLinks`.
- If `PublicLinkList` receives `publicLinks` before `FormShow` has received its resources, then `PublicLinkList` will no longer show its loading message, but the user will continue to see the loading message in `FormShow`.

Unlike `PublicLinkList, `FormAttachmentList` historically hasn't requested any additional resource. However, now it will request `datasets`, so it will need to render a loading message. I would start by approaching it like this:

```diff
 <div v-show="attachments.dataExists" id="form-attachment-list" ref="dropZone">
   <!-- ^^^ We use v-show rather than v-if so that this.$refs.dropZone is in
   the DOM from the first render. -->
+  <loading :state="datasets.initiallyLoading"/>
+  <div v-show="datasets.dataExists">
     <div class="heading-with-button">
       <!-- ... -->
     </div>
     <table id="form-attachment-list-table" class="table">
       <!-- ... -->
     </table>
+  </div>
   <form-attachment-popups
```

(By the way, the `v-show` that's currently on `#form-attachment-list` is a historical artifact and could probably be removed. `FormShow` already specifies a `v-show` that will hide `FormAttachmentList` if `!attachments.dataExists`.)

I think it's important to think about loading messages, because there are Central users on slow connections. However, don't worry too much about testing them. Unless there's a fair amount of complexity involved, I don't usually test loading messages.

matthew-white: To patch this test, feel free to navigate to a different page or to create different test data (for example, a published form rather than a draft one).

src/components/form-attachment/restore-link.vue
```diff
@@ -0,0 +1,97 @@
...
+import Modal from '../modal.vue';
+import Spinner from '../spinner.vue';
+
+import request from '../../mixins/request';
+import { apiPaths } from '../../util/request';
+import { noop } from '../../util/util';
+import { useRequestData } from '../../request-data';
+
+export default {
+  name: 'RestoreLink',
```
matthew-white, line 44: I would use a longer, full name, `FormAttachmentRestoreLink`.

src/components/form-attachment/list.vue
```diff
@@ -357,6 +386,30 @@ export default {
...
+    showRestoreModal(name) {
+      this.restoreModal.attachmentName = name;
+      this.showModal('restoreModal');
+    },
+    afterRestore() {
+      this.alert.success(this.$t('alert.restore', {
+        attachmentName: this.restoreModal.attachmentName
+      }));
+
+      this.attachments.patch(() => {
```
matthew-white, line 405: Nice use of `patch()`! This will group these changes together within Vue Devtools.

src/components/form-attachment/row.vue
```diff
@@ -13,15 +13,28 @@ except according to the terms contained in the LICENSE file.
...
       <template v-else>
         {{ attachment.name }}
       </template>
     </td>
     <td class="form-attachment-list-uploaded">
-      <template v-if="attachment.exists">
+      <template v-if="attachment.datasetExists">
+        <span class="icon-link"></span>
+        <span class="dataset-label">
+          <i18n-t keypath="linkedToDataset">
```
matthew-white, line 27: By default, `<i18n-t>` doesn't render an element, but it's possible to make it do so, and that's something we do a lot. You'd effectively be specifying the `<span>` and the `<i18n-t>` together. Feel free to take that approach if it feels more ergonomic!

```vue
<i18n-t tag="span" keypath="linkedToDataset" class="dataset-label">
```

matthew-white (commented):
Mostly just took a look at i18n and at resources/requests. I didn't take a close look, but from what I can tell, it seems like the PR isn't far off from being ready to merge. You could consider trying to merge this one even before #641.

test/components/form/show.spec.js
matthew-white: I thought about this a little more, and I'm not sure it's essential that `FormAttachmentList` show a loading message during the request for `datasets`. `datasets` is only used to decide whether to render the "Restore Dataset Link" button, and I think it's fine if there's a delay between when the rest of the page appears and that button appears. In other words, I think we can show the page as soon as it's otherwise ready, then allow it to change slightly once `datasets` comes in (possibly showing the button). Just a thought — feel free to add the loading message if that seems better!

src/components/form-attachment/list.vue
```diff
@@ -135,15 +151,25 @@ export default {
...
   computed: {
     disabled() {
       return this.uploadStatus.total !== 0;
+    },
+    dsHashset() {
+      return this.datasets.dataExists ? new Set(this.datasets.map(d => `${d.name}.csv`)) : null;
     }
   },
+  created() {
+    this.fetchDatasets();
```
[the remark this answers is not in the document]
sadiqkhoja: `formDraft.entityRelated` means that form has dataset definition (creates / updates a dataset), so it is not necessary that entityRelated will be true for forms where user wants to link dataset to an attachment.

However, I can use `project.datasets > 0` to determine if I need to make fetchDatasets call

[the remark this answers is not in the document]
matthew-white: Ah good point. That sounds like a good approach, but also totally fine to come back to this later.

test/components/form/show.spec.js
sadiqkhoja: Thanks for the help, I elected not to show loading message if getDatasets is still loading in FormAttachments component

src/components/form-attachment/list.vue
```diff
@@ -62,27 +68,37 @@ import FormAttachmentNameMismatch from './name-mismatch.vue';
...
   inject: ['alert'],
+  props: {
+    projectId: {
+      type: String,
+      required: true
+    }
+  },
   setup() {
     const { form, resourceView } = useRequestData();
+    const { datasets } = useProject();
```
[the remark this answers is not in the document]
sadiqkhoja: thanks for explaining this, it was helpful

matthew-white:
@ktuite and I chatted yesterday, and we're thinking that it'd be the best use of our remaining time before QA for me to do the fuller review of this PR once it's ready. @sadiqkhoja, let me know when I should take another look!

matthew-white (approved):
I've left a number of comments, but they're mostly small changes. Overall, the PR is looking great. ✨ Information is communicated between components effectively, resources are managed well, and I like how you patched `testData`.

Since we're getting close to QA, I think it's OK if you don't make all these changes right away. I think it'd be nice if the changes were addressed before release, but I don't think they're necessarily the highest priority right now. There are some changes that I think are more important to make before merging, including some related to i18n/translations, and I've marked those as [Before merge]. There are also things that I wanted to put on your radar but that really don't need to be changed, and I've marked those as [Just FYI]. Feel free to merge after addressing the [Before merge] comments! We could file an issue about addressing any remaining comments.

I'm remembering that we were thinking about adding some text to the top of the page. (Maybe just if one or more of the attachments are linkable?) That could happen in a follow-up PR, but I wanted to ask about it.

I also thought that there were places where there could be more tests. Again, that's something that could come later. Specifically:

- If an attachment is linked to a dataset, its name (in the Name column) should be just text, not a link.
- The link in `.dataset-label` should have the correct `href` attribute.
- Clicking the "Restore Dataset Link" button should show the modal. (Check out `.testModalToggles()`.)
- Clicking the Restore button in the modal should send the correct request. (Check out `.testRequests()`.)
- The Restore button should do [standard button things](https://github.com/getodk/central-frontend/blob/ca7103799f0810682b430ec875591d2b02585eab/CONTRIBUTING.md#standard-button-things). (Check out `.testStandardButton()`.)
- A success alert should be shown after a dataset link is restored.
- If a dataset link is overridden with a file, the row should render correctly. You already test the existence of the Override label before the attachment is overridden, but because overriding modifies the attachment's `datasetExists` property, it'd be good to test the end result of overriding as well.

src/components/form-attachment/list.vue
```diff
@@ -28,6 +28,7 @@ except according to the terms contained in the LICENSE file.
           <th class="form-attachment-list-type">{{ $t('header.type') }}</th>
           <th class="form-attachment-list-name">{{ $t('header.name') }}</th>
           <th class="form-attachment-list-uploaded">{{ $t('header.uploaded') }}</th>
+          <th class="form-attachment-list-action"></th>
```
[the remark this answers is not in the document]
sadiqkhoja: I have added a wrapper div in form-attachment-list-action with the width of 200px, also changed styling of uploaded column so that it doesn't grow infinitely

src/components/form-attachment/link-dataset.vue
```diff
@@ -0,0 +1,104 @@
...
+-->
+<template>
+  <modal id="form-attachment-link-dataset" :state="state" :hideable="!awaitingResponse"
+    backdrop @hide="$emit('hide')">
+    <template #title>{{ $t('title') }}</template>
+    <template #body>
+      <div class="modal-introduction">
+          <p>
+            {{ $t('introduction[0]') }}
+            <template v-if="blobExists"> {{ $t('introduction[1]') }}</template>
```
matthew-white, line 20: This is a great case for `<sentence-separator>`. That'd be something like:

```vue
<span>{{ $t('introduction[0]') }}</span>
<sentence-separator/>
<template v-if="blobExists">{{ $t('introduction[1]') }}</template>
```

That way, the only way a space character comes in is via `<sentence-separator>`.

matthew-white (commented):
I took a look at what's changed ([`a40eec1..96c0b6b`](https://github.com/getodk/central-frontend/compare/a40eec1232e0c04391c682debd88838978c6a343..96c0b6b158416bea0b7036d2940960f0f5a5b307`)), and the changes there are looking great.

src/components/form-attachment/row.vue
```diff
@@ -34,17 +45,31 @@ except according to the terms contained in the LICENSE file.
...
+    <td class="form-attachment-list-action text-right">
+      <template v-if="attachment.datasetExists">
+        {{ $t('uploadToOverride') }}
+      </template>
+      <template v-else-if="linkable && !attachment.datasetExists">
+        <button type="button" class="btn btn-primary btn-restore" @click="$emit('restore', {
+          name: attachment.name,
+          action: attachment.blobExists ? 'restore' : 'link'
+        })">
+          <span class="icon-link"></span>{{ attachment.blobExists ? $t('restoreDatasetLink') : $t('linkDataset') }}
```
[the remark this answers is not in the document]
matthew-white: I like how the button and modal have turned out! I think we should run with what you have.

PR #643, opened 2022-11-09
Dataset and Entity audit log events

ktuite:
This PR shows new Dataset and Entity audit log events in two places:

### In the Submission activity feed
When a submission is approved and turns into an entity successfully, the `entity.create` is shown with some brief info about the entity.
Screen Shot 2022-11-09 at 10 58 25 AM

If the submission should become an entity but has a validation error, the `entity.create.error` is shown with the problem reason visible as a tool tip.
Screen Shot 2022-11-09 at 10 59 12 AM

### In the system Audit Log
Dataset events are shown in the audit log. Like submission events, entity events are filtered out in the API and NOT shown in the system audit log because they might drown out other events.
Screen Shot 2022-11-09 at 1 43 03 PM

TODO:
- [ ] Update the icon to the 'wand-magic-sparkles'
- [ ] Finalize what to show when there is an error creating an entity that is due to a validation problem that the form designer should know about and fix

### DIFF
--- a/src/components/submission/feed-entry.vue
+++ b/src/components/submission/feed-entry.vue
@@ -31,6 +31,17 @@ except according to the terms contained in the LICENSE file.
             <template #name><actor-link :actor="entry.actor"/></template>
           </i18n-t>
         </template>
+        <template v-else-if="entry.action === 'entity.create'">
+          <span class="icon-check entity-icon submission-feed-entry-icon"></span>
+          <i18n-t keypath="title.entity.create">
+            <template #label><span class="submission-feed-entry-entity-data">{{ entityLabel(entry) }}</span></template>
+            <template #dataset><span class="submission-feed-entry-entity-data">{{ entityDataset(entry) }}</span></template>
+          </i18n-t>
+        </template>
+        <template v-else-if="entry.action === 'entity.create.error'">
+          <span class="icon-warning entity-icon submission-feed-entry-icon"></span>
+          <span class="submission-feed-entry-entity-error" :title="entityProblem(entry)">{{ $t('title.entity.error') }}</span>
+        </template>
         <template v-else>
           <span class="icon-comment submission-feed-entry-icon"></span>
           <i18n-t keypath="title.comment">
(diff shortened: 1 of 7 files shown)


sadiqkhoja (approved):
looks good to me

src/components/submission/feed-entry.vue
```diff
@@ -173,9 +200,19 @@ export default {
       text-align: center;
       width: 18px;
     }
+
+    .submission-feed-entry-entity-data {
+      font-weight: normal;
+      color: $color-action-foreground;
```
matthew-white, line 206: I think this color in the release criteria is aspirational linking. Since things won't link yet, probably no need to color it like a link yet.

src/components/submission/feed-entry.vue
```diff
@@ -31,6 +31,17 @@ except according to the terms contained in the LICENSE file.
...
+        <template v-else-if="entry.action === 'entity.create'">
+          <span class="icon-check entity-icon submission-feed-entry-icon"></span>
+          <i18n-t keypath="title.entity.create">
+            <template #label><span class="submission-feed-entry-entity-data">{{ entityLabel(entry) }}</span></template>
+            <template #dataset><span class="submission-feed-entry-entity-data">{{ entityDataset(entry) }}</span></template>
+          </i18n-t>
+        </template>
+        <template v-else-if="entry.action === 'entity.create.error'">
+          <span class="icon-warning entity-icon submission-feed-entry-icon"></span>
+          <span class="submission-feed-entry-entity-error" :title="entityProblem(entry)">{{ $t('title.entity.error') }}</span>
```
matthew-white, line 43: It might be nice to make the `entry.details.problem.problemDetails.reason` more visible at some point, maybe by putting it in the `.body` below. But I think putting it in the `title` is a good place to start.

src/components/submission/feed-entry.vue
```diff
@@ -135,13 +146,29 @@ export default {
...
       return `${this.reviewStateIcon(state)} submission-feed-entry-icon`;
+    },
+    entityLabel(entry) {
+      if ('entity' in entry.details)
+        return entry.details.entity.label;
+      return '';
+    },
+    entityDataset(entry) {
+      if ('entity' in entry.details)
+        return entry.details.entity.dataset;
```
matthew-white, line 157: This is also available on the actee, right? `entry.actee.name`?

matthew-white (approved):
Looks great! I think we should go ahead and get this merged and not wait for the wand icon.

src/components/submission/feed-entry.vue
ktuite: It's not actually available here because the submission-specific audit endpoint `v1/projects/1/forms/<form>/submissions/<uuid>/audits` doesn't currently expand `acteeId` to `actee` like the general audit endpoint `v1/audits` does.

src/components/submission/feed-entry.vue
ktuite: Linking to this backend issue about what to do when submission-to-entity processing fails: https://github.com/getodk/central-backend/issues/661

PR #645, opened 2022-11-15
Make two submission filters multiselect

### LINKED ISSUE #396: Show warning if user selects 100+ columns
Above the submissions table, we show a dropdown that allows the user to select the columns to display in the table. We only allow the user to select 100 columns at a time, because anything above that takes some time to render. For example, on my computer it takes about 3 seconds to render 100 columns for 250 submissions.

However, instead of disabling additional selection after 100 columns, it would be better to simply warn the user about the potential slowdown. That way, the user can select additional columns if they want to while knowing what to expect. Removing the logic in the `SubmissionFieldDropdown` component around disabling should also simplify that component.

If the user selects 100+ columns, they should see a warning beneath the list of fields. The warning should show an `.icon-exclamation-triangle` as we do elsewhere, alongside the text "Selecting too many columns might slow down your computer."

### LINKED ISSUE #559: Make review state filter a multiselect
Right now the review state filter on the submissions page is a single-select. However, we want it to be a multiselect. We tried using a native `<select multiple>`, but that made for a confusing user experience. One possibility that we're considering is a dropdown like what we have for `SubmissionFieldDropdown`. (Could we genericize that selection dropdown?) As an alternative, maybe there's a nice external component out there that we could use?

This functionality was also requested on the forum: https://forum.getodk.org/t/multiple-filter-odk-central/36905

matthew-white:
Closes #559 and closes #396.


matthew-white:
Submitter and review state filters:

- All options are selected.
- Select none.
- Close dropdown.
- "None" falls back to "all". ✅
- But if you open the dropdown again, it's still the case that none are selected.

matthew-white (commented):

src/components/multiselect.vue
```diff
@@ -0,0 +1,479 @@
...
+    ? `${titleToLowerCase}${textToLowerCase}`
+    : textToLowerCase;
+}));
+// We may need to add debouncing here at some point.
+watch(searchValue, (value) => {
+  searchMatches.clear();
+  if (value === '') return;
+  const searchToLowerCase = value.toLocaleLowerCase(i18n.locale);
+  for (let i = 0; i < props.options.length; i += 1) {
+    if (textToSearch.value[i].includes(searchToLowerCase))
```
matthew-white, line 270:
```js
textToSearch.value[i][0].includes(searchToLowerCase) || textToSearch.value[i][1].includes(searchToLowerCase)
```

matthew-white (commented):
Interactive code review with @ktuite

ktuite (approved):
Just finished an interactive review with Matt - will review tests after thanksgiving!

matthew-white:
@ktuite and I just reviewed 793aae163c57475d8f48396cd203aac8ef7e996a interactively. @ktuite will take a look at tests asynchronously.

matthew-white:
I've noticed an issue here, which I've filed at #664. I think we can come back to that later.

PR #651, opened 2022-11-16
Add text about datasets/entities

matthew-white:
This PR adds text about datasets/entities to three places:

- The top of the Datasets page
- The "Enable Encryption" modal
- The "Delete Form" modal

The top of the Datasets page still needs some work, but I think that can come after this PR. (I'm eager to get this text to translators.) Most importantly, we need to add the image there. I believe this will be our first image in Central (!). Given that, I think we should wait to add that until after I either upgrade Vue CLI or switch us to Vite, since that will affect how the image is imported.

I've written tests for the change to the "Delete Form" modal, since that text is shown only conditionally. I didn't write tests for the other text, since there's nothing dynamic there. We don't usually test static text.

I think this PR can be reviewed asynchronously.

Screenshots:

Screen Shot 2022-11-16 at 3 38 04 PM
Screen Shot 2022-11-16 at 3 35 12 PM
Screen Shot 2022-11-16 at 3 36 12 PM

### DIFF
--- a/src/components/dataset/introduction.vue
+++ b/src/components/dataset/introduction.vue
@@ -1,0 +1,79 @@
+<!--
+Copyright 2022 ODK Central Developers
+See the NOTICE file at the top-level directory of this distribution and at
+https://github.com/getodk/central-frontend/blob/master/NOTICE.
+
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+<template>
+  <div id="dataset-introduction" class="panel-dialog">
+    <div class="panel-heading">
+      <span class="panel-title">{{ $t('title') }}</span>
+    </div>
+    <div class="panel-body">
+      <p>{{ $t('body[0]') }}</p>
+      <i18n-t tag="p" keypath="body[1]">
+        <template #icon1><span class="step-number">(1)</span></template>
+      </i18n-t>
+      <i18n-t tag="p" keypath="body[2]">
+        <template #icon2><span class="step-number">(2)</span></template>
+        <template #icon3><span class="step-number">(3)</span></template>
+      </i18n-t>
+      <i18n-t tag="p" keypath="body[3]">
+        <template #icon4><span class="step-number">(4)</span></template>
+      </i18n-t>
+      <p>
+        <span>{{ $t('body[4]') }}</span>
+        <sentence-separator/>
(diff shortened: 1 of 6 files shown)


src/components/dataset/introduction.vue
```diff
@@ -0,0 +1,79 @@
...
+      <i18n-t tag="p" keypath="body[3]">
+        <template #icon4><span class="step-number">(4)</span></template>
+      </i18n-t>
+      <p>
+        <span>{{ $t('body[4]') }}</span>
+        <sentence-separator/>
+        <i18n-t keypath="getStarted.full">
+          <template #clickHere>
+            <!-- TODO. Specify the `to` prop. -->
+            <doc-link>{{ $t('getStarted.clickHere') }}</doc-link>
```
matthew-white, line 35: I've filed an issue for this at #652.

src/components/dataset/introduction.vue
```diff
@@ -0,0 +1,79 @@
...
+-->
+<template>
+  <div id="dataset-introduction" class="panel-dialog">
+    <div class="panel-heading">
+      <span class="panel-title">{{ $t('title') }}</span>
+    </div>
+    <div class="panel-body">
+      <p>{{ $t('body[0]') }}</p>
+      <i18n-t tag="p" keypath="body[1]">
+        <template #icon1><span class="step-number">(1)</span></template>
```
matthew-white, line 20: The number and parentheses are intentionally not internationalized. See the [discussion](https://docs.google.com/document/d/1naxyzFU8EjcWKfYzE_mEF5cY38l1QRNUZ_0VW2Gp-r0/edit?disco=AAAAjSbSdz0) in the release criteria.

src/components/form/delete.vue
```diff
@@ -21,6 +21,9 @@ except according to the terms contained in the LICENSE file.
           </template>
         </i18n-t>
         <p>{{ $t('introduction[1]') }}</p>
+        <p v-if="form.dataExists && form.entityRelated">
+          {{ $t('noDeleteEntities') }}
+        </p>
```
matthew-white, line 26: I initially made this the second paragraph, but after looking at it, I think it actually works better as the last paragraph.

src/components/dataset/introduction.vue
```diff
@@ -0,0 +1,79 @@
...
+  .step-number { font-style: italic; }
+}
+</style>
+
+<i18n lang="json5">
+{
+  "en": {
+    "title": "Welcome to the Datasets preview!",
+    "body": [
+      "Datasets are a new way to create and manage data attachments in ODK. They allow you to use data gathered in Submissions to be downloaded and used as part of another Form.",
```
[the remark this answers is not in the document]
lognaturel: We also don't say what it means that this is a "preview". I think it's ok and we can highlight that in the longer docs but did want to call it out.

PR #655, opened 2022-11-18
Don't allow property names to contain spaces in testing

matthew-white:

### DIFF
--- a/test/data/form-dataset-diff.js
+++ b/test/data/form-dataset-diff.js
@@ -9,15 +9,15 @@ export const formDatasetDiffs = dataStore({
   factory: ({
     properties
   }) => ({
-    name: faker.hacker.noun(),
+    name: faker.random.alphaNumeric(10),
     properties: properties.map(p => {
       switch (p) {
         case Property.InFormProperty:
-          return { name: faker.hacker.noun(), inForm: true };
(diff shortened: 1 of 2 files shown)


test/data/form-dataset-diff.js
```diff
@@ -9,15 +9,15 @@ export const formDatasetDiffs = dataStore({
   factory: ({
     properties
   }) => ({
-    name: faker.hacker.noun(),
+    name: faker.random.alphaNumeric(10),
     properties: properties.map(p => {
       switch (p) {
         case Property.InFormProperty:
-          return { name: faker.hacker.noun(), inForm: true };
+          return { name: faker.random.alphaNumeric(10), inForm: true };
```
matthew-white, line 16: I first tried `faker.internet.userName()`, but the test still failed because that can return a string that contains `.`. I then tried `faker.random.alpha()`, but the version of Faker that we're on doesn't include `alpha()`. I landed on `alphaNumeric()` because it seems pretty close and because it allows the test to consistently pass.

matthew-white:
OK, I went through a few versions of this (should have run things locally first!), but this should now be ready for review.

sadiqkhoja (approved):
oh, I didn't know faker can return two words for hacker.noun - strange :(

PR #656, opened 2022-11-22
new: Added Dataset preview hint box on attachment page

sadiqkhoja:
## Changes:

- Added a hint box for the "Dataset Preview" on the attachment page. The hint box is displayed only if there is a linkable Dataset in the Project
- Created integration tests to verify that the hint box is displayed only when there is a linkable Dataset.

image

### DIFF
--- a/src/components/form-attachment/list.vue
+++ b/src/components/form-attachment/list.vue
@@ -23,16 +23,20 @@ except according to the terms contained in the LICENSE file.
       <div class="panel-heading">
         <span class="panel-title">
           <span class="icon-database"></span>
-          {{ $t('datasetsPreview.title') }}
+          {{ $t('common.datasetsPreview') }}
         </span>
       </div>
       <div class="panel-body">
-        <i18n-t tag="p" keypath="datasetsPreview.body.full">
-          <template #documentation>
-            <!-- TODO. Specify the `to` prop. -->
-            <doc-link>{{ $t('datasetsPreview.body.documentation') }}</doc-link>
-          </template>
-        </i18n-t>
+        <p>
+          <span>{{ $t('datasetsPreview.body[0]') }}</span>
+          <sentence-separator/>
+          <i18n-t keypath="moreInfo.clickHere.full">
+            <template #clickHere>
+              <!-- TODO. Specify the `to` prop. -->
+              <doc-link>{{ $t('moreInfo.clickHere.clickHere') }}</doc-link>
+            </template>
+          </i18n-t>
+        </p>
       </div>
     </div>
     <table id="form-attachment-list-table" class="table">
@@ -51,7 +55,7 @@ except according to the terms contained in the LICENSE file.
           :dragover-attachment="dragoverAttachment"
           :planned-uploads="plannedUploads"
           :updated-attachments="updatedAttachments" :data-name="attachment.name"
-          :linkable="!!dsHashset && dsHashset.has(attachment.name)"
+          :linkable="attachment.type === 'file' && !!dsHashset && dsHashset.has(attachment.name.replace(/\.[^.]+$/i,''))"
           @link="showLinkDatasetModal($event)"/>
       </tbody>
     </table>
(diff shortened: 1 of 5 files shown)


src/components/form-attachment/list.vue
```diff
@@ -482,6 +505,13 @@ export default {
       "readError": "Something went wrong while reading “{filename}”.",
       "success": "{count} file has been successfully uploaded. | {count} files have been successfully uploaded.",
       "link": "Dataset linked successfully."
+    },
+    "datasetsPreview": {
+      // This is a title shown above a section of the page.
+      "title": "Datasets Preview",
```
matthew-white, line 511: To avoid this being translated twice (here and in `FormDraftTesting`), could you move this message to en.json5 and remove it from `FormDraftTesting`? Maybe it could be `common.datasetsPreview`.

src/components/form-attachment/list.vue
```diff
@@ -420,6 +439,10 @@ export default {
   // Extend to the bottom of the page (or slightly beyond it) so that the drop
   // zone includes the entire page below the PageHead.
   min-height: calc(100vh - 146px);
+
+  .panel-dialog {
+    margin-bottom: 20px;
```
matthew-white, line 444: Right now, the `.panel-dialog` is closer to the table than to the text above it, when my instinct is that it should be the other way around. What do you think about something like:

```scss
margin-top: -5px;
margin-bottom: 25px;
```

src/components/form-attachment/list.vue
```diff
@@ -160,6 +174,11 @@ export default {
     },
     dsHashset() {
       return this.datasets.dataExists ? new Set(this.datasets.map(d => `${d.name}.csv`)) : null;
+    },
+    datasetLinkable() {
+      return this.attachments.dataExists &&
+        this.datasets.dataExists &&
+        any(d => this.attachments.data.has(`${d.name}.csv`), this.datasets.data);
```
matthew-white, line 181: Could we do this in a way that's case-insensitive with respect to `.csv`? Maybe the easiest thing would just be to check for both `${d.name}.csv` and `${d.name}.CSV`.

Also, no need to change anything, but if you used the `some()` method instead of `any()`, you wouldn't need to access the `data` property:

```js
this.datasets.some(d => this.attachments.has(`${d.name}.csv`))
```

src/components/form-attachment/list.vue
```diff
@@ -482,6 +505,13 @@ export default {
       "readError": "Something went wrong while reading “{filename}”.",
       "success": "{count} file has been successfully uploaded. | {count} files have been successfully uploaded.",
       "link": "Dataset linked successfully."
+    },
+    "datasetsPreview": {
+      // This is a title shown above a section of the page.
+      "title": "Datasets Preview",
+      "body": [
+        "This Form can use Datasets of the Project. For now, we recommend testing the use of your Datasets by uploading temporary data as .csv files while testing, then link the Datasets once you are ready."
```
matthew-white, line 513: Let's discuss this text during the meeting, I wonder whether @issa-tseng or @lognaturel have any quick thoughts.

matthew-white (approved):

sadiqkhoja:
Text has been updated:

image

src/components/form-attachment/list.vue
```diff
@@ -19,6 +19,22 @@ except according to the terms contained in the LICENSE file.
...
+          {{ $t('datasetsPreview.title') }}
+        </span>
+      </div>
+      <div class="panel-body">
+        <i18n-t tag="p" keypath="datasetsPreview.body.full">
+          <template #documentation>
+            <!-- TODO. Specify the `to` prop. -->
+            <doc-link>{{ $t('datasetsPreview.body.documentation') }}</doc-link>
+          </template>
+        </i18n-t>
```
matthew-white, line 35: I mentioned this during the meeting, but could you use `moreInfo.clickHere` or `moreInfo.helpArticle` for the last sentence? In that case, `datasetsPreview.body` won't use component interpolation (it won't include `{documentation}`) and will just be a simple message. You could then use `<sentence-separator>` between `$t('datasetsPreview.body')` and the `<i18n-t>` block for `moreInfo.clickHere`/`moreInfo.helpArticle`.

matthew-white (approved):

PR #657, opened 2022-11-22
Make small changes to ProjectEnableEncryption

matthew-white:
Screenshot:

Screen Shot 2022-11-22 at 6 24 52 PM

### DIFF
--- a/src/components/project/enable-encryption.vue
+++ b/src/components/project/enable-encryption.vue
@@ -101,9 +101,9 @@ except according to the terms contained in the LICENSE file.
           <form-group v-model="hint" :placeholder="$t('field.hint')"
             autocomplete="off"/>
           <div class="modal-actions">
-            <button type="submit" class="btn btn-primary"
+            <button type="submit" class="btn btn-danger"
               :disabled="awaitingResponse">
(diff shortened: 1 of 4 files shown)


lognaturel:
Yay! Given that there's no string for finish, did you consider "Enable Encryption"? I guess you think "Finish" might get used elsewhere?

"there are no length restrictions" is a lie -- we now enforce 10 chars. Maybe fix that too while here? Fine if you'd rather separate out.

matthew-white:
> Given that there's no string for finish, did you consider "Enable Encryption"? I guess you think "Finish" might get used elsewhere?

I wouldn't be surprised if "Finish" gets reused, but I'd also be happy to run with "Enable Encryption" if that seems better to you!

> "there are no length restrictions" is a lie -- we now enforce 10 chars. Maybe fix that too while here?

On it!

lognaturel:
> wouldn't be surprised if "Finish" gets reused

Ok, let's do that for now!

matthew-white:
Screen Shot 2022-11-22 at 6 33 04 PM

lognaturel (approved):

lognaturel:
I noticed that the “First” also implies this is a multi step process. But I think these changes are enough for now.

PR #659, opened 2022-11-23
Replace "Media File" with "Attachment" in text

matthew-white:


src/components/form-draft/checklist.vue
```diff
@@ -136,11 +136,11 @@ export default {
       },
       {
         // This is the title of a checklist item.
-        "title": "Upload Form Media Files",
+        "title": "Upload Form Attachments",
         "body": [
           {
-            "full": "Your Form design references files that are needed in order to present your Form. You can upload new or updated copies of these for distribution under the {mediaFiles} tab.",
-            "mediaFiles": "Media Files"
+            "full": "Your Form design references files that are needed in order to present your Form. You can upload new or updated copies of these for distribution under the {formAttachments} tab.",
```
matthew-white, line 142: This text also needs to account for the possibility of datasets (the user might not need to upload anything, just link a dataset). But we can also make that change later.

src/components/form-draft/publish.vue
```diff
@@ -242,7 +242,7 @@ export default {
       }
     },
     "introduction": [
-      "You are about to make this Draft the published version of your Form. This will finalize any changes you have made to the Form definition and its attached Media Files.",
+      "You are about to make this Draft the published version of your Form. This will finalize any changes you have made to the Form definition and Form Attachments.",
```
matthew-white, line 245: Hmm maybe I should have kept "its": the text could be "its Form Attachments" or maybe even just "its Attachments".

src/components/submission/decrypt.vue
```diff
@@ -369,7 +369,8 @@ $actions-padding-left: $label-icon-max-width + $margin-right-icon;
         // This is the text of a button. "Repeats" refers to repeat groups.
         "mainTable": "Main data table (no repeats)",
         "allTables": "All data tables",
-        "withMedia": "All data and media files"
+        // "Attachments" refers to Submission Attachments.
+        "withAttachments": "All data and Attachments"
```
matthew-white, line 373: "All data and Submission Attachments" might be clearer, but too long I think.

src/components/form-draft/publish.vue
lognaturel: I think it's ok as-is. It's not clear whether "its" refrences Form or Form definition anyway.

lognaturel (commented):
I've verified all of the changes made. I have NOT verified that the changes are exhaustive though I couldn't think of any other affected places.

matthew-white:
Maybe this is another one to consider replacing?

> Drag over the file entry you wish to replace with the file and drop to upload.

I'm thinking we should replace "file entry" with "Form Attachment". Let me know if I should make that change!

matthew-white:
@lognaturel, would you mind leaving a ✅? I can merge once the PR is approved.

lognaturel:
Agreed with that additional replacement. Meant to approve before! 😫

lognaturel (approved):

PR #663, opened 2022-11-28
fix: css fix for entity feature

sadiqkhoja:
## Changes:

- Add `text-overflow-ellipsis` for Dataset summary:
image

- Add auto hyphens for Dataset details on publish modal:
Screen Shot 2022-11-28 at 11 19 53 AM

- Fix css bleeding from Dataset summary row component

### DIFF
--- a/src/components/dataset/summary/row.vue
+++ b/src/components/dataset/summary/row.vue
@@ -10,10 +10,10 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <div class="dataset-row">
+  <div class="dataset-summary-row">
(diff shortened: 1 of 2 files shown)


src/components/dataset/summary/row.vue
```diff
@@ -75,12 +75,16 @@ export default {
 @import '../../../assets/scss/_variables.scss';
 @import '../../../assets/scss/mixins';
 
-.dataset-row {
+.dataset-summary-row {
```
matthew-white, line 78: Ah good call. Actually, I think this will fix issue 3 in the QA document: once loaded, these styles were getting applied to the rows of the dataset table.

matthew-white (approved):

PR #666, opened 2022-11-29
add word-break for Dataset summary and publish modal

sadiqkhoja:
Added `word-break: break-word;` for Dataset summary and publish modal

### DIFF
--- a/src/components/dataset/summary/row.vue
+++ b/src/components/dataset/summary/row.vue
@@ -116,6 +116,7 @@ export default {
 
     .property-list {
       hyphens: auto;
+      overflow-wrap: break-word;
     }
 }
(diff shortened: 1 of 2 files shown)


src/components/dataset/summary/row.vue
```diff
@@ -116,6 +116,7 @@ export default {
 
     .property-list {
       hyphens: auto;
+      word-break: break-word;
```
matthew-white, line 119: I'm seeing that `break-word` is deprecated. What do you think about using `overflow-wrap: break-word;`? We use that elsewhere, and it seems to do the right thing.

matthew-white (approved):

PR #667, opened 2022-11-30
Update translations for v2022.3

matthew-white:


matthew-white:
@ktuite and @sadiqkhoja, would one of you mind glancing over this PR? I just pulled from Transifex and ran destructure.js to convert the Transifex JSON to Vue I18n JSON, so I don't think the PR needs a close look. You should only see changes to `<i18n>` blocks and files in src/locales/ and transifex/.

ktuite (approved):
looks like translations!

PR #669, opened 2022-12-01
Upgrade to Vue CLI v5

matthew-white:
I want to revisit some of these commits, so please don't merge.

Upgrading Vue CLI usually requires upgrading other build/lint/test dependencies, so I went ahead and updated a bunch of those dependencies without checking that that was strictly necessary.


.eslintrc.js
```diff
@@ -74,7 +75,11 @@ module.exports = {
     }],
     'vue/max-attributes-per-line': 'off',
     'vue/multi-word-component-names': 'off',
+    'vue/no-template-target-blank': 'off',
     'vue/require-default-prop': 'off',
-    'vue/singleline-html-element-content-newline': 'off'
+    'vue/singleline-html-element-content-newline': 'off',
+    'vuejs-accessibility/anchor-has-content': 'off',
+    'vuejs-accessibility/click-events-have-key-events': 'off',
+    'vuejs-accessibility/mouse-events-have-key-events': 'off'
```
matthew-white, line 83: I think we should actually address some of these later. I'll file an issue about that later.

src/components/date-range-picker.vue
```diff
@@ -10,7 +10,7 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <label class="form-group">
+  <label class="form-group"><!-- eslint-disable-line vuejs-accessibility/label-has-for -->
```
matthew-white, line 13: This `<label>` does contain an `<input>` via flatpickr, but ESLint doesn't know that.

src/components/form-attachment/upload-files.vue
```diff
@@ -22,6 +22,7 @@ except according to the terms contained in the LICENSE file.
         </i18n-t>
         <i18n-t tag="p" keypath="introduction[1].full">
           <template #clickHere>
+            <!-- eslint-disable-next-line vuejs-accessibility/form-control-has-label -->
             <input v-show="false" ref="input" type="file" multiple
```
matthew-white, line 26: This input is hidden, so I think it doesn't need a label.

src/components/submission/decrypt.vue
```diff
@@ -104,6 +104,7 @@ except according to the terms contained in the LICENSE file.
       </div>
       <!-- We specify a Frontend page for src so that any cookies are sent when
       the iframe form is submitted. -->
+      <!-- eslint-disable-next-line vuejs-accessibility/iframe-has-title -->
       <iframe v-show="false" ref="iframe" src="/blank.html"></iframe>
```
matthew-white, line 108: This iframe is hidden, so I think it doesn't need a title.

vue.config.js
```diff
@@ -33,6 +33,11 @@ module.exports = {
       config.resolve.alias.set('vue$', 'vue/dist/vue.esm-bundler.js');
     }
   },
+  css: {
+    loaderOptions: {
+      css: { url: false }
+    }
+  },
```
matthew-white, line 40: The build didn't like icomoon.css: something about the asset URLs. I made this change based on vuejs/vue-cli#7032, but I'll take another look at this later.

matthew-white:
`npm run test` seems to be complaining about module type. I wanted to note that since upgrading to Vue 3, I've been running into similar issues. In vue.config.js, I've used `config.resolve.alias` in a few places, but I'm not sure why that's been needed / maybe that could change.

package.json
```diff
@@ -51,12 +51,13 @@
...
-    "karma-webpack": "~4.0",
-    "mocha": "~9",
+    "karma-webpack": "^5.0.0",
+    "mocha": "^10.1.0",
     "sass": "~1",
-    "sass-loader": "~10",
+    "sass-loader": "^13.2.0",
     "should": "~13.2",
     "sinon": "~9",
+    "util": "^0.12.5",
```
sadiqkhoja, line 60: Webpack v5 doesn't include node core modules, we have to add it manually here

vue.config.js
```diff
@@ -26,13 +26,14 @@ module.exports = {
       config.resolve.symlinks(false);
 
     if (process.env.NODE_ENV === 'test') {
-      config.resolve.alias.set(
-        '@vue/test-utils',
-        '@vue/test-utils/dist/vue-test-utils.esm-bundler.js'
-      );
```
sadiqkhoja, line 32: no longer necessary, causing tests to fail

package.json
```diff
@@ -51,12 +51,13 @@
...
     "vue-cli-plugin-i18n": "~2"
```
sadiqkhoja, line 61: We are using latest version of this, and it has moderate level vulnerability. 

This is a PR to fix this but the maintainer is MIA since July: https://github.com/intlify/vue-cli-plugin-i18n/pull/279

lognaturel (approved):

vue.config.js
```diff
@@ -11,9 +11,6 @@ except according to the terms contained in the LICENSE file.
 */
 module.exports = {
   chainWebpack: (config) => {
-    // TODO/requestData. Revisit this.
-    config.resolve.alias.set('pinia', 'pinia/dist/pinia.cjs');
```
matthew-white, line 15: Previously, I was having trouble using Pinia without this line, but everything seems fine now without it (building, linting, testing). We don't use Pinia in production, only development, so it feels OK to make this change now.

PR #670, opened 2022-12-02
docs: compress big images

alxndrsn:
I ran these images through `pngout`, and they're now ~750KB smaller.  Their dimensions are still very big.


matthew-white:
Are smaller images useful because they make it faster to load the contributing guide? Am I right in thinking that smaller images won't reduce the size of the repository / the time to `git clone`, because the larger images will continue to live on in the repository? Smaller images seem like a good thing, but I'm wondering what benefits we would see by making this change.

alxndrsn:
> Are smaller images useful because they make it faster to load the contribution guide? Am I right in thinking that smaller images won't reduce the size of the repository / the time to git clone, because the larger images will continue to live on in the repository?

All correct.  Ideally we could travel back in time and compress before initial upload.  This change will actually make the repo bigger :disappointed_relieved:

matthew-white (approved):

matthew-white:
I don't see much downside to merging: the repo will get bigger, but not by much. And it'd be nice to have smaller images in the contribution guide. Let's make sure that future images are smaller!

> Their dimensions are still very big.

Feel free to try scaling them down if that's of interest. But I think we could also merge what you have here.

Issue #671, opened 2022-12-04
Move from Vue CLI to Vite


matthew-white:
Vue CLI is now in maintenance mode, so I think we should start planning to move to Vite. Some of the performance issues described in [Why Vite?](https://vitejs.dev/guide/why.html) sound familiar, so hopefully we'll see some improvements in that area.

I had interpreted a comment on a Vue I18n repository [here](https://github.com/intlify/bundle-tools/issues/13#issuecomment-1316159391) as meaning that Vue I18n will stop working with Vue CLI. But looking at it slightly more closely, I think they're still planning to support webpack, just with something other than `@intlify/vue-i18n-loader`. Though as @sadiqkhoja has noted, `vue-cli-plugin-i18n` also doesn't seem to be well maintained.

We don't have much custom webpack configuration: I think what little there is is just in vue.config.js. Hopefully that will ease the transition to Vite.

Probably the hardest part of transitioning to Vite will be figuring out testing. Will we still be able to use Karma? Right now we use `karma-webpack` to get Karma working with webpack. Vite seems to use Rollup, and from a glance, there do seem to be ways to get Karma working with Rollup. That said, I have no particular love of Karma. Since Karma is no longer recommended within the Vue community, some testing things haven't always worked out-of-the-box for us. Looking at the Vue 3 [testing guide](https://vuejs.org/guide/scaling-up/testing.html), it looks like Vitest is the current recommendation. It'd be nice though if we could somehow sequence things so that we don't have to transition off Vue CLI and Karma at the same time. Could we move from Vue CLI to Vite in one step, initially setting up Karma to work with Rollup, then in a next step move from Karma to something like Vitest?

matthew-white:
I've been able to move over the build from Vue CLI to Vite, but I'm still working on moving testing over to Vitest. I don't think I'm going to finish that part in v2025.1, so I've filed a separate issue for it: #1211. I also wanted to address some comments about testing that are in the issue description:

> Will we still be able to use Karma?

Karma itself is deprecated. The goal isn't just to move off of Vue CLI, but also to move off of Karma.

> Could we move from Vue CLI to Vite in one step, initially setting up Karma to work with Rollup, then in a next step move from Karma to something like Vitest?

I'm still working through some questions about Vitest, but I'm confident that we'll be able to move directly from Karma to Vitest. There's no need for an intermediate step. Vitest browser mode is pretty similar to Karma.

PR #678, opened 2022-12-05
Don't create Pinia outside development

matthew-white:


src/container.js
```diff
@@ -45,7 +48,7 @@ export default ({
   logger = console
 } = {}) => {
   const container = {
-    pinia: createPinia(),
+    pinia: process.env.NODE_ENV === 'development' ? createPinia() : piniaMock,
```
matthew-white, line 51: Originally, my strategy was going to be to use webpack to alias the `pinia` package to src/util/pinia.js. That way, only our webpack config would change, not this file. However, at the end of the day, I think this approach is clearer: there's less webpack magic involved. Plus, now that we're planning to move to Vite (#671), it seems good to have as little webpack-specific code as possible. I checked the build, and even though this file still imports Pinia, if it doesn't use Pinia at all, which it won't outside development, the import is removed from the bundle.

Since we won't use it after all, I've removed src/util/pinia.js.

ktuite (approved):

Issue #682, opened 2022-12-15
Filter form upload options by file type


alxndrsn:
The form uploader says files should be XML or XLS forms, but the form input accepts any file types:

![Screenshot_2022-12-15_07-59-22](https://user-images.githubusercontent.com/191496/207776792-e6d4df0d-4328-40a5-913a-f54630c5b7eb.png)

![Screenshot_2022-12-15_08-13-11](https://user-images.githubusercontent.com/191496/207778030-be5e158f-b022-4c78-a396-36c8b37f9408.png)

Would it be helpful to add an `accepts` attribute to filter files in the client's file picker?  E.g.

![Screenshot_2022-12-15_08-12-56](https://user-images.githubusercontent.com/191496/207778040-050a1b09-27e1-4780-98d4-6b5d1ef6e216.png)

srujner:
Tested with success!

dbemke:
Tested with success!

PR #683, opened 2022-12-15
form/new: filter accepted file types

alxndrsn:
Fix for #682


matthew-white (approved):
Nice, thank you! I think this makes the file picker more useful. Users will still be able to drag-and-drop a file of a different type, but I think that that's OK/good.

PR #689, opened 2023-01-12
Make small changes to white space

matthew-white:
Just making a few small changes. I noticed these while working on tooltips, which brought me to many components, but I wanted to commit these changes separately.


sadiqkhoja (approved):
👍

PR #691, opened 2023-01-12
Change "Total" title to "Total Submissions"

matthew-white:
These will soon be changed from titles to instant tooltips, but I think it's better to first make this change in a separate commit.


ktuite (approved):
totally great!

PR #696, opened 2023-01-12
Prevent Karma from timing out when an assertion fails on a VTU wrapper

matthew-white:
Sometimes, an assertion failure will cause Karma to time out and exit, which can add friction to testing. I thought it was maybe a bug with Should.js, but I think it's something more specific, that Should.js is doing something that Vue or Vue Test Utils doesn't like. These timeouts seem to happen when an assertion fails on a Vue Test Utils wrapper specifically (or an array of wrappers). I'm not quite sure why that is (maybe Should.js is traversing the wrapper, and the wrapper contains a recursive cycle?). And my solution here still needs more validation. But so far, it looks like this change prevents these timeouts.

### DIFF
--- a/test/assertions.js
+++ b/test/assertions.js
@@ -1,8 +1,58 @@
 import should from 'should';
+import { DOMWrapper, VueWrapper } from '@vue/test-utils';
 
-const unwrapElement = (elementOrWrapper) => (elementOrWrapper instanceof HTMLElement
-  ? elementOrWrapper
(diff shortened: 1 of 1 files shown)


matthew-white:
We're going to be replacing Should.js with Chai as part of #671. This issue doesn't seem to be a problem with Chai. If an assertion fails on a Vue Test Utils wrapper, the test doesn't time out, and the error message is what I'd expect.

Issue #686, opened 2022-12-29
Keyboard can be used to navigate to disabled tab


matthew-white:
When a tab is disabled, its `<li>` element has `pointer-events: none;`, so clicking it has no effect. However, the tab link can still be focused and navigated into using the keyboard (using the tab and enter keys).

This is different from how a disabled link works. A disabled link can be focused using the keyboard, but it cannot be navigated into because we prevent default in that case. We don't have a generic mechanism for calling `preventDefault()` on a disabled link, but in each case in which a link is disabled, we prevent default on `click` (or at least I think we do). Doing so can be as simple as specifying `@click.prevent` in the template. For a link, I'm pretty sure that preventing default on `click` also prevents default when the enter key is pressed.

However, we don't do anything like that for the links of disabled tabs, at least in the cases I'm seeing. In `FormHead` and `UserHome`, it is possible to use the keyboard to navigate into the links of disabled tabs. I think we should prevent that possibility.

To be clear, I don't think it's a huge deal that that's possible, because the user could always use the URL bar to attempt the same navigation, and we redirect the user to the homepage when they attempt to visit a page that they don't have access to. However, I think it'd be confusing for a user using the keyboard to accidentally navigate into one of these links and be redirected to the homepage. I think it'd be better for keyboard use to be disabled in this case just as mouse use is.

I also wanted to note one other thing I'm noticing, which is that disabled buttons seem to work slightly differently from both disabled links and the links of disabled tabs. While both types of links can be focused, it seems like disabled buttons can't be focused (at least in Chrome). Should we match that behavior for disabled links and the links of disabled tabs?

matthew-white:
It's ending up that I'm resolving this as part of resolving #648. Disabled elements intersect with tooltips for reasons I'll explain in that PR.

matthew-white:
> For a link, I'm pretty sure that preventing default on click also prevents default when the enter key is pressed.

I've confirmed that this is the case. See https://css-tricks.com/when-a-click-is-not-just-a-click/

srujner:
**Tested with Success!**

dbemke:
**Tested with Success!**

PR #700, opened 2023-01-17
Make changes related to disabled elements

matthew-white:
The primary purpose of this PR is to set things up so that it is possible to add instant tooltips to disabled elements. I think that the PR is also a useful change on its own, for example, closing #686.

One challenge that I have encountered while working on instant tooltips is elements with a `disabled` attribute. We have a fair number of cases in which a disabled element has a `title` attribute. The browser will gladly show that `title` attribute, but it isn't straightforward to add an instant tooltip to such an element. That's because the element no longer triggers the events to show/hide the tooltip (`mouseenter`/`mouseleave` + `focus`/`blur`).

Researching online, I read about a few different approaches:

- **Use the [`aria-disabled`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled) attribute rather than `disabled`.** This is the approach I ended up going with. By using `aria-disabled`, disabled elements will continue to trigger `mouseenter`/`mouseleave` events and to be focusable, allowing us to add tooltips. The fact that they will continue to be focusable is a change from our behavior today: it's not currently possible to tab to an element with a `disabled` attribute. However, I think it's a good change in terms of accessibility. For example, with this change, a user using the keyboard will be able to focus a disabled element and see its tooltip. I found [this article](https://css-tricks.com/making-disabled-buttons-more-inclusive/) helpful when thinking about accessibility. Using `aria-disabled` seemed like a relatively straightforward way to enable us to add tooltips and also seemed like a good approach in terms of accessibility.
- **Position an overlay over each disabled element** or wrap the element in another element, then listen for events on the overlay/wrapper rather than the element itself. Because the overlay/wrapper wouldn't be focusable, this would result in the closest behavior to what we have today. The overlay idea felt complex to me though, and the wrapper idea also didn't seem ideal.
- **Put a help icon next to each disabled button and add the tooltip to that** (or replace the disabled button with the help icon entirely). This wouldn't be hard to do, but it would be the biggest change from our current behavior.

I also wanted to leave some notes about disabled links, to which we sometimes add tooltips:

- `<a>` elements don't use the `disabled` attribute, so there weren't the same issues for tooltips as elements with that attribute. We use the `disabled` class to disable links, which styles them to look disabled and (for tab links only) sets `pointer-events` to `none`.
- We actually don't want to set `pointer-events` to `none` because like the `disabled` attribute, it prevents `mouseenter`/`mouseleave` events. Also, it still allows the link to be navigated into via the keyboard. I set `pointer-events` back to `auto` and used JavaScript to disable these links, including via the keyboard.
- Disabled links have been and will continue to be focusable.
- I would love to set `aria-disabled="true"` on disabled links so that screen readers will announce them as disabled and so that we have fewer different ways to disable things. However, according to [this article](https://www.scottohara.me/blog/2021/05/28/disabled-links.html), it's not recommended to set `aria-disabled="true"` on an `<a>` element with an `href` attribute (which we always set in Frontend).
  - Among `<a>` elements that we disable, we have many with `href="#"`. We sometimes set `role="button"` on those and sometimes don't, though probably we almost always should. As long as we set `role="button"`, I don't think we need to set `href="#"`. And if we don't set `href="#"`, I think that would allow us to set `aria-disabled` on these `<a>` elments. I think that would be an improvement that we could consider for the future.
  - There are other `<a>` elements that we disable that set `href` to something other than `#`. But there aren't many: the ones in `FormHead` are the only ones that come to mind. According to the article above, you can disable an `<a>` element by removing its `href` attribute and setting `role="link"` and `aria-disabled="true"`. The resulting element isn't focusable, so I think we would add a `tabindex` attribute as well. Again, I think that's something we could consider for the future.
  - I found this [other article](https://css-tricks.com/how-to-disable-links/) helpful as well. It seemed consistent with the article above.
- In several places where a link is styled as a button, the link is rendered as an `<a>` element when it is enabled and rendered as a `<button>` when it is disabled (for example, `EnketoPreview`). I think I did it that way because it felt like the easiest way to fully disable the "link" and because the link looked like a button anyway. I'm also not sure that `a.btn.disabled` elements have been styled the same as `button.btn[disabled]` elements. However, I am starting to question whether this pattern works well for screen readers. It seems funny that an element would sometimes be announced as a link and sometimes as a button. If we make the improvements to disabled links in the bullet point above, we could also look into consistently rendering these links as `<a>` elements.

I tried to adjust our styles so that `aria-disabled` elements appear the same way that elements with a `disabled` attribute appeared before. However, there are some combinations that are possible now that weren't possible before, for example, focus + disabled. I fixed related issues that I noticed and also took a look through app.scss for styles to adjust there. However, I'm not certain that I have accounted for all new combinations in all places. I'll keep my eyes open for related issues.

### DIFF
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -784,13 +784,11 @@ becomes more complicated.
       }
 
       > a {
-        pointer-events: none;
-
-        &, &:focus {
+        &, &:hover, &:focus {
           color: #999;
         }
 
-        &:focus {
+        &:hover, &:focus {
           border-bottom-color: transparent;
         }
       }
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -120,7 +120,7 @@ class. */
     + .caret { margin-left: 0; }
   }
 
-  &[disabled], &.disabled, fieldset[disabled] & {
+  &[aria-disabled="true"], &.disabled, fieldset[disabled] & {
     cursor: not-allowed;
     opacity: 0.65;
   }
@@ -138,20 +138,20 @@ $btn-focus-box-shadow: 0 0 0 3px rgba(0, 0, 0, 0.18);
 
   &:focus, &:active:focus { outline: none; }
   &:focus { box-shadow: $btn-focus-box-shadow; }
-  &[disabled], &.disabled, fieldset[disabled] & { box-shadow: none; }
+  &[aria-disabled="true"], &.disabled, fieldset[disabled] & { box-shadow: none; }
 }
 
 .btn-primary {
   background-color: $color-action-background;
 
-  &[disabled], &.disabled, fieldset[disabled] & {
+  &[aria-disabled="true"], &.disabled, fieldset[disabled] & {
     background-color: $color-action-background-disabled;
   }
 
   &:hover, &:focus, &:active:focus {
     background-color: $color-action-background-hover;
 
-    &[disabled], &.disabled, fieldset[disabled] & {
+    &[aria-disabled="true"], &.disabled, fieldset[disabled] & {
       background-color: $color-action-background-disabled;
     }
   }
@@ -162,7 +162,7 @@ $btn-focus-box-shadow: 0 0 0 3px rgba(0, 0, 0, 0.18);
 
   &:hover, &:focus, &:active:focus {
     background-color: #ccc;
-    &[disabled], &.disabled, fieldset[disabled] & { background-color: #ddd; }
+    &[aria-disabled="true"], &.disabled, fieldset[disabled] & { background-color: #ddd; }
   }
 
   > [class^="icon-"]:first-child { color: $color-action-background; }
@@ -173,7 +173,10 @@ $btn-focus-box-shadow: 0 0 0 3px rgba(0, 0, 0, 0.18);
 
   &:hover, &:focus, &:active:focus {
     background-color: $color-danger-dark;
-    &[disabled], &.disabled, fieldset[disabled] { background-color: $color-danger; }
+
+    &[aria-disabled="true"], &.disabled, fieldset[disabled] {
+      background-color: $color-danger;
+    }
   }
 }
 
@@ -187,7 +190,7 @@ a, .btn-link {
     text-decoration: underline;
   }
 
-  &.disabled, &[disabled], fieldset[disabled] & {
+  &.disabled, &[aria-disabled="true"], fieldset[disabled] & {
     color: #888;
     text-decoration: none;
   }
@@ -223,7 +226,7 @@ a {
 .btn-link {
   border-radius: 0;
 
-  &, &:active, &[disabled] fieldset[disabled] & {
+  &, &:active, &[aria-disabled="true"], fieldset[disabled] & {
     background-color: transparent;
   }
 
@@ -324,35 +327,29 @@ $form-group-padding-bottom: 15px;
   padding-bottom: $form-group-padding-bottom;
   position: relative;
 
-  &:focus-within {
-    .form-label {
-      color: $color-action-foreground;
-    }
-  }
-
   $form-control-z-index: 1;
   .form-control {
     background: $color-input-background;
     border: none;
     border-bottom: 1px solid $color-input-inactive;
     border-radius: 0;
-    box-shadow: none;
     color: $color-input;
     position: relative;
     z-index: $form-control-z-index;
 
-    &:focus {
+    &, &:focus { box-shadow: none; }
+
+    &:focus:not([aria-disabled="true"]) {
       border-bottom-color: $color-action-foreground;
-      box-shadow: none;
     }
 
     &::placeholder { color: $color-text; }
 
-    &[disabled], &[readonly], fieldset[disabled] & {
+    &[aria-disabled="true"], &[readonly], fieldset[disabled] & {
       &, &::placeholder { color: $color-input-inactive; }
     }
 
-    &[disabled], fieldset[disabled] & { cursor: not-allowed; }
+    &[aria-disabled="true"], fieldset[disabled] & { cursor: not-allowed; }
   }
 
   .form-label {
@@ -364,8 +361,10 @@ $form-group-padding-bottom: 15px;
     transform: translateY(2px);
     transition: 0.15s transform, 0.15s color;
   }
+  &:focus-within .form-label { color: $color-action-foreground; }
   .form-control {
     &:placeholder-shown ~ .form-label { transform: translateY(-15px); }
+    &[aria-disabled="true"] ~ .form-label { color: $color-input-inactive; }
   }
 
   .close {
@@ -479,8 +478,9 @@ select {
     border-color: $color-danger;
 
     &:focus {
-      border-color: $color-danger-dark;
       box-shadow: none;
+
+      &:not([aria-disabled="true"]) { border-color: $color-danger-dark; }
     }
   }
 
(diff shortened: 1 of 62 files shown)


src/assets/scss/app.scss
```diff
@@ -324,35 +327,29 @@ $form-group-padding-bottom: 15px;
   padding-bottom: $form-group-padding-bottom;
   position: relative;
 
-  &:focus-within {
-    .form-label {
-      color: $color-action-foreground;
-    }
-  }
```
matthew-white, line 331: I moved this below, closer to the other `.form-label` styles.

src/components/field-key/row.vue
```diff
@@ -69,12 +71,6 @@ export default {
   methods: {
     showCode() {
       this.$emit('show-code', this.fieldKey, this.$refs.popoverLink);
-    },
-    revoke() {
-      // Bootstrap does not truly disable dropdown menu items marked as
-      // disabled.
-      if (this.fieldKey.token == null) return;
-      this.$emit('revoke', this.fieldKey);
```
matthew-white, line 77: We no longer need the `if` because `useDisabled()` will stop the handler from being called. Without the `if`, the method is just one line, so I moved it into the template.

src/components/project/user/row.vue
```diff
@@ -20,7 +20,8 @@ except according to the terms contained in the LICENSE file.
       <form>
         <div class="form-group">
           <select class="form-control" :value="selectedRoleId"
-            :disabled="disabled" :title="selectTitle"
+            :aria-disabled="disabled || awaitingResponse"
+            :title="disabled ? $t('cannotAssignRole') : null"
```
matthew-white, line 24: Light refactor here. I changed `disabled` so that it doesn't have anything to do with `awaitingResponse`, just whether the user is the current user. I think that simplifies things a little, including the `selectTitle` computed property. That computation is simple enough now that I think it can be in the template itself rather than a computed property.

src/components/submission/decrypt.vue
```diff
@@ -277,11 +277,9 @@ export default {
...
-      const willDownload = !a.classList.contains('disabled') &&
+      const willDownload = this.managedKey == null ||
         this.$refs.form.reportValidity();
-      if (this.managedKey == null) {
-        if (!willDownload) event.preventDefault();
-      } else {
+      if (this.managedKey != null) {
         event.preventDefault();
         if (willDownload) this.decrypt(a.getAttribute('href'));
       }
```
matthew-white, line 285: Because of `useDisabled()`, it's now guaranteed that `a` doesn't have the `disabled` class. That means that if `this.managedKey == null`, we never need to prevent default now: there's never a reason to stop the download now. (If `this.managedKey != null`, we always prevent default, because we initiate the download another way, within `this.decrypt()`.) Note that if `this.managedKey == null`, we don't need to check form validity, because there's no decryption passphrase input to validate.

src/components/user/edit/basic-details.vue
```diff
@@ -20,8 +20,8 @@ except according to the terms contained in the LICENSE file.
           :placeholder="$t('field.email')" required autocomplete="off"/>
         <form-group v-model.trim="displayName" type="text"
           :placeholder="$t('field.displayName')" required autocomplete="off"/>
-        <button :disabled="awaitingResponse" type="submit"
-          class="btn btn-primary">
+        <button type="submit" class="btn btn-primary"
+          :aria-disabled="awaitingResponse">
```
matthew-white, line 24: Also taking the chance to reorder these attributes. (Vue ESLint rules used to require bound attributes to come first, but thankfully no longer.)

test/components/field-key/revoke.spec.js
```diff
@@ -23,8 +23,6 @@ describe('FieldKeyRevoke', () => {
       const app = await load('/projects/1/app-users');
       const li = app.get('.field-key-row .dropdown-menu li');
       li.classes('disabled').should.be.true();
-      await li.get('a').trigger('click');
-      app.getComponent(FieldKeyRevoke).props().state.should.be.false();
```
matthew-white, line 27: We no longer need to test this, and in general going forward, trying to test this sort of thing would be a little difficult. Once we verify that `li` has the `disabled` class, we know from the tests of `useDisabled()` that clicking the link will have no effect. The root `App` component calls `useDisabled()`, but most tests don't mount `App` — so `useDisabled()` won't be called in most tests. `useDisabled()` also requires that the component be attached to the document body. Altogether, this means that testing that a `.disabled` link is actually disabled is more difficult. I think it's OK to remove this assertion though, because we have good test coverage of `useDisabled()`.

test/components/project/user/list.spec.js
```diff
@@ -155,20 +155,19 @@ describe('ProjectUserList', () => {
         });
     });
 
-    it('renders the select correctly for the current user', () => {
+    it('renders the select correctly for the current user', async () => {
```
matthew-white, line 158: Making this async because that will make things easier in the PR for instant tooltips.

test/components/submission/decrypt.spec.js
```diff
@@ -307,16 +307,6 @@ describe('SubmissionDownload', () => {
       modal.get('a').trigger('click');
       modal.should.alert('info');
     });
-
-    it('does nothing if the link is disabled', () => {
```
matthew-white, line 311: We test above that the link has the `disabled` class, so we know from the tests of `useDisabled()` that the link will do nothing. I think this test isn't needed anymore (and no longer passes because `useDisabled()` isn't called).

src/composables/disabled.js
```diff
@@ -0,0 +1,96 @@
...
+    preventAndStop(event);
+};
+
+// Prevents the menu of a <select> element from being shown if the element has
+// aria-disabled="true". event.type will be either 'mousedown' or 'keydown'.
+const preventDisabledSelect = (event) => {
+  if (event.target.tagName === 'SELECT' &&
+    event.target.getAttribute('aria-disabled') === 'true' &&
+    (event.type === 'mousedown' ||
+    event.key === ' ' || event.key === 'ArrowDown' || event.key === 'ArrowUp'))
```
sadiqkhoja, line 79: I think we need to check for Enter key here as well

matthew-white: I thought so at first as well, but after trying it, it doesn't look like Enter opens a select menu in Chrome, Firefox, or Safari. Actually, for our `Multiselect` component, we had to add that behavior ourselves.

sadiqkhoja (approved):
Looks good. Just a couple of things:

- Do we need to use aria-disabled in analytics/form.vue:55 `<fieldset :disabled="enabled !== true">` ?
- User can press enter on select control to open it, so we should check for Enter key event on as well

matthew-white:
> * Do we need to use aria-disabled in analytics/form.vue:55 `<fieldset :disabled="enabled !== true">` ?

I wondered that myself, and I'm not 100% sure. I needed to change `disabled` to `aria-disabled` for buttons, inputs, and selects in order to get tooltips working because we add tooltips to those elements. However, we don't add tooltips to fieldsets, and it's a little hard to imagine why we would. We only use fieldsets in one component, and partly I didn't change them because I didn't need to for tooltips. But I also wasn't sure that doing so would end up being a helpful change to users navigating the page using a keyboard or screen reader. For buttons and form controls, it seems like it's a good thing to keep the element focusable even if it's disabled, especially since we usually add a tooltip to a disabled button or form control. However, given how we use fieldsets, I'm not sure that it'd be useful to keep the form controls in the fieldset focusable. We use fieldsets when there are form controls that only become relevant based on the value of a prior form control. If a set of form controls aren't relevant, I'm not sure it's useful to make them focusable, which would cause the Tab key to navigate through each of them. I could be wrong about that though! But I also figure that that's a change we could make later if we needed to.

src/composables/disabled.js
sadiqkhoja: you are right, Enter doesn't work on default select

PR #702, opened 2023-01-17
enhance: see more context on datasets list

sadiqkhoja:
Added "Total Entities" and "Latest Entities" column on Dataset list page.

image

### DIFF
--- a/src/components/dataset/list.vue
+++ b/src/components/dataset/list.vue
@@ -10,10 +10,10 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <div id="dataset-list-page">
+  <div id="dataset-list">
     <page-section>
       <template #heading>
-        <h1>Datasets</h1>
+        <h1>{{ $t('resource.datasets') }}</h1>
       </template>
       <template #body>
         <dataset-table/>
@@ -53,14 +53,3 @@ datasets.request({
   resend: false
 }).catch(noop);
 </script>
-
-<style lang="scss">
-@import '../../assets/scss/_variables.scss';
-
-#dataset-list-page{
-  h1{
-    color: $color-accent-primary;
-    margin-top: 0;
-  }
-}
-</style>
--- a/src/components/dataset/list.vue
+++ b/src/components/dataset/list.vue
@@ -11,9 +11,9 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <div>
-    <dataset-introduction/>
     <dataset-table/>
     <loading :state="datasets.initiallyLoading"/>
+    <dataset-introduction/>
   </div>
 </template>
 
@@ -41,6 +41,7 @@ const props = defineProps({
 const { datasets } = useRequestData();
 datasets.request({
   url: apiPaths.datasets(props.projectId),
+  extended: true,
   resend: false
 }).catch(noop);
 </script>
--- a/src/components/dataset/list.vue
+++ b/src/components/dataset/list.vue
@@ -10,10 +10,17 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <div>
-    <dataset-table/>
-    <loading :state="datasets.initiallyLoading"/>
-    <dataset-introduction/>
+  <div id="dataset-list-page">
+    <page-section>
+      <template #heading>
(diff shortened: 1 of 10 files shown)


src/components/dataset/list.vue
```diff
@@ -11,9 +11,9 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <div>
-    <dataset-introduction/>
     <dataset-table/>
     <loading :state="datasets.initiallyLoading"/>
+    <dataset-introduction/>
```
matthew-white, line 16: Is there any space between `DatasetIntroduction` and the loading message above it? You can check this by adding [network throttling](https://css-tricks.com/throttling-the-network/) in Chrome devtools. (I use custom network throttling profiles with various latency numbers.) If not, how about adding some `margin-top` to `DatasetIntroduction`?

src/components/dataset/row.vue
```diff
@@ -46,6 +57,10 @@ export default {
 
 .dataset-row {
   .name { @include text-overflow-ellipsis; }
+  .entities {
+    text-align: right;
+    padding-right: 20px;
```
matthew-white, line 62: In the screenshot in the release criteria, there looks to be more space between the "Total Entities" and "Latest Entity" columns. What do you think about increasing the padding? Does it help to make the padding a percentage?

src/components/dataset/list.vue
matthew-white: Actually, looking at the screenshot in the release criteria, there seems to be a bit of space between `DatasetTable` and `DatasetIntroduction`. So that might be another reason to add some `margin-top` to `DatasetIntroduction`.

src/components/dataset/list.vue
```diff
@@ -11,9 +11,9 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <div>
-    <dataset-introduction/>
```
matthew-white, line 14: Even though `DatasetIntroduction` should be moved below, the release criteria also says:

> at the top I should see a Datasets header

You can use the `PageSection` component for that. (Everything here would go in the `body` slot of the `PageSection`.)

src/components/dataset/list.vue
sadiqkhoja: Our tables have `margin-bottom: 20px`. And loading message with the image look fine to me:

image

I can hide the image during loading though

matthew-white: Either way, I think it's OK to show the image during loading.

src/components/dataset/list.vue
```diff
@@ -10,10 +10,17 @@ including this file, may be copied, modified, propagated, or distributed
...
 -->
 <template>
-  <div>
-    <dataset-introduction/>
-    <dataset-table/>
-    <loading :state="datasets.initiallyLoading"/>
+  <div id="dataset-list-page">
+    <page-section>
+      <template #heading>
+        <h1>Datasets</h1>
```
matthew-white, line 16: The `PageSection` component actually expects this to be a `<span>`. `PageSection` has styles to set the color and font size. Maybe we should change `PageSection` though so that it works with `<h*>` elements: using those elements makes a lot of sense.

src/components/dataset/list.vue
```diff
@@ -10,10 +10,17 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <div>
-    <dataset-introduction/>
-    <dataset-table/>
-    <loading :state="datasets.initiallyLoading"/>
+  <div id="dataset-list-page">
```
matthew-white, line 13: Would you mind making it just `dataset-list`? I've been trying to keep the HTML id and the component name in sync.

src/components/dataset/list.vue
matthew-white, line 16: `$t('resource.datasets')`

src/components/dataset/table.vue
```diff
@@ -40,5 +42,12 @@ const { datasets } = useRequestData();
 </script>
 
 <style lang="scss">
-#dataset-table { table-layout: fixed; }
+#dataset-table {
+  table-layout: fixed;
+
+  .entities {
+    text-align: right;
+    padding-right: 20px;
```
matthew-white, line 50: Does this need to be 10% to match `DatasetRow`?

matthew-white (approved):
Looks great! ✨

PR #703, opened 2023-01-17
new: Dataset overview page

sadiqkhoja:
Created new Dataset overview page

image


src/components/dataset/overview.vue
```diff
@@ -0,0 +1,185 @@
...
+      <template #body>
+        <div class="row">
+          <div class="col-md-6 creation-forms">
+            <summary-item icon="magic-wand">
+              <template #heading>
+                {{ propertiesByForm.length }}
+              </template>
+              <template #body>
+                <p>{{ $t('formsCreateEntities') }}</p>
+                <table class="table">
```
matthew-white, line 28: How about adding `v-if="propertiesByForm.length > 0"` ?

src/components/dataset/overview.vue
```diff
@@ -0,0 +1,185 @@
...
+                    </tr>
+                  </tbody>
+                </table>
+              </template>
+            </summary-item>
+          </div>
+        </div>
+      </template>
+    </page-section>
+    <page-section id="dataset-overview-properties">
```
matthew-white, line 70: `v-if="dataset.properties.length > 0`?

src/components/dataset/overview.vue
```diff
@@ -0,0 +1,185 @@
...
+            <tr>
+              <!-- we have to show property name even if there is no associated form -->
+              <td :rowspan="property.forms.length || 1">
+                {{ property.name }}
+              </td>
+              <td>
+                <router-link v-if="property.forms.length > 0" :to="formPath(project.id, property.forms[0].xmlFormId)">{{ property.forms[0].name }}</router-link>
+              </td>
+            </tr>
+            <template v-for="(form, index) in property.forms" :key="form">
```
matthew-white, line 93: `key` should be a primitive, but you could specify `key="form.xmlFormId"`.

src/components/expandable-row.vue
```diff
@@ -0,0 +1,76 @@
...
+https://github.com/getodk/central-frontend/blob/master/NOTICE.
+
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+<template>
+  <tr class="expandable-row">
```
matthew-white, line 13: It'd be great to add unit tests of this component that verify that the component toggles correctly and uses the provided slots. That might have the result of simplifying other component tests: in those other tests, you will be able to just test that this component has been passed the correct content, not anything about the toggling.

src/components/dataset/overview.vue
```diff
@@ -0,0 +1,185 @@
...
+            </summary-item>
+          </div>
+          <div class="col-md-6 linked-forms">
+            <summary-item icon="link">
+              <template #heading>
+                {{ dataset.linkedForms.length }}
+              </template>
+              <template #body>
+                <p>{{ $tc('formsConsumeData', dataset.linkedForms.length) }}</p>
+                <table class="table">
```
matthew-white, line 55: How about adding `v-if="dataset.linkedForms.length > 0"` ?

matthew-white (commented):
I've left a number of comments (many small), but I think this is definitely headed in the right direction. Adding a new page can be a complex task in Frontend because it touches so many layers, but things are set up well here. I haven't looked at quite everything yet, but I thought we should start with these comments, then I'll do another round of review once those are resolved.

src/components/dataset/overview.vue
```diff
@@ -0,0 +1,185 @@
...
+              <template #heading>
+                {{ propertiesByForm.length }}
+              </template>
+              <template #body>
+                <p>{{ $t('formsCreateEntities') }}</p>
+                <table class="table">
+                  <tbody>
+                    <expandable-row v-for="(form) in propertiesByForm" :key="form.name">
+                      <template #title>
+                        <router-link :to="formPath(project.id, form.xmlFormId)">{{ form.name }}</router-link>
```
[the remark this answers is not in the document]
sadiqkhoja: From the API, I am returning coalesce(name, xmlFormId) so name is guaranteed to have a value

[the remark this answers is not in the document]
matthew-white: Oh interesting! It's a little surprising to me that the API would have `name` fall back to `xmlFormId` for this endpoint when it doesn't for other endpoints. But if that's the way the endpoint works, then it makes sense that there doesn't need to be any fallback in Frontend.

src/components/dataset/overview.vue
```diff
@@ -0,0 +1,78 @@
...
+    LinkedForms,
+    DatasetProperties
+  },
+  setup() {
+    const route = useRoute();
+    // The component does not assume that this data will exist when the
+    // component is created.
+    const { dataset, resourceStates } = useRequestData();
+    const { dataExists } = resourceStates([dataset]);
+    return { dataset, dataExists, projectId: Number(route.params.projectId) };
```
matthew-white, line 64: Since you specified `props: true` for the route, you can also access `projectId` from the component's props if you prefer.

matthew-white, line 64: When we first added projects, I started out doing a lot of project ID string-to-number conversion, but at some point I stopped. Ultimately, I think it's not a necessary conversion because we only use the project ID to construct URLs.

src/components/dataset/overview/connection-to-forms.vue
```diff
@@ -0,0 +1,116 @@
...
+except according to the terms contained in the LICENSE file.
+-->
+
+<template>
+  <summary-item id="connection-to-forms" icon="magic-wand">
+    <template #heading>
+      {{ propertiesByForm.length }}
+    </template>
+    <template #body>
+      <p>{{ $t('formsCreateEntities') }}</p>
```
matthew-white, line 19: `$tc()` since the message uses plurals.

src/components/dataset/overview/connection-to-forms.vue
```diff
@@ -0,0 +1,116 @@
...
+      <div class="div">
+        <expandable-row v-for="(form) in propertiesByForm" :key="form.xmlFormId">
+          <template #title>
+            <router-link :to="publishedFormPath(projectId, form.xmlFormId)" :title="form.name">{{ form.name }}</router-link>
+          </template>
+          <template #caption>
+            {{ $tcn('common.propertiesCount', totalProperties, { inform: $n(form.properties.length, 'default') }) }}
+          </template>
+          <template #details>
+            <span v-for="(property, index) in form.properties" :key="property">
```
matthew-white, line 29: Should we add the same styles as `.property-list` in `DatasetSummaryRow`?

```css
.property-list {
  hyphens: auto;
  overflow-wrap: break-word;
}
```

src/components/dataset/overview/linked-forms.vue
```diff
@@ -0,0 +1,74 @@
...
+    <template #heading>
+      {{ linkedForms.length }}
+    </template>
+    <template #body>
+      <p>{{ $tc('formsConsumeData', linkedForms.length) }}</p>
+      <table v-if="linkedForms.length > 0" class="table">
+        <tbody>
+          <tr v-for="(form) in linkedForms" :key="form">
+            <td>
+              <router-link :to="publishedFormPath(projectId, form.xmlFormId)" :title="form.name">{{ form.name }}</router-link>
```
matthew-white, line 24: If we're going to add a `title` attribute with the possibility of text truncation in mind, then the `<td>` element should include the `text-overflow-ellipsis` mixin.

src/components/dataset/overview/connection-to-forms.vue
```diff
@@ -0,0 +1,116 @@
...
+  name: 'ConnectionToForms',
+  components: {
+    SentenceSeparator,
+    SummaryItem,
+    ExpandableRow
+  },
+  mixins: [routes()],
+  props: {
+    properties: {
+      type: Object,
```
matthew-white, line 57: `type: Array`?

src/components/dataset/overview/dataset-properties.vue
```diff
@@ -0,0 +1,87 @@
...
+
+<script>
+import routes from '../../../mixins/routes';
+
+export default {
+  name: 'DatasetProperties',
+  mixins: [routes()],
+  props: {
+    properties: {
+      type: Object,
```
matthew-white, line 58: `type: Array`?

src/components/dataset/overview/linked-forms.vue
```diff
@@ -0,0 +1,74 @@
...
+
+export default {
+  name: 'LinkedForms',
+  components: {
+    SummaryItem
+  },
+  mixins: [routes()],
+  props: {
+    linkedForms: {
+      type: Object,
```
matthew-white, line 45: `type: Array`?

src/components/dataset/overview/dataset-properties.vue
```diff
@@ -0,0 +1,87 @@
...
+    </thead>
+    <tbody>
+      <template v-for="(property) in properties" :key="property.name">
+        <tr>
+          <!-- we have to show property name even if there is no associated form -->
+          <td :rowspan="property.forms.length || 1">
+            {{ property.name }}
+          </td>
+          <td>
+            <router-link v-if="property.forms.length > 0" :to="publishedFormPath(projectId, property.forms[0].xmlFormId)" :title="property.forms[0].name">
```
matthew-white, line 29: Similar to above, the `<td>` here should probably include `text-overflow-ellipsis` if the form name might be truncated.

matthew-white (commented):
I still need to take a look at tests, and I'm leaving some small comments, but this is looking great so far.

src/components/expandable-row.vue
```diff
@@ -0,0 +1,85 @@
...
+      <slot name="title"></slot>
+    </div>
+    <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events, vuejs-accessibility/interactive-supports-focus -->
+    <div role="button" class="caption-cell" @click.prevent="toggleExpanded">
+      <slot name="caption"></slot>
+    </div>
+    <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events, vuejs-accessibility/interactive-supports-focus -->
+    <div role="button" class="button-cell" @click.prevent="toggleExpanded">
+      <span v-if="!expanded" class="icon-caret-left"></span>
+      <span v-else class="icon-caret-down"></span>
```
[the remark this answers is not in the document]
matthew-white: Just to be a little more clear, my thought is to add an `<a role="button">` element within the `<div>`, move the `@click` from the `<div>` to the `<a>`, remove `role="button"` from the `<div>`, and insert the icons inside the `<a>`.

src/components/dataset/overview/dataset-properties.vue
```diff
@@ -0,0 +1,87 @@
...
+        <th>{{ $t('header.name') }}</th>
+        <th>{{ $t('header.updatedBy') }}</th>
+      </tr>
+    </thead>
+    <tbody>
+      <template v-for="(property) in properties" :key="property.name">
+        <tr>
+          <!-- we have to show property name even if there is no associated form -->
+          <td :rowspan="property.forms.length || 1">
+            {{ property.name }}
```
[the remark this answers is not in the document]
sadiqkhoja: yes, it is wrapping. We have enough space at the bottom to show quite a lengthy property name but I can change it if it is important

test/components/expandable-row.spec.js
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it('collapses details', async () => {
+    const c = mount(ExpandableRow, {
+      data() {
+        return {
+          expanded: true
+        };
+      }
+    });
```
matthew-white, line 54: Instead of overriding the `data()` function, I think you could test what happens when `.button-cell` is clicked twice.

src/components/dataset/show.vue
```diff
@@ -0,0 +1,124 @@
...
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+
+<template>
+  <div id="dataset-show">
+    <page-back v-show="dataExists" :to="projectPath('datasets')">
+      <template #title>{{ project.dataExists ? project.nameWithArchived : '' }}</template>
+      <template #back>{{ $t('back') }}</template>
+    </page-back>
```
matthew-white, line 18: Could you add tests of the back link?

- The link points to the right page. (You can just check the `to` prop of the `PageBack` component. Check out the parallel test in `SubmissionShow` for inspiration.)
- The back link show's the project's name. (`FormHead` has a parallel test.)
- The back link appends "(archived)" to the project's name if the project is archived. (`FormHead` has a parallel test.)

src/routes.js
```diff
@@ -494,7 +494,28 @@ const routes = [
...
+    key: ({ projectId, datasetName }) =>
+      `/projects/${projectId}/datasets/${encodeURIComponent(datasetName)}`,
+    children: [
+      asyncRoute({
+        path: '',
+        component: 'DatasetOverview',
+        props: true,
+        loading: 'tab',
+        meta: {
+          title: () => [dataset.name],
```
matthew-white, line 511: It'd be great to add a test of this in test/router.spec.js.

src/routes.js
```diff
@@ -494,7 +494,28 @@ const routes = [
...
+          validateData: {
+            project: () => project.permits('dataset.list')
```
matthew-white, line 513: It'd be great to add a test of this in test/router.spec.js. A Data Collector should be redirected if they try to navigate to this route.

matthew-white, line 513: Shouldn't the verb be `dataset.read` instead of `dataset.list`? You can add new verbs in test/data/seed.js.

src/routes.js
```diff
@@ -494,7 +494,28 @@ const routes = [
...
+    path: '/projects/:projectId([1-9]\\d*)/datasets/:datasetName',
+    component: 'DatasetShow',
+    props: true,
+    loading: 'page',
+    key: ({ projectId, datasetName }) =>
+      `/projects/${projectId}/datasets/${encodeURIComponent(datasetName)}`,
+    children: [
+      asyncRoute({
+        path: '',
+        component: 'DatasetOverview',
```
matthew-white, line 507: It'd be great to add a test that a user who is not logged in is redirected to login if they try to navigate to this route. (Look for `requireLogin` in test/router.spec.js.)

src/components/dataset/overview/connection-to-forms.vue
```diff
@@ -0,0 +1,116 @@
...
+  <summary-item id="connection-to-forms" icon="magic-wand">
+    <template #heading>
+      {{ propertiesByForm.length }}
+    </template>
+    <template #body>
+      <p>{{ $t('formsCreateEntities') }}</p>
+      <div class="div">
+        <expandable-row v-for="(form) in propertiesByForm" :key="form.xmlFormId">
+          <template #title>
+            <router-link :to="publishedFormPath(projectId, form.xmlFormId)" :title="form.name">{{ form.name }}</router-link>
```
matthew-white, line 23: Could you add assertions about this link in testing, of its `to` prop and its `title` attribute? And add another test that the `to` prop is correct if the user is a project viewer?

test/components/expandable-row.spec.js
```diff
@@ -0,0 +1,58 @@
...
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+*/
+
+import ExpandableRow from '../../src/components/expandable-row.vue';
+import { mount } from '../util/lifecycle';
+
+describe('Dataset summary row', () => {
```
matthew-white, line 16: `'ExpandableRow'` instead of `'Dataset summary row'` I think.

src/components/expandable-row.vue
```diff
@@ -0,0 +1,104 @@
...
+    </div>
+    <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events, vuejs-accessibility/interactive-supports-focus -->
+    <div role="button" class="caption-cell" @click.prevent="toggleExpanded">
+      <slot name="caption"></slot>
+    </div>
+
+    <div class="button-cell">
+      <a href="javascript:void(0)" role="button" @click.prevent="toggleExpanded">
+        <span v-if="!expanded" class="sr-only">{{ $t('common.expand') }}</span>
+        <span v-else class="sr-only">{{ $t('common.collapse') }}</span>
```
matthew-white, line 25: Good call on adding these `.sr-only` elements!

src/locales/en.json5
```diff
@@ -361,7 +362,12 @@
       "comma": ","
     },
     // This is a title shown above a section of the page.
-    "datasetsPreview": "Datasets Preview"
+    "datasetsPreview": "Datasets Preview",
+    // {inform} is the number of dataset properties defined by the form.
+    // {count} is the total number of properties. The string will be pluralized based on {count}.
+    "propertiesCount": "{inform} of {count} property | {inform} of {count} properties",
+    "expand": "Expand",
+    "collapse": "Collapse"
```
matthew-white, line 370: Would you mind moving these into the `action` block?

src/components/dataset/overview/dataset-properties.vue
```diff
@@ -0,0 +1,93 @@
...
+          </td>
+          <td>
+            <router-link v-if="property.forms.length > 0" :to="publishedFormPath(projectId, property.forms[0].xmlFormId)" :title="property.forms[0].name">
```
matthew-white, line 29: Could you add assertions about the links in this component, of the `to` prop and the `title` attribute? And add another test that the `to` prop is correct if the user is a project viewer?

src/components/dataset/overview/linked-forms.vue
```diff
@@ -0,0 +1,79 @@
...
+    <template #heading>
+      {{ linkedForms.length }}
+    </template>
+    <template #body>
+      <p>{{ $tc('formsConsumeData', linkedForms.length) }}</p>
+      <table v-if="linkedForms.length > 0" class="table">
+        <tbody>
+          <tr v-for="(form) in linkedForms" :key="form.xmlFormId">
+            <td>
+              <router-link :to="publishedFormPath(projectId, form.xmlFormId)" :title="form.name">{{ form.name }}</router-link>
```
matthew-white, line 24: Could you add assertions about this link in testing, of its `to` prop and its `title` attribute? And add another test that the `to` prop is correct if the user is a project viewer?

matthew-white (approved):
More small comments, largely about tests to add. Things are looking good! 🎉 Feel free to merge once the comments are resolved.

src/components/dataset/overview/dataset-properties.vue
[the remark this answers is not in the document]
matthew-white: If it's wrapping, that sounds good to me!

src/routes.js
sadiqkhoja: need to add `dataset.read` in backend before changing here, will create separate PR for both repositories

src/components/dataset/overview/connection-to-forms.vue
sadiqkhoja: Project viewer also has 'form.read', so link will be same. Instead I have added unit tests in mixins/routes.js for validating `publishedFormPath`

test/mixins/routes.spec.js
```diff
@@ -64,4 +64,16 @@ describe('mixins/routes', () => {
...
   });
+
+  describe('publishedFormPath', () => {
+    it('returns form overview URL when user can route', () => {
+      mixin.methods.canRoute = () => true;
+      mixin.methods.publishedFormPath(1, 'f').should.equal('/projects/1/forms/f');
+    });
+
+    it('returns submissions page URL when user can not route', () => {
+      mixin.methods.canRoute = () => false;
```
matthew-white, line 75: This is a great use case for Sinon:

```js
sinon.replace(mixin.methods, 'canRoute', () => false);
```

That will automatically restore `mixin.methods.canRoute` to its original value after the test. Currently, if we ever add tests after this one, `mixin.methods.canRoute` will still be `() => false`.

No need to change anything, just wanted to put Sinon on your radar as a useful tool!

src/routes.js
matthew-white: Should we also check for `form.list` in that endpoint since it returns forms? It probably doesn't matter that much, since right now anyone who can `dataset.list` can also `form.list`. But I have this comment in Backend in mind: https://github.com/getodk/central-backend/blob/f5ccb2611c2ae283df458943aef9ef8bb1680742/lib/resources/forms.js#L135-L150

Issue #648, opened 2022-11-16
Instant tooltips


matthew-white:
Moved from the v2022.3 release criteria:

- On anything with title text, show our own popup with the text right away, rather than make people wait for the browser.
  - GitHub has done this forever, for example. It’s a big big usability win.
- Focusing on a title-text element with keyboard navigation should ideally also pop the tip.
- (Nice to have:) Today there are elements where if their text content overflows, we truncate the text with an ellipsis; we also add title text. The title text is added even if the text content does not actually end up overflowing: the title text is added if the text content might overflow. However, ideally we show an instant tooltip only if the text content actually overflows.

srujner:
**Tested wit Success!**

dbemke:
**Tested with Success!**

matthew-white:
I've filed an issue with design review feedback at #753. We didn't end up making those changes for .2, but we're planning to do so for .3.

PR #704, opened 2023-01-19
Replace title attribute with instant tooltips

matthew-white:
This PR adds instant tooltips, closing #648. My goals were to:

- Add a tooltip wherever we currently use the `title` attribute.
- Avoid redundant tooltips. That is, don't show tooltips that just duplicate visible text.
- Make tooltips accessible. The `title` attribute is associated with [accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#accessibility_concerns), so I wanted our new tooltips to address those concerns.
- Make it easy to use tooltips, ideally as easy as the `title` attribute was. Adding a tooltip shouldn't require a component or a composable. It should be easy for the tooltip to be dynamic or conditional.

The solution I landed on was to add our own `v-tooltip` [directive](https://vuejs.org/guide/reusability/custom-directives.html). The directive uses the [Floating UI](https://floating-ui.com/) package and the Bootstrap `tooltip` class under the hood. I will show a few examples, then discuss some of the choices I made.

### Case 1: Truncated text

Our most common use of tooltips is for truncated text. Before, we would do something like:

```vue
<span :title="someText">{{ someText }}</span>
```

Now it will be:

```vue
<span v-tooltip.text>{{ someText }}</span>
```

This will only show a tooltip if the text is actually truncated. If a tooltip is shown, it won't be read by screen readers because it would be redundant.

### Case 2: Disabled buttons and links

Our next most common use of tooltips is for disabled buttons and links. Before, we would do something like:

```vue
<button disabled title="This button is disabled." ...>
  Do something
</button>

<a class="disabled" title="This link is disabled." ...>
  Link to something
</a>
```

Now:

```vue
<button aria-disabled="true" v-tooltip.aria-describedby="This button is disabled." ...>
  Do something
</button>

<a class="disabled" v-tooltip.aria-describedby="This link is disabled." ...>
  Link to something
</a>
```

Here, `v-tooltip` will set an [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) attribute on the button or link. The `aria-describedby` attribute will link to the tooltip, allowing screen readers to know that the button/link and tooltip are connected. I found [this page](https://inclusive-components.design/tooltips-toggletips/) helpful when thinking about tooltips and accessibility.

In #700, I replaced the `disabled` attribute with `aria-disabled="true"` for most elements. See that PR for why that was needed.

### Case 3: Buttons and links without text

In some cases, we don't show text in buttons and links, only icons. We often add a `title` attribute in those cases, for example:

```vue
<button title="Edit" ...>
</button>
```

We could use `v-tooltip.aria-describedby` here, but I think the [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) attribute is a better fit. `v-tooltip` allows you to pull from `aria-label`:

```vue
<button aria-label="Edit" v-tooltip.aria-label ...>
</button>
```

Note that like `v-tooltip.text`, `v-tooltip.aria-label` isn't provided a value: both types of tooltips retrieve their text from somewhere in the DOM. `v-tooltip.aria-describedby` has to be provided a value because its text isn't otherwise in the DOM. I was tempted to use the [`aria-description`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description) attribute instead of `aria-describedby` to make these uses more consistent. However, `aria-description` appears to still be a draft attribute. I also could have used [`aria-labelledby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) instead of `aria-label` in order to make things more consistent. However, I think that would require rendering more tooltip elements immediately, which we have to do for `aria-describedby`, and I wanted to avoid that.

### Case 4: Icons with tooltips

It is fairly common for us to add a tooltip to an icon that is outside a button or link. For example:

```vue
<span>Chelsea</span>
<span class="icon-plus-circle" title="New user"></span>
```

However, `aria-describedby` and `aria-label` don't always seem to play well with non-interactive/static/text elements (see [here](https://www.tpgi.com/short-note-on-aria-label-aria-labelledby-and-aria-describedby/) and [here](https://www.davidmacd.com/blog/does-aria-label-override-static-text.html)). Instead of using these ARIA attributes, one approach is to add text that will be read by a screen reader but that isn't initially visible on-screen. We can use the Bootstrap `sr-only` class for that: adding an `.sr-only` element after the icon, then telling `v-tooltip` to show the text of the `.sr-only` element for the tooltip. That way, the text will be shown on-screen as a tooltip and will also be read by screen readers. We end up with:

```vue
<span>Chelsea</span>
<span class="icon-plus-circle" v-tooltip.sr-only></span>
<span class="sr-only">New user</span>
```

### Other cases

In rare cases, none of these patterns really work. You can use `v-tooltip.no-aria` to add a tooltip that won't be read by screen readers. For example, I wasn't sure how to adjust the `DateTime` component so that screen readers will read both the relative and absolute date/time formats (or even whether that would always be the best user experience).

## External packages

Bootstrap 3 has a plugin for tooltips. However, I'm looking for ways to decrease our use of Bootstrap plugins, so I didn't want to use the Bootstrap tooltip plugin unless there was a strong reason to do so. The plugin also didn't do everything that I wanted. However, I was happy to use the Bootstrap tooltip CSS, which is small.

I checked out the tooltip packages for Vue listed [here](https://github.com/vuejs/awesome-vue#tooltip). Some didn't support Vue 3, and others didn't seem widely used, but there were a couple of interesting options.

One thing that I noticed while checking out the packages is that many packages use Popper under the hood. Bootstrap 5 also uses Popper.

One Vue package that seems to have gained a lot of traction is [Floating Vue](https://floating-vue.starpad.dev/). The author of the package is a member of the Vue core team. They use [Floating UI](https://floating-ui.com/) under the hood, which forked off Popper at some point.

Besides Floating Vue, the main package that stood out to me was [VueTippy](https://vue-tippy.netlify.app/).

Ultimately, those packages didn't do quite what I wanted. For example, I really wanted to be able to use something like `v-tooltip.text`. However, it was useful to see how these packages approached things.

After taking a quick look at Popper and Floating UI, Floating UI seemed a little simpler, so I went with that. Floating UI has a minimal Vue package of their own that exports a composable. But it wasn't very useful, so I just used the main Floating UI package.

## Performance questions

One thing that we should look out for is performance issues around the submissions table, because in theory, every cell outside metadata columns could have a tooltip to show truncated text. That would result in adding individual event listeners to every cell. However, I have tried to put up some limits here:

- We don't add tooltips for field types whose text shouldn't be truncated: `int`, `decimal`, `date`, `time`, `dateTime`, and `geopoint`. (`geopoint` columns are styled to be up to twice as wide as others.)
- In general, `v-tooltip.text` adds event listeners only if it seems at all possible that the text could be truncated. Specifically, `v-tooltip.text` adds event listeners only if the text is more than 5 characters, which means that empty cells in particular are excluded. (I chose 5 to be very conservative, knowing that we use `v-tooltip.text` in a lot of different components. A larger number would very likely also be fine.)

I'm hopeful that we won't encounter performance issues. However, I wanted to mention an alternative implementation that I considered. Right now, we add individual event listeners on `mouseenter` and `mouseleave`. Instead, we could add a single event listener on `mouseover` to the body and a single event listener on `mouseout`, then use event bubbling. However, I decided to use `mouseenter` and `mouseleave` because:

- `mouseover` and `mouseout` are associated with their own performance concerns, noted [here](https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseenter_event#behavior_of_mouseover_events).
- Most tooltip packages I looked at use `mouseenter` and `mouseleave`.
- In practice, I'm not sure that the memory cost of using `mouseenter` and `mouseleave` is as high as it may seem. Even though we add individual event listeners, the event handlers are shared: there is a single function across tooltips to handle show (on `mouseenter`) and a single function to handle hide (on `mouseleave`).

That said, I have previously used `mouseover` in `SubmissionTable` without much issue: there, hovering over a `SubmissionDataRow` component causes actions to be shown in the corresponding `SubmissionMetadataRow` component. If we end up experiencing performance issues, I'd be happy to experiment with `mouseover` and `mouseout`. There are also some parallel considerations around `focus`/`blur` (which we use) vs. `focusin`/`focusout` (which bubble).

I also considered a different angle. For elements that use `v-tooltip.text` and have text of more than 5 characters, we evaluate on hover/focus whether the text is truncated and show a tooltip if so. Instead, we could try to evaluate that eagerly, then only add event listeners to elements whose text is truncated. However, I wouldn't be surprised if that approach came with its own performance concerns, since that evaluation is cheap but not free. There is also the complication that there may be time between when an element is rendered and when it is visible. For example, on the submissions page, if the request to &hellip;/submissions/keys takes longer than the OData request, then the submissions table will be fully rendered but hidden between the two responses. During that time, we don't have a way to test for text truncation, so we would somehow need to wait for the table to be visible before evaluating whether each cell has truncated text.

Again, I'm hopeful that we won't encounter performance issues. If we do, then there are things we can try out.

## Things I didn't do / other risks

I ran out of time to add unit tests of `v-tooltip`. I think there is decent test coverage of `v-tooltip` from our existing tests: I added custom assertions so that existing tests can continue to make assertions about tooltips. However, whenever there's time or whenever we make a change to tooltips, I think it'd be a good idea to add unit tests as well, given how much is happening in that code.

Floating UI has a way to auto-update the position of a floating element, but I didn't use that. I don't think we really need that with tooltips. We can add it later if we need to.

In some cases, I added a wrapper element, such as an extra `<span>`, in order to position a tooltip correctly. There's a chance that that could lead to issues, for example, CSS selectors that no longer match. However, I think that's unlikely, and I think regression testing would catch most of those cases.


src/components/form/head.vue
```diff
@@ -33,34 +33,37 @@ except according to the terms contained in the LICENSE file.
               to render the tabs even if the form does not have a published
               version (in which case canRoute() will return `false`). -->
               <li v-if="rendersFormTabs" :class="formTabClass('')"
-                :title="formTabTitle" role="presentation">
-                <router-link :to="tabPath('')">
+                role="presentation">
+                <router-link :to="tabPath('')"
+                  v-tooltip.aria-describedby="formTabDescription">
```
matthew-white, line 38: Before, the `<li>` had the `title` attribute, but now I've put the tooltip on the link. I don't think it makes a big difference for users using a mouse, but for screen readers, I think it's helpful for the link to get the `aria-describedby` attribute.

src/components/form/row.vue
```diff
@@ -21,7 +21,7 @@ except according to the terms contained in the LICENSE file.
     <td v-for="reviewState of visibleReviewStates" :key="reviewState" class="review-state">
       <template v-if="form.publishedAt != null">
         <link-if-can :to="formPath(form.projectId, form.xmlFormId, `submissions?reviewState=${urlFilterEncode.get(reviewState)}`)"
-          :title="$t(`reviewState.${reviewState}`)">
+          v-tooltip.no-aria="$t(`reviewState.${reviewState}`)">
```
matthew-white, line 24: I think screen readers will read the column header when you enter a cell, so as long as the column headers are set up usefully, there doesn't need to be additional text describing the cell. Because of that, I chose `v-tooltip.no-aria` here. (For this table, we could probably update the column headers so that they're more useful to screen readers, but I think that's something we can do later.)

src/components/form/row.vue
```diff
@@ -59,10 +59,11 @@ except according to the terms contained in the LICENSE file.
...
       <template v-if="form.state !== 'closed'">
         <template v-if="form.state === 'closing'">
-          <span :title="$t('formClosingTip')">
-          <span class="icon-clock-o closing-icon"></span>
-          <span>{{ $t('formState.closing') }}</span>
+          <span aria-hidden="true" v-tooltip.sr-only>
+            <span class="icon-clock-o closing-icon"></span>
+            <span>{{ $t('formState.closing') }}</span>
           </span>
+          <span class="sr-only">{{ $t('formClosingTip') }}</span>
```
matthew-white, line 66: Since screen readers will read `$t('formClosingTip')`, I don't think it's useful for them to also read `$t('formState.closing')` immediately beforehand, which is why I set `aria-hidden="true"` on the first `<span>` and didn't add a `&nbsp;` to the `.sr-only` element.

src/components/multiselect.vue
```diff
@@ -20,7 +20,7 @@ except according to the terms contained in the LICENSE file.
       @keydown="toggleAfterEnter" @mousedown.prevent>
       <option value="">{{ selectOption }}</option>
     </select>
-    <span class="form-label">{{ label }}</span>
+    <span class="form-label" aria-hidden="true">{{ label }}</span>
```
matthew-white, line 23: The `aria-label` of the `<select>` element is set to `label`, so `label` doesn't need to be surfaced to screen readers here.

src/components/multiselect.vue
```diff
@@ -49,13 +49,16 @@ except according to the terms contained in the LICENSE file.
...
+            <li v-for="({ value, key = value, text = value, description }, i) in options"
               :key="key" :class="{ 'search-match': searchMatches.has(value) }">
               <div class="checkbox">
                 <label>
-                  <input type="checkbox" :data-index="i">
-                  <span :title="title">{{ text }}</span>
+                  <input type="checkbox" :data-index="i"
+                    :aria-describedby="description != null ? descriptionId(i) : null">
+                  <span v-if="description == null" v-tooltip.text>{{ text }}</span>
+                  <span v-else v-tooltip.no-aria="description">{{ text }}</span>
```
matthew-white, line 59: In `SubmissionDecrypt`, the tooltip is over the checkbox as a whole: both the label text and the checkbox `<input>`. Here, we only show the tooltip over the text. That's how we had it set up before and I think is a reasonable user experience.

matthew-white, line 59: This is `v-tooltip.no-aria`, but the description will be surfaced to screen readers using the `aria-describedby` attribute above. This is an example of how `v-tooltip.no-aria` can be used in complex/rare situations.

src/components/project/home-block.vue
```diff
@@ -13,11 +13,13 @@ except according to the terms contained in the LICENSE file.
...
-      <span v-if="project.keyId" class="encrypted badge"
-        :title="$t('encryptionTip')">
-        <span class="icon-lock"></span>
-        {{ $t('encrypted') }}
+      <template v-if="project.keyId">
+        <span class="encrypted badge" aria-hidden="true" v-tooltip.sr-only>
+          <span class="icon-lock"></span>
+          {{ $t('encrypted') }}
         </span>
+        <span class="sr-only">{{ $t('encryptionTip') }}</span>
```
matthew-white, line 21: Similar to my comment above in `FormRow`, since screen readers will read `$t('encryptionTip')`, I don't think it's useful for them to also read `$t('encrypted')` immediately beforehand — hence `aria-hidden="true"` on the first `<span>` and no `&nbsp;` in the `.sr-only` element.

src/components/project/user/row.vue
```diff
@@ -12,17 +12,15 @@ except according to the terms contained in the LICENSE file.
...
     </td>
     <td>
       <form>
         <div class="form-group">
           <select class="form-control" :value="selectedRoleId"
-            :aria-disabled="disabled || awaitingResponse"
-            :title="disabled ? $t('cannotAssignRole') : null"
             :aria-label="$t('field.projectRole')"
+            :aria-disabled="disabled || awaitingResponse"
+            v-tooltip.aria-describedby="disabled ? $t('cannotAssignRole') : null"
```
matthew-white, line 23: I reordered these attributes so that `aria-label` comes before `aria-describedby`.

src/components/submission/basic-details.vue
```diff
@@ -45,13 +45,13 @@ except according to the terms contained in the LICENSE file.
...
           <dd>
-            <span :title="submission.__system.deviceId">{{ submission.__system.deviceId }}</span>
+            <span v-tooltip.text>{{ submission.__system.deviceId }}</span>
           </dd>
         </div>
         <div v-if="submissionVersion.userAgent != null">
           <dt>{{ $t('userAgent') }}</dt>
           <dd id="submission-basic-details-user-agent">
-            <span :title="submissionVersion.userAgent">{{ submissionVersion.userAgent }}</span>
+            {{ submissionVersion.userAgent }}
```
matthew-white, line 54: We actually don't need a tooltip here because we added word-wrap to this specific element.

src/components/submission/data-row.vue
```diff
@@ -32,7 +33,7 @@ except according to the terms contained in the LICENSE file.
         <span class="encryption-overlay"></span>
       </td>
     </template>
-    <td :title="submission.__id">{{ submission.__id }}</td>
+    <td>{{ submission.__id }}</td>
```
matthew-white, line 36: The instance ID column can go wider than others (it has a larger `max-width`), so it shouldn't need a tooltip.

src/components/submission/decrypt.vue
```diff
@@ -21,27 +21,36 @@ except according to the terms contained in the LICENSE file.
...
+                :disabled="splitSelectMultiplesDescription != null"
+                :aria-describedby="splitSelectMultiplesDescribedBy">
               {{ $t('field.splitSelectMultiples') }}
             </label>
+            <p v-if="splitSelectMultiplesDescription != null"
+              id="submission-download-split-select-multiples-disabled"
+              class="sr-only">
+              {{ splitSelectMultiplesDescription }}
+            </p>
           </div>
```
matthew-white, line 36: This is a more complicated case, because we want `aria-describedby` on the `<input>`, but we want the tooltip on the wrapping `<label>`. It turns that `v-tooltip.sr-only` works well for this use.

src/components/submission/decrypt.vue
```diff
@@ -74,14 +83,13 @@ except according to the terms contained in the LICENSE file.
...
           <div class="submission-download-action"
             :class="{ disabled: noRepeat }">
-            <span class="submission-download-action-label"
-              :title="noRepeat ? $t('noRepeat') : null">
+            <span class="submission-download-action-label">
```
matthew-white, line 86: I tried using `v-tooltip` here, but that gives the label a separate tooltip (separately positioned) from the link below it. That seemed a little funny to me, but I think it's fine for just the link to get the tooltip.

src/components/user/row.vue
```diff
@@ -52,8 +52,9 @@ except according to the terms contained in the LICENSE file.
               {{ $t('action.resetPassword') }}&hellip;
             </a>
           </li>
-          <li :class="{ disabled }" :title="disabled ? $t('cannotRetire') : null">
+          <li :class="{ disabled }">
             <a class="retire-user" href="#"
+              v-tooltip.aria-describedby="disabled ? $t('cannotRetire') : null"
               @click.prevent="$emit('retire', user)">
               {{ $t('action.retire') }}&hellip;
             </a>
```
matthew-white, line 60: Similar to the tabs in `FormHead`, the `<li>` had the `title` attribute before, but now I've put the tooltip on the link. I don't think it makes a big difference for users using a mouse, but for screen readers, I think it's helpful for the link to get the `aria-describedby` attribute.

test/assertions.js
```diff
@@ -1,8 +1,48 @@
...
+import { wait } from './util/util';
+
 const unwrapElement = (elementOrWrapper) => (elementOrWrapper instanceof HTMLElement
   ? elementOrWrapper
   : elementOrWrapper.element);
+
+// Assertion.addSync() is like Assertion.add(), but the specified function can
+// be async.
+const { Assertion, AssertionError } = should;
+Assertion.addAsync = (name, f) => {
```
matthew-white, line 12: Should.js has async assertions which are `await`ed, for example:

```js
await somePromise.should.be.resolved();
await anotherPromise.should.be.rejectedWith('some error');
```

However, while Should.js provides a way to add custom synchronous assertions, I don't see a way to add custom async assertions. The tooltip assertions need to be async because `computePosition()` from Floating UI is async. I added this method as a way to add custom async assertions. If the assertion fails, a useful Should.js `AssertionError` is thrown.

test/assertions.js
```diff
@@ -35,13 +75,155 @@ should.Assertion.add('hidden', function hidden(computed = false) {
...
+/* Asserts that an element shows the entirety of its text in a tooltip if and
+only if its text is truncated. The assertion works by replacing the text with a
+long string, checking that that shows a tooltip, then replacing the text with a
+short string and checking that that does not show a tooltip. If the element does
+not truncate its text (by setting text-overflow) and does not have an ancestor
+that truncates its text, the assertion will try to insert a new ancestor into
+the tree to truncate its text. In other words, the assertion does not test that
+the element has truncated text, only that if it has truncated text, it will show
+that text in a tooltip. */
+Assertion.addAsync('textTooltip', async function textTooltip() {
```
matthew-white, line 157: I have mixed feelings about this assertion. It's used in a lot of tests as a replacement for assertions about the `title` attribute. The assertion in the test itself is simple to use. However, under the hood, things are pretty complicated here. Given how simple `v-tooltip.text` is, part of me wonders how important it is to be verifying these tooltips. Maybe this assertion isn't really needed? I guess we have it, and it's not likely to change, so we might as well keep it. 🤔

test/components/actor-link.spec.js
```diff
@@ -8,21 +8,21 @@ import { mount } from '../util/lifecycle';
 
 describe('ActorLink', () => {
   beforeEach(() => {
-    mockLogin({ displayName: 'Alice' });
+    mockLogin({ displayName: 'Alice Allison' });
```
matthew-white, line 11: Changed to be more than 5 characters so that there is a tooltip.

test/components/enketo/fill.spec.js
```diff
@@ -29,12 +29,13 @@ describe('EnketoFill', () => {
     });
     button.element.tagName.should.equal('BUTTON');
     button.attributes('aria-disabled').should.equal('true');
-    button.attributes().title.should.equal('Web Form is not available yet. It has not finished being processed. Please refresh later and try again.');
+    button.should.have.ariaDescription('Web Form is not available yet. It has not finished being processed. Please refresh later and try again.');
+    await button.should.have.tooltip();
```
matthew-white, line 33: It seemed useful to have separate assertions about the `aria-describedby` attribute and about the tooltip even though the two go hand-in-hand.

test/components/form-attachment/list.spec.js
```diff
@@ -45,13 +45,17 @@ describe('FormAttachmentList', () => {
...
+      });
       const component = await load('/projects/1/forms/f/draft/attachments', {
         root: false
       });
-      const td = component.get('td.form-attachment-list-name');
-      td.attributes().title.should.equal('foo.jpg');
+      const span = component.get('td.form-attachment-list-name span');
+      span.text().should.equal('foo.jpg');
+      await span.should.have.textTooltip();
     });
```
matthew-white, line 59: This is a new test. I realized that we didn't test the case where the Name column renders just text, not a link. In both cases, there should be a tooltip.

test/components/form/head.spec.js
```diff
@@ -91,31 +89,33 @@ describe('FormHead', () => {
       });
     });
 
-    it('disables tabs for a form without a published version', () => {
+    it('disables tabs for a form without a published version', async () => {
```
matthew-white, line 92: We want to `await` the tooltip assertion, so I made the whole test async.

test/components/multiselect.spec.js
```diff
@@ -275,7 +278,7 @@ describe('Multiselect', () => {
       });
       await toggle(component);
       await component.get('.search input').setValue('c');
-      const matches = component.findAll('.search-match');
+      const matches = component.findAll('.search-match label');
```
matthew-white, line 281: `.search-match` now contains the `p.sr-only` element, so we need to select on `label` as well.

test/components/submission/basic-details.spec.js
```diff
@@ -117,9 +117,7 @@ describe('SubmissionBasicDetails', () => {
           }
         }
       });
-      const span = component.get('dl :nth-child(6) dd span');
-      span.text().should.equal('Collect');
-      span.attributes().title.should.equal('Collect');
```
matthew-white, line 122: This element doesn't have a tooltip anymore because it word-wraps.

test/components/submission/data-row.spec.js
```diff
@@ -299,7 +333,6 @@ describe('SubmissionDataRow', () => {
     const td = mountComponent().findAll('td');
     td.length.should.equal(2);
     td[1].text().should.equal('foo');
-    td[1].attributes().title.should.equal('foo');
```
matthew-white, line 302: Instance IDs no longer have tooltips in this component.

src/directives/tooltip.js
```diff
@@ -0,0 +1,329 @@
...
+except according to the terms contained in the LICENSE file.
+*/
+
+/*
+Use the v-tooltip directive to add a tooltip to an element. The tooltip will be
+shown when the user hovers over the element or focuses it.
+
+There are a few ways to specify the text of a tooltip. In many cases, the text
+will be copied from some place where it already exists in the DOM.
+
```
sadiqkhoja, line 19: Can we list all available usage as a list here:

- v-tooltip.text
- v-tooltip.aria-describedby="someValue"
- v-tooltip.aria-label
- v-tooltip.sr-only
- v-tooltip.no-aria="someValue"

src/directives/tooltip.js
```diff
@@ -0,0 +1,329 @@
...
+const show = (event) => {
+  const tooltip = tooltips.get(event.target);
+  if (!tooltip.shown) tooltip.show();
+};
+const hide = (event) => {
+  const tooltip = tooltips.get(event.target);
+  if (tooltip.shown) tooltip.hide();
+};
+
+class Tooltip {
```
sadiqkhoja, line 111: I wish there was an alias `v-tooltip` that would default to `v-tooltip.text`. Something for later on

src/directives/tooltip.js
```diff
@@ -0,0 +1,329 @@
...
+  static maybeCreate(anchor, { modifiers, value }) {
+    const type = types.find(name => modifiers[name] === true);
```
sadiqkhoja, line 113: how about just 
`const type = types.find(name => modifiers[name]);` or 
`const type = types.find(name => !!modifiers[name]);`

src/directives/tooltip.js
```diff
@@ -0,0 +1,329 @@
...
+    For v-tooltip.sr-only, we always add event listeners because we can't easily
+    hook into the lifecycle of the .sr-only element. If/when an event listener
+    is triggered for show, we will evaluate then whether there is an .sr-only
+    element and whether it has text and show the tooltip if so.
+
+    For other tooltip types, we can know right away whether a tooltip should be
+    shown: we just need to check tooltip.text().
+    */
+    if (type === 'text'
+      ? anchor.textContent.trim().length > 5
```
sadiqkhoja, line 141: should we use innerText? textContent return content of the hidden children as well

src/directives/tooltip.js
```diff
@@ -0,0 +1,329 @@
...
+      // clientWidth === 0 if the element is inline.
+      while (closestBlock.clientWidth === 0) {
+        closestBlock = closestBlock.parentNode;
+        // The `null` case is possible in testing, because many components are
+        // not attached to the document, in which case clientWidth is also 0.
+        // The document.body case shouldn't happen, but it seems like a good
+        // safeguard.
+        if (closestBlock == null || closestBlock === document.body) return null;
+      }
+      return closestBlock.scrollWidth > closestBlock.clientWidth
```
sadiqkhoja, line 174: is this checking if text is truncated?

sadiqkhoja (approved):
I'm impressed 🎉

I have added few minor comments/questions, rest looks great.

src/directives/tooltip.js
matthew-white: Yes, that's the intention! Maybe I should leave a comment along those lines? 🤔

src/directives/tooltip.js
matthew-white: `v-tooltip.text` is definitely our most common use of `v-tooltip`, though `v-tooltip.aria-describedby` is also quite common. I feel like requiring a `.text` modifier makes things more explicit, and it's just a few extra characters. But if we start adding uses of `v-tooltip` and feel like it'd be nice to have a default, I think I'd be up for that change.

src/directives/tooltip.js
matthew-white: Just personally, I try not to rely on truthiness and try to use explicit boolean expressions. Here, `modifiers[name]` may be `undefined`, so `modifiers[name]` isn't consistently boolean. I think issa has had a similar preference, so you'll see a lot of `=== true` and `=== false` in Frontend and maybe even more so in Backend. It's not a strong preference though! I certainly don't have any issues with others reaching for that pattern.

src/directives/tooltip.js
matthew-white: Ooh good question. As we use `v-tooltip` today, I can't think of cases where the two would differ. For example, I'm pretty sure we only use `v-tooltip` on elements without children. It usually works best to place `v-tooltip` on the deepest possible element so that the tooltip is positioned correctly. `.should.have.textTooltip()` even makes the assertion that the element has no children.

That said, I could imagine cases in the future where the two would be different and where it'd be better to use `innerText`. I'll go ahead and use `innerText` instead of `textContent` when computing the tooltip text in `_text()`.

On this particular line, I think I want to continue using `textContent`. MDN [notes](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext):

> since `innerText` takes CSS styles into account, reading the value of `innerText` triggers a reflow to ensure up-to-date computed styles. (Reflows can be computationally expensive, and thus should be avoided when possible.)

This line is part of the `Tooltip.maybeCreate()` method, which is called for every element that is a candidate to have a tooltip. In some cases, the number of candidates could be much greater than the number of elements to which we ultimately add a tooltip. I think we should make this candidate evaluation step as efficient as possible. On the other hand, `_text()` will only use `innerText` when a tooltip is being shown or when a tooltip that is currently shown is updated. Only one or two tooltips should be shown at a time, so there aren't the same performance considerations there.

src/directives/tooltip.js
matthew-white: I've left a comment about this.

matthew-white:
I merged #702 and #703 into this PR and replaced the `title` attributes there with `v-tooltip`. Just a few lines changed, so I don't think those changes necessarily need to be reviewed.

PR #708, opened 2023-01-22
Prevent admin from invalidating their own password

matthew-white:

### DIFF
--- a/src/components/user/row.vue
+++ b/src/components/user/row.vue
@@ -46,8 +46,9 @@ except according to the terms contained in the LICENSE file.
               {{ $t('action.editProfile') }}
             </router-link>
           </li>
-          <li>
+          <li :class="{ disabled }">
             <a class="reset-password" href="#"
+              v-tooltip.aria-describedby="disabled ? $t('cannotResetPassword') : null"
               @click.prevent="$emit('reset-password', user)">
               {{ $t('action.resetPassword') }}&hellip;
             </a>
@@ -158,6 +159,9 @@ export default {
     "field": {
       "sitewideRole": "Sitewide Role"
     },
+    // An Administrator may reset the password for another Web User, but not for
+    // their own account.
+    "cannotResetPassword": "You may not reset your own password on this page. To change your password, edit your profile.",
     // An Administrator may retire other Web Users, but not their own account.
     "cannotRetire": "You may not retire yourself.",
     "action": {
--- a/test/components/user/reset-password.spec.js
+++ b/test/components/user/reset-password.spec.js
@@ -13,12 +13,22 @@ describe('UserResetPassword', () => {
     mockLogin({ email: 'alice@getodk.org', displayName: 'Alice' });
   });
 
-  it('toggles the modal', () =>
-    load('/users', { root: false }).testModalToggles({
-      modal: UserResetPassword,
-      show: '.user-row .reset-password',
-      hide: '.btn-link'
-    }));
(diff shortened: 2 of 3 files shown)


src/components/user/row.vue
```diff
@@ -46,8 +46,9 @@ except according to the terms contained in the LICENSE file.
               {{ $t('action.editProfile') }}
             </router-link>
           </li>
-          <li>
+          <li :class="{ disabled }">
             <a class="reset-password" href="#"
+              v-tooltip.aria-describedby="disabled ? $t('cannotResetPassword') : null"
               @click.prevent="$emit('reset-password', user)">
```
matthew-white, line 52: This is very similar to how we disable the retire button directly below.

test/components/user/reset-password.spec.js
```diff
@@ -13,12 +13,22 @@ describe('UserResetPassword', () => {
...
+        hide: '.btn-link'
+      }));
+
+    it('is disabled for the current user', async () => {
+      const component = await load('/users', { root: false });
+      const a = component.get('.user-row .reset-password');
+      a.element.parentNode.classList.contains('disabled').should.be.true();
+      a.should.have.ariaDescription(/^You may not reset your own password/);
+      await a.should.have.tooltip();
+    });
```
matthew-white, line 30: This is very similar to the test in test/components/user/retire.spec.js that the retire button is disabled for the current user.

matthew-white:
@alxndrsn, I've tagged you for review here with the thought that the changes here are very similar to how things are done elsewhere in the codebase. I figure you could probably verify that even without having written a lot of component code in Frontend, but let me know if it'd be better for me to tag someone else.

alxndrsn (approved):
LGTM.  Hopefully there is corresponding code in `odk-backend` to prevent this on the API side as well.

matthew-white:
> Hopefully there is corresponding code in `odk-backend` to prevent this on the API side as well.

I figure that if an admin wants to use the API to invalidate their own password, YOLO. Their session token won't work for future requests, but that's an outcome they should be prepared for after invalidating their password. The API will invalidate a password during a password reset only if `?invalidate=true` is specified, so I think it would be pretty unusual for an admin to do that for their own account. If they really want to, I don't think we necessarily need to stop them. But I wanted to stop them in Frontend because invalidating their own password would lead to the issues described in #705, and that issue is something that we still need to address on our end. I could be convinced to do this in Backend as well, but my instinct is that it's probably not needed.

srujner:
@matthew-white
What about resetting the admin password from the login screen? Should it be possible to reset it from the email token?
Screenshot(56)

matthew-white:
Good question. It should be possible to reset anyone's password using that process. The difference between requesting a password reset from the page in the screenshot above and requesting it from the Users page is that when a reset is initiated from the Users page, the user's password is immediately invalidated. In contrast, the page above is only available to someone who isn't logged in already. If you aren't logged in, you can request a password reset for anyone, but that won't invalidate anyone's existing password. (Otherwise you could invalidate someone's password just by knowing their email address.)

An admin shouldn't be able to request a reset of their own password from the Users page because that would immediately invalidate their password, which would cause other issues. However, if the admin isn't logged in, they should be able to request a password reset from the page above; it just won't invalidate their password. Does that make sense?

srujner:
Yes, It's pretty clear now :)

dbemke:
Tested with success!

srujner:
**Tested with success!**

PR #710, opened 2023-01-23
Fix test of DatasetRow that can fail during certain times of day

matthew-white:
See the commit message for notes. It's a funny issue!

### DIFF
--- a/test/components/dataset/row.spec.js
+++ b/test/components/dataset/row.spec.js
@@ -34,7 +34,7 @@ describe('DatasetRow', () => {
   });
 
   it('shows the newest entity timestamp', () => {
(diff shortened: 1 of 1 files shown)


matthew-white:
Another way to solve this would be to mock the time zone:

```js
setLuxon({ defaultZoneName: 'UTC' });
```

Then Luxon will act as if the machine is in UTC, causing the UTC date and the date in the local time zone to match.

You could also check the `iso` prop of the `DateTime` component rather than the text of the `<time>` element. That's usually what I end up doing.

sadiqkhoja (approved):
it was failing for me earlier, it's working now. I must have been doing something wrong.

PR #711, opened 2023-01-23
Address warnings in test of useDisabled()

matthew-white:
See the commit message for notes. We should probably try to find a way for CircleCI to check for warnings like these. I'll file an issue for that.


sadiqkhoja (approved):
eslint should be able to catch these.

matthew-white:
> eslint should be able to catch these.

Good point, I think you're right that ESLint would have caught this if the template was in a .vue file. I wonder if there's a way to signal to ESLint that "this string is HTML" or "this object is a Vue component".

PR #714, opened 2023-01-23
Specify $select query parameter when requesting OData

### LINKED ISSUE #638: Use $select query parameter
This issue is for the Frontend aspect of getodk/central#205. @lognaturel writes:

> it won’t make a noticeable difference with small forms and fast connections but it will with big forms and slow connections

Now that getodk/central-backend#494 is merged, the next step is to think about Frontend. I'll use this issue for related notes. The `$select` query parameter will be used in two components:

# `SubmissionShow` (submission details page)

From the submission OData, I think `SubmissionShow` just uses `__id`, `__system`, and `meta/instanceName`.

However, things aren't as simple as requesting `$select=__id, __system, meta/instanceName`. I believe that some forms don't have a `meta/instanceName` field. For such a form, requesting `$select=meta/instanceName` would result in a 400.23 Problem response from Backend.

Another idea might be to `$select` all of `meta` rather than just `meta/instanceName`. If the resulting `meta` group has an `instanceName` field, then we'll use it; otherwise we'll know that there isn't an `instanceName` field. However, I also think it's possible that some forms won't have a `meta` group. For such a form, requesting `$select=meta` would again result in a 400.23 Problem. Backend also doesn't currently support `$select=meta`: right now, the only `ComplexType` that can be `$select`ed is `__system`.

Question: does the OData spec require that `$select`ing a field that doesn't exist result in an error response? Could we have Frontend request `$select=__id, __system, meta/instanceName`, and if `meta/instanceName` doesn't exist, Backend will ignore the `$select` or select just `__id` and `__system`?

When `SubmissionShow` sends a request for the submission OData, it also sends a request for the form fields. We could serialize those requests such that we request the form fields first, then request the submission OData once there's a response, `$select`ing `meta/instanceName` or not depending on whether it is among the form fields. However, that would require two round trips to the server before rendering the page, so I don't think we should take that approach.

**Maybe the easiest thing would be to pull the instance name off the latest submission version**, which `SubmissionShow` also requests, rather than off the OData.

Navigation to `SubmissionShow` will often happen from `SubmissionList` (the submissions page). By the time it renders a link to `SubmissionShow`, `SubmissionList` will know whether the form has a `meta/instanceName` field. It could add a query parameter to the link if the form has a `meta/instanceName` field (something like `?instanceName=true`). We could then have `SubmissionShow` send a request with `$select=__id, __system, meta/instanceName` if `instanceName=true`, send a request with `$select=id, __system` if `instanceName=false`, and send a request without `$select` if `instanceName` is not specified (for example, if the user navigates directly to `SubmissionShow`).

# `SubmissionList` (submissions + draft testing pages)

`SubmissionList` has a dropdown to select the fields to view. By default, this is the first 10 fields (excluding repeat groups). It'd be great if when we request the submission OData, we `$select` just those 10 fields (+ `__id` and `__system`). However, as with `SubmissionShow`, right now we request the first chunk of submission OData at the same time as we request the form fields. We could serialize those requests, but that doesn't seem ideal.

One approach might be to not specify `$select` for the first chunk of OData, but then specify it for later OData requests: for later chunks in the infinite scroll, when the field selection changes, or when a filter changes. Only the first OData request wouldn't specify `$select`.

Somewhat separately, I think we could also considering saving the column selection in a Frontend query parameter. Right now, when a filter changes, that change is reflected in the Frontend URL: a query parameter changes. That allows users to bookmark a filtered view of the submissions page. If we added the column selection to the Frontend URL as well, then users could bookmark a view of the submissions page with the columns that they wish to see. We could have `SubmissionList` look to that Frontend query parameter when it sends the request for the first chunk of OData: if the query parameter is specified, then Frontend will pass it along to Backend via `$select`. (Something to consider: the query parameter could be quite long. Could we encode or compress it somehow?)

@lognaturel also asked:

> issue request using most recently available selected fields falling back to no $select; cancel and reissue if fields come in and list doesn’t match?

I think something like that would work well once we start storing user preferences in Backend. (Though possibly wouldn't be needed if we make the other changes above?) We've also talked about adding saveable views of the submissions table, which I think would intersect with this.

matthew-white:
Closes #638. See the issue for a few different approaches I considered. The approach I take here ended up feeling not overly complicated and didn't require much code to change in src/. I also made changes to testing helper functions that are used to test the `$select` query parameter. I've split those changes into their own commits.


src/components/submission/list.vue
```diff
@@ -219,6 +219,12 @@ export default {
       }
       return conditions.length !== 0 ? conditions.join(' and ') : null;
     },
+    odataSelect() {
+      if (this.selectedFields == null) return null;
+      const paths = this.selectedFields.map(({ path }) => path.replace('/', ''));
```
matthew-white, line 224: The fields' paths are already mostly valid OData paths because Frontend requests &hellip;/fields?odata=true. However, we still need to remove the leading slash.

src/components/submission/list.vue
```diff
@@ -269,15 +275,19 @@ export default {
...
           this.projectId,
           this.xmlFormId,
           this.draft,
-          query
+          {
+            $top: this.top(skip),
+            $skip: skip,
+            $count: true,
+            $wkt: true,
+            $filter: this.odataFilter,
```
matthew-white, line 288: Previously, `$filter` was not specified at all if `this.odataFilter == null`. However, there's no harm in always specifying it. If the value of a query parameter is `null` or `undefined`, then `apiPaths.odataSubmission()` won't include it in the query string. I think the code reads more easily with all the query parameters grouped together like this.

test/components/submission/filters.spec.js
```diff
@@ -229,7 +220,7 @@ describe('SubmissionFilters', () => {
           ]);
         })
         .beforeEachResponse((_, { url }) => {
-          const match = url.match(/&%24filter=__system%2FsubmissionDate\+ge\+([^+]+)\+and\+__system%2FsubmissionDate\+le\+([^+]+)(&|$)/);
+          const match = url.match(/&%24filter=__system%2FsubmissionDate\+ge\+([^+]+)\+and\+__system%2FsubmissionDate\+le\+([^&]+)(&|$)/);
```
matthew-white, line 223: A small mistake in the regular expression. It only started being an issue now that there is a query parameter that can come after `$filter`. Instead of parsing the query string like this, we could use something like `relativeUrl(url).searchParams.get('$filter')`. However, I think that's something we can do in the future; I just wanted to make this small change for now.

test/components/submission/list.spec.js
```diff
@@ -1,21 +1,39 @@
...
+  document.dispatchEvent(new Event('scroll'));
+  // eslint-disable-next-line no-param-reassign
+  component.vm.scrolledToBottom = method;
+};
+// eslint-disable-next-line consistent-return
+const scroll = (componentOrBoolean) => {
+  if (componentOrBoolean === true || componentOrBoolean === false)
+    return (component) => _scroll(component, componentOrBoolean);
+  _scroll(componentOrBoolean, true);
+};
```
matthew-white, line 36: A somewhat ugly helper function that makes the tests below much cleaner

test/components/submission/list.spec.js
```diff
@@ -386,10 +353,7 @@ describe('SubmissionList', () => {
             props: { top: () => 2 }
           })
             .complete()
-            .request(component => {
-              sinon.replace(component.vm, 'scrolledToBottom', () => true);
-              document.dispatchEvent(new Event('scroll'));
-            });
+            .testNoRequest(scroll);
```
matthew-white, line 356: Replacing `.request()` with `.testNoRequest()`. The two are basically aliases, and I think that makes the intent of the test clearer.

test/components/submission/list.spec.js
```diff
@@ -489,4 +437,72 @@ describe('SubmissionList', () => {
...
+        .complete()
+        .request(component =>
+          component.get('#submission-list-refresh-button').trigger('click'))
+        .beforeEachResponse((_, { url }) => {
+          $select(url).should.equal('__id,__system,g/s1,s2,s3,s4,s5,s6,s7,s8,s9,s10');
+        })
+        .respondWithData(() => testData.submissionOData(1, 0)));
+
+    it('specifies $select if a filter is changed', () =>
+      load('/projects/1/forms/f/submissions', { attachTo: document.body })
```
matthew-white, line 482: This test uses `load()`, mounting `App` and injecting a real router, instead of using `loadSubmissionList()`. That's because changing the filter will change the route.

ktuite (approved):

PR #715, opened 2023-01-23
new: show linked datasets on form overview page

sadiqkhoja:
Added linked dataset widget on Form overview page:

### DIFF
--- a/src/components/dataset/summary/row.vue
+++ b/src/components/dataset/summary/row.vue
@@ -11,16 +11,20 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <div class="dataset-summary-row">
-    <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events, vuejs-accessibility/interactive-supports-focus -->
-    <div class="row" role="button" @click.prevent="toggleExpanded">
+    <div class="row">
       <div class="col-xs-6 dataset-name-wrap">
-        <div class="dataset-name text-overflow-ellipsis" v-tooltip.text>{{ dataset.name }}</div>
+        <div class="dataset-name text-overflow-ellipsis" v-tooltip.text>
+          <router-link :to="datasetPath(projectId, dataset.name)" v-tooltip.text>
+            {{ dataset.name }}
+          </router-link>
+        </div>
         <div v-if="dataset.isNew" class="dataset-new">
           <span class="icon-plus-circle"></span>
           {{ $t('new') }}
         </div>
       </div>
-      <div class="col-xs-6 properties-count">
+      <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events, vuejs-accessibility/interactive-supports-focus -->
+      <div class="col-xs-6 properties-count" role="button" @click.prevent="toggleExpanded">
         {{ $tcn('common.propertiesCount', dataset.properties.length, { inform: $n(inFormProperties.length, 'default') }) }}
         <!-- eslint-disable-next-line vuejs-accessibility/anchor-has-content -->
         <a href="javascript:void(0)" class="expand-button">
@@ -42,13 +46,19 @@ except according to the terms contained in the LICENSE file.
 
 <script>
 import SentenceSeparator from '../../sentence-separator.vue';
+import routes from '../../../mixins/routes';
 
 export default {
   name: 'DatasetSummaryRow',
   components: { SentenceSeparator },
+  mixins: [routes()],
   props: {
     dataset: {
       type: Object,
+      required: true
+    },
+    projectId: {
+      type: Number,
       required: true
     }
   },
@@ -110,7 +120,7 @@ export default {
         line-height: 28px;
     }
 
-    a {
+    a.expand-button {
         color: #666;
         font-size: 20px;
         vertical-align: middle;
--- a/src/components/form/overview.vue
+++ b/src/components/form/overview.vue
@@ -13,7 +13,7 @@ except according to the terms contained in the LICENSE file.
   <div id="form-overview">
     <div class="row">
       <div class="col-xs-6">
-        <form-overview-right-now v-if="form.dataExists"
+        <form-overview-right-now v-if="dataExists"
           @view-xml="showModal('viewXml')"/>
       </div>
       <div v-if="formDraft.dataExists" id="form-overview-draft" class="col-xs-6">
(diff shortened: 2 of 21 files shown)


src/request-data/resources.js
```diff
@@ -67,6 +67,10 @@ export default ({ i18n }, createResource) => {
 
   const formDraft = createResource('formDraft', () =>
     setupOption(data => shallowReactive(transformForm(data))));
+
+  // Form attachments (published)
+  createResource('formAttachments');
```
matthew-white, line 72: I think this should be a local resource rather than one available app-wide. (I really need to work on #674 so that I can document the difference!) The other resources in this file are used in the router and other places that mean that they can't be created in components. But the default is for a resource to be created in a component (locally); it will be removed when the component is unmounted. The `useRequestData()` composable returns a `createResource` function that can be used to create a local resource.

There's a question though of which component should create the local resource. The most common choice would be `FormShow`, in which case `formAttachments` should be added to `useForm()` along with the other form subresources there. For example, that's where `formDatasetDiff` is created. The advantage of creating a resource in `FormShow` is that it will persist across tabs (across child routes). The downside is that because it persists, there's more opportunity for it to become out-of-sync with other data. For example, `formDatasetDiff` is reset after a form draft is published, and we'd probably need to do the same for `formAttachments`.

Alternatively, you could create the resource in `FormOverview` or even `FormOverviewRightNow`. It would be removed once the component is unmounted as the user navigates to a different page. That would entail less complexity (no keeping data in sync, also fewer things to test). But it would result in an additional request if the user navigates to the form overview, then another form page, then back to the form overview.

Curious what you think makes the most sense here!

src/util/request.js
```diff
@@ -86,6 +86,7 @@ export const apiPaths = {
     return `/v1/test/${token}/projects/${projectId}/forms/${encodedFormId}/draft`;
   },
   publishFormDraft: formPath('/draft/publish'),
+  formAttachments: formPath('/attachments'),
```
matthew-white, line 89: Could you add a unit test for this?

src/components/form/overview/right-now.vue
```diff
@@ -128,7 +168,8 @@ export default {
       // The count of Submissions is shown separately above this text.
       "full": "{submissions} has been saved for this Form. | {submissions} have been saved for this Form.",
       "submissions": "Submission | Submissions"
-    }
+    },
+    "datasetsLinked": "Dataset attach data into this Form: | Datasets attach data into this Form:"
```
matthew-white, line 172: "Dataset attaches" instead of "Dataset attach" I think.

matthew-white, line 172: Could you add a comment that a number is shown separately above this text?

src/components/dataset/summary/row.vue
```diff
@@ -42,14 +46,20 @@ except according to the terms contained in the LICENSE file.
...
 export default {
   name: 'DatasetSummaryRow',
   components: { SentenceSeparator },
+  mixins: [routes()],
   props: {
     dataset: {
       type: Object,
       required: true
+    },
+    projectId: {
```
matthew-white, line 60: Do the tests of `DatasetSummaryRow` need to specify this prop now?

src/components/dataset/summary/row.vue
```diff
@@ -11,16 +11,20 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <div class="dataset-summary-row">
-    <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events, vuejs-accessibility/interactive-supports-focus -->
-    <div class="row" role="button" @click.prevent="toggleExpanded">
+    <div class="row">
       <div class="col-xs-6 dataset-name-wrap">
-        <div class="dataset-name text-overflow-ellipsis" v-tooltip.text>{{ dataset.name }}</div>
+        <div class="dataset-name text-overflow-ellipsis" v-tooltip.text>
+          <router-link :to="datasetPath(projectId, dataset.name)" v-tooltip.text>
```
matthew-white, line 17: It'd be great if you could test this `to` prop in the tests of `DatasetSummaryRow`.

src/components/form/overview/right-now.vue
```diff
@@ -10,8 +10,7 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <page-section id="form-overview-right-now"
-    :class="{ 'open-form': form.state === 'open' }">
+  <page-section id="form-overview-right-now" :class="{ 'open-form': form.state === 'open' }">
```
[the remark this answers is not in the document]
sadiqkhoja: vscode document formatter must have done this. What IDE and formatter are you using?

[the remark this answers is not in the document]
matthew-white: Ah that makes sense that it's from an IDE. I just use Sublime Text personally, but I know Kathleen uses VSCode.

src/components/form/overview/right-now.vue
```diff
@@ -27,24 +26,40 @@ except according to the terms contained in the LICENSE file.
...
         </template>
       </summary-item>
       <summary-item id="form-overview-right-now-state" :icon="stateIcon">
         <template #heading>
-          {{ $t(`formState.${form.state}`) }}
+          {{ $t(`formState.${form.state }`) }}
         </template>
         <template #body>
-          <p>{{ $t(`stateCaption.${form.state}`) }}</p>
+          <p>{{ $t(`stateCaption.${ form.state }`) }}</p>
```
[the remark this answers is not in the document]
matthew-white: Hmm that's weird. 🤔 ESLint seems to be passing on `master`. Why does ESLint say?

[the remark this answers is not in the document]
sadiqkhoja: oh it's vscode 😒

matthew-white (approved):
Looks great! Things are making a lot of sense to me around how the `publishedAttachments` resource is used.

src/components/form/overview.vue
```diff
@@ -13,7 +13,7 @@ except according to the terms contained in the LICENSE file.
   <div id="form-overview">
     <div class="row">
       <div class="col-xs-6">
-        <form-overview-right-now v-if="form.dataExists"
+        <form-overview-right-now v-if="dataExists"
```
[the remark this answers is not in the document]
matthew-white: To be clear, I think you could add a `<loading>` component right below `<form-overview-right-now>`. It's OK and even good if the right half of the page loads and is visible before the left half.

PR #719, opened 2023-01-25
Smaller form ID on form list when name duplicated

### LINKED ISSUE #677: form list: duplicate-name formId fallback display text is huge
image

probably it should be a bit smaller than the main text

ktuite:
Closes #677.

Font size is now 12px. On the form list page, it used to be 18px. Too small, or just right?

(on project page)
Screen Shot 2023-01-24 at 4 11 20 PM

(on project list / home page)
Screen Shot 2023-01-24 at 4 12 21 PM

### DIFF
--- a/src/components/form/row.vue
+++ b/src/components/form/row.vue
@@ -164,6 +164,7 @@ export default {
 
   .duplicate-form-id {
     font-family: $font-family-monospace;
+    font-size: 12px;
     padding-left: 6px;
   }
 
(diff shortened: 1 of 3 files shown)


matthew-white (approved):
Looks like a good size to me! Would be great to get @issa-tseng's feedback either before or after merging.

Something about the vertical alignment looks a little off to me though, like the larger form name and the smaller form ID aren't in line. Not sure how much that matters though / might just be me.

PR #721, opened 2023-01-26
css fixes for 2023.1

sadiqkhoja:
- Add page icon with the form name on form overview page
- Minor css fixes on Dataset overview page

### DIFF
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -120,7 +120,11 @@ class. */
   background-color: #888;
 }
 
-
+.title-icon{
+  padding-left: 6px;
+  font-size: 24px;
+  color: #666;
+}
 
 ////////////////////////////////////////////////////////////////////////////////
 // LINKS AND BUTTONS
@@ -813,10 +817,4 @@ becomes more complicated.
       }
     }
   }
-}
-
-.title-icon{
-  padding-left: 6px;
-  font-size: 24px;
-  color: #666;
 }
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -814,3 +814,9 @@ becomes more complicated.
     }
   }
 }
+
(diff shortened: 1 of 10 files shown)


src/assets/scss/app.scss
```diff
@@ -814,3 +814,9 @@ becomes more complicated.
     }
   }
 }
+
+.title-icon{
```
matthew-white, line 818: There's a section above for icons, would you mind removing this there?

src/components/form/new.vue
```diff
@@ -248,13 +248,13 @@ export default {
         });
     },
     removeLearnMore(value) {
-      return value.replace(/Learn more: (https?:\/\/xlsform\.org\/[^.]*)\.?/, '').trim();
+      return value.replace(/Learn more: (https?:\/\/xlsform\.org[^.]*)\.?/, '').trim();
```
matthew-white, line 251: I think it'd be nice to verify the character after .org. (Wouldn't want to match on xlsform.org.uk. 🤣) How about `\.org[/#]`?

src/components/dataset/overview/dataset-properties.vue
```diff
@@ -70,17 +70,23 @@ export default {
...
-  }
+  &.table{
+    table-layout: fixed;
 
-  &.table > tbody > tr:first-child > td {
-    border-top: none;
-  }
+    td {
+      // @include text-overflow-ellipsis;
+      word-wrap: break-word;
```
matthew-white, line 78: I think this is deprecated, how about `overflow-wrap: break-word;` ?

matthew-white (approved):

src/components/form/new.vue
sadiqkhoja: then it won't match with just `https://xlsform.org`

PR #726, opened 2023-01-30
Add background color to ExpandableRow button when it is focused

matthew-white:
The QA team noted:

> When a user uses keyboard navigation other parts are visibly highlighted (when chosen) except from “4 of 7 properties” in Connections to Forms and the form overview.

We usually add a background color on focus to buttons and links, so I think we should do so here as well.

In `ExpandableRow`, the properties count and the toggle button are in two different blocks. At least in Chrome, the properties count isn't focusable via the keyboard. So I've added a background color just to the toggle button, which is focusable. I think that's an improvement and all we need to do for now.

I didn't update the form overview, since that uses `DatasetSummaryRow`, which we will probably replace with `ExpandableRow` at some point in the future.

Other than the background color on focus, everything else about the appearance of the toggle button should be the same as before. I checked things locally in Chrome, and it looked the same.

I'll also add some comments about individual lines.

### DIFF
--- a/src/components/expandable-row.vue
+++ b/src/components/expandable-row.vue
@@ -14,13 +14,14 @@ except according to the terms contained in the LICENSE file.
     <div class="title-cell">
       <slot name="title"></slot>
     </div>
-    <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events, vuejs-accessibility/interactive-supports-focus -->
-    <div role="button" class="caption-cell" @click.prevent="toggleExpanded">
+    <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events -->
+    <div class="caption-cell" @click="toggleExpanded">
       <slot name="caption"></slot>
     </div>
 
-    <div class="button-cell">
-      <button type="button" class="btn btn-link" @click="toggleExpanded">
+    <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events -->
+    <div class="button-cell" @click="toggleExpanded">
+      <button type="button" class="btn btn-link">
         <span v-if="!expanded" class="sr-only">{{ $t('action.expand') }}</span>
         <span v-else class="sr-only">{{ $t('action.collapse') }}</span>
         <span v-if="!expanded" class="icon-caret-left"></span>
@@ -66,6 +67,8 @@ export default {
   .title-cell {
     flex-grow: 1;
   }
+
+  .caption-cell, .button-cell { cursor: pointer; }
 
   .caption-cell {
     text-align: right;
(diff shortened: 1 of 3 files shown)


src/components/expandable-row.vue
```diff
@@ -20,12 +20,12 @@ except according to the terms contained in the LICENSE file.
     </div>
 
     <div class="button-cell">
-      <a href="javascript:void(0)" role="button" @click.prevent="toggleExpanded">
+      <button type="button" class="btn btn-link" @click="toggleExpanded">
```
matthew-white, line 23: I had forgotten that we have `.btn-link` and that it's useful for cases like this. We have a similar `<a role="button">` in `ProjectHomeBlock`, which we should probably replace with a `<button class="btn btn-link">` at some point.

src/components/expandable-row.vue
```diff
@@ -73,25 +73,15 @@ export default {
...
-    line-height: normal;
+    align-self: center;
+    text-align: center;
     padding: 0px;
+    width: 30px;
 
-    a {
-      color: #888;
-      display: block;
-      width: 30px;
```
matthew-white, line 84: I moved the width to the `.button-cell` so that only the button will have a background color on focus, not the whole cell.

src/components/expandable-row.vue
```diff
@@ -73,25 +73,15 @@ export default {
...
-    font-size: 20px;
-    text-align: right;
-    line-height: normal;
+    align-self: center;
+    text-align: center;
     padding: 0px;
+    width: 30px;
 
-    a {
-      color: #888;
```
matthew-white, line 82: It looks like this style is currently overridden. On the QA server, the link's color is `#666`.

src/components/expandable-row.vue
```diff
@@ -73,25 +73,15 @@ export default {
   }
 
   .button-cell {
-    font-size: 20px;
```
matthew-white, line 76: `ConnectionToForms` makes the button 16px, so I don't think we need this. `align-self` ends up working out as a way to vertically align.

src/components/expandable-row.vue
```diff
@@ -14,18 +14,19 @@ except according to the terms contained in the LICENSE file.
     <div class="title-cell">
       <slot name="title"></slot>
     </div>
-    <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events, vuejs-accessibility/interactive-supports-focus -->
-    <div role="button" class="caption-cell" @click.prevent="toggleExpanded">
+    <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events -->
+    <div class="caption-cell" @click="toggleExpanded">
```
matthew-white, line 18: I added a similar click handler to `.button-cell` below. I thought I'd try to make the two divs similar, so I removed `.prevent` here, which I thought probably wasn't needed. I also removed `role="button"`, since users using a screen reader will use the button itself. I don't think we need to announce the div as a button. I want it to be clear though to users using a mouse that they can click, so I added `cursor: pointer;` to the CSS below.

PR #730, opened 2023-02-03
Added arrow in page head component

sadiqkhoja:
- Added arrow in page head component
- Removed Form and Dataset icon from the page heading
- Removed unused translation strings and css


ktuite (commented):
Overall, I think it's a nice change and looks good.

One thing I don't know about is the translations -- if we take out these translations here, are there other broader translation steps we have to take before the release?
- Would it be fine to take them out without doing that broader translation step?
- Would it be fine to leave these strings in and plan to take them out of the translations in the next release?

sadiqkhoja:
added translation keys back, we can remove them in next release

ktuite (approved):
Looks good!

Todo for the future: remove these unused translation keys.

src/components/dataset/show.vue
```diff
@@ -14,12 +14,11 @@ except according to the terms contained in the LICENSE file.
   <div id="dataset-show">
     <page-back v-show="dataExists" :to="projectPath('datasets')">
       <template #title>{{ project.dataExists ? project.nameWithArchived : '' }}</template>
-      <template #back>{{ $t('back') }}</template>
+      <template #back>{{ $t('resource.datasets') }}</template>
     </page-back>
     <page-head v-show="dataExists">
       <template #title>
         {{ datasetName }}
-        <span class="icon-cube title-icon"></span>
```
matthew-white, line 22: Can the CSS for `.title-icon` be removed in app.scss?

Issue #676, opened 2022-12-05
Replace remaining mixins with composables


matthew-white:
Vue 3 strongly encourages the use of composables over mixins. Composables can be used in any component, but mixins can only be used within components that use the Options API, not the new Composition API. I've been writing more components using the Composition API, which can be very convenient. However, right now many new components can't be written using the Composition API, simply because they use a mixin that we haven't yet converted to a composable. I've converted a few mixins to composables already, but there are still a few key mixins left to convert. I started working on that during v2022.3, and I'm close to being done, but I haven't finished yet. In particular, I'm not sure whether I should complete #675 before or after converting the `request` mixin to a composable. But even if we don't convert the `request` mixin to a composable immediately, I think it'd be good to finish converting the other mixins during v2023.1 in order to continue making progress on this issue.

PR #736, opened 2023-02-10
Add ability to cache composable results

matthew-white:
This PR is a precursor to #676.

### DIFF
--- a/src/composables/audit.js
+++ b/src/composables/audit.js
@@ -12,7 +12,7 @@ except according to the terms contained in the LICENSE file.
 
 // useAudit() returns functions related to audit log entries.
 
-import { inject } from 'vue';
+import { memoizeForContainer } from '../util/composable';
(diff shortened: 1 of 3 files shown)


src/composables/audit.js
```diff
@@ -24,13 +24,10 @@ const actionPath = (action) => {
...
-    }
-  };
-};
+export default memoizeForContainer(({ i18n }) => ({
+  // Returns a message describing an audit action.
+  actionMessage: (action) => {
+    const path = actionPath(action);
+    return i18n.te(path, i18n.fallbackLocale) ? i18n.t(path) : null;
+  }
+}));
```
matthew-white, line 33: Today, each `AuditRow` component that uses the `useAudit()` composable ends up creating its own `actionMessage()` function. After this change, the same function will be returned to each component.

PR #737, opened 2023-02-12
Replace routes mixin with composable

matthew-white:
Progress toward #676. Depends on #736.

In a few cases, I also converted a simple component using the Options API to the Composition API. That's possible now that the component is using a composable instead of a mixin.


test/components/dataset/row.spec.js
```diff
@@ -1,14 +1,14 @@
...
 import testData from '../../data';
+import { mockRouter } from '../../util/router';
 import { mount } from '../../util/lifecycle';
 
 const mountComponent = () => mount(DatasetRow, {
   props: { dataset: testData.extendedDatasets.last() },
-  global: {
-    stubs: { RouterLink: RouterLinkStub }
-  }
+  container: { router: mockRouter('/projects/1/datasets') }
```
matthew-white, line 11: `mockRouter()` automatically registers `RouterLinkStub`. `mockRouter()` also allows Vue Router composables to be called, which is needed now.

sadiqkhoja (approved):
looks perfect.

PR #745, opened 2023-02-28
removed gzipping large csv file on form attachments page

sadiqkhoja:
Closes #

- Manually tested by uploading file and multiple files on form attachments page
- All unit and integration tests are passing

NA

NA

NA

### DIFF
--- a/src/components/form-attachment/list.vue
+++ b/src/components/form-attachment/list.vue
@@ -79,7 +79,6 @@ except according to the terms contained in the LICENSE file.
 
 <script>
 import { any } from 'ramda';
-import pako from 'pako/lib/deflate';
 import { markRaw } from 'vue';
 import FormAttachmentNameMismatch from './name-mismatch.vue';
 import FormAttachmentPopups from './popups.vue';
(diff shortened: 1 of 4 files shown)


src/components/form-attachment/list.vue
```diff
@@ -281,52 +280,6 @@ export default {
...
-        return Promise.resolve({ data: file, encoding: 'identity' });
-      return new Promise((resolve, reject) => {
-        const reader = new FileReader();
-        const initialRoute = this.$route;
-        reader.onload = () => {
-          resolve({ data: pako.gzip(reader.result), encoding: 'gzip' });
-        };
-        reader.onerror = () => {
-          if (this.$route === initialRoute) {
-            this.alert.danger(this.$t('alert.readError', {
```
matthew-white, line 321: Would you mind removing this message in the `<i18n>` block below?

matthew-white (approved):

matthew-white:
Just in case our future selves wonder why we made this change, we're updating nginx in getodk/central#391. As part of the update, it looks like we're going to drop support for request body decompression. That requires us to make this change to Frontend.

PR #752, opened 2023-03-02
Add back password strength meter

### LINKED ISSUE #528: Add back password strength meter
In #526, the password strength meter is removed because the package we use doesn't support Vue 3 (apertureless/vue-password-strength-meter#54). ~v1.4 won't ship with a password strength meter, but I think we will add one back in v1.5.~

When we add it back, we'll have to decide what kind of password strength meter to use. I think we have two main options:

- Use a dictionary approach.
  - The main issue here is that zxcvbn is no longer maintained. However, there is a fork of it rewritten in TypeScript that we could look into (https://github.com/zxcvbn-ts/zxcvbn).
  - I'm not sure that there's a Vue component out there that uses this fork, but we could probably write our own without too much difficulty.
  - How does this interact with internationalization? Would we want to use dictionaries for languages other than English?
- Use a pattern-based approach.
  - Given that our only requirement for passwords/passphrases is that they are at least 10 characters long, this approach might work fine for our needs.
  - This approach is ligher-weight insomuch as it doesn't require the user to download a large dictionary.
  - One component to look into: https://github.com/miladd3/vue-simple-password-meter

matthew-white:
Closes #528.

Tests.

This PR effectively reverts #586. We still can't use vue-password-strength-meter, which doesn't support Vue 3. But I've extracted parts of vue-password-strength-meter into a standalone component, `PasswordStrength`. The new component evaluates passwords based on password length and doesn't use zxcvbn, which vue-password-strength-meter did. I also fixed a small CSS issue whereby the five bars weren't equally sized.

Because `PasswordStrength` is an isolated component, I think the regression risks are low.

### DIFF
--- a/src/components/password-strength.vue
+++ b/src/components/password-strength.vue
@@ -1,0 +1,109 @@
+<!--
+Copyright 2023 ODK Central Developers
+See the NOTICE file at the top-level directory of this distribution and at
+https://github.com/getodk/central-frontend/blob/master/NOTICE.
+
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+
+<!-- Parts of this component are based on the npm package
+vue-password-strength-meter 1.7.2, which uses the MIT license.
+https://github.com/apertureless/vue-password-strength-meter -->
+<template>
+  <div class="password-strength">
+    <div :data-score="score"></div>
+  </div>
+</template>
+
+<script setup>
+import { computed } from 'vue';
+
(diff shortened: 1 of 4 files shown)


src/components/password-strength.vue
```diff
@@ -0,0 +1,109 @@
...
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+
+<!-- Parts of this component are based on the npm package
+vue-password-strength-meter 1.7.2, which uses the MIT license.
+https://github.com/apertureless/vue-password-strength-meter -->
```
matthew-white, line 15: @lognaturel, are these comments enough to credit the original package?

ktuite (approved):
Looks great!

PR #754, opened 2023-03-07
Remove reference to Form in generic odata modal

ktuite:
Closes design feedback issue from issa.

Used to say "To connect to this Form's OData feed..." even on an entity page. Now it says "To connect to this OData feed..."

Screen Shot 2023-03-07 at 2 19 47 PM


matthew-white (approved):
This looks reasonable to me. The alternative would be to have two separate messages, one for forms and one for datasets. I think that would be slightly more clear, but I'm not sure that it's worth the extra logic and translation. The main goal here is to provide enough information so that users who aren't familiar with OData will give it a try in the tool of their choice, and I think that's accomplished here.

ktuite:
I thought a bit about making a more differentiated message but went with the easiest route 😅 
Hopefully seeing `forms` or `datasets` in the URL will also be a bit of a clue. Also, maybe we'll have even more OData feeds in the future!

PR #756, opened 2023-03-08
Linkify project name in DatasetShow

### LINKED ISSUE #748: Linkify project name in DatasetShow
Dataset pages show the project name at the top:

My Project > Datasets

This is similar to form pages, which likewise show the project name at the top:

My Project > Forms

On form pages, the project name links to the project overview. However, the project name does not link on dataset pages. I think it's intuitive to try clicking the project name (and have tried to do so myself), so I think we should linkify it.

Question: Should the project name link to the dataset list or to the project overview? (If the former, we can just use the existing `linkTitle` prop of the `PageBack` component. If the latter, then we should probably modify `PageBack`.) My instinct is that it should link to the project overview.

matthew-white:
Closes #748.

I tried out this change locally and verified that the project name is a link. I also checked the link to &hellip;/datasets that existed previously and observed that it is still working. I also did a quick check of the other two uses of `PageBack`, in `FormHead` and `SubmissionShow`, and everything seemed fine in those places.

`PageBack` has two parts/elements where a link can be rendered. `PageBack` is used in slightly different ways in `FormHead`, `SubmissionShow`, and `PageBack`:

- `FormHead`: The first and second elements are both links and link to the same location.
- `SubmissionShow`: The second element is a link, but the first is not.
- `DatasetShow`: The first and second elements are both links, but they link to different locations.

The way that `PageBack` was set up before didn't work with the desired behavior in `DatasetShow`, so I modified `PageBack`. I removed the `linkTitle` prop and instead made the `to` prop more flexible by allowing it to be an array. In the simple case where only the second element is a link, the `to` prop is still allowed to be string. (`SubmissionShow` doesn't need to change in this PR.)

Right now the `title` and `back` slots of `PageBack` take text, but I thought about having them take links. Then we could do away with the `to` prop entirely. However, that limits the control that `PageBack` has over those elements. Right now, `PageBack` adds an `id` attribute to the link that is the parent element of the `back` slot.

I also thought about making the `to` prop an object instead of an array, something like:

```js
{ to: '/some/location', back: '/another/location' }
```

However, that felt unnecessarily verbose to me.

Because we have unit tests of `PageBack`, I think the risk of regression is low.

### DIFF
--- a/src/components/dataset/show.vue
+++ b/src/components/dataset/show.vue
@@ -12,7 +12,7 @@ except according to the terms contained in the LICENSE file.
 
 <template>
   <div id="dataset-show">
-    <page-back v-show="dataExists" :to="projectPath('datasets')">
+    <page-back v-show="dataExists" :to="[projectPath(), projectPath('datasets')]">
       <template #title>{{ project.dataExists ? project.nameWithArchived : '' }}</template>
       <template #back>{{ $t('resource.datasets') }}</template>
     </page-back>
--- a/src/components/form/head.vue
+++ b/src/components/form/head.vue
@@ -11,7 +11,7 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <div id="form-head">
-    <page-back :to="projectPath()" link-title>
+    <page-back :to="[projectPath(), projectPath()]">
       <template #title>{{ project.dataExists ? project.nameWithArchived : '' }}</template>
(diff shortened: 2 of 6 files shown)


ktuite (approved):
Looks good! I see how you pass it two links now in `:to`, or if you still just pass one not in an array (like in submission show, which didn't change), it will put the link on the second subtitle slot in Title > Subtitle.

PR #761, opened 2023-03-25
Replace tab mixin with useTabs() composable

matthew-white:
See #676.

Existing tests continue to pass. There weren't unit tests of the mixin, so I didn't have to update any such tests. I think it's fine that there aren't unit tests, because the mixin/composable is pretty simple, and there are a number of integration tests that get at its behavior.


sadiqkhoja (approved):
perfect

PR #762, opened 2023-03-26
Check more dataset and entity verbs in routes

### LINKED ISSUE #662: Check dataset/entity verbs in more places
Most routes require that the current user be able to do certain verbs (either sitewide or on a particular project). If the user tries to navigate to a route that requires a verb that they can't do, then they are redirected to the homepage. I think it's a good idea for the route to check every verb that the route component might require, unless the component itself checks the verb as part of rendering something conditionally (for example, using `LinkIfCan`). With that in mind, I think we should check dataset and entity verbs for more routes:

- `FormOverview` and `FormDraftStatus` should check `dataset.list` because they request &hellip;/dataset-diff.
- `FormAttachmentList` should check `entity.list` because it allows the user to link a form attachment to a dataset.
- Maybe others?

Given that this doesn't come up for our current roles and we don't have custom roles, and we rely on the API to do these checks anyway, I don't think this is something we need to do for v2022.3. However, I think it'd be a good idea to add this at some point.

matthew-white:
Closes #662. Depends on getodk/central-backend#808.

I determined which routes needed to change by searching for components that call a dataset or entity `apiPaths` method.

Tests. Even though this PR changes the specific verbs that are checked, tests don't need to change. That's because tests are based on roles, not verbs, and which roles can access which routes hasn't changed. The only thing that needed to change was adding the `dataset.read` verb to `testData`.

This change shouldn't be visible to users. Mostly it's for code clarity and future-proofing.

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -359,7 +359,7 @@ const routes = [
           validateData: {
             // Including form.update in order to exclude project viewers and
             // Data Collectors.
-            project: () => project.permits(['form.read', 'form.update']),
+            project: () => project.permits(['form.read', 'form.update', 'dataset.list']),
             form: () => form.publishedAt != null
           },
           title: () => [form.nameOrId]
@@ -440,7 +440,8 @@ const routes = [
             project: () => project.permits([
               'form.read',
               'form.update',
-              'form.delete'
+              'form.delete',
+              'dataset.list'
             ]),
             formDraft: () => formDraft.isDefined()
           },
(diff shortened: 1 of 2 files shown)


ktuite (approved):
The changes make sense -- the pages like the form detail page and the form draft page show dataset information, so it sounds right to check for dataset access. As we discussed on https://github.com/getodk/central-backend/pull/808, those dataset diff endpoints on backend use `dataset.list` instead of `dataset.read` in case multiple datasets are (someday) involved. 

I've tried it with the new frontend verb changes and it works!

PR #763, opened 2023-03-27
Replace request mixin with useRequest() composable

matthew-white:
This PR makes progress on #676, replacing the `request` mixin with a composable.

Existing tests continue to pass.

At some point, I'd like to make a few changes to the composable to make it more similar to the `request()` method of `requestData` resources. I'm planning to do that as part of #675, and I may make some of those changes in .3 in a follow-up PR. However, since this PR touches so many files, I avoided making functional changes to the composable and mostly just tried to complete the conversion from mixin to composable.

I've updated the comments above the composable. Those comments now mention `requestData` and mention that the composable is allowed to send GET requests (not that that should be common). However, there is other documentation that I need to update, particularly the contribution guide. I'll do that as part of #674.

One change that I did make is that the composable now returns a single function, `request()`. In contrast, the mixin defined a number of other methods: `post()`, `put()`, `patch()`, and `delete()`. I wanted to move away from that pattern for a few reasons:

- Consistency across components. There weren't that many components that used a method other than `request()`, and it seems simpler to have fewer ways to send requests.
- Consistency with `requestData` resources. Resources only have a `request()` method.
- The names `get`, `put`, and `delete` seem very generic: without context, they don't seem like they necessarily have to do with requests. Also, while `delete` was a valid method (property) name, it's not a valid name for a variable, so it doesn't work as well with the Composition API.


src/components/navbar/actions.vue
```diff
@@ -38,14 +38,12 @@ except according to the terms contained in the LICENSE file.
...
 
 <script>
-import request from '../../mixins/request';
 import { logOut } from '../../util/session';
 import { noop } from '../../util/util';
 import { useRequestData } from '../../request-data';
 
 export default {
   name: 'NavbarActions',
-  mixins: [request()],
```
matthew-white, line 48: This component didn't actually use the mixin, so I just removed it. I think what happened here is that the component used to use the mixin when it was the one to send the logout request. At some point, that request was moved to the `logOut()` function, which this component calls, but the mixin wasn't removed at the same time.

src/composables/request.js
```diff
@@ -66,86 +63,77 @@ following options:
...
-      return this.request({ ...config, method: 'DELETE', url });
-    }
-  }
 };
 
-export default () => mixin;
+export default () => {
+  const container = inject('container');
+  const awaitingResponse = ref(false);
+  const request = _request(container, awaitingResponse);
```
sadiqkhoja, line 131: can we use `.apply` here and use update value of awaitingResponse using `this`? that why we don't have to disable eslint above. Not totally sure if that will work

src/composables/request.js
```diff
@@ -66,86 +63,77 @@ following options:
...
+
+  const { router } = container;
+  // `router` may be `null` in testing.
+  if (router != null)
+    watch(router.currentRoute, () => { awaitingResponse.value = false; });
+
+  return { request, awaitingResponse: readonly(awaitingResponse) };
```
sadiqkhoja, line 138: Not super important but I like http helper functions `get`, `post`, 'put`, etc. One thing that we can do is to return an object like:
```
{
  send: _request(container, awaitingResponse),
  post: _request(container, awaitingResponse, 'POST'), // _request should take optional argument for HTTP method
  get: _request(container, awaitingResponse, 'GET'),
  ...
}
```

and then in components we would call `this.request.send` or `this.request.post`, that way we would have context of `get`, `post` and also having `delete` would be possible as well.

sadiqkhoja (approved):
I have added a couple of comments, which are reflection of my preference and not necessary to be incorporated.

src/composables/request.js
matthew-white: What do you think about the idea of adding these as properties on the `request()` function? That's what axios does: `axios` is a function with additional properties. That way, you can just call `request()` if you want, but you could also call `this.request.post()`, etc.

sadiqkhoja: that would work too

src/composables/request.js
matthew-white: Good question, I was thinking about something like this as well. I think something like what you describe would be possible. But it sort of feels like mostly a way to bypass ESLint. The same mutation would be happening, just via `this`.

Another strategy I considered was to pass a function to `_request()` to set `awaitingResponse.value`. `_request()` only ever needs to set `awaitingResponse.value`, not get it. So we could do something like:

```
const setAwaitingResponse = (value) => { awaitingResponse.value = value; };
const request = _request(container, setAwaitingResponse);
```

ESLint wouldn't object in that case. But I don't think that code is really clearer.

In general, I think this is going to come up sometimes with the Composition API. Within the Composition API, refs will be passed from function to function. And the way to change a ref is to set its `value` property, which ESLint will sometimes object to. But my sense is that that's a normal pattern within the Composition API. If something that creates a ref wants to prevent it from being set, it can call `readonly()` on it. That's what happens here: `awaitingResponse.value` is set internally within the composable, but when the composable returns it to the component, it calls `readonly()` on it first.

sadiqkhoja: make sense

src/composables/request.js
matthew-white: This ended up being involved enough that I made a separate PR for it: #765.

PR #764, opened 2023-03-27
Add max width to Frontend content

### LINKED ISSUE #447: Set maximum width for frontend
Right now, Frontend will use the full width of the page. However, after a certain width, that detracts from the user experience. Instead, we want to set a maximum width for Frontend to use. (Let's try 1400px.) The content should be centered on the page.

One caveat has to do with the Submissions and Testing pages. On those pages, we want the submissions table to use nearly the full width of the page. We could set it up so that even while the rest of the page has a max width of 1400px, the table somehow breaks out of the container and uses more width. Or we could set it up so that on the Submissions and Testing pages, Frontend as a whole uses the full width of the page.

<hr>

Right now we use the Bootstrap full-width container (`.container-fluid`). However, we could use the Bootstrap responsive fixed-width container instead (`.container`). Or we could update `.container-fluid` or write our own container.

One thing that we know will be an issue is elements with `position: fixed;`. However, those are not common in Frontend. The main example I can think of is `FormAttachmentPopups`.

matthew-white:
This PR adds a max width of 1280px to the root component `App`, closing #447. `App` is centered on the page.

I've tried it out locally. If we decide to run with this approach, I'll write tests.

When I wrote #447, I imagined adding a max width for all Frontend content. However, other approaches are possible: an alternative to this PR, #775, only adds a max width to the `PageBody` component.

Assuming for this PR that the max width is on `App`, there are still open questions. One of them is whether `App` should be centered or left-aligned on the page. I didn't think it'd work well for the entire app, including the navbar, to be truncated just on the right, so I opted for centering `App`.

#447 calls for the submissions table to use the full page width:

> We could set it up so that even while the rest of the page has a max width of 1400px, the table somehow breaks out of the container and uses more width. Or we could set it up so that on the Submissions and Testing pages, Frontend as a whole uses the full width of the page.

After centering `App`, I couldn't really imagine the first option. Instead, I made is so that the submissions and draft testing pages as a whole use the full page width. I thought it'd be too abrupt to suddenly change the width of `App`, so I added a CSS animation for that.

The issue suggested a max width of 1400px. After trying 1400px, issa suggested we try 1280px.

The risk of regression is hopefully low, because the main change is the CSS for `App` and the document body: almost nothing about the content of `App` had to change. The exception is elements that have `position: fixed;`, but I think the only element like that is in `FormAttachmentPopups`. That said, this is a change to the app as a whole, so we'll want to keep our eyes out for unexpected behavior. There's probably some risk in particular in the transition to/from full width.

No changes needed.


matthew-white:
Closing in favor of #775.

PR #765, opened 2023-03-31
Add convenience methods to request() functions for HTTP verbs

matthew-white:
This PR adds convenience methods for HTTP verbs to the `request()` function returned by `useRequest()`, as discussed [here](https://github.com/getodk/central-frontend/pull/763#discussion_r1153846715). For example, the following two calls will send the same request:

```js
const { request } = useRequest();
request({
  method: 'POST',
  url: '/v1/projects',
  data: { name: 'My Project' }
});
request.post('/v1/projects', { name: 'My Project' });
```

The parameters of the convenience methods are the same as they were in the `request` mixin.

I want to make the `request()` function returned by `useRequest()` and the `request()` method of `requestData` resources as similar as possible (see #675). With that in mind, I've also added the convenience methods to the `request()` method. The `request()` method usually sends GET requests, but there are several cases where it sends a non-GET request.

New tests.

In this PR, `useRequest()` and `requestData` call a shared function named `withHttpMethods()`, which adds convenience methods to any `request()` function/method. The main question I had was how to implement `withHttpMethods()`. I wanted to avoid eagerly creating five additional functions for every component that calls `useRequest()`, because most components will use at most one convenience method (and many components won't use a convenience method at all). I tried two different working approaches, which are both in the commit history.

One approach I considered was to subclass `Function`. The convenience methods would go on the subclass's prototype. However, after briefly investigating this option, it sounded like extending `Function` involved complications.

The next approach I considered was to wrap the `request()` function in a proxy. The proxy would behave the same as the `request()` function except when getting one of the convenience methods. In that case, the proxy would create the convenience method on the fly. I got this approach to work, and you can see it in the commit history.

I then realized that if the convenience methods are properties of the `request()` function, they can access the `request()` function via `this`. That means that we don't need to create different convenience methods for each `request()` function: all `request()` functions can use the same convenience methods. So the approach I ended up taking was to define the convenience methods once, then have `withHttpMethods()` assign them on the `request()` function.

`this` ended up coming up in another issue I encountered, which is that calling a convenience method results in a different `this` compared to calling `request()` directly. That isn't an issue for `useRequest()`, but it comes up for `requestData`, because the `request()` method references `this`. For example, if you call `someResource.request()`, then `this` is `someResource`. But if you call `someResource.request.post()`, then `this` is `someResource.request`. I was able to resolve this by binding `someResource.request` to `someResource`.

No changes needed.

### DIFF
--- a/src/request-data/resource.js
+++ b/src/request-data/resource.js
@@ -11,7 +11,7 @@ except according to the terms contained in the LICENSE file.
 */
 import { computed, isRef, readonly, toRef } from 'vue';
 
-import { isProblem, logAxiosError, requestAlertMessage, withAuth } from '../util/request';
+import { isProblem, logAxiosError, requestAlertMessage, withAuth, withHttpMethods } from '../util/request';
 import { noop } from '../util/util';
 import { setCurrentResource } from './util';
 import { unlessFailure } from '../util/router';
@@ -279,8 +279,20 @@ class Resource extends BaseResource {
 }
 
 const proxyHandler = {
-  get: (resource, prop) => {
-    if (prop in resource) return resource[prop];
+  get: (resource, prop, proxy) => {
+    // First check the resource for the property.
+    if (prop in resource) {
+      // We want to call withHttpMethods() on Resource.prototype.request().
+      // However, because the request() method references `this`, we must first
+      // bind it to the proxy. For example, if resource.request.post() is
+      // called, we need `this` to be the proxy in post().
+      if (prop === 'request')
+        return withHttpMethods(resource.request.bind(proxy));
+
+      return resource[prop];
+    }
+
+    // Fall back to getting the property from resource.data.
     const { data } = resource;
     if (data == null) return undefined;
     const value = data[prop];
--- a/src/util/request.js
+++ b/src/util/request.js
@@ -166,6 +166,37 @@ export const apiPaths = {
 ////////////////////////////////////////////////////////////////////////////////
 // SENDING REQUESTS
 
+const methods = {
+  // The boolean indicates whether the method accepts a `data` argument.
+  get: false,
+  delete: false,
+  post: true,
+  put: true,
+  patch: true
+};
+const proxyHandler = {
+  get: (request, prop) => {
+    if (prop in request) return request[prop];
+
+    const acceptsData = methods[prop];
+    if (acceptsData == null) return undefined;
+
+    const upper = prop.toUpperCase();
+    return !acceptsData
+      ? (url, config = undefined) => request({ ...config, method: upper, url })
+      : (url, data = undefined, config = undefined) => {
+        const full = { ...config, method: upper, url };
+        if (data != null) full.data = data;
+        return request(full);
+      };
+  }
+};
+// Adds convenience methods for HTTP verbs to a request() function by wrapping
+// it in a proxy. `this` cannot be passed through to the convenience methods, so
+// if the request() function references `this`, you must bind `this` for it
+// first.
+export const withHttpMethods = (f) => new Proxy(f, proxyHandler);
+
 export const withAuth = (config, token) => {
   const { headers } = config;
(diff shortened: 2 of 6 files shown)


src/request-data/resource.js
```diff
@@ -279,8 +279,20 @@ class Resource extends BaseResource {
...
-    if (prop in resource) return resource[prop];
+  get: (resource, prop, proxy) => {
+    // First check the resource for the property.
+    if (prop in resource) {
+      // We want to call withHttpMethods() on Resource.prototype.request().
+      // However, because the request() method references `this`, we must first
+      // bind it to the proxy. For example, if resource.request.post() is
+      // called, we need `this` to be the proxy in post().
+      if (prop === 'request')
+        return withHttpMethods(resource.request.bind(proxy));
```
matthew-white, line 290: I wasn't sure at first where to do this binding. Initially, I put it in the constructor of the `Resource` class above. However, that binds the `request()` method to the `Resource` instance. Instead, we want to bind it to the resource proxy. The resource proxy is what is returned from this file: the file only uses the `Resource` instance internally.

src/util/request.js
```diff
@@ -152,6 +161,34 @@ export const apiPaths = {
...
+  // eslint-disable-next-line func-names
+  httpMethods[prop] = function(url, config = undefined) {
+    return this({ ...config, method, url });
+  };
+}
+for (const prop of ['post', 'put', 'patch']) {
+  const method = prop.toUpperCase();
+  // eslint-disable-next-line func-names
+  httpMethods[prop] = function(url, data = undefined, config = undefined) {
+    const full = { ...config, method, url };
```
sadiqkhoja, line 181: what happens if data is null or undefined?

sadiqkhoja (approved):
looks good.

src/util/request.js
matthew-white: Not specifying `data`, specifying `undefined` for `data`, and specifying `null` for `data` all have the same effect, which is that no data will be passed to axios. That's pretty rare, but it happens for POST /v1/assignments/:role/:actorId and POST /v1/projects/:projectId/assignments/:role/:actorId.

I suppose `null` is valid JSON, so it might make sense to distinguish between `null` and `undefined` here: I could just check for `undefined` below (`!== undefined` instead of `!= null`). I also think we could leave it as-is and change it if/when we need to POST a `null` value. Before they were removed, the convenience methods of the `request` mixin didn't distinguish between `null` and `undefined` values of `data`.

sadiqkhoja: I was thinking to merge these two for loops into one, if `if (data != null) full.data = data;` is not necessary.

Also `= undefined` is redundant in the function signature. By default it's `undefined` if consumer doesn't pass any value

matthew-white: > I was thinking to merge these two for loops into one, if `if (data != null) full.data = data;` is not necessary.

The functions in the two loops have a different number of arguments, so I don't think they can be merged (unless I'm misunderstanding).

> Also `= undefined` is redundant in the function signature. By default it's `undefined` if consumer doesn't pass any value

It has some effect (on the `length` property of the function), but it's definitely not needed. I've gotten it in the habit of specifying `= undefined` as a way to signal that a parameter is optional.

sadiqkhoja: What is the implications of this?

```
const httpMethods = {};
for (const prop of ['get', 'delete', 'post', 'put', 'patch']) {
  const method = prop.toUpperCase();
  // eslint-disable-next-line func-names
  httpMethods[prop] = function(url, config, data) {
    return this({ ...config, method, url, data });
  };
}
```

matthew-white: Ah I see what you mean. For POST/PUT/PATCH, I think it's nice to order `data` before `config`. Those methods will almost always receive `data`, but it's rare for them to receive `config`. `url, data, config` is also the same order as the axios convenience methods.

sadiqkhoja: right.

PR #766, opened 2023-03-31
Add useRenderTimer() composable to help measure performance

matthew-white:
This PR adds a simple composable to help measure how long rendering takes. Usually I use Vue devtools for that. However, this composable can be helpful in a couple of cases:

- After each render, the composable `console.log()`s how long the render took. That can be useful if there's other `console.log()`-ing happening: you can see the render times around the the other logging that you're doing.
- The composable returns summary statistics about the render times.

I wrote this composable while debugging performance issues in another PR (not yet pushed), so I think it works. The composable is a tool for debugging and performance evaluation and isn't used in production. Given that, I think that tests aren't needed.

There may be external packages out there with similar functionality, but I just wanted something simple. In most cases, Vue devtools will be the most useful thing.

No changes needed.

### DIFF
--- a/src/util/debug.js
+++ b/src/util/debug.js
@@ -85,11 +85,12 @@ export const useRenderTimer = () => {
   for (const [name, f] of hooks) {
     f(() => { // eslint-disable-line no-loop-func
       const duration = Date.now() - timeBefore;
-      timer._push(duration);
 
       const onOrOff = timer.state ? 'on' : 'off';
       // eslint-disable-next-line no-console
       console.log(`[useRenderTimer] ${name} (timer ${onOrOff})`, duration);
+
+      timer._push(duration);
     });
   }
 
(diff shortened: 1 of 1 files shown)


matthew-white:
Ah I realized that I made a last-minute change that wasn't quite right. I had reordered `timer._push()` before `console.log()`, but it's actually important that `console.log()` happens first. (`timer.state` is read as part of the `console.log()`, but `timer._push()` may change `timer.state`.)

src/util/debug.js
```diff
@@ -23,3 +23,76 @@ export const ticking = (count, callback = logTick) => {
...
+      max: Math.max(...this.durations),
+      sum: sum(this.durations)
+    };
+    stats.avg = stats.sum / stats.count;
+    return stats;
+  }
+}
+
+// useRenderTimer() is a composable. It returns a timer object that can help
+// measure performance around rendering.
```
matthew-white, line 75: Add an example:

```js
setup() {
  const timer = useRenderTimer();
  // Turn the timer on to store the render times in timer.durations.
  timer.on();
  window.timer = timer;
}
```

matthew-white, line 75: Mention Vue devtools here.

PR #767, opened 2023-04-02
Render two potentially large tables in chunks

matthew-white:
While working on #764, I noticed that there can be a short delay in rendering the submissions table during which Frontend is nonresponsive. The delay is long enough that the CSS transition in which the app becomes full width is interrupted. I'm not sure how often the delay would be seen in production: it would require the OData response to be returned more quickly than the transition completes (0.5 seconds). I checked a few projects on the QA server, and most OData responses took longer than that, but some took less. The issue with the CSS transition is quite noticeable locally.

The first version of #764 attempted to resolve the issue by introducing a `widening` flag that indicates whether the CSS transition is in progress (https://github.com/getodk/central-frontend/pull/764/commits/cd4fd63fcef8a7b0aaf06023fce0e6097ece93b9). The submissions table will not be rendered during that time. However, I think that's not an ideal solution:

- It adds more state to the container.
- It delays the initial rendering of the submissions table.
- The flag might be needed not just for navigation to the submissions page, but also navigation from the submissions page, if navigating to a separate page where initial rendering takes longer. That's because there will be another CSS transition as the width becomes more narrow.

Instead of using a flag, this PR renders the submissions table in chunks. Because there are fewer rows to render at first, the initial render of the submissions table is faster. When I try it locally, I don't see the same issue with the CSS transition.

This is an idea I've had for a while, and it can be dropped into other potentially large tables. We've seen on the QA server that the projects page takes a moment to render even after the response is received from Backend, especially on slower machines. That may be a problem specific to the QA server, which has many projects. But it was easy to use the same approach as the submissions table to render the project list in chunks.

At some point, I think it'd be a good idea to introduce pagination to solve issues like this. With pagination, there would be less data to render at once. However, introducing pagination will likely be a heavier lift. We will also need more search functionality before moving to pagination. I think of this PR as a stopgap for until we have pagination. Until then, it will improve the rendering of large tables.

It's pretty easy to drop in this approach, so if anyone has noticed other tables that can take longer to render, I'd be happy to try adding it to those.

New tests. I've tried out the submissions table locally. Once the PR is merged, I'll put it on the QA server. I'm interested to see the effect on the projects page there.

I wanted to make this change as small as possible, especially since I don't imagine us using it forever.

I thought about only rendering additional chunks as the user scrolled through the page. But that would add more complexity and wouldn't work with native browser search. I also thought about list virtualization: there are [existing packages](https://vuejs.org/guide/best-practices/performance.html#virtualize-large-lists) for that. But again, I don't think that works with browser search. (Search keeps coming up!)

I tried a few approaches to the chunking. Right now, there are 10 equal chunks, and Frontend will wait 25 milliseconds after rendering each chunk before rendering the next one. If the array is smaller (<250 elements), then each chunk will be 25 elements, and there will be fewer than 10 chunks. The intention is for the chunking to happen fast enough that users won't notice it. Yet it will still improve the user experience by helping to prevent Frontend from becoming nonresponsive.

I used #766 to look at render times. I saw that chunking increases the total render time, but that each render is shorter. I landed on 10 chunks after trying a few different approaches. I chose 25 as the minimum chunk size with the thought that even on a monitor, the user probably won't see much more than 25 rows at first. We wouldn't want the user to see the chunking happening on-screen.

Rendering in chunks is more complicated than rendering all at once. I'm not sure that this approach would work in all cases: in theory, there could be broken assumptions if the size of the table/list differs from the length of the array. However, I don't think that's an issue for the two tables/lists I modified.

No changes needed.

### DIFF
--- a/src/composables/chunky-array.js
+++ b/src/composables/chunky-array.js
@@ -1,0 +1,119 @@
+/*
+Copyright 2023 ODK Central Developers
+See the NOTICE file at the top-level directory of this distribution and at
+https://github.com/getodk/central-frontend/blob/master/NOTICE.
+
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+*/
+import { computed, onUnmounted, ref, watch } from 'vue';
+
+/*
+useChunkyArray() creates a proxy of an array such that the length of the proxy
+is less than the actual length of the array. The length of the proxy will
+increase in steps over the course of a second or so, until it equals the length
+of the array. The length of the proxy is reactive. useChunkyArray() is helpful
+if an array may be large and is used in a potentially expensive v-for, because
+it reduces the amount of time during which Frontend is busy rendering and is
+nonresponsive. (There will be more renders, but each render will be shorter.)
+
+useChunkyArray() takes a ref whose value is either an array or nullish,
+returning a computed ref. If the value of the ref is an array, then the value of
+the computed ref will be a proxy as described above. If the value of the ref is
+nullish, then the value of the computed ref will be `null`.
+*/
+export default (arrayRef, minChunkSize = 25) => {
+  // The length of the proxy
+  const length = ref(0);
+  let chunkSize;
+  let intervalId;
+  const stopChunks = () => {
+    if (chunkSize == null) return;
+    chunkSize = null;
+    clearInterval(intervalId);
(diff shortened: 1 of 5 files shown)


src/composables/chunky-array.js
```diff
@@ -0,0 +1,119 @@
...
+    clearInterval(intervalId);
+    intervalId = null;
+  };
+  const addChunk = () => {
+    length.value += chunkSize;
+    if (length.value >= arrayRef.value.length) {
+      length.value = arrayRef.value.length;
+      stopChunks();
+    }
+    return length.value;
```
matthew-white, line 45: Doesn't need to return.

matthew-white (commented):
Notes from interactive review

Issue #760, opened 2023-03-25
See Entity detail page


matthew-white:
See the release criteria [here](https://docs.google.com/document/d/1tOBNJaWZSJjkZjTYxV7vYBOFSkiZWKsnp48uWb_L5Co/edit#heading=h.lurckciqqz5h).

srujner:
Tested with Success!

dbemke:
Tested with Success!

PR #768, opened 2023-04-03
Add generic FeedEntry component to use for submissions and entities

matthew-white:
This PR is a first step toward #760. It genericizes the `SubmissionFeedEntry` component, which we will need to use in the entity detail page. The new generic component is `FeedEntry`. I may end up adding more to `FeedEntry` later, but for now, I've just moved things from `SubmissionFeedEntry` to `FeedEntry` that felt like they clearly belonged in `FeedEntry`.

I patched tests of `SubmissionFeedEntry` that failed and wrote new tests of `FeedEntry`. I also took a look at a few types of feed entries locally. I think the thing that's most likely to have gone wrong is the Sass, since many styles were moved from `SubmissionFeedEntry` to `FeedEntry`, and some styles became less specific. However, I reviewed my style-related code changes closely.

As mentioned above, I'm not quite sure where `FeedEntry` should end and `SubmissionFeedEntry` should begin. For now, I've moved the most generic things to `FeedEntry`. Almost all the specifics of what goes in the title and the body of the `FeedEntry` are still in `SubmissionFeedEntry`. That may change in the future though.

This is a refactor that isn't supposed to have user-facing effects.

No changes needed.

### DIFF
--- a/src/components/feed-entry.vue
+++ b/src/components/feed-entry.vue
@@ -1,0 +1,78 @@
+<!--
+Copyright 2023 ODK Central Developers
+See the NOTICE file at the top-level directory of this distribution and at
+https://github.com/getodk/central-frontend/blob/master/NOTICE.
+
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+<template>
+  <div class="feed-entry">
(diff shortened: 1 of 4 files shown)


src/components/feed-entry.vue
```diff
@@ -0,0 +1,78 @@
...
+    padding-top: 3px;
+    text-align: center;
+    width: 18px;
+  }
+
+  a { font-weight: normal; }
+}
+
+.feed-entry-body {
+  background-color: #f9f9f9;
```
matthew-white, line 74: This style is applied more widely than before. Before, only the `MarkdownView` had this background, but now the body will always have this background. As a result, the background of `SubmissionDiffItem` has changed ever so slightly from one gray to another.

PR #769, opened 2023-04-05
Check for warnings when running tests

### LINKED ISSUE #712: Check for Vue warnings in CircleCI
We have had a couple cases of tests passing despite Vue warnings being logged:

- #653
- #711

Warnings like these point to something that needs to be addressed in a test even though the test is passing. I think it'd be a good idea to have CircleCI check for Vue warnings logged during `npm run test` and fail if it finds any.

matthew-white:
Closes #712. This PR adds a script that runs tests, and if they're successful, will search the test output for two types of warnings:

- Warnings that contain `WARN LOG:`. Vue warnings fall in this category.
- Warnings that contain `WARN [web-server]:`. We've seen these from Karma.

If a warning is found, then the script will display the warning, display a message, then exit with a non-zero status. That will cause CircleCI to fail.

I caused both types of warnings locally and verified that the script caught them both. If there are no warnings, then the script is successful.

My initial plan was to only make this change for CircleCI, but I think it's nice for this to happen locally as well.

I took a quick look for Karma options to do this, but I didn't see anything relevant.

No changes needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,5 +30,3 @@ npm-debug.log*
 
 /.nginx/*
 !/.nginx/.gitkeep
-
-/test/output.txt
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ npm-debug.log*
 
 /.nginx/*
 !/.nginx/.gitkeep
+
+/test/output.txt
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "dev": "nf start",
     "build": "vue-cli-service build",
     "lint": "vue-cli-service lint --no-fix --max-warnings 0 src/ bin/ test/ *.js",
-    "test": "NODE_ENV=test karma start"
+    "test": "./test/run.sh"
   },
   "volta": {
     "node": "16.19.1"
--- a/test/run.sh
+++ b/test/run.sh
@@ -1,0 +1,10 @@
+#!/bin/bash -eu
+set -o pipefail
+
+NODE_ENV=test karma start | tee test/output.txt
+if grep -F -e 'WARN LOG:' -e 'WARN [web-server]:' -C5 test/output.txt; then
+	echo
+	echo 'All tests passed, but there were warnings: see above.'
+	exit 1
+fi
+rm test/output.txt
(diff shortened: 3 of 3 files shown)


sadiqkhoja (commented):
I think this can be done without writing to output.txt
```
NODE_ENV=test karma start | tee /dev/tty | grep -F -e 'WARN LOG:' -e 'WARN [web-server]:' -C5 | tee /dev/tty | grep -c -F -e 'WARN LOG:' -e 'WARN [web-server]:' | grep -q 0 || echo 'All tests passed, but there were warnings: see above.'
```

matthew-white:
Ooh I was trying to find something like piping from `tee /dev/tty`! I'll remember that one for the future. ✨

One thing that I think would be nice is to only show warnings at the bottom if all tests passed. If a test failed, then I think we can stop after `karma start`. If there's a Vue warning, it's often the case that the test is failing, and in that case, I don't think displaying the warning at the bottom is helpful. It only becomes important to highlight these warnings if tests are passing.

I don't think there's an easy way to have that sort of logic in a pipeline, is there? (The exit status from `karma start` won't be readily available until the end of the pipeline.) What do you think about writing to a file, but deleting the file at the end of the script more consistently? Also, maybe I could use `mktemp` and write to a temporary file rather than writing to test/output.txt.

sadiqkhoja (approved):

matthew-white:
> What do you think about writing to a file, but deleting the file at the end of the script more consistently? Also, maybe I could use `mktemp` and write to a temporary file rather than writing to test/output.txt.

I've made both of these changes. I'm going to go ahead and merge, but let me know if you think of other improvements!

matthew-white:
I tried the latest version of the script in a few different scenarios. In all cases, I checked (1) whether the exit status was zero or non-zero, (2) whether a warning was shown, and (3) whether the temporary file was removed. (Everything seemed OK.)

- All tests are passing, and there are no warnings.
- A test is failing, but there are no warnings.
- All tests are passing, but there is a warning.
- A test is failing, and there is a warning.
- `karma start` is interrupted.

PR #775, opened 2023-04-10
Add max width to PageBody

### LINKED ISSUE #447: Set maximum width for frontend

matthew-white:
Closes #447.

This PR is an alternative to #764. Rather than adding a max width to the entire app, this PR adds a max width to only the `PageBody` component. As a result, the navbar is still full-width, as are navigation tabs and the page title above them. By default, the main content of the page (the `PageBody`) has a max width of 1280px. A few pages are exceptions and are still full-width:

- The submissions and draft testing pages, since we want the submissions table to be full-width.
- The entities page, since we want the entities table to be full-width.
- The form access page. Similar to the submissions and entities tables, the table on this page can have an arbitrary number of columns, so I think it's nice for it to be full-width.
- The form overview page. This is needed so that the section about "Your Current Draft" is aligned with the navigation tabs for the draft.
- The audit log and the usage information page. I think it's nice for there to be more room for the details column of the audit log table. I could maybe also just reduce the width of the other columns. Definitely no strong feelings on this one.

Most but not all pages have a `PageBody`. I added a `PageBody` to `Home` so that the content below the summary at the top of the homepage will have a max width. That just left pages with a `.panel-main` element: the login page, the reset password page, the set password page, and the not found page. Those pages take a different approach from `PageBody`, wrapping the `.panel-main` in a `.col-sm-6`, and I didn't want to change them much. Instead of adding a `PageBody` to those pages, I added a max width to `.panel-main` that matches its width when the page width is 1280px.

~In #764, the app is centered on the page. I tried centering `PageBody`, but it looked odd to me. I think it's nice for `PageBody` to continue to be left-aligned.~

#764 only set the max width on the app when the page width was above 1500px — when the page width was quite a bit wider than the max width of 1280px on the app. That's because we wanted to avoid narrow gutters on the sides of the app: the margin to the left and right of the app had a different background from the app. In contrast, this PR always sets a max width of 1280px on `PageBody` (except for the pages that are full-width). I think that works fine because the margin has the same background as `PageBody`.

I've tried it out locally, and we had it on the QA server for a while. I've also written a couple of tests of `PageBody`.

The first approach to #447 that I tried was #764, and I compare the two approaches above.

In terms of the implementation of this PR, my main question was how to indicate to `PageBody` whether it should be full-width. At first, I added a `fullWidth` prop that was set by parent components (`ProjectShow`, `FormShow`, `DatasetShow`, `UserHome`, `SystemHome`). I had those parent components compute the prop based on the route path, but it didn't feel right to have route-related logic duplicated across a number of parent components. Ultimately, I concluded that whether the `PageBody` is full-width is a concern of the route, so it makes sense to set it on the route as a meta field, as was done in #764. It's guaranteed that there is only one `PageBody` on the page. Once that information is set on the route, `PageBody` can get it from there and doesn't need to be passed anything from its parent component.

We should quickly check each page with a max width (most pages) and make sure that the max width looks fine. If there are other pages that we want to make full-width, that's easy to do.

No changes needed.

### DIFF
--- a/src/assets/scss/_mixins.scss
+++ b/src/assets/scss/_mixins.scss
@@ -28,9 +28,10 @@
 }
 
 @mixin clearfix {
-  &::after {
-    clear: both;
+  &::before, &::after {
     content: " ";
     display: table;
   }
+
+  &::after { clear: both; }
 }
--- a/src/assets/scss/_variables.scss
+++ b/src/assets/scss/_variables.scss
@@ -94,6 +94,7 @@ $padding-modal-body: 15px;
 
 // PageBody
 $margin-top-page-body: 20px;
+$max-width-page-body: 1280px;
 
 // Popover
 $box-shadow-popover: 0 5px 10px rgba(0, 0, 0, 0.2);
--- a/src/assets/scss/_variables.scss
+++ b/src/assets/scss/_variables.scss
@@ -15,6 +15,12 @@ $color-accent-primary: #bd006b;
 $color-accent-secondary: #8d0050;
 $color-page-background: #f7f7f7;
 
+// Media query breakpoints - copied from bootstrap
+$screen-xs-min: 480px;
+$screen-sm-min: 768px;
+$screen-md-min: 992px;
+$screen-lg-min: 1200px;
+
 // Text
 $color-text: #333;
 $font-family-monospace: Monaco, Menlo, Consolas, "Courier New", monospace;
@@ -91,9 +97,3 @@ $margin-top-page-body: 20px;
 
 // Popover
 $box-shadow-popover: 0 5px 10px rgba(0, 0, 0, 0.2);
-
-// Media query breakpoints - copied from bootstrap
-$screen-xs-min: 480px;
-$screen-sm-min: 768px;
-$screen-md-min: 992px;
-$screen-lg-min: 1200px;
--- a/src/components/page/body.vue
+++ b/src/components/page/body.vue
@@ -26,8 +26,7 @@ import { computed, inject } from 'vue';
 const { router } = inject('container');
 const htmlClass = computed(() => ({
   // `router` may be `null` in testing.
-  bound: router == null || !router.currentRoute.value.meta.fullWidth,
-  centered: window.centralOptions?.left !== true
+  'full-width': router != null && router.currentRoute.value.meta.fullWidth
 }));
 </script>
 
@@ -36,13 +35,11 @@ const htmlClass = computed(() => ({
 
 #page-body {
   @include clearfix;
+  margin-left: auto;
+  margin-right: auto;
   margin-top: $margin-top-page-body;
+  max-width: $max-width-page-body;
 
-  &.bound { max-width: $max-width-page-body; }
-
-  &.centered {
-    margin-left: auto;
-    margin-right: auto;
-  }
+  &.full-width { max-width: none; }
 }
 </style>
--- a/src/components/page/body.vue
+++ b/src/components/page/body.vue
@@ -10,7 +10,7 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <div id="page-body" class="row">
+  <div id="page-body" :class="htmlClass">
     <div class="col-xs-12">
       <slot></slot>
     </div>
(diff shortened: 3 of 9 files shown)


matthew-white:
I've looked at it more, and centering `PageBody` is growing on me. I've pushed a commit that centers `PageBody` by default and provides an option to left-align.

matthew-white:
This PR is now ready for review! I'll leave a few comments now.

There is discussion above about whether to center `PageBody` or left-align it. We landed on centering it.

src/components/page/body.vue
```diff
@@ -10,10 +10,8 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <div id="page-body" class="row">
```
matthew-white, line 13: We need to add margins to the element, but `.row` already has margins. And actually, I don't think this element even needs to use `.row`: I don't think `.row` + `.col-xs-12` has much of an effect.

src/components/page/body.vue
```diff
@@ -22,11 +20,26 @@ export default {
...
+  'full-width': router != null && router.currentRoute.value.meta.fullWidth
+}));
+</script>
 
 <style lang="scss">
-@import '../../assets/scss/variables';
+@import '../../assets/scss/mixins';
 
 #page-body {
+  @include clearfix;
```
matthew-white, line 37: `.row` has clearfix styles. Now that `PageBody` won't use `.row`, I thought I'd add a clearfix to `PageBody`.

src/assets/scss/_mixins.scss
```diff
@@ -28,9 +28,10 @@
 }
 
 @mixin clearfix {
-  &::after {
-    clear: both;
+  &::before, &::after {
```
matthew-white, line 31: I was looking at the Bootstrap clearfix and realized that this mixin was missing `::before` here.

src/components/page/body.vue
```diff
@@ -22,11 +20,26 @@ export default {
   name: 'PageBody'
 };
 </script>
+<script setup>
+import { computed, inject } from 'vue';
+
+const { router } = inject('container');
+const htmlClass = computed(() => ({
+  // `router` may be `null` in testing.
```
matthew-white, line 28: It's rare for a test to render a `PageBody` component but not inject a router. However, that happens in the tests of `AsyncRoute`.

src/components/page/body.vue
```diff
@@ -22,11 +20,26 @@ export default {
   name: 'PageBody'
 };
 </script>
+<script setup>
+import { computed, inject } from 'vue';
+
+const { router } = inject('container');
```
matthew-white, line 26: We can't use `useRouter()` in case the test didn't inject a router.

sadiqkhoja (approved):

src/assets/scss/_variables.scss
```diff
@@ -15,6 +15,12 @@ $color-accent-primary: #bd006b;
 $color-accent-secondary: #8d0050;
 $color-page-background: #f7f7f7;
 
+// Media query breakpoints - copied from bootstrap
+$screen-xs-min: 480px;
+$screen-sm-min: 768px;
+$screen-md-min: 992px;
+$screen-lg-min: 1200px;
```
sadiqkhoja, line 22: some day we can import bootstrap sass file instead of its compiled css file

matthew-white: I agree that that would be a useful thing. :+1:

PR #483, opened 2021-08-06
Preview Markdown while commenting on a submission

ktuite:
The comments in the Submission activity feed support Markdown, and now that Markdown can be previewed as it is typed!

Screen Shot 2021-08-06 at 2 40 34 PM

There is a new reusable component called `MarkdownTextarea` that is used in the two places we accept comments on Submissions:
- Submissions table when changing the review state
- Submission activity feed

There are a few configurable properties on this `MarkdownTextarea`:
1. The prop in the parent component that will be bound to a prop called `value`. This is so the outer component can continue to handle the comment submission/POST request.
2. The parent compnent's `awaitingRepsonse` prop so some of the UI elements will still be displayed during requests.
3. If the footer with the link "Markdown supported" and "Preview" button should be shown when the component is initialized, or if it should wait until the user starts typing text.
4. Default placerholder text
5. Additional `<slot>` HTML for any buttons that appear to the right of the "Preview" button. 

### Markdown Preview on Submissions table comments
(Comment submit button is the "Update" button on bottom left, not next to the "Preview" button)
Screen Shot 2021-08-06 at 2 43 33 PM
Screen Shot 2021-08-06 at 2 43 54 PM
Screen Shot 2021-08-06 at 2 43 58 PM

### Markdown Preview on Submission activity feed when Comment for an Edit is needed
(Shows footer and active blue border by default)
Screen Shot 2021-08-06 at 2 38 17 PM
Screen Shot 2021-08-06 at 2 39 05 PM
Screen Shot 2021-08-06 at 2 39 09 PM

### Markdown Preview on Submission activity feed otherwise
(Doesn't show the footer by default)
Screen Shot 2021-08-06 at 2 59 48 PM
Screen Shot 2021-08-06 at 2 59 57 PM
Screen Shot 2021-08-06 at 3 00 22 PM
Screen Shot 2021-08-06 at 3 00 28 PM

PR #776, opened 2023-04-10
Move styles from SubmissionComment to MarkdownTextarea

matthew-white:
I was randomly looking at `SubmissionComment` and noticed that it sets styles on `#submission-comment-actions`. However, it looks like there isn't a `#submission-comment-actions` element anymore: I think that element was moved into `MarkdownTextarea` as part of #483 more than a year ago. This PR moves the styles from `SubmissionComment` into `MarkdownTextarea`, though given that they haven't been in use for a while, we'd probably also be fine removing them.

The styles do two things:

- Adds 5px of margin-top.
- Adds an animation on margin-bottom.

I've recorded the current behavior (without these styes) and the behavior with these styles. I slowed down the animation in the recording so that it's easier to see.

Current behavior (without styles):

https://user-images.githubusercontent.com/5970131/230971625-ee844d33-1bd7-4fad-bfbc-19786c4495be.mov

<br>With styles:<br>

https://user-images.githubusercontent.com/5970131/230971695-912d7d17-17d3-4d92-adad-647695c5a745.mov

<br>@ktuite, do you remember these styles at all? Do you know whether we intended to remove them?


src/components/markdown/textarea.vue
matthew-white, line 1: Let's not add the animation styles here; let's just remove them.

ktuite (approved):

matthew-white (commented):
Notes from interactive review

PR #777, opened 2023-04-10
Check for Sass warnings when running tests

matthew-white:
Sass can display warnings, and I think it'd be nice to automatically catch those warnings. Sass warnings seem to be shown whenever Frontend is built, including during testing, so I think we can add this check to our existing test/run.sh (introduced in #769).

I intentionally caused a Sass warning (by replacing `math.div()` with the `/` operator), then saw what the resulting warning looked like. It looks like sass-loader outputs a `Module Warning` with additional details, so that's what I search for.

I caused a Sass warning and verified that test/run.sh caught it.

We could check for Sass warnings separately, but I think it's convenient to do so in test/run.sh, where we're already checking for other warnings.

This is not a user-facing change.

No changes needed.

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -4,7 +4,9 @@ set -o pipefail
 output=$(mktemp)
 trap 'rm -- "$output"' EXIT
 NODE_ENV=test karma start | tee "$output"
-if grep -F -e 'WARN LOG:' -e 'WARN [web-server]:' -C5 "$output"; then
+# Search for: warnings from console.warn(), including Vue warnings; Sass
+# warnings; and warnings from Karma.
+if grep -F -e 'WARN LOG:' -e 'Module Warning' -e 'WARN [web-server]:' -C5 "$output"; then
 	echo
 	echo 'All tests passed, but there were warnings: see above.'
 	exit 1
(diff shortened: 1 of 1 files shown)


test/run.sh
```diff
@@ -4,7 +4,11 @@ set -o pipefail
 output=$(mktemp)
 trap 'rm -- "$output"' EXIT
 NODE_ENV=test karma start | tee "$output"
-if grep -F -e 'WARN LOG:' -e 'WARN [web-server]:' -C5 "$output"; then
+# Search for: warnings from console.warn(), including Vue warnings; Sass
+# warnings; and warnings from Karma.
+if grep -F -e 'WARN LOG:' -e 'Module Warning' -e 'WARN [web-server]:' -C5 "$output"; then
+	# Reset the text format in case the search results contained formatted text.
+	tput sgr0
```
matthew-white, line 11: The Sass warning I caused was pretty long and was also bold. `grep` displayed the start of the bold text but not the end, so before I added `tput sgr0`, my terminal was still bold after running `npm run test`. Given that `npm run test` displays formatted text, I think this is a possibility to account for in general (not just for Sass warnings).

sadiqkhoja (approved):

PR #783, opened 2023-04-18
Update link to user-facing docs

yanokwa:


matthew-white (approved):

matthew-white:
@lognaturel, you mentioned possibly moving the pyODK docs to docs.getodk.org. Would that happen before .3 is released? If so, maybe we should hold off on merging this PR?

yanokwa:
We could use `https://docs.getodk.org/pyodk` as the URL. That redirects now to Github and be the likely URL. What say ye, @lognaturel.

matthew-white:
It looks like there's a conflict now that we've merged #784. The component was moved to src/components/odata/analyze.vue.

In terms of which link to use, either approach seems reasonable to me!

yanokwa:
Fixed conflict and updated link to docs.getodk.org

matthew-white (approved):
Thanks, @yanokwa! I'll go ahead and merge now.

PR #788, opened 2023-05-09
Move required asterisk to before label/placeholder

matthew-white:
In Collect and Enketo, the asterisk for a required form field appears before the text, but in Central, it's after. This PR makes things more consistent by changing it to before the text in Central as well. Here's an example from the login page:

I searched src/components for `*`, and I'm pretty sure the only places to change are `FormGroup` and `DateRangePicker`. I've viewed the change locally as well.

Probably some screenshots will be out-of-date, but I wouldn't think we need to update those just for this.


matthew-white:
@lognaturel, would you mind taking a look at the screenshot just to confirm that this is a change we want to make?

matthew-white:
After discussion, we prefer the current placement of the asterisk.

PR #791, opened 2023-05-10
Update Entities diagram and description

lognaturel:
Before:
Screen Shot 2023-05-10 at 3 23 40 PM

After:
Screen Shot 2023-05-16 at 11 40 53 AM

Ran it.

The goals for this change are:
- Match language in the documentation which has evolved based on user feedback
- Simplify the visual and make it easier to follow
- Start the visual workflow with Entity creation

I'm opening this PR as a draft to get feedback. @alyblenkin is it looking as you expect in context? Do you remember why the (2) line is dotted? @yanokwa does it match the docs sufficiently?

The most annoying thing is that translators will need to translate again. I think it's worth it, though. The worst case would be that this change makes Entities harder to understand.

No.


yanokwa:
I think the copy is good. If we could find some way to combine the copy and the image, that'd feel more human. 

It'd be nice to have some words in the diagram like we do at https://docs.getodk.org/_images/intro-diagram.png. The 2 and 4 are also horizontally misaligned. I also kinda liked the paper clip icons in the old one.

lognaturel:
> find some way to combine the copy and the image

Localization is the main barrier to that.

> "A Dataset is attached to a form just like a CSV" requires a bit too much context

Sounds good

> words in the diagram

I think it would be weird to have text in one language and words on the image in another (English).

> kinda liked the paper clip icons in the old one.

Either looks good to me. I think the black background did bring focus to attaching which is important.

alyblenkin:
Sounds good. 

Here is the updated SVG:
Central_light

lognaturel:
@yanokwa asked me to try putting the diagram above and the text below:
Screen Shot 2023-05-16 at 11 35 17 AM

I think that to go further in that direction we'd need to reconsider the layout more deeply.

I've tried making the text container a little wider and that feels like a nice balance:
Screen Shot 2023-05-16 at 11 37 20 AM

But I also think the layout from before is fine:
Screen Shot 2023-05-16 at 11 35 45 AM

transifex/strings_en.json
```diff
@@ -1253,23 +1253,23 @@
...
+          "string": "A Form {icon1} can specify that its Submissions create Entities in a Dataset {icon2}. By default, Entities are created when a Submission is received. They can also be configured to require Submission approval.",
+          "developer_comment": "{icon1} and {icon2} are icons that display the numbers 1 and 2. They refer to the first and second step in an image that shows the relationship between Forms and Entities."
         },
         "2": {
-          "string": "When a Submission related to a Dataset is approved {icon2}, the selected data is added {icon3} to the Dataset, so that the next time a client downloads the Form, the new data will be included in the data download.",
-          "developer_comment": "{icon2} and {icon3} are icons that display the numbers 2 and 3. They refer to the second and third steps in an image that depicts this workflow."
+          "string": "A Dataset is attached {icon3} to a Form just like a CSV. When data collectors update Forms with an attached Dataset, they will automatically download all Entities created so far.",
```
sadiqkhoja, line 1267: Don't we want to drop "A Dataset is attached to a form just like a CSV"?

public/images/entities-intro.svg
sadiqkhoja, line 1: icons for form and dataset looks same, can we use database cylinder icon ⛁ for dataset?

lognaturel: I agree that they're pretty similar. @alyblenkin and I decided against a database icon because we don't think it would have meaning to most non-technical folks. A barrel? A stack of coins?

Maybe something more like a table would be more clearly different (e.g. https://www.shutterstock.com/image-vector/database-table-entries-tabular-data-line-1874648533). Or playing with shading and line weight?

alyblenkin: Thanks for the feedback @sadiqkhoja! We tried to customize the icon, but I agree I think it's still too similar. 

I've played around with a few more options using the [Phosphors icon library](https://phosphoricons.com/). What do you think? I'm liking the 3 diagram ☺️


Screenshot 2023-05-18 at 9 11 55 AM

yanokwa: I think the table icon (3) is worse than the database. I vote database icon.

alyblenkin: I'm happy to go with the database icon if that's what the group thinks is best. The text will hopefully make it clear if anyone hasn't seen that icon before.

matthew-white: > I think we can add translatable text in the image as well if that helps. Since svg is just xml we can place it in the vue template with the $t functions.

Yeah, I bet there's some answer there! At the very least, we could put the text in separate elements and position them over the image.

lognaturel:
Screen Shot 2023-05-18 at 11 06 07 AM

Agreed putting the numbers first is clearer. Having bold titles adds complexity and doesn't wrap very well so hopefully you're ok with this. And hopefully it's now clear enough without the overlaid text.

alyblenkin (approved):

sadiqkhoja (approved):

PR #792, opened 2023-05-11
Add ability to update entity

matthew-white:


test/components/entity/update.spec.js
```diff
@@ -0,0 +1,237 @@
...
+    });
+  });
+
+  it('focuses the entity label textarea', () => {
+    testData.extendedEntities.createPast(1);
+    const modal = mountComponent({ attachTo: document.body });
+    modal.get('textarea').should.be.focused();
+  });
+
+  it('resizes the textarea elements', async () => {
```
matthew-white, line 75: Not reviewing this test.

ktuite (approved):

matthew-white (commented):
@ktuite and I didn't review some tests related to dynamic styles, for example, tests of `TextareaAutosize`. If there's an issue there, its scope should be limited, and we'll hopefully catch it in QA. I've also tried things out locally.

test/components/entity/show.spec.js
```diff
@@ -53,4 +54,75 @@ describe('EntityShow', () => {
...
+      const component = await submit();
+      component.get('#page-head-title').text().should.equal('Updated Entity');
+    });
+
+    it('updates the entity data', async () => {
+      const component = await submit();
+      component.get('#entity-data dd').text().should.equal('2');
+    });
+
+    it('updates the number of entries in the feed');
```
matthew-white, line 126: This will be tested in a follow-up PR.

PR #795, opened 2023-05-16
new: added dataset settings page

sadiqkhoja:

Manual testing + integration tests

NA

None

https://github.com/getodk/docs/issues/1611


test/components/dataset/settings.spec.js
```diff
@@ -0,0 +1,102 @@
...
+import testData from '../../data';
+import { load } from '../../util/http';
+import { mockLogin } from '../../util/session';
+
+describe('DatasetSettings', () => {
+  beforeEach(mockLogin);
+
+  it('should have onReceipt selected', async () => {
+    testData.extendedProjects.createPast(1);
+    testData.extendedDatasets.createPast(1);
```
matthew-white, line 12: Just FYI, creating a dataset automatically creates a project if there isn't one already.

test/components/dataset/settings.spec.js
```diff
@@ -0,0 +1,102 @@
...
+    component.get('input[value="true"]').element.checked.should.be.true();
+    component.get('input[value="false"]').element.checked.should.be.false();
+  });
+
+  it('should sends the correct request', async () => {
+    testData.extendedProjects.createPast(1);
+    testData.extendedDatasets.createPast(1);
+    await load('/projects/1/datasets/trees/settings', { root: false })
+      .complete()
+      .request(async (component) => component.get('input[value="true"]').trigger('change'))
```
matthew-white, line 31: I think `setValue(true)` would be better here than `trigger('change')`. `setValue(true)` will trigger a `change` event, but it will also set the `checked` property of the element. As far as I know, `trigger()` just triggers the event.

src/components/dataset/settings.vue
```diff
@@ -0,0 +1,133 @@
...
+    url: apiPaths.dataset(project.id, dataset.name, { convert }),
+    data: { approvalRequired: approvalRequired.value },
+    fulfillProblem: ({ code }) => code === 400.29,
+    patch: ({ data }) => {
+      if (isProblem(data)) {
+        pendingSubmissionModal.value.pendingSubmissions = data.details.count;
+        pendingSubmissionModal.value.state = true;
+      } else {
+        dataset.approvalRequired = data.approvalRequired;
+        alert.success(dataset.approvalRequired ? t('onApproval.successMessage') : t('onReceipt.successMessage'));
```
matthew-white, line 96: I think it'd be good to test this alert.

src/components/dataset/settings.vue
```diff
@@ -0,0 +1,133 @@
...
+
+const update = (convert) => {
+  dataset.request({
+    method: 'PATCH',
+    url: apiPaths.dataset(project.id, dataset.name, { convert }),
+    data: { approvalRequired: approvalRequired.value },
+    fulfillProblem: ({ code }) => code === 400.29,
+    patch: ({ data }) => {
+      if (isProblem(data)) {
+        pendingSubmissionModal.value.pendingSubmissions = data.details.count;
```
matthew-white, line 92: Could you add a test that after a 400.29 Problem response, this count is shown somewhere in the modal? I think it'd be good to have a test that that value is being passed through correctly.

matthew-white (commented):
Leaving a bunch of small comments, but this is looking very solid to me. The flow of data and the communication between components is effective. I also like the Composition API components!

test/components/dataset/pending-submissions.spec.js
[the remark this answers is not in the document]
sadiqkhoja: this modal can't be opened with simple click, so I am not sure what should I provide in `show` argument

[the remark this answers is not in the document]
matthew-white: Ah good point. Maybe at some point we allow `show` to be a function for cases like this. I think we can skip for now!

src/components/dataset/pending-submissions.vue
```diff
@@ -0,0 +1,110 @@
...
+<i18n lang="json5">
+{
+  "en": {
+    "title": "Pending Submissions",
+    "action": {
+      "changeSetting": "Change setting"
+    },
+    "explanation": {
+      "first": "You are setting Entity creation to occur when Submissions are first received by Central.",
+      "second": "Once this takes effect, Entities will no longer be generated when Submissions are marked Approved, including {records} we found that have neither been marked Approved or Rejected. | Once this takes effect, Entities will no longer be generated when Submissions are marked Approved, including {records} we found that have neither been marked Approved or Rejected.",
```
matthew-white, line 97: "has" instead of "have" in the first case. Also "nor" instead of "or" in both cases.

src/components/dataset/settings.vue
```diff
@@ -0,0 +1,133 @@
...
+            <label>
+              <input v-model="approvalRequired" name="approvalRequired" type="radio" :value="false"
+                aria-describedby="dataset-setting-on-receipt" :aria-disabled="dataset.awaitingResponse" @change="update()">
+              <strong>{{ $t('onReceipt.label') }}</strong>
+            </label>
+            <p id="dataset-setting-on-receipt" class="help-block">
+              {{ $t('onReceipt.description') }}
+            </p>
+          </div>
+          <div class="radio">
```
matthew-white, line 30: For radios and checkboxes, add a `disabled` class to the `<div>` if the input is disabled so that the whole thing looks disabled:

```vue
<div class="radio" :class="{ disabled: dataset.awaitingResponse }">
```

src/components/dataset/settings.vue
```diff
@@ -0,0 +1,133 @@
...
+            <label>
+              <input v-model="approvalRequired" name="approvalRequired" type="radio" :value="true"
```
matthew-white, line 32: I'm actually not sure whether `aria-disabled` works here or if it should be `disabled`. This is our first disabled radio input, so you are breaking new ground. For most inputs, it's `aria-disabled`, but in this case, I think it should probably be `disabled`. `disabled` is what's needed for checkboxes: see the comments about checkboxes in src/composables/disabled.js. Some of those comments are relevant to radio inputs as well. In app.scss, it looks like we style radio inputs based on `disabled`, not `aria-disabled`.

src/components/dataset/settings.vue
```diff
@@ -0,0 +1,133 @@
...
+      if (isProblem(data)) {
+        pendingSubmissionModal.value.pendingSubmissions = data.details.count;
+        pendingSubmissionModal.value.state = true;
+      } else {
+        dataset.approvalRequired = data.approvalRequired;
+        alert.success(dataset.approvalRequired ? t('onApproval.successMessage') : t('onReceipt.successMessage'));
+      }
+    }
+  })
+    .catch(noop);
```
matthew-white, line 99: Should we reset `approvalRequired.value` to `dataset.approvalRequired` if there's an error? Otherwise, to retry the request, the user would first have to change the radio back to its previous state, which would send a request.

test/data/seed.js
matthew-white, line 1: It looks like there's a conflict with this file after I merged all those PRs.

matthew-white (approved):
Looks great! ✨ The communication between Frontend and Backend works really nicely, I like the `convert` query parameter.

src/components/dataset/settings.vue
```diff
@@ -0,0 +1,133 @@
...
+    method: 'PATCH',
+    url: apiPaths.dataset(project.id, dataset.name, { convert }),
+    data: { approvalRequired: approvalRequired.value },
+    fulfillProblem: ({ code }) => code === 400.29,
+    patch: ({ data }) => {
+      if (isProblem(data)) {
+        pendingSubmissionModal.value.pendingSubmissions = data.details.count;
+        pendingSubmissionModal.value.state = true;
+      } else {
+        dataset.approvalRequired = data.approvalRequired;
```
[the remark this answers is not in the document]
matthew-white: Also feel free to skip for now!

PR #807, opened 2023-05-23
Use property odataName in entities table

### LINKED ISSUE #793: Dataset properties that are not valid OData identifiers are always empty in entities table
If the name of a dataset property is not a valid OData identifier, then a different name will be used for the property in the OData feed (via `sanitizeOdataIdentifier()` in Backend). In Frontend, the entities table uses the OData feed, but right now, if the name of a property differs from its OData identifier, the property's value is never shown. For example, I created a dataset with a property named `circumference.cm`, and the property is empty in the table: https://staging.getodk.cloud/#/projects/23/datasets/trees/entities. Yet if you access this entity using the REST API, you can see that its `circumference.cm` property is not empty: https://staging.getodk.cloud/v1/projects/23/datasets/trees/entities/c3e0d5f0-2af5-44bb-ad3f-26fb9f93288b. I think the issue is that we try to access property data using the name of the property, not its OData identifier:

https://github.com/getodk/central-frontend/blob/1e7902b988ed7ba8cbaf661d5415ab849ff32506/src/components/entity/data-row.vue#L14-L15

This isn't an issue in the submissions table because Frontend specifies a query parameter of odata=true when it sends a request to the form fields endpoint. As a result, the endpoint returns OData identifiers for Frontend to use. I think we could have Backend do something similar for dataset properties, or we could import `sanitizeOdataIdentifier()` to Frontend and use it here.

matthew-white:
Closes #793.

A new test. Once it's on staging, I can verify it against the entity mentioned in the issue.

Backend returns the `odataName`, so Frontend just needs to use it.

No changes needed.

### DIFF
--- a/test/data/datasets.js
+++ b/test/data/datasets.js
@@ -3,6 +3,12 @@ import { comparator } from 'ramda';
 import { dataStore } from './data-store';
 import { extendedProjects } from './projects';
 
+const normalizeProperty = (property) => ({
+  odataName: property.name,
+  forms: [],
+  ...property
+});
+
 // eslint-disable-next-line import/prefer-default-export
 export const extendedDatasets = dataStore({
   factory: ({
@@ -23,7 +29,7 @@ export const extendedDatasets = dataStore({
     name,
     entities,
     lastEntity,
-    properties,
+    properties: properties.map(normalizeProperty),
     linkedForms,
     approvalRequired
   }),
--- a/src/components/entity/data-row.vue
+++ b/src/components/entity/data-row.vue
@@ -12,7 +12,7 @@ except according to the terms contained in the LICENSE file.
 <template>
   <tr>
     <td v-for="property of properties" :key="property.id">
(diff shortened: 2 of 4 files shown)


test/data/datasets.js
```diff
@@ -3,6 +3,12 @@ import { comparator } from 'ramda';
 import { dataStore } from './data-store';
 import { extendedProjects } from './projects';
 
+const normalizeProperty = (property) => ({
+  odataName: property.name,
```
sadiqkhoja, line 7: `odataName: property.odataName ?? property.name`

sadiqkhoja (approved):
a minor change, rest looks good

test/data/datasets.js
matthew-white: `property.name` is the default `odataName`. The defaults are merged with `property` using `...property`, so if `property` specifies its own `odataName`, it will overwrite the default.

PR #808, opened 2023-05-23
Make changes to tooltip styles

### LINKED ISSUE #753: Small changes to tooltips
@issa-tseng writes on Slack:

> - instant tooltips are nice but i'd eliminate the fade.
>   - it's also a little bottom heavy? like either a px or two needs to be added to the top or taken from the bottom or something between but there's more space at the bottom than the top rn

matthew-white:
Closes #753.

I viewed the results locally. Once the PR is on staging, I'll ask issa to take a look.

I'm going to leave comments about individual changes I made.

No changes needed.

### DIFF
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -86,7 +86,9 @@ em:lang(ja) {
   left: -1000px;
 }
 .tooltip-inner {
+  border-radius: 2px;
   overflow-wrap: break-word;
+  padding-bottom: 2px;
   // Allow the text to have line breaks.
   white-space: pre-line;
 }
--- a/src/directives/tooltip.js
+++ b/src/directives/tooltip.js
@@ -265,7 +265,7 @@ class Tooltip {
     }
 
     this.position()
-      .then(() => { this.element.classList.add('in', 'fade'); })
+      .then(() => { this.element.classList.add('in'); })
       .catch(noop);
     this.shown = true;
   }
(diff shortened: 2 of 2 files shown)


src/assets/scss/app.scss
```diff
@@ -86,7 +86,9 @@ em:lang(ja) {
   left: -1000px;
 }
 .tooltip-inner {
+  border-radius: 2px;
```
matthew-white, line 89: The Bootstrap default is 4px. 2px matches the `border-radius` of `.btn`.

src/assets/scss/app.scss
```diff
@@ -86,7 +86,9 @@ em:lang(ja) {
...
   overflow-wrap: break-word;
+  padding-bottom: 2px;
```
matthew-white, line 91: The Bootstrap default is 3px, so this is a reduction of 1px.

issa wrote:

> either a px or two needs to be added to the top or taken from the bottom or something between

I tried adding 1px to the top, but even that felt like possibly too much padding. I think it's nice for tooltips to be smaller and to not have too much padding.

I also tried removing 2px from the bottom, but that felt like too much.

src/directives/tooltip.js
```diff
@@ -265,7 +265,7 @@ class Tooltip {
     }
 
     this.position()
-      .then(() => { this.element.classList.add('in', 'fade'); })
+      .then(() => { this.element.classList.add('in'); })
```
matthew-white, line 268: We need to keep the `in` class because it sets the tooltip's opacity.

sadiqkhoja (approved):
Code looks good. I wish there was a screenshot in the PR as well

PR #817, opened 2023-05-29
link dataset to attachment URL

sadiqkhoja:
## Changes:
- Dataset hyperlink on attachments page links to attachment URL instead of entities.csv

Verified locally, all tests are passing

NA

No

No


matthew-white (approved):
Change looks good! I'm a little surprised though that no test had to change. Do we not test the `href` attribute of the dataset link? If you have a moment, would you mind quickly adding a test?

PR #818, opened 2023-05-30
Make changes for v2023.3 after design review

matthew-white:
See the commits for the specific changes. Here are a few screenshots:

**Change is indicated on the entire row:**

<br>
<br>

**"Current value" and "Updated value" have the same width:**

Tests and/or viewing things locally.

I'm going to leave some comments on individual lines.

We should maybe ask the QA team to take another quick pass on the entity update modal.

N/A

### DIFF
--- a/src/components/entity/update.vue
+++ b/src/components/entity/update.vue
@@ -22,7 +22,9 @@ except according to the terms contained in the LICENSE file.
                 <th ref="labelCellHeader" class="label-cell">
                   <span class="sr-only">{{ $t('resource.property') }}</span>
                 </th>
-                <th class="old-value">{{ $t('header.currentValue') }}</th>
+                <th ref="oldValueHeader" class="old-value">
+                  {{ $t('header.currentValue') }}
+                </th>
                 <th class="new-value">{{ $t('header.updatedValue') }}</th>
               </tr>
             </thead>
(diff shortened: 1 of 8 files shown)


src/components/entity/update.vue
```diff
@@ -155,7 +173,6 @@ const currentVersion = computed(() =>
   table { table-layout: fixed; }
   thead {
     .label-cell { width: 16.66666667%; }
-    .old-value { width: 33.33333333%; }
```
matthew-white, line 158: By removing this, `.old-value` and `.new-value` get the remaining width not consumed by `.label-cell`. The remaining width will be equally divided between the two because the table has `table-layout: fixed;`.

src/components/entity/update/row.vue
```diff
@@ -85,10 +85,7 @@ const resize = () => {
   textarea.value.resize();
   if (minHeightOutdated) setMinHeight();
 };
-const focus = () => textarea.value.focus();
-defineExpose({
-  textarea: { resize, focus }
-});
+defineExpose({ textarea: computed(() => ({ ...textarea.value, resize })) });
```
matthew-white, line 88: We need to expose `el` from the textarea. But that's a little harder than it was for `resize()` and `focus()`, because `el` isn't a function: `textarea.value` is a reactive object, and `textarea.value.el` is the element (not a ref of the element). I decided to take a different approach, copying `textarea.value` and just overwriting the `resize()` function. That actually feels like a better approach in general than what it was before.

ktuite (approved):

PR #819, opened 2023-05-31
 Support Single Sign-on with OpenID Connect

alxndrsn:
Frontend portion of getodk/central#449. These are Frontend changes required for https://github.com/getodk/central-backend/pull/910

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
 #vue: vue-cli-service build --mode development --watch
-vue: vue-cli-service serve
+vue: vue-cli-service serve # FIXME revert?
 nginx: nginx -c "$PWD/main.nginx.conf" -p "$PWD"
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,2 @@
-#vue: vue-cli-service build --mode development --watch
-vue: vue-cli-service serve
+vue: vue-cli-service build --mode development --watch
 nginx: nginx -c "$PWD/main.nginx.conf" -p "$PWD"
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
 #vue: vue-cli-service build --mode development --watch
-vue: vue-cli-service serve # FIXME revert?
+vue: PORT=5000 vue-cli-service serve # FIXME revert?
 nginx: nginx -c "$PWD/main.nginx.conf" -p "$PWD"
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,3 @@
-vue: vue-cli-service build --mode development --watch
+#vue: vue-cli-service build --mode development --watch
+vue: vue-cli-service serve
 nginx: nginx -c "$PWD/main.nginx.conf" -p "$PWD"
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,2 @@
-#vue: vue-cli-service build --mode development --watch
-vue: PORT=5000 vue-cli-service serve # FIXME revert?
+vue: vue-cli-service build --mode development --watch
 nginx: nginx -c "$PWD/main.nginx.conf" -p "$PWD"
--- a/main.nginx.conf
+++ b/main.nginx.conf
@@ -63,14 +63,12 @@ http {
 
     location /- {
       proxy_pass http://localhost:8005/-;
-      proxy_cache off; # FIXME revert?
       proxy_redirect off;
       proxy_set_header Host $host;
     }
 
     location ~ ^/v\d {
       proxy_pass http://localhost:8383;
-      proxy_cache off; # FIXME revert?
       proxy_redirect off;
 
       # buffer requests, but not responses, so streaming out works.
@@ -97,9 +95,7 @@ http {
     }
 
     location / {
-      proxy_pass http://localhost:5000/; # FIXME revert?
-      proxy_cache off; # FIXME revert?
-      proxy_redirect off; # FIXME revert?
+      root ./dist;
 
       include ./common-headers.nginx.conf;
       # We return this header for more files in development than we do in
--- a/main.nginx.conf
+++ b/main.nginx.conf
@@ -63,14 +63,12 @@ http {
 
     location /- {
       proxy_pass http://localhost:8005/-;
-      proxy_cache off;
       proxy_redirect off;
       proxy_set_header Host $host;
     }
 
     location ~ ^/v\d {
       proxy_pass http://localhost:8383;
-      proxy_cache off;
       proxy_redirect off;
 
       # buffer requests, but not responses, so streaming out works.
@@ -97,9 +95,7 @@ http {
     }
 
     location / {
-      proxy_pass http://localhost:5000/;
-      proxy_cache off;
-      proxy_redirect off;
+      root ./dist;
 
       include ./common-headers.nginx.conf;
       # We return this header for more files in development than we do in
--- a/main.nginx.conf
+++ b/main.nginx.conf
@@ -63,14 +63,14 @@ http {
 
     location /- {
       proxy_pass http://localhost:8005/-;
-      proxy_cache off;
+      proxy_cache off; // FIXME revert?
       proxy_redirect off;
       proxy_set_header Host $host;
     }
 
     location ~ ^/v\d {
       proxy_pass http://localhost:8383;
-      proxy_cache off;
+      proxy_cache off; // FIXME revert?
       proxy_redirect off;
 
       # buffer requests, but not responses, so streaming out works.
@@ -97,9 +97,9 @@ http {
     }
 
     location / {
-      proxy_pass http://localhost:5000/;
-      proxy_cache off;
-      proxy_redirect off;
+      proxy_pass http://localhost:5000/; // FIXME revert?
+      proxy_cache off; // FIXME revert?
+      proxy_redirect off; // FIXME revert?
 
       include ./common-headers.nginx.conf;
       # We return this header for more files in development than we do in
--- a/main.nginx.conf
+++ b/main.nginx.conf
@@ -63,12 +63,14 @@ http {
 
     location /- {
       proxy_pass http://localhost:8005/-;
+      proxy_cache off;
       proxy_redirect off;
       proxy_set_header Host $host;
     }
 
     location ~ ^/v\d {
       proxy_pass http://localhost:8383;
+      proxy_cache off;
       proxy_redirect off;
 
       # buffer requests, but not responses, so streaming out works.
@@ -95,7 +97,9 @@ http {
     }
 
     location / {
-      root ./dist;
+      proxy_pass http://localhost:5000/;
+      proxy_cache off;
+      proxy_redirect off;
 
       include ./common-headers.nginx.conf;
       # We return this header for more files in development than we do in
--- a/main.nginx.conf
+++ b/main.nginx.conf
@@ -63,14 +63,14 @@ http {
 
     location /- {
       proxy_pass http://localhost:8005/-;
-      proxy_cache off; // FIXME revert?
+      proxy_cache off; # FIXME revert?
       proxy_redirect off;
       proxy_set_header Host $host;
     }
 
     location ~ ^/v\d {
       proxy_pass http://localhost:8383;
-      proxy_cache off; // FIXME revert?
+      proxy_cache off; # FIXME revert?
       proxy_redirect off;
 
       # buffer requests, but not responses, so streaming out works.
@@ -97,9 +97,9 @@ http {
     }
 
     location / {
-      proxy_pass http://localhost:5000/; // FIXME revert?
-      proxy_cache off; // FIXME revert?
-      proxy_redirect off; // FIXME revert?
+      proxy_pass http://localhost:5000/; # FIXME revert?
+      proxy_cache off; # FIXME revert?
+      proxy_redirect off; # FIXME revert?
 
       include ./common-headers.nginx.conf;
       # We return this header for more files in development than we do in
--- a/src/components/account/login.vue
+++ b/src/components/account/login.vue
@@ -88,13 +88,7 @@ export default {
     }
   },
   created() {
-    const { oidcError } = this.$route.query;
-    if (this.config.oidcEnabled && typeof oidcError === 'string' &&
-      /^[\w-]+$/.test(oidcError)) {
-      const path = `oidc.error.${oidcError}`;
-      if (this.$te(path, this.$i18n.fallbackLocale))
-        this.alert.danger(this.$t(path));
-    }
+    this.handleOIDCError();
   },
   mounted() {
     if (this.config.oidcEnabled)
@@ -121,6 +115,15 @@ export default {
       }
 
       this.disabled = true;
+    },
+    handleOIDCError() {
+      const { oidcError } = this.$route.query;
+      if (this.config.oidcEnabled && typeof oidcError === 'string' &&
+        /^[\w-]+$/.test(oidcError)) {
+        const path = `oidc.error.${oidcError}`;
+        if (this.$te(path, this.$i18n.fallbackLocale))
+          this.alert.danger(this.$t(path));
+      }
     },
     /* Pressing the back button on the IdP login page may restore Frontend from
     the back/forward cache. In that case, this.disabled would still be `true` --
--- a/src/components/account/login.vue
+++ b/src/components/account/login.vue
@@ -186,9 +186,9 @@ export default {
     "oidc": {
       "body": "Click Continue to proceed to the login page.",
       "error": {
-        "auth-ok-user-not-found": "Authentication successful, but your user does not exist in ODK Central.",
-        "email-not-verified": "Your email address has not been verified by the authentication provider.",
-        "provider-misconfigured": "There was a problem with your authentication provider. Please contact the site administrator.",
+        "auth-ok-user-not-found": "There is no Central account associated with your email address. Please ask your Central administrator to create an account for you to continue.",
+        "email-not-verified": "Your email address has not been verified by your login server. Please contact your server administrator.",
+        "provider-misconfigured": "Central could not access the email address associated with your account. This could be because your server administrator has configured something incorrectly, or has not set an email address for your account. It could also be the result of privacy options that you can choose during the login process. If so, please try again and ensure that your email is shared."
       },
     },
     "problem": {
--- a/src/components/account/login.vue
+++ b/src/components/account/login.vue
@@ -18,10 +18,8 @@ except according to the terms contained in the LICENSE file.
         </div>
         <div v-if="config.oidcEnabled" class="panel-body">
           <div class="panel-footer">
-            <a href="/v1/oidc/login">
-              <button class="btn btn-primary">
-                {{ $t('action.logIn') }} (TODO add spinner when clicked)
-              </button>
+            <a href="/v1/oidc/login" class="btn btn-primary">
+              {{ $t('action.logIn') }} (TODO add spinner when clicked)
             </a>
           </div>
         </div>
--- a/src/components/account/login.vue
+++ b/src/components/account/login.vue
@@ -79,6 +79,9 @@ export default {
   },
   mounted() {
     if (!this.config.oidcEnabled) this.$refs.email.focus();
+
+    const { oidcError } = this.$route.query;
+    if (oidcError) this.alert.danger(this.$t(`oidc.error.${oidcError}`));
   },
   methods: {
     navigateToNext(
@@ -154,6 +157,12 @@ export default {
       "alreadyLoggedIn": "A user is already logged in. Please refresh the page to continue.",
       "changePassword": "Your password is shorter than 10 characters. To protect your account, please change your password to make it longer."
     },
+    "oidc": {
+      "error": {
+        "email-not-verified": "Your email address has not been verified by the authentication provider.",
+        "provider-misconfigured": "There was a problem with your authentication provider.  Please contact the site administrator.",
+      },
+    },
     "problem": {
       "401_2": "Incorrect email address and/or password."
     }
@@ -186,6 +195,12 @@ export default {
     "alert": {
       "alreadyLoggedIn": "Un usuario ya ha iniciado sesión. Actualice la página para continuar.",
       "changePassword": "Su contraseña tiene menos de 10 caracteres. Para proteger su cuenta, cambie su contraseña para que sea más larga."
+    },
+    "oidc": {
+      "error": {
+        "email-not-verified": "Tu email no ha estado verificado por el proveedor de autenticación.",
+        "provider-misconfigured": "Había un problema con el proveedor de autenticación. Contactar el administrador del sitio.",
+      },
     },
     "problem": {
       "401_2": "Dirección de correo electrónico y/o contraseña incorrecta."
--- a/src/components/account/login.vue
+++ b/src/components/account/login.vue
@@ -161,7 +161,7 @@ export default {
     },
     "oidc": {
       "error": {
-        "auth-ok-user-not-found": "Authentication was successful, but your user does not exist in ODK Central.",
+        "auth-ok-user-not-found": "Authentication successful, but your user does not exist in ODK Central.",
         "email-not-verified": "Your email address has not been verified by the authentication provider.",
         "provider-misconfigured": "There was a problem with your authentication provider.  Please contact the site administrator.",
       },
--- a/src/components/account/login.vue
+++ b/src/components/account/login.vue
@@ -216,7 +216,7 @@ export default {
         "auth-ok-user-not-found": "There is no Central account associated with your email address. Please ask your Central administrator to create an account for you to continue.",
         "email-not-verified": "Your email address has not been verified by your login server. Please contact your server administrator.",
         "email-claim-not-provided": "Central could not access the email address associated with your account. This could be because your server administrator has configured something incorrectly, or has not set an email address for your account. It could also be the result of privacy options that you can choose during the login process. If so, please try again and ensure that your email is shared.",
-        "internal-server-error": "There was an unhandled error on the server. Please try again later."
+        "internal-server-error": "Something went wrong during login. Please contact your server administrator."
       }
     },
     "problem": {
--- a/src/components/account/login.vue
+++ b/src/components/account/login.vue
@@ -199,13 +199,6 @@ export default {
       "alreadyLoggedIn": "Un usuario ya ha iniciado sesión. Actualice la página para continuar.",
       "changePassword": "Su contraseña tiene menos de 10 caracteres. Para proteger su cuenta, cambie su contraseña para que sea más larga."
     },
-    "oidc": {
-      "error": {
-        "auth-ok-user-not-found": "Autenticado, pero no tienes cuenta en ODK Central.",
-        "email-not-verified": "Tu email no ha estado verificado por el proveedor de autenticación.",
-        "provider-misconfigured": "Había un problema con el proveedor de autenticación. Contactar el administrador del sitio.",
-      },
-    },
     "problem": {
       "401_2": "Dirección de correo electrónico y/o contraseña incorrecta."
     }
--- a/src/components/account/login.vue
+++ b/src/components/account/login.vue
@@ -16,7 +16,16 @@ except according to the terms contained in the LICENSE file.
         <div class="panel-heading">
           <h1 class="panel-title">{{ $t('action.logIn') }}</h1>
         </div>
-        <div class="panel-body">
+        <div class="panel-body" v-if="config.oidcEnabled">
+          <div class="panel-footer">
+            <a href="/v1/oidc/login">
+              <button class="btn btn-primary">
+                {{ $t('action.logIn') }} (TODO add spinner when clicked)
+              </button>
+            </a>
+          </div>
+        </div>
+        <div class="panel-body" v-if="!config.oidcEnabled">
           <form @submit.prevent="submit">
             <form-group ref="email" v-model.trim="email" type="email"
               :placeholder="$t('field.email')" required autocomplete="off"/>
@@ -54,7 +63,7 @@ import { useRequestData } from '../../request-data';
 export default {
   name: 'AccountLogin',
   components: { FormGroup, Spinner },
-  inject: ['container', 'alert'],
+  inject: ['container', 'alert', 'config'],
   beforeRouteLeave() {
     return !this.disabled;
   },
--- a/src/components/account/login.vue
+++ b/src/components/account/login.vue
@@ -163,7 +163,7 @@ export default {
       "error": {
         "auth-ok-user-not-found": "Authentication successful, but your user does not exist in ODK Central.",
         "email-not-verified": "Your email address has not been verified by the authentication provider.",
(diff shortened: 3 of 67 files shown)


src/components/account/login.vue
```diff
@@ -16,7 +16,16 @@ except according to the terms contained in the LICENSE file.
         <div class="panel-heading">
           <h1 class="panel-title">{{ $t('action.logIn') }}</h1>
         </div>
-        <div class="panel-body">
+        <div class="panel-body" v-if="config.oidcEnabled">
+          <div class="panel-footer">
+            <a href="/v1/oidc/login">
+              <button class="btn btn-primary">
```
matthew-white, line 22: You can style a link like a button with this class:

```html
<a class="btn btn-primary" ...
```

src/components/user/edit.vue
```diff
@@ -20,7 +20,8 @@ except according to the terms contained in the LICENSE file.
         <div class="col-xs-7">
           <user-edit-basic-details v-if="user.dataExists"/>
         </div>
-        <div class="col-xs-5">
+        <pre>Config: {{config}}</pre>
+        <div class="col-xs-5" vi-if="!config.oidcEnabled"> <!-- REVIEW: this doesn't seem to work -->
```
matthew-white, line 24:
```diff
-vi-if
+v-if
```

matthew-white: Also, if there's a `v-if`, that's usually the first thing after the tag name.

Procfile
```diff
@@ -1,2 +1,3 @@
-vue: vue-cli-service build --mode development --watch
+#vue: vue-cli-service build --mode development --watch
+vue: vue-cli-service serve
```
matthew-white, line 2: Did `serve` instead of `build` end up working better for you? Have you had the chance to try #805?

src/components/user/edit.vue
```diff
@@ -51,7 +52,8 @@ const props = defineProps({
   id: {
     type: String,
     required: true
-  }
+  },
+  inject: ['config'],
```
matthew-white, line 56: Unlike Backend, I think ESLint still yells about dangling commas in Frontend. Feel free to turn off that rule!

src/components/user/list.vue
```diff
@@ -47,7 +47,7 @@ either is an Administrator or has no role. -->
     <user-new v-bind="newUser" @hide="hideModal('newUser')"
       @success="afterCreate"/>
     <user-reset-password v-bind="resetPassword" @hide="hideResetPassword"
-      @success="afterResetPassword"/>
+      @success="afterResetPassword" v-if="!config.oidcEnabled"/> <!-- REVIEW this may not be necessary; could help performance? -->
```
matthew-white, line 50: Agreed, not necessary, but I think it's a nice change. We could one step further and not even download this modal component, but that's slightly harder (though still not hard, I could show you how if it's of interest).

matthew-white (commented):
Just random thoughts for now!

matthew-white:
Should we also disable the /reset-password and /account/claim routes if `config.oidcEnabled`? If so, probably the easiest thing would be to add a `beforeEnter` guard on the route, similar to what we do for /system/analytics:

https://github.com/getodk/central-frontend/blob/3fc66c57c02bbd48eb471fb6beddc95db3e29dce/src/routes.js#L641-L660

main.nginx.conf
```diff
@@ -95,12 +97,15 @@ http {
     }
 
     location / {
-      root ./dist;
-
-      location /index.html {
-        include ./common-headers.nginx.conf;
-        add_header Cache-Control no-cache;
-      }
+      proxy_pass http://localhost:5000/;
```
sadiqkhoja, line 100: I had to change this to 5002 to run the app

src/components/user/edit.vue
```diff
@@ -51,8 +52,10 @@ const props = defineProps({
   id: {
     type: String,
     required: true
-  }
+  },
 });
+const config = inject('config');
+const hi = 3;
```
sadiqkhoja, line 58: probably dead code

src/router.js
```diff
@@ -71,7 +71,7 @@ router.afterEach(unlessFailure(to => {
       // Implements the restoreSession meta field. A test can skip this request
       // by setting the session before the initial navigation.
       async (to) => {
-        if (to.meta.restoreSession && !session.dataExists) {
+        if (/*to.meta.restoreSession &&*/ !session.dataExists) {
```
sadiqkhoja, line 74: I don't understand this line, is this change just for development purpose?

sadiqkhoja (commented):
Few observations after running this PR:

- When I created a new web user, I saw "User  Update Details" event in system logs with "{"data":{"password":true}}" details. What does this mean, do we need to log this in case of oidc is enabled?
- Frontend is respecting `next` query parameter. When an anonymous user go to a private link, frontend redirects to login page with route of private page in the `next` query parameter, after login I would expect user to be redirect to the original private page, that's not happening with OIDC enabled
- OIDC logged in user can change their email address from the frontend and after changing email address understandably they are unable to go in, I think we should disable change of email address.
- Can we do something about the Central login splash after successful login on IDP page?

alxndrsn:
@sadiqkhoja:

> * OIDC logged in user can change their email address from the frontend and after changing email address understandably they are unable to go in, I think we should disable change of email address.

The `PUT /v1/users/{id}/password` endpoint should be explicitly disabled in the backend by https://github.com/getodk/central-backend/pull/910/files#diff-4901d3a28e9b995b0b225b9c735a9020b33d902dd3ecd48a04b531bcd7d87a97R96, so I'm surprised it's working :thinking: 

> * Can we do something about the Central login splash after successful login on IDP page?

Can you clarify what this is referring to?

src/components/user/edit.vue
alxndrsn: PR at https://github.com/getodk/central-frontend/pull/826

main.nginx.conf
alxndrsn: Do you have other services running on ports 5000 & 5001?  I'll add a note that this change will not be included in final PR.

src/router.js
alxndrsn: I'll add a comment.

src/components/account/login.vue
```diff
@@ -16,7 +16,16 @@ except according to the terms contained in the LICENSE file.
...
+                {{ $t('action.logIn') }} (TODO add spinner when clicked)
+              </button>
+            </a>
+          </div>
+        </div>
+        <div class="panel-body" v-if="!config.oidcEnabled">
```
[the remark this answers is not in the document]
alxndrsn: Fixed.

src/components/account/login.vue
matthew-white, line 1: This component wasn't simple to begin with, and with the OIDC logic, it's getting more complex. I'm leaning toward moving the OIDC logic into its own component so that this component won't have to change. The router would then render this component or the OIDC component based on `config.oidcEnabled`. I don't think the existing `AccountLogin` component provides much that the OIDC component needs, so there's not much gained by adding OIDC to this component.

I may wait to make that change until we've finished discussing the [possibility](https://github.com/getodk/central-backend/pull/910#discussion_r1287474982) of adding a generic redirect route to Frontend. If we added such a route, that would make this component less complex, because the logic in `navigateToNext()` would get moved elsewhere.

matthew-white:
One thing I'm thinking about is the `sessionExpires` item in local storage. In the normal login flow, `AccountLogin` sets it by specifying `true` for `newSession` when calling `logIn()`.

One thing that's nice about `sessionExpires` is that `restoreSession()` can check it before sending its request. If the session is already expired, such that a request to /v1/sessions/restore would result in an error, then `restoreSession()` won't even send its request.

A more important thing that `sessionExpires` is used for is the prevention of multiple, concurrent sessions in the same browser. Normally, if the user opens the login page in two tabs, then logs in in one tab, they won't be able to log in in the other tab (which would create a second session). That's in large part due to how `sessionExpires` is set up. Does Backend prevent that possibility in the OIDC case? Specifically, I'm wondering whether a sequence like the following is currently possible:

- Open the login page in two tabs in the same browser.
- Log in in one tab.
- Log out in the IdP. This won't log out Central.
- Log in in the other tab as a different user.
- The first tab isn't logged out, because `sessionExpires` isn't reset.
- The first tab will use the cookie acquired by the second tab.

I'll think more about this. For now, I just wanted to mention that I'm thinking about scenarios like these and about whether there's a way for us to set `sessionExpires` in the OIDC case.

matthew-white:
> One thing I'm thinking about is the `sessionExpires` item in local storage.

I thought more about this, and I think we can set `sessionExpires` when the session is restored after OIDC login. That's the moment when Frontend first has access to the session created by OIDC login and the first time that Frontend knows the session expiration timestamp, so it's a natural time to set `sessionExpires`.

This approach takes reasonable measures to prevent multiple, concurrent sessions:

- Open the login page in two tabs. Log in in one tab. Now try to log in in the other tab (click the Continue button). You are not redirected to the IdP. Instead, an alert is shown like the one you see in the default login flow.
- Open the login page. Log in. Now log out in your IdP. Navigate to /v1/oidc/login directly. Log in there. After a moment, both tabs will end up logged out. (That's because the second login will set `sessionExpires` to a new value, causing the first tab to log out, which then causes the second tab to log out.)

One case that this approach doesn't handle is multiple, concurrent sessions with the exact same expiration timestamp. That's a case that is handled by the default login flow today. However, that's a very unlikely case, and I don't think we need to worry about it at this point.

src/components/account/login.vue
matthew-white: I'm having second thoughts about this: the two login flows have gotten more similar, because they now both check whether `sessionExpires` in local storage indicates that there is an existing session.

matthew-white: We're leaning toward not adding a generic redirect route to Frontend. Even so, I think I'll keep OIDC in the same component as the normal login flow.

matthew-white (commented):
I think this PR is now ready for review! I'm adding a few comments to facilitate review. @alxndrsn, I think we should start out with you taking a look at the changes I made. It's OK if you don't understand all the Vue mechanics, and you don't need to review tests. More I think it'd be helpful for you to take a look at things from the perspective of how you've set up Backend, thinking about the login flow at a higher level. Of course, please leave questions about anything you're curious about. Once you feel comfortable with the PR, I'll ask @sadiqkhoja to take a more detailed look at the whole PR.

src/components/user/edit/basic-details.vue
```diff
@@ -17,7 +17,10 @@ except according to the terms contained in the LICENSE file.
     <div class="panel-body">
       <form @submit.prevent="submit">
         <form-group v-model.trim="email" type="email"
-          :placeholder="$t('field.email')" required autocomplete="off"/>
+          :placeholder="$t('field.email')" required
+          :aria-disabled="emailDisabled"
+          :tooltip="emailDisabled ? $t('emailDisabled') : null"
+          autocomplete="off"/>
```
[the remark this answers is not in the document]
alxndrsn: Which part is surprising?

alxndrsn:
The frontend doesn't load for me on this branch:

Screenshot_2023-08-21_14-33-23

src/components/user/edit/basic-details.vue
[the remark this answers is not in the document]
matthew-white: It feels like disabling inputs is a common thing to be needed in general. `FormGroup` is maybe the most common way we use inputs, so I'm surprised we don't already have an example of disabling inputs with `FormGroup`. That's not a big deal, and I made a related change. But I wanted to note that this is the first example of this pattern.

matthew-white (commented):
@alxndrsn observed:

> pressing back from the oauth provider takes you back to the login screen, but with the button disabled

I see that behavior as well. I think it's due to the [back/forward cache](https://web.dev/bfcache/). I've pushed a commit that handles that case by listening for a `pageshow` event. I've tried it out locally, and the button is now enabled after navigation back from the IdP login page. 🎉

src/util/session.js
```diff
@@ -203,33 +203,25 @@ export const useSessions = () => {
...
   // is unlikely, and the worst case should be that the user sees 401 messages.
-  return session.request({ url: '/v1/sessions/restore', alert: false })
+  session.request({ url: '/v1/sessions/restore', alert: false })
     .catch(error => {
-      // The user's session may be removed without the user logging out, for
+      // The user's session may be deleted without the user logging out, for
       // example, if a backup is restored. In that case, the request will result
-      // in a 404. sessionExpires may need to be removed from local storage in
-      // order for the user to log in again.
-      if (sessionExpires != null) {
```
matthew-white, line 222: Removing this `if` means that we will always remove `sessionExpires` after a 404, not just if `sessionExpires` was set to a future timestamp before the request was sent. However, I don't think that's a problem:

- If somehow there's a login in another tab between when this tab sends the request and when it receives the 404 (due to some race condition), any open tab that is logged in will be logged out after the 404. That may be inconvenient, but that sequence of events should also be rare.
- Even if the tab that logged in is closed before the 404 is received, things should mostly be OK. In that case, `sessionExpires` will not be set even though there is a session cookie. However, Frontend will still be able to coordinate logout among tabs.
- In that case, there would be more of an issue in the tab that receives the 404, because `AccountLogin` would allow a login attempt to proceed. However, that safeguard is there just to prevent a confusing user experience. If SSO is not enabled, the user would see a cryptic error message from Backend. If SSO is enabled, and if another tab is opened after the 404 is received, the user will be immediately logged out after logging in.

Given that there weren't tests about this or comments in the code, I think the reason we checked `sessionExpires` before wasn't to prevent these possibilities, but rather to prevent unneeded `storage` events from being triggered (not that those are a big deal either).

I went back and forth on this one, which you can see in the commit history. But given our current setup, including the fact that `restoreSession()` does need to remove `sessionExpires` in some cases, I can't think of a way to prevent these possibilities entirely. And again, this sort of sequence should be unlikely. I suspect we'll come back to this area soon as part of getodk/central#400. For now, I think it's nice for `sessionExpires` to be used in fewer ways.

matthew-white (commented):
I observed the following while trying the SSO login flow on dev:

> I initially had difficulty logging in. I clicked Continue, selected my account, saw the redirect happen, but then ended back on the login page. I cleared cookies and local storage, then things started working.

&hellip;

<blockquote>

OK, I have an idea about how this sequence could happen:

- Log into a server that doesn’t use SSO.
- Close the tab and let your session expire (24 hours). You will no longer have a cookie, but the `sessionExpires` timestamp will still exist in local storage.
- Because `sessionExpires` indicates that the session is expired, Frontend won’t try to restore the session after the redirect back from IdP login.
- And actually, I think this will be a problem even on a server that uses SSO. If you let your session expire, then you won’t be able to log in again.
</blockquote>

I resolved this by having `restoreSession()` always send a request. Previously, it didn't send a request if it could infer from `sessionExpires` that the session was expired. I'm pretty sure I added that logic just as an optimization, but here it's ended up causing issues for SSO.

The optimization was helpful in some cases, but there were also several logout cases in which the request was still sent:

- If the user visited Frontend for the very first time.
- If the user clicked "Log out", then opened Frontend in a new tab (perhaps some time later).
- If Frontend automatically logged the user out before their session expired, then the user opened Frontend in a new tab.

Those are all cases in which `sessionExpires` is not set in local storage. There are times when it is important for the request to be sent even when `sessionExpires` isn't set, for example, after a logout error.

The optimization helped when `sessionExpires` was set, and its value was in the past. That would happen if the user's session expired, but they weren't logged out through Frontend.

I considered alternatives to the approach I took:

- _After the user clicks the Continue button in the `AccountLogin` component, have the component remove `sessionExpires` if its value is in the past._ I didn't opt for this approach because we don't currently export `removeSessionFromStorage()` from src/util/session.js, and I'm reluctant to add to that interface. This also felt like additional complexity in the component and in the login flow as a whole.
- _When Frontend determines whether to send the session restore request in `restoreSession()`, remove `sessionExpires` if its value is in the past (and don't send a request)._ This felt like one more piece of logic in an already complex flow: I'm happy to be removing an `if` block here. Ultimately, the optimization doesn't seem helpful enough to me to justify additional complexity.

With this change, `sessionExpires` is only used for two things:

- Coordinating logout among tabs
- Showing an alert on the login page and preventing login if a user is already logged in

We may change the latter as part of getodk/central#400, further simplifying our use of `sessionExpires`.

src/components/account/login.vue
```diff
@@ -69,10 +79,60 @@ export default {
...
+      /^[\w-]+$/.test(oidcError)) {
+      const path = `oidc.error.${oidcError}`;
+      if (this.$te(path, this.$i18n.fallbackLocale))
+        this.alert.danger(this.$t(path));
+    }
+  },
   mounted() {
-    this.$refs.email.focus();
+    if (this.config.oidcEnabled)
+      window.addEventListener('pageshow', this.reenableIfPersisted);
```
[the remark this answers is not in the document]
alxndrsn: Works for me :+1: checked on safari mobile, and firefox & chrome desktop.

src/components/account/login.vue
```diff
@@ -69,10 +79,60 @@ export default {
...
+      const { query } = this.$route;
+      const next = typeof query.next === 'string' ? query.next : null;
+      const qs = queryString({ next });
+      return `/v1/oidc/login${qs}`;
+    }
+  },
+  created() {
+    const { oidcError } = this.$route.query;
+    if (this.config.oidcEnabled && typeof oidcError === 'string' &&
+      /^[\w-]+$/.test(oidcError)) {
```
alxndrsn, line 93: I guess regex could be more precise if wanted, e.g. `/^[a-z-]+$`.

src/components/account/login.vue
```diff
@@ -146,6 +201,14 @@ export default {
       "alreadyLoggedIn": "A user is already logged in. Please refresh the page to continue.",
       "changePassword": "Your password is shorter than 10 characters. To protect your account, please change your password to make it longer."
     },
+    "oidc": {
+      "body": "Click Continue to proceed to the login page.",
+      "error": {
+        "auth-ok-user-not-found": "There is no Central account associated with your email address. Please ask your Central administrator to create an account for you to continue.",
+        "email-not-verified": "Your email address has not been verified by your login server. Please contact your server administrator.",
+        "provider-misconfigured": "Central could not access the email address associated with your account. This could be because your server administrator has configured something incorrectly, or has not set an email address for your account. It could also be the result of privacy options that you can choose during the login process. If so, please try again and ensure that your email is shared."
```
alxndrsn, line 209: This needs to change to `email-claim-not-provided` to account for possibility suggested in https://github.com/getodk/central-backend/pull/910/files#r1289487577

src/components/user/edit/password.vue
```diff
@@ -32,9 +33,7 @@ except according to the terms contained in the LICENSE file.
           {{ $t('action.change') }} <spinner :state="awaitingResponse"/>
         </button>
       </form>
-      <template v-else>
-        {{ $t('cannotChange') }}
-      </template>
+      <p v-else>{{ $t('cannotChange') }}</p>
```
alxndrsn, line 36: what's the significance of changing from `<template>` to `<p>`?

transifex/strings_en.json
```diff
@@ -924,6 +928,22 @@
...
+          "string": "Click Continue to proceed to the login page."
+        },
+        "error": {
+          "auth-ok-user-not-found": {
+            "string": "There is no Central account associated with your email address. Please ask your Central administrator to create an account for you to continue."
+          },
+          "email-not-verified": {
+            "string": "Your email address has not been verified by your login server. Please contact your server administrator."
+          },
+          "provider-misconfigured": {
```
alxndrsn, line 942: As above - this has changed name to `email-claim-not-provided`.

src/routes.js
```diff
@@ -637,26 +641,27 @@ const routes = [
...
-          },
-          title: () => [
-            i18n.t('systemHome.tab.analytics'),
-            i18n.t('systemHome.title')
-          ],
-          fullWidth: true
-        },
-        beforeEnter: () => (config.showsAnalytics ? true : '/')
-      })
+      ...conditionalRoutes(config.showsAnalytics, [
```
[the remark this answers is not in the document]
matthew-white: I think what's on `master` works OK for /system/analytics, but I wanted to take a different approach for /reset-password and /account/claim. I left some notes in the commit message for 7335412ca6fbf419e413ac6e7ad0a127b4c51ee2:

> Making this change because different disabled routes were behaving
slightly differently depending on their meta fields. That's because
global beforeEach guards are called before a beforeEnter guard in the
route config. For example, navigating to /reset-password sent a request
to restore the session, while navigating to /account/claim did not. Now
disabled routes are removed entirely, so all resolve to the NotFound
route.

After introducing `conditionalRoutes()` for /reset-password and /account/claim, I also used `conditionalRoutes()` for /system/analytics so that the same approach is used for all three routes.

test/components/user/reset-password.spec.js
```diff
@@ -21,6 +21,16 @@ describe('UserResetPassword', () => {
...
 
+    it('is not rendered if OIDC is enabled', async () => {
+      const component = await load('/users', {
+        container: {
+          config: { oidcEnabled: true },
+        },
+        root: false
+      });
+      component.find('.user-row .reset-password').exists().should.be.false();
+    });
```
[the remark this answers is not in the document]
matthew-white: I think the test above verifies that:

https://github.com/getodk/central-frontend/blob/f6aefcc71621a394943f91613234ba295e34cd86/test/components/user/reset-password.spec.js#L17-L22

If there isn't an element that matches the selector specified to `show`, the test will fail.

src/components/account/login.vue
matthew-white: Who knows, maybe we'll have an error message with a number one day. 🤣 In some ways, the primary check is the next one, that there is an i18n message associated with the value of `oidcError`. To complete that check, we call `$te()`, but there are some values that I think would cause issues with `$te()`, for example, `..`. So this regular expression is here to check, "is this value well-formed enough to pass to `$te()`?"

src/components/user/edit/password.vue
matthew-white: Not a ton. We generally use `<p>` in a `.panel-body`, because there are styles on `<p>`, for example, for `max-width`. I'm not sure those styles end up mattering here, but `<p>` certainly doesn't hurt. I used `<p>` for the OIDC case since I think it's a little better, then wanted to be consistent and changed `<template>` to `<p>` in this case as well.

sadiqkhoja (commented):
Few things I noticed on dev:

- After showing error message, should we clear the `oidcError` query parameter? Currently, when I refresh error page after dismissing the error, I see the error again which looks odd to me
- Email address is still editable for currently logged in user, though it's not changed at the backend. I think frontend should block this

sadiqkhoja:
image

Steps to reproduce:
- Login using id1@gmail.com
- Press back button
- Select another user i.e. id2@gmail.com

matthew-white:
> After showing error message, should we clear the `oidcError` query parameter? Currently, when I refresh error page after dismissing the error, I see the error again which looks odd to me

Sounds good, I'll make that change!

> Email address is still editable for currently logged in user, though it's not changed at the backend. I think frontend should block this

The release criteria say that admins should still be update email addresses in case someone's email address changes in the IdP (say, after a name change). However, a user who is not an admin should not be able to update their own email address (or anyone's for that matter). So I think we actually need to make a change to Backend here so that admins can continue to update email addresses.

matthew-white:
> image

Ooh that's an interesting one! That looks like a page that Backend is rendering: I don't think we have an error component like that in Frontend. If Backend can't handle this sequence, I think it'd be better for Backend to redirect to Frontend, specifying an `oidcError` query parameter.

Maybe there's a way for Frontend to stop the user from going back in this case? Is that something I should look into? (Or do you know a way to do that?)

src/routes.js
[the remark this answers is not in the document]
sadiqkhoja: I like this new approach, looks cleaner

src/components/account/login.vue
```diff
@@ -69,10 +79,63 @@ export default {
...
+  beforeUnmount() {
+    if (this.config.oidcEnabled)
+      window.removeEventListener('pageshow', this.reenableIfPersisted);
   },
   methods: {
+    verifyNewSession() {
+      const sessionExpires = localStore.getItem('sessionExpires');
+      const newSession = sessionExpires == null ||
+        Number.parseInt(sessionExpires, 10) < Date.now();
+      if (!newSession) this.alert.info(this.$t('alert.alreadyLoggedIn'));
```
sadiqkhoja, line 108: how about navigating to home page instead of show alert?

src/components/account/login.vue
```diff
@@ -69,10 +79,63 @@ export default {
...
+      if (!this.verifyNewSession()) {
+        event.preventDefault();
+        return;
+      }
+
+      this.disabled = true;
+    },
+    handleOIDCError() {
+      const { oidcError } = this.$route.query;
+      if (this.config.oidcEnabled && typeof oidcError === 'string' &&
```
sadiqkhoja, line 121: `typeof oidcError === 'string'` looks unnecessary to me

src/components/user/edit/basic-details.vue
```diff
@@ -46,12 +49,16 @@ defineOptions({
...
-const { user } = useRequestData();
+const { currentUser, user } = useRequestData();
 const { awaitingResponse } = user.toRefs();
 
 const email = ref(user.email);
 const displayName = ref(user.displayName);
 
+const config = inject('config');
+const emailDisabled = computed(() =>
+  config.oidcEnabled && !currentUser.can('user.update'));
```
sadiqkhoja, line 60: email should be disabled whenever `config.oidcEnabled` **or** when current user can't update user

test/components/account/login.spec.js
```diff
@@ -233,4 +243,175 @@ describe('AccountLogin', () => {
...
+    it('does not append ?next if it has no value', async () => {
+      const component = await load('/login?next', {
+        container: oidcContainer,
+        root: false
+      });
+      component.get('a').attributes('href').should.equal('/v1/oidc/login');
+    });
+  });
+
+  describe('OIDC error', () => {
```
sadiqkhoja, line 317: can we test these in a loop?

test/components/account/login.spec.js
```diff
@@ -233,4 +243,175 @@ describe('AccountLogin', () => {
...
+      const component = await load('/login?oidcError=email-not-verified', {
+        container: oidcContainer,
+        root: false
+      });
+      component.should.alert('danger', (message) => {
+        message.should.startWith('Your email address has not been verified by your login server.');
+      });
+    });
+
+    it('does not show an alert if there are two errors', async () => {
```
sadiqkhoja, line 348: That's not gonna happen the way backend works but I think we should give first error message the priority

test/components/account/login.spec.js
```diff
@@ -233,4 +243,175 @@ describe('AccountLogin', () => {
...
+
+    it('does not show an alert if query parameter has no value', async () => {
+      const component = await load('/login?oidcError', {
+        container: oidcContainer,
+        root: false
+      });
+      component.should.not.alert();
+    });
+
+    it('does not show an alert for an error whose name is invalid', async () => {
```
sadiqkhoja, line 364: how about showing generic error message?

sadiqkhoja (approved):
Besides minor comments about the functionality, code looks clean to me

src/components/account/login.vue
matthew-white: `this.$route.query.oidcError` can also be an array (if the query parameter is specified twice) or `null` (if it's specified without a value, i.e., `?oidcError`). We're about to insert the value into a template string before passing the result to `$te()`, so I just want to make sure that nothing breaks if something is wrong with the query parameter.

test/components/account/login.spec.js
matthew-white: If Backend doesn't work that way, I think we can just ignore it: it would be extra code to use either error message. This test is mostly here just to verify that nothing breaks if the query parameter is specified twice. I thought about skipping the test entirely — like you said, it'd be pretty strange for that to happen!

src/components/user/edit/basic-details.vue
matthew-white: I thought about it some more, but I really think it's "and", not "or"!

- If OIDC is not enabled, and the user is updating their own email on /account/edit, that is allowed. We don't need to check for `user.update` or disable the input.
- If OIDC is not enabled, and the user is updating someone's email on /users/:id/edit, then the `validateData` route meta field checks that the user can `user.update`. That means that if the user is able to see the input at all, we know they can use it, so it doesn't need to be separately disabled.
- If OIDC is enabled, and the user wants to update their own email on /account/edit, that should only be allowed if they are an admin (if they can `user.update`). I mentioned in a comment above that admins should still be able to update email addresses. In other words, the input should be disabled if OIDC is enabled and the user can't `user.update`.
- If OIDC is enabled, and the user wants to update someone's email on /users/:id/edit, the `validateData` meta field will again check for `user.update`.

test/components/account/login.spec.js
matthew-white: We don't actually really have a generic error message. We could add one, but it's something that would need to be translated. This test is mostly here just to verify that nothing breaks in this case. I don't expect that users will encounter this case in normal usage.

src/components/account/login.vue
matthew-white: We thought about doing that in the regular/non-OIDC case, but we thought it'd be confusing if the credentials entered didn't match the user who was already logged in, or if the credentials were incorrect (because they wouldn't be checked, yet it'd seem like they were). It's definitely a confusing user experience though, and I've filed getodk/central#400 to improve it.

That said, your comment is making me realize that we're able to do something different in the OIDC case. There are no credentials entered in that case, so nothing to be mismatched or incorrect, and I think what you describe would be helpful behavior.

sadiqkhoja: this is not a blocker, I feel this can happen in next release or later

matthew-white: Sounds good. I probably won't get to it for .4, but I've linked to this discussion from getodk/central#400. It'd be great to make that change at some point.

matthew-white:
@alxndrsn added a new error, `internal-server-error`. I reworded the message, but I'd be happy to revert it if others prefer.

> image

This now works differently, though I'm still not sure it's quite right. @sadiqkhoja, would you mind trying the new behavior on dev and seeing what you think? Here's what's happens now:

- Log in to Central using one account.
- Press back.
- Select a different account.
- Backend redirects to /#/login?oidcError=internal-server-error
- However, because there's still a session, the user is not brought to the login page: the session is restored, and the user is brought to the homepage. The `oidcError` is ignored, and the first account remains logged in.

I looked into it a little, and we could probably prevent the user from pressing the back button to return to the IdP login page. However, I'm not seeing that it's possible to restrict all navigation back, for example, if the user holds down on the back button and selects the IdP login page. If there's an approach we could take without restricting navigation, I think that'd probably be better.

matthew-white:
OK, I think I'm ready to merge! I've filed an issue about the sequence with pressing back, etc.: getodk/central#477. Let's continue related discussion there. If I've missed something else here, or if you think of something additional, please leave a comment on this PR or file an issue. Excited to get this merged!

PR #820, opened 2023-05-31
Simplify the Download route

matthew-white:


sadiqkhoja (approved):
(Y) for composition API

PR #827, opened 2023-07-12
lint: switch to using eslint directly

alxndrsn:
* this allows the use of the --cache flag for faster repeated linting (not supported in `vue-cli-service lint`) see: https://cli.vuejs.org/core-plugins/eslint.html#injected-commands
* `vue-cli-service` is deprecated, and using `eslint` directly is recommended for vue 3 see: https://cli.vuejs.org/core-plugins/eslint.html see: https://vuejs.org/guide/scaling-up/tooling.html#linting


alxndrsn:
For me this cuts repeated lint run duration from 25s to < 2s.

matthew-white:
> For me this cuts repeated lint run duration from ~25s to <2s.

I've also been noticing that linting has been pretty slow lately. Cutting down the duration like this would be a big win.

> `vue-cli-service` is deprecated

Hoping to move to Vite (#641) before too long!

> using `eslint` directly is recommended for vue 3 see: https://cli.vuejs.org/core-plugins/eslint.html see: https://vuejs.org/guide/scaling-up/tooling.html#linting

Are you sure this doesn't require Vite? Before listing recommendations, the second link says:

> when using a Vite-based build setup, our general recommendation is:

Can you tell whether .vue files are still being linted?

- Are templates being linted? For example, if you remove `<!-- eslint-disable vue/no-v-html -->` from src/components/collect-qr.vue, does ESLint display an error?
- Is the JS in `<script>` tags being linted?

matthew-white:
I'm in the process of moving from Vue CLI to Vite (#671). As mentioned above, under Vite, `eslint` is used directly, so we're about to be moving in the direction of this PR anyway. Given that linting continues to work with the changes here, I think we should try to get this PR merged. There's a conflicting file, but I can resolve that now.

package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "nf start",
     "build": "vue-cli-service build",
-    "lint": "vue-cli-service lint --no-fix --max-warnings 0 src/ bin/ test/ *.js",
+    "lint": "eslint --cache --no-fix --max-warnings 0 --ext .js,.vue src/ bin/ test/ *.js",
```
matthew-white, line 8: I don't see a `--no-fix` flag for `eslint`. I think it's an option from Vue CLI.

matthew-white, line 8: I see that Vite also specifies `.cjs,.mjs` for `--ext`. We don't have any of those files currently, but maybe we will in the future?

package-lock.json
```diff
@@ -12679,85 +12543,6 @@
       "funding": {
         "url": "https://github.com/sponsors/sindresorhus"
       }
-    },
-    "node_modules/yorkie": {
```
matthew-white, line 12683: Even though `yorkie` is a dependency of `@vue/cli-plugin-eslint` and not a top-level dependency, we also use it to ensure that bin/transifex/restructure.js is run:

https://github.com/getodk/central-frontend/blob/a54cf7bfd93c91cf6edd1126cddc2c7e365eac25/package.json#L16-L18

When it's installed, `yorkie` adds Git hooks locally. However, it doesn't seem to remove the hooks when it's uninstalled. When I try to make a commit after `yorkie` is uninstalled, I see:

```
node:internal/modules/cjs/loader:1147
  throw err;
  ^

Error: Cannot find module 'node_modules/yorkie/src/runner.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
    at Module._load (node:internal/modules/cjs/loader:985:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:135:12)
    at node:internal/main/run_main_module:28:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v20.10.0

pre-commit hook failed (add --no-verify to bypass)
```

I'm surprised that you were able to make your later commits on this branch. Maybe you had switched to a different branch and reinstalled `yorkie`?

Interestingly, I do see that there's an uninstall script in the `yorkie` [`package.json`](https://github.com/yyx990803/yorkie/blob/a4cf01d789da2633a33a888b496fa35395e72109/package.json#L12). However, it [looks like](https://docs.npmjs.com/cli/v10/using-npm/scripts#a-note-on-a-lack-of-npm-uninstall-scripts) uninstall scripts are no longer run automatically.

In any case, since we want to continue using `yorkie` at least for now, I think we should install it as a top-level dependency.

package.json
matthew-white: I thought about adding these, but I think we can wait until we have examples of them in the codebase. We haven't needed them yet.

vue.config.js
```diff
@@ -31,7 +31,6 @@ module.exports = {
       css: { url: false }
     }
   },
-  lintOnSave: false,
```
matthew-white, line 34: This looks to be an option of the ESLint plugin for Vue CLI, so I think we can remove it. From https://cli.vuejs.org/core-plugins/eslint.html#configuration:

> The following option is under the section of `vue.config.js`. It is respected only when `@vue/cli-plugin-eslint` is installed.

matthew-white (approved):
I think this PR is ready to merge, so I'm going to go ahead and approve and merge. I've made some changes myself, though nothing big. @alxndrsn, it'd be great if you could take a look at my changes once you have the chance just to confirm that they make sense.

package-lock.json
matthew-white: I've filed a related issue about possibly replacing `yorkie`: see #992.

alxndrsn: > I'm surprised that you were able to make your later commits on this branch. Maybe you had switched to a different branch and reinstalled `yorkie`?

It's likely I deleted the git hooks manually.

PR #829, opened 2023-07-20
enhances #448: show datasets tab always

sadiqkhoja:
Closes getodk/central#448 

Test

NA

None

No

### DIFF
--- a/src/components/dataset/table.vue
+++ b/src/components/dataset/table.vue
@@ -52,10 +52,10 @@ const { datasets } = useRequestData();
     text-align: right;
     padding-right: 10%;
   }
-}
-.empty-table-message {
-  margin-bottom: 20px;
-  margin-top: -20px;
+
+  &+ .empty-table-message {
+    margin-bottom: 20px;
+  }
 }
 </style>
 
--- a/src/components/dataset/table.vue
+++ b/src/components/dataset/table.vue
@@ -24,6 +24,9 @@ except according to the terms contained in the LICENSE file.
         :dataset="dataset"/>
     </tbody>
   </table>
+  <p v-if="datasets.dataExists && datasets.length === 0" class="empty-table-message">
+    {{ $t('emptyTable') }}
+  </p>
 </template>
 
 <script>
@@ -50,4 +53,16 @@ const { datasets } = useRequestData();
     padding-right: 10%;
   }
 }
+.empty-table-message {
+  margin-bottom: 20px;
+  margin-top: -20px;
+}
 </style>
+
+<i18n lang="json5">
+  {
+    "en": {
+      "emptyTable": "No Datasets have been created for this Project yet."
+    }
+  }
+</i18n>
--- a/src/components/project/show.vue
+++ b/src/components/project/show.vue
@@ -25,7 +25,7 @@ except according to the terms contained in the LICENSE file.
           role="presentation">
           <router-link :to="tabPath('datasets')">
             {{ $t('resource.datasets') }}
-            <span class="chip">New</span>
+            <span class="chip">{{ $t('common.new') }}</span>
           </router-link>
         </li>
         <li v-if="canRoute(tabPath('users'))" :class="tabClass('users')"
--- a/src/components/project/show.vue
+++ b/src/components/project/show.vue
@@ -25,6 +25,7 @@ except according to the terms contained in the LICENSE file.
           role="presentation">
           <router-link :to="tabPath('datasets')">
             {{ $t('resource.datasets') }}
+            <span class="chip">New</span>
           </router-link>
         </li>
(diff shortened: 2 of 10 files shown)


src/components/dataset/table.vue
```diff
@@ -50,4 +53,16 @@ const { datasets } = useRequestData();
     padding-right: 10%;
   }
 }
+.empty-table-message {
+  margin-bottom: 20px;
+  margin-top: -20px;
```
matthew-white, line 58: Does this mean that there won't be space between the column headers of the table and the `.empty-table-message`? We usually leave some space between the two.

src/components/project/show.vue
```diff
@@ -25,6 +25,7 @@ except according to the terms contained in the LICENSE file.
           role="presentation">
           <router-link :to="tabPath('datasets')">
             {{ $t('resource.datasets') }}
+            <span class="chip">New</span>
```
matthew-white, line 28: "New" needs an i18n message I think. I think it could go in en.json5 with a comment mentioning that it will be shown next to new features.

Also, based on the screenshot in getodk/central#450, I think this should be lowercase.

test/components/project/show.spec.js
```diff
@@ -166,19 +166,9 @@ describe('ProjectShow', () => {
         });
         const li = app.findAll('#page-head-tabs li');
         const text = li.map(wrapper => wrapper.get('a').text());
-        text.should.eql(['Overview', 'Datasets']);
+        text.should.eql(['Overview', 'Datasets New']);
         li[0].should.be.visible(true);
       });
-
-      it('does not show tabs if project does not have a dataset', async () => {
```
matthew-white, line 173: Could you also update the title of the previous test (removing "if the project has a dataset")?

matthew-white (approved):

src/components/dataset/table.vue
sadiqkhoja: I was following Forms list and it doesn't have space. Now I have added margin for Datasets list as well as Forms list

image

src/components/project/show.vue
sadiqkhoja: Aly requested to capitalize it

src/components/dataset/table.vue
matthew-white: Ah great, I'd been thinking that it'd be great to get space above the forms list message.

lognaturel:
Looks good! I have a couple nits on the chip and its placement that you may want to track somewhere. I think they're important enough to address in this release. CC @alyblenkin
Screen Shot 2023-07-21 at 8 55 35 AM
* The vertical alignment feels awkward. It feels to me like either the text baseline should be the same or the chip should be significantly higher.
* It doesn't feel linked to the Datasets tab. Maybe reducing the horizontal spacing would help?
* A note that may not require action but should at least be considered: "new" may be much longer in other languages.

This is very similar but feels subtly more polished to me:

(https://userpilot.com/blog/new-feature-onboarding/)

lognaturel:
Just randomly came across this from Ghost:
Screen Shot 2023-07-21 at 9 42 19 AM

I think the caps help! Also so weird that both of these examples I'm just seeing for the first time use magenta!

alyblenkin:
I love that we are prioritizing this! These are great examples. I've been collecting some [examples here](https://www.notion.so/New-feature-or-Beta-chip-all-tools-7febd0bc945f4bd3b4d307bb79da81bf?pvs=4) as well. I would like to do the [anchor message](https://www.figma.com/file/KFi9hIQdrRgo0rLqIC7Sg8/ODK-user-flows?type=design&node-id=658%3A40862&mode=design&t=I3glp9ydlnamBR5U-1aske we talked about with Collect, but I wasn't sure if that would make it in!  

I agree; the text baseline should be the same. All caps aren't great for accessibility, but we could put an exclamation mark to make it more exciting :) We tried the filled chip, but it didn't look as good. I didn't mock it up in figma; we just chatted about it, but here is a [quick sketches ⬇️  ](https://www.figma.com/file/KFi9hIQdrRgo0rLqIC7Sg8/ODK-user-flows?type=design&node-id=2221%3A50222&mode=design&t=I3glp9ydlnamBR5U-1). @sadiqkhoja let me know if you want to talk about tightening this up a bit!

Screenshot 2023-07-21 at 12 28 29 PM


P.S. I may need to get my eyes checked. I see pink with those examples! 👀

lognaturel:
Looking forward to small improvements to make it pop! I don't have strong feelings on filled vs not, font size, tab highlight, etc, just spacing and alignment.

> I see pink with those examples! 👀

Haha, ok, they're definitely not exactly the same color but in the same family, right??

alyblenkin:
I prefer option 4, but the leanest thing is likely option 1, and we tweak the radius of the chip and tighten the spacing.

PR #831, opened 2023-07-25
Update Vue to 3.3.4

matthew-white:
This PR updates Vue to the latest version. I had to wrangle `npm` a little to make the update happen, but I was able to do so: see the commit message for details.

In #830, I thought I had tested my change against the latest Vue. However, while I had run `npm update vue`, I don't think I confirmed that the update was successful, so I actually wasn't running against an updated Vue. When I did manage to update Vue, I saw an error that seems related to the change in #830. Here's part of the error:

```
WARN LOG: '[Vue warn]: Property "$container" was accessed during render but is not defined on instance.', '
', ' at <I18nT', 'tag="li"', 'keypath="introduction[1].full"', '>', '
', ' at <Modal', 'state=false', 'hideable=""', 'backdrop=""', ' ...', '>', '
', ' at <ProjectSubmissionOptions', 'state=false', 'onHide=fn', '>', '
', ' at <FieldKeyList', 'key=0', 'projectId="1"', 'onFetchProject=fn<bound fetchProject>', ' ...', '>', '
', ' at <AsyncRoute', 'componentName="FieldKeyList"', 'props=', Object{projectId: '1'}, 'loading="tab"', ' ...', '>', '
', ' at <RouterView', 'onFetchProject=fn<bound fetchProject>', 'onFetchForms=fn<bound fetchForms>', 'onFetchFieldKeys=fn<bound fetchFieldKeys>', '>', '
', ' at <PageBody>', '
', ' at <ProjectShow', 'key=2', 'projectId="1"', '>', '
', ' at <AsyncRoute', 'componentName="ProjectShow"', 'props=', Object{projectId: '1'}, 'loading="page"', ' ...', '>', '
', ' at <RouterView>', '
', ' at <App', 'ref="VTU_COMPONENT"', '>', '
', ' at <VTUROOT>'
...

  FieldKeyList
    ✗ toggles the "Submission Options" modal
```

I don't have a good explanation for why this test of `FieldKeyList` fails, but tests before `FieldKeyList` pass. Maybe because `FieldKeyList` renders `ProjectSubmissionOptions`, which accesses `$route` in its template? Could it be that `$container`, which is a Vue Test Utils mock, isn't being injected until later in the component's lifecycle compared to before?

Ultimately, I don't think it's very important that we understand this error. Instead of computing `$router` based on `$container`, we can copy the global property to each component. That way, components won't ever be accessing `$container`; only tests will. When I make that change, tests pass again.

### DIFF
--- a/test/util/lifecycle.js
+++ b/test/util/lifecycle.js
@@ -20,11 +20,11 @@ $route based on $router. */
 const shadowRouterProps = {
   created() {
     // eslint-disable-next-line no-self-assign
-    this.$router = this.$router;
+    if (this.$router != null) this.$router = this.$router;
   },
   computed: {
     $route() {
-      return this.$router.currentRoute.value;
+      return this.$router != null ? this.$router.currentRoute.value : undefined;
     }
   }
 };
--- a/test/util/lifecycle.js
+++ b/test/util/lifecycle.js
@@ -15,13 +15,14 @@ import createTestContainer from './container';
 from the vm property of the Vue Test Utils wrapper: see
 https://github.com/vuejs/test-utils/issues/2140. In practice, the only global
 properties that we try to access in testing are $router and $route. We can
-provide access to them by shadowing the global properties with computed
-properties from a mixin. */
+provide access to them by copying $router to each component, then computing
+$route based on $router. */
 const shadowRouterProps = {
+  created() {
+    // eslint-disable-next-line no-self-assign
+    this.$router = this.$router;
+  },
   computed: {
-    $router() {
-      return this.$container.router;
-    },
     $route() {
       return this.$router.currentRoute.value;
     }
--- a/package-lock.json
+++ b/package-lock.json
@@ -7673,9 +7673,9 @@
       }
     },
     "node_modules/magic-string": {
-      "version": "0.30.1",
-      "resolved": "https://registry.npmjs.org/magic-string/-/magic-string-0.30.1.tgz",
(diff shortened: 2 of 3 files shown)


test/util/lifecycle.js
```diff
@@ -15,15 +15,16 @@ import createTestContainer from './container';
...
 https://github.com/vuejs/test-utils/issues/2140. In practice, the only global
 properties that we try to access in testing are $router and $route. We can
-provide access to them by shadowing the global properties with computed
-properties from a mixin. */
+provide access to them by copying $router to each component, then computing
+$route based on $router. */
 const shadowRouterProps = {
+  created() {
+    // eslint-disable-next-line no-self-assign
+    if (this.$router != null) this.$router = this.$router;
```
matthew-white, line 23: I think this works because `$router` is a property on the prototype chain. By assigning it to `this`, we give `this` its own `$router` property. The `vm` property of the Vue Test Utils wrapper seems to provide access to an instance's own properties, but not global properties.

test/util/lifecycle.js
```diff
@@ -15,15 +15,16 @@ import createTestContainer from './container';
...
+  },
   computed: {
-    $router() {
-      return this.$container.router;
-    },
     $route() {
-      return this.$router.currentRoute.value;
+      return this.$router != null ? this.$router.currentRoute.value : undefined;
     }
```
matthew-white, line 28: We don't assign `$route` to `this` like we do with `$router` because the value of `$route` can change (whenever there's a navigation and the route changes). Unlike `$router`, `$route` as a global property seems to be a getter: https://github.com/vuejs/router/blob/a1611b6099403cfe9539ee8d4e9308b3a6a0175c/packages/router/src/router.ts#L1211-L1215

sadiqkhoja (approved):
looks good

PR #834, opened 2023-07-31
Remove "Role Settings" tab in UserHome

matthew-white:
Closes getodk/central#450.

I viewed the page locally. The "Role Settings" tab is gone, but everything seems to be the same. I also wrote a new test.

It's a pretty small PR, so not a lot of different options to consider. However, I'll leave a comment about a particular line.

No changes needed.

### DIFF
--- a/src/components/user/home.vue
+++ b/src/components/user/home.vue
@@ -15,13 +15,7 @@ except according to the terms contained in the LICENSE file.
       <template #title>{{ $t('title') }}</template>
       <template #tabs>
         <li :class="tabClass('')" role="presentation">
-          <router-link :to="tabPath('')">{{ $t('tab.users') }}</router-link>
-        </li>
-        <li class="disabled" role="presentation">
-          <a href="#">
-            {{ $t('tab.roles') }}
-            <span class="coming-soon">{{ $t('comingSoon') }}</span>
-          </a>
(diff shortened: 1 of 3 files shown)


src/components/user/home.vue
```diff
@@ -46,26 +40,16 @@ const { tabPath, tabClass } = useTabs('/users');
 </script>
 
 <style lang="scss">
-@import '../../assets/scss/variables';
-
-#user-home .coming-soon {
-  color: $color-text;
-  font-size: 10px;
+#user-home {
+  #page-head-tabs li { display: block; }
```
matthew-white, line 44: By default, `PageHome` hides the tab if it's the only one. Usually that's a useful thing, but here we do want to show the tab (see the discussion in the issue).

src/components/user/home.vue
```diff
@@ -46,26 +40,16 @@ const { tabPath, tabClass } = useTabs('/users');
...
 }
 </style>
 
 <i18n lang="json5">
 {
   "en": {
     // This is shown as the title at the top of the page.
-    "title": "User Settings",
-    "tab": {
-      "users": "Web Users",
```
matthew-white, line 63: This is a duplicate message, so I removed it.

ktuite (approved):

PR #836, opened 2023-07-31
Update usage reporting preview to be able to handle empty projects

ktuite:
Closes https://github.com/getodk/central-backend/issues/763 

The main issue with the usage reporting preview not working on a fresh install of Central was that the data wasn't what frontend expected:
1. projects could be empty, so it couldn't pick a first project to show
2. some fields, namely `num_questions_biggest_form` were `null` because with no projects or forms, that value wasn't calculated. I believe the real reporting of the metrics is fine with `null`, but frontend was trying to display that it as a number and running into an error. 

Trying it on a local version of central with a fresh database with no projects, just an admin user.

Small changes.

Allows users to see usage reporting preview even when there are no projects.
Screen Shot 2023-07-31 at 3 41 07 PM

No.

### DIFF
--- a/src/components/analytics/metrics-table.vue
+++ b/src/components/analytics/metrics-table.vue
@@ -21,7 +21,10 @@ except according to the terms contained in the LICENSE file.
     <tbody>
       <tr v-for="(value, name) in metrics" :key="name">
         <td>{{ $t(`fields.${name}`) }}</td>
-        <template v-if="value.recent != null">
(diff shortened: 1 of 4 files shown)


matthew-white (commented):
This is looking good to me! I want to think through one part of the modal a little more, but things are otherwise looking great.

matthew-white (approved):

PR #841, opened 2023-08-09
Replace "Dataset" with "Entity List" in frontend UI

lognaturel:
Part of getodk/central#466

Updates the UI strings to replace Dataset with Entity List, makes the introduction text short and high-level, matching the pattern on other pages.

I searched in `strings_en.json` to identify strings that needed updating. I tried to cluster changes by feature area so we could consider them separately if needed. This should be squashed upon merge.

I wanted to go through this myself because I thought there might be some strings to make decisions about but for the most part it turned into a 1:1 mapping between Dataset and Entity List. I wasn't sure about making Entity List a first-class concept with capitalization but I've come around to it.

<details>
<summary>Original writeup when this was just about the intro text</summary>
Now that the Datasets tab will always be shown, I want to make sure that the information provided there is useful even for people who aren't ready to try using the functionality yet. I think that no matter how much howto information we put in, users are going to have to reference longer docs so the focus of this message should be helping users understand what this functionality can do for them.

Goals:
- focus on what users want to accomplish so they can get excited to learn more
- reduce overall length of text

Opening as a draft so we can see this in context and decide on next steps. I'd like to get something merged sooner than later so translators can have some time.

I didn't want to start messing around with css. If we like this direction, I think that at the very least we should reduce the font size of the title and make the box a tighter fit around the text.

Screen Shot 2023-08-08 at 5 39 37 PM

CC @yanokwa 
</details>


alyblenkin:
This looks pretty good as a first iteration! I agree we should reduce the font size. And now that I see it in context, I wonder if it should go under the datasets header? Maybe we don't change this too much and think about using the pop-out or something in the next iteration. 

I also wonder if we should make the title the subtext instead? Maybe something like this: 

**About datasets** 
Sharing information between forms  

I also just noticed the "Help improve Central" is that part of the new designs issa did?

matthew-white:
> I also just noticed the "Help improve Central" is that part of the new designs issa did?

That's existing behavior on servers where a selection has not been made for [usage reporting](https://docs.getodk.org/central-usage-reporting/). I'll invite you to a server where you can see that behavior!

alyblenkin:
I played around with the copy. I know this might not be part of this iteration, but I think the short text will work well in the tooltip/ popout. That way we keep their workspace clean and you don't need to continue seeing it once you've read the information. 

New feature hover_link out

matthew-white:
> I wonder if it should go under the datasets header?

That's my instinct actually. I feel like there are a number of pages where there is text above a table (the Project Roles page, App Users, Form Access, etc).

Actually, do we even need a header here? Could we just put the text above the table? I feel like we often don't have a header like this. Right now, the header is the same text as the tab, so I'm not sure it adds much.

> I know this might not be part of this iteration, but I think the short text will work well in the tooltip/ popout.

Just wanted to note that a static popout like this is pretty cheap: we already have an existing popout component. (The component is limited in some ways, but it is up to the task here.) Again, I'm personally inclined toward text above the table, but I'm also happy to run with other options.

alyblenkin:
I agree; either above the table or in a popout. 

For the text above the table, I was thinking something like this. Does that align with what you were imagining @matthew-white? I think this could work. It's just harder to read long sentences.  
About datasets

matthew-white:
Yeah, that's about what I was thinking! We don't have an info box with white background like that in other places, but I see how that makes it possible to have a &times; button at the top right corner of the box.

> It's just harder to read long sentences.

I think the text would wrap a little earlier, so wouldn't be quite so long, because text is set up to wrap at about 600px. For example, the Project Roles page and the App Users page both have big blocks of text that wrap. Happy to decrease the wrap width from 600px if that helps. Also happy to run with a popout, just wanted to throw out some thoughts!

matthew-white:
During the meeting today, @lognaturel said that she too thinks a popout work well, so I'll plan on implementing that.

lognaturel:
The goal here is to make the landing experience friendlier for people who were just seeing our entities concept for the first time. We think that will be most Central users. As I've worked on copy with @alyblenkin and @yanokwa, we're realized that the challenges go deeper than the explanation blurb. @matthew-white, we will likely hold off on the popout for now. I'm going to take a first pass at a deeper change that eliminates "dataset" entirely so that we can further discuss.

src/components/dataset/overview/dataset-properties.vue
```diff
@@ -95,7 +95,7 @@ export default {
 <i18n lang="json5">
   {
     "en": {
-      "emptyTable": "There are no Properties in this Dataset."
+      "emptyTable": "The Entities in this Entity List do not have any user-defined Properties."
```
matthew-white, line 98: Should we capitalize "properties"? I had been thinking that we wouldn't, because we generally don't capitalize "fields", which is a similar concept for forms. I guess we were capitalizing here already. But I see other places where we aren't capitalizing:

- `propertiesCount` in en.json5
- `FormDraftPublish` component ("a new property")
- `FormDraftTesting` ("the desired properties")

No preference here around whether we capitalize or not. Just wanted to note that we don't for fields and that there are a few places where we don't for properties.

src/components/dataset/overview/linked-forms.vue
```diff
@@ -85,7 +85,7 @@ export default {
 {
   "en": {
     // Number of form(s) is shown separately above this text
-    "formsConsumeData": "Form consumes data from this Dataset | Forms consume data from this Dataset",
+    "formsConsumeData": "Form consumes Entities from this Entity List | Forms consume Entities from this Entity List",
```
matthew-white, line 88: Maybe even just "Forms consume this Entity List"?

src/components/dataset/summary/row.vue
```diff
@@ -149,9 +149,9 @@ export default {
 <i18n lang="json5">
 {
   "en": {
-    // This is shown when a dataset is new
+    // This is shown when an Entity List is new
     "new": "new!",
```
matthew-white, line 153: If the noun has changed, does "new!" need to be retranslated? You could rename the key (e.g., to `new2` or `newEntityList`) to force retranslation.

src/locales/en.json5
```diff
@@ -235,6 +235,7 @@
     "instanceId": "Instance ID",
     "lastSubmission": "Latest Submission",
     "name": "Name",
+    "listName": "List name",
```
matthew-white, line 238: We generally capitalize column headers, so I think this should be "List Name".

src/components/dataset/overview/dataset-properties.vue
matthew-white, line 98: Just an idea, what do you think about "custom" properties instead of "user-defined"?

matthew-white (approved):
Tests updated and everything! ✨ This is looking great.

I made a few small changes in commits that I just pushed. It'd be awesome if you could take a quick look at those changes just to make sure they're reviewed. I didn't modify any text in my commits. I'm leaving a few questions/comments about text, which you should feel free to take or leave. Please feel free to merge once you think it's ready!

src/components/dataset/overview/dataset-properties.vue
lognaturel: Let's not! I went back and forth on this one.

src/components/dataset/summary/row.vue
lognaturel: I think we should take our chances. None of the languages I know would require a change. Maybe German? I'm going to send out a message to translators anyway so I can highlight that.

src/components/dataset/overview/dataset-properties.vue
lognaturel: I went with "user-defined" to match terminology we've used at https://getodk.github.io/xforms-spec/entities and elsewhere. I feel unease about custom but I'm not sure why. user-defined doesn't actually provide more info, does it? How about we leave it for now and see whether how we use the "user-defined properties" concept and whether we want to rename it across the board.

src/components/dataset/overview/linked-forms.vue
lognaturel: Ooh, how about "Forms use this Entity List"? I think we were doing "consume" to try to mention "Entity" as well as "Dataset". But we probably don't need to do that anymore.

PR #843, opened 2023-08-16
Enable concurrent update for Entities

sadiqkhoja:
image

Automated tests and manually testes

NA

None

NA

### DIFF
--- a/src/components/entity/list.vue
+++ b/src/components/entity/list.vue
@@ -124,7 +124,7 @@ export default {
         data[name] = odataEntity[odataName];
       this.update.entity = {
         uuid: odataEntity.__id,
-        currentVersion: { label: odataEntity.label, data }
+        currentVersion: { label: odataEntity.label, version: odataEntity.__system.version, data }
       };
       this.showModal('update');
     },
--- a/src/components/entity/list.vue
+++ b/src/components/entity/list.vue
@@ -145,6 +145,7 @@ export default {
         label: updatedEntity.currentVersion.label,
         __system: {
           ...oldOData.__system,
+          version: updatedEntity.currentVersion.version,
           updates: oldOData.__system.updates + 1,
           updatedAt: updatedEntity.updatedAt
         }
--- a/src/components/entity/update.vue
+++ b/src/components/entity/update.vue
@@ -57,6 +57,7 @@ except according to the terms contained in the LICENSE file.
 
 <script setup>
 import { computed, nextTick, ref, watch } from 'vue';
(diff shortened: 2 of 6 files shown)


src/components/entity/list.vue
matthew-white, line 1: I think we need to change `afterUpdate()` as well so that the OData is updated with the correct version after the update. Right now, I think it wouldn't be possible to update the same entity twice in a row.

matthew-white (commented):
Just a one-off thought, happy to also take a closer look if @ktuite prefers.

matthew-white:
There's a parallel change to Backend (getodk/central-backend#832), but the Backend issue is more about the API and database and won't need testing by the QA team. I think I'll slot this PR in the project separately, then add the "needs testing" label to just this PR. Then the QA team can focus on verifying the changes here.

src/components/entity/list.vue
ktuite: tried this (with sadiq's recent commit that makes this fix) and it works to update an entity multiple times

ktuite (approved):
My only nitpick (and I'm not sure what to suggest instead) is with the styling around the conflict alert: there's a little gap above the top red border, and the X to close the alert doesn't have a useful function there's nothing to "retry" there.

Screen Shot 2023-08-25 at 4 13 04 PM

Here's the same alert in a different modal with more padding.

Screen Shot 2023-08-25 at 4 17 24 PM

sadiqkhoja:
@getodk/testers this feature can be tested on staging

## How to test

- Open entities data page on two browser tabs / windows
- Modify a row on one tab
- Try to modify same row on another tab, you should see error message

srujner:
@sadiqkhoja 
I found one thing that might be an issue. 

Scenario:
User A opens Edit from Data tab first and starts editing the Entity.
User B opens Edit from Data tab second and is unable to edit the Entity.

Now as User B I'm able to close the modal -> click on More in the "Actions" tab -> Click on Edit next to Entity Data -> Edit any information in the Entity and Update it while User A is still editing the same Entity.

After that User A is unable to finish editing the Entity.

Maybe it'll be better to block editing from both tabs? Because right now it's easy to take over.

lognaturel:
@srujner, I think your comment is somewhat related to https://github.com/getodk/central/issues/476. Ideally clicking on "Edit" would immediately check the entity version. It wouldn't help with a case in which two people/tabs already had the edit modal open but it would help in the much more common case where two people have the entity list or entity detail page open at the same time.

srujner:
**Tested with Success!**

PR #846, opened 2023-08-24
Enhances c#466: changes datasets routes to entity-list

sadiqkhoja:
Part of getodk/central#466

Updated tests and verified manually

NA

Existing regression test suite should be sufficient

NA


matthew-white:
If a "Dataset" is now an "Entity List", I feel like &hellip;/datasets in the route should become &hellip;/entity-lists (plural).

I'm also noticing that there are now conflicts to resolve (maybe with #841).

matthew-white (approved):

PR #850, opened 2023-09-01
Change rendering of entity.create audit event in submission feed

ktuite:
Frontend part of https://github.com/getodk/central-backend/pull/959 for issue https://github.com/getodk/central-backend/issues/822

`label` was removed from the audit details for `entity.create` events but the backend PR populates the label from the entity's current version's label under `event.details.entity.currentVersion.label`. 

If the entity has been deleted or purged, only `uuid` and `dataset` are available for that entity (the basic information stored in the audit log details) so the `uuid` is used instead and no link is rendered.

Screen Shot 2023-09-01 at 1 59 14 PM

Tests, checking it directly.

Won't show links to deleted entities, will make it easier to create entity deletion tools in the future.

No.

### DIFF
--- a/src/components/submission/feed-entry.vue
+++ b/src/components/submission/feed-entry.vue
@@ -35,9 +35,12 @@ except according to the terms contained in the LICENSE file.
         <span class="icon-magic-wand entity-icon"></span>
         <i18n-t keypath="title.entity.create">
           <template #label>
-            <router-link :to="entityPath(projectId, entityDataset(entry), entityUuid(entry))">
+            <router-link v-if="entityLabel(entry) != null" :to="entityPath(projectId, entityDataset(entry), entityUuid(entry))">
               {{ entityLabel(entry) }}
             </router-link>
+            <template v-else>
+              <span class="entity-label">{{ entityUuid(entry) }}</span>
+            </template>
           </template>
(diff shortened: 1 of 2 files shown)


src/components/submission/feed-entry.vue
```diff
@@ -35,9 +35,12 @@ except according to the terms contained in the LICENSE file.
         <span class="icon-magic-wand entity-icon"></span>
         <i18n-t keypath="title.entity.create">
           <template #label>
-            <router-link :to="entityPath(projectId, entityDataset(entry), entityUuid(entry))">
+            <router-link v-if="entityLabel(entry) != null" :to="entityPath(projectId, entityDataset(entry), entityUuid(entry))">
```
sadiqkhoja, line 38: I think we can use optional chaining here and entityLabel function is not required at all:

`entry.details?.entity?. currentVersion?. label`

src/components/submission/feed-entry.vue
```diff
@@ -157,11 +160,9 @@ export default {
   },
   methods: {
```
sadiqkhoja, line 161: if you are up for it then you can get rid of all these function by using optional chaining

sadiqkhoja (approved):

PR #851, opened 2023-09-01
Feature #741: Use opaque cursor on submission list and entity list

### LINKED ISSUE #741: Add infinite scroll to Entities table
For v2023.2, we are just loading the entire block of entities to show in the entity table. For the following release, when we likely add bulk CSV managing of entities, we will want to add the same kind of chunking and infinite scroll that we have for submissions. 

We will likely want to reuse functionality from the submission chunking/scrolling and make a generic infinite scroll odata table. 

Note about infinite scroll vs. paging: one use case we see in practice is loading a bunch of submissions and then searching on the page itself for specific content. This is a pretty good case for sticking with infinite scroll until we have the ability to search submission/entity content.

sadiqkhoja:
Closes #741

Integration tests and manual verification

Used the same approach as Submissions List

NA

No

### DIFF
--- a/src/components/entity/list.vue
+++ b/src/components/entity/list.vue
@@ -14,7 +14,7 @@ except according to the terms contained in the LICENSE file.
     <div id="entity-list-actions">
       <button id="entity-list-refresh-button" type="button"
             class="btn btn-default" :aria-disabled="refreshing"
-            @click="fetchData(false)">
+            @click="fetchChunk(true, true)">
             <span class="icon-refresh"></span>{{ $t('action.refresh') }}
             <spinner :state="refreshing"/>
           </button>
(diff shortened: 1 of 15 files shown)


matthew-white (commented):
This is a great simplification! I love that we no longer need to track instance IDs or have challenges related to scrolling in chunks while there's concurrent creation or deletion. The code here definitely looks like it's on the right track. I've left some suggestions, but I think @ktuite could do the final review without me if that ends up working best. Excited for `$skiptoken` and the changes here!

src/components/entity/list.vue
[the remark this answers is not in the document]
sadiqkhoja: I like the idea but thinking to do this in a separate PR later

src/components/entity/list.vue
```diff
@@ -67,13 +73,25 @@ export default {
     datasetName: {
       type: String,
       required: true
+    },
+    // Returns the value of the $top query parameter.
+    top: {
+      type: Function,
+      default: (loaded) => (loaded < 1000 ? 10 : 1000)
```
sadiqkhoja, line 80: 250

src/components/odata-loading-message.vue
```diff
@@ -0,0 +1,137 @@
...
+    if (props.filter)
+      return t(`${props.type}.filtered.withoutCount`);
+
+    if (props.totalCount === 0)
+      return t(`${props.type}.withoutCount`);
+
+    if (props.totalCount <= props.top)
+      return tcn(`${props.type}.all`, props.totalCount);
+
+    // console.log( t('submission.first', { count: 'asdf', top: 10 }, 11 ) );
```
sadiqkhoja, line 67: delete this

src/util/i18n.js
```diff
@@ -43,3 +43,8 @@ export const loadLocale = ({ i18n, logger }, locale) => {
 export function $tcn(path, count, values = undefined) {
   return this.$tc(path, count, { count: this.$n(count, 'default'), ...values });
 }
+
+// same as above but for composition API
+export function tcn(path, count, values) {
```
sadiqkhoja, line 48: double check if I really need this function

ktuite (approved):
Looked at this after a synchronous review with Sadiq earlier and I think it's good!

PR #855, opened 2023-09-12
Remove link to entity list on summary if new and unpublished

ktuite:
Frontend part of https://github.com/getodk/central/issues/467 to de-linkify the name of a new entity list.

Screen Shot 2023-09-12 at 2 59 11 PM

The corresponding backend change https://github.com/getodk/central-backend/pull/974 makes it so if the user gets to the page that used to be in this link, it will say "resource not found" w/o any changes because backend now returns a 404 for getting details about an unpublished dataset.

Trying it, tests.

I looked at the details coming back from the dataset-diff endpoint for a form draft, and `isNew` on the dataset itself seemed like a good one to use to decided to create a link or not.

Less access to broken parts of frontend, more consistent behavior.

n/a

### DIFF
--- a/src/components/dataset/summary/row.vue
+++ b/src/components/dataset/summary/row.vue
@@ -14,9 +14,12 @@ except according to the terms contained in the LICENSE file.
     <div class="row">
       <div class="col-xs-6 dataset-name-wrap">
         <div class="dataset-name text-overflow-ellipsis" v-tooltip.text>
-          <router-link :to="datasetPath(projectId, dataset.name)" v-tooltip.text>
+          <router-link v-if="!dataset.isNew" :to="datasetPath(projectId, dataset.name)" v-tooltip.text>
             {{ dataset.name }}
           </router-link>
+          <template v-else>
+            {{ dataset.name }}
+          </template>
         </div>
         <div v-if="dataset.isNew" class="dataset-new">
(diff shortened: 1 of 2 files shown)


src/components/dataset/summary/row.vue
```diff
@@ -14,9 +14,12 @@ except according to the terms contained in the LICENSE file.
     <div class="row">
       <div class="col-xs-6 dataset-name-wrap">
         <div class="dataset-name text-overflow-ellipsis" v-tooltip.text>
-          <router-link :to="datasetPath(projectId, dataset.name)" v-tooltip.text>
+          <router-link v-if="!dataset.isNew" :to="datasetPath(projectId, dataset.name)" v-tooltip.text>
             {{ dataset.name }}
           </router-link>
+          <template v-else>
```
matthew-white, line 20: Instead of `<template>`, how about `<span v-tooltip.text>` in case the dataset name is very long?

matthew-white (approved):

src/components/dataset/summary/row.vue
ktuite: the outer div `.dataset-name` seems to take care of this already without changes!

Screen Shot 2023-09-12 at 5 37 11 PM

PR #856, opened 2023-09-12
Do not render OdataDataAccess if SSO is enabled

matthew-white:
With this change, `OdataDataAccess` will not be rendered if SSO is enabled. That component contains the "Analyze via OData" button, as well as a link to the API docs. If SSO is enabled, then OData and the API will no longer be accessible to users. Hopefully that will change in the near future, at which time we can render this component again. Two pages are affected: the Submissions page and the Entities Data page.

I wrote a new test. I also checked locally that the pages look OK if `OdataDataAccess` isn't rendered. I didn't verify this change with the full SSO flow, but I also think that the risk is low.

I added a `v-if` to `OdataDataAccess`. I first thought about adding the `v-if` inside `FormSubmissions` and `DatasetEntities`. But since this is behavior shared by the two pages, it feels better to put the `v-if` in `OdataDataAccess`.

I forgot where exactly this component was rendered and was a little worried that I would need to adjust styles. But it's rendered in a `PageSection` heading, not in `SubmissionList`, so I don't think any styles need to change.

### DIFF
--- a/src/components/dataset/entities.vue
+++ b/src/components/dataset/entities.vue
@@ -11,7 +11,7 @@ except according to the terms contained in the LICENSE file.
 -->
 
 <template>
-    <div>
+  <div>
     <page-section>
       <template #heading>
         <span>{{ $t('resource.entities') }}</span>
--- a/src/components/odata/analyze.vue
+++ b/src/components/odata/analyze.vue
@@ -10,7 +10,7 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <modal id="odata-analyze" :state="state" hideable backdrop
+  <modal v-if="!config.oidcEnabled" id="odata-analyze" :state="state" hideable backdrop
     @hide="$emit('hide')">
     <template #title>{{ $t('title') }}</template>
     <template #body>
@@ -93,6 +93,7 @@ import Selectable from '../selectable.vue';
 export default {
   name: 'OdataAnalyze',
   components: { Modal, Selectable },
+  inject: ['config'],
   props: {
     state: Boolean,
     odataUrl: String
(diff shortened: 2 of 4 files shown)


src/components/odata/analyze.vue
```diff
@@ -10,7 +10,7 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <modal id="odata-analyze" :state="state" hideable backdrop
+  <modal v-if="!config.oidcEnabled" id="odata-analyze" :state="state" hideable backdrop
```
matthew-white, line 13: This `v-if` is a nice-to-have. With the change to `OdataDataAccess`, the `state` prop of this component will never be toggled to `true`, so this modal will never be shown. What this `v-if` does is also remove this unnecessary content from the DOM if SSO is enabled. We've done something similar in another case for SSO, with the `UserResetPassword` modal.

Relatedly, I didn't add a test for this change: I test the change to `OdataDataAccess`, but not that the content here isn't rendered. I don't think such a test is really needed, because this change is more of an optimization and isn't user-facing. What's most important to test is that the button and link aren't shown.

src/components/dataset/entities.vue
```diff
@@ -11,7 +11,7 @@ except according to the terms contained in the LICENSE file.
 -->
 
 <template>
-    <div>
```
matthew-white, line 14: Just noticed this while I was looking at this component! The `<div>` was indented a little too much.

src/components/odata/data-access.vue
```diff
@@ -25,15 +25,19 @@ except according to the terms contained in the LICENSE file.
...
-<script>
-export default {
-  name: 'OdataDataAccess',
-  props: {
-    analyzeDisabled: Boolean,
-    analyzeDisabledMessage: String
-  },
-  emits: ['analyze']
-};
+<script setup>
```
matthew-white, line 28: This component was very simple, so I also took the opportunity to convert it to a Composition API component.

ktuite (approved):

PR #857, opened 2023-09-13
Add npm script: dev:oidc

alxndrsn:
This script adds the environment variable required for running the project for OIDC development.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "nf start",
+    "dev:oidc": "VUE_APP_OIDC_ENABLED=true npm run dev",
     "build": "vue-cli-service build",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "nf start",
+    "dev:oidc": "VUE_APP_OIDC_ENABLED=true npm run dev",
```
matthew-white, line 7: Does this also need to start nginx? As-is, I feel like this would only work if you've already run `npm run dev`. Would it work to run `nf start` here?

matthew-white: Ooh never mind, it calls `npm run dev`, which starts nginx. 🤦

matthew-white (approved):

PR #859, opened 2023-09-14
Document exact Transifex CLI command

matthew-white:


matthew-white:
We also saw that an error is shown if a language is downloaded that's in Transifex but not Frontend. However, if `--force` is used and not `--all`, I think `tx pull` won't pull such languages.

ktuite (approved):

Issue #823, opened 2023-06-30
Update translations for v2023.4


matthew-white:

matthew-white:
I'm going to move this to "in progress" since we'll need to remember to pull translations again, but we've done everything we need to for now. :+1:

PR #860, opened 2023-09-15
update translations for v2023.4

sadiqkhoja:
Progress toward [#823](https://github.com/getodk/central-frontend/issues/823). We'll pull translations again right before release.


sadiqkhoja:
I don't know why `locale/ja.json` is not updated?

matthew-white:
I don't think there's been much translation of Japanese lately, which would explain why there aren't new translations in ja.json. A few messages have been changed or removed in en.json5 as part of .4, and if they were previously translated, you'd expect to see them removed in ja.json. But it could be that they just weren't translated to begin with.

matthew-white (approved):

PR #862, opened 2023-09-15
fixes c#480: set required engine version to major level (18)

sadiqkhoja:
Closes https://github.com/getodk/central/issues/480

Reverted back engine requirement to major version i.e. 18 instead of patch level

Verified on test env

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "vue-cli-plugin-i18n": "~2"
   },
   "engines": {
-    "node": "18.17.0",
+    "node": "18",
     "npm": "9"
   }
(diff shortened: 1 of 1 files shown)


lognaturel:
Any reason not to match infrastructure which uses major and minor (no patch)? The reasoning is that patches are less likely to introduce accidental regressions and more likely to target things we really care about like security fixes.

And I think with this change we'd still end up with warnings (or maybe errors?) because at some point this will require 18.18 but we'll still be using an 18.17 image.

matthew-white (approved):

PR #873, opened 2023-10-13
Change entity error message in submission feed

ktuite:
Frontend piece of https://github.com/getodk/central/issues/520

"Problem creating Entity" --> "Problem processing Entity" 

In the context of the submission feed, problems processing submissions about entities now look like this, whether they are failing to create or update an entity.

Screen Shot 2023-10-16 at 4 29 59 PM

Tests and trying it on frontend.

Needed generic error message.

Makes update entity errors clearer. 

I don't think so.

### DIFF
--- a/src/components/submission/feed-entry.vue
+++ b/src/components/submission/feed-entry.vue
@@ -49,7 +49,25 @@ except according to the terms contained in the LICENSE file.
           </template>
         </i18n-t>
       </template>
-      <template v-else-if="entry.action === 'entity.create.error'">
+      <template v-else-if="entry.action === 'entity.update.version'">
+        <span class="icon-magic-wand entity-icon"></span>
+        <i18n-t keypath="title.entity.update">
+          <template #label>
+            <router-link v-if="entry.details?.entity?.currentVersion?.label != null" :to="entityPath(projectId, entry.details?.entity?.dataset, entry.details?.entity?.uuid)">
+              {{ entry.details?.entity?.currentVersion?.label }}
+            </router-link>
+            <template v-else>
+              <span class="entity-label">{{ entry.details?.entity?.uuid }}</span>
+            </template>
+          </template>
+          <template #dataset>
+            <router-link :to="datasetPath(projectId, entry.details?.entity?.dataset)">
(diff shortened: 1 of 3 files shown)


src/components/submission/feed-entry.vue
```diff
@@ -49,7 +49,25 @@ except according to the terms contained in the LICENSE file.
           </template>
         </i18n-t>
       </template>
-      <template v-else-if="entry.action === 'entity.create.error'">
+      <template v-else-if="entry.action === 'entity.update.version'">
+        <span class="icon-magic-wand entity-icon"></span>
+        <i18n-t keypath="title.entity.update">
+          <template #label>
+            <router-link v-if="entry.details?.entity?.currentVersion?.label != null" :to="entityPath(projectId, entry.details?.entity?.dataset, entry.details?.entity?.uuid)">
```
ktuite, line 56: get rid of `?.` or are they for deleted entities? look into it more for deleted entities specifically.

test/components/submission/feed-entry.spec.js
```diff
@@ -216,28 +216,28 @@ describe('SubmissionFeedEntry', () => {
       });
     });
 
-    describe('entity.create.error audit', () => {
+    describe('entity.error audit', () => {
```
ktuite, line 219: Add separate tests for the entity update (success) case.

matthew-white (approved):

ktuite (commented):
interactive review

src/components/submission/feed-entry.vue
ktuite: I think we can stop worrying about (most) of these `?.`s... they're from the central versions 2023.1-2023.3 that we stored the entity label in the audit details and used it for rendering but couldn't guarantee it was there because entities created with 2022.3 didn't save this field. 

In 2023.4, we removed 'label' from the entity audit details again (to go along with entity deletion). This introduced a different way of calculating the entity label and changed frontend to look for it in a different place or fall back to the uuid.

The only important check now is whether or not `details.entity.currentVersion` exists, which has a label if it does exist.

PR #874, opened 2023-10-16
Add conflict filter to EntityList

### LINKED ISSUE #525: Include filter selections in URLs
Moved from the block 14 release criteria:

> - If I am viewing the Submissions table or the Server Audit Log, and I make one or more filter selections
>   - then my filter selections should appear in the URL
>   - and if I load that URL in a different tab, I should see the same filter selections applied

I took a first look at this, and mostly it seemed quite doable. We will very likely want to reuse components after the URL changes, but I don't think any changes will be needed to `asyncRoute()` or `createRoutes()`. Often when a component is reused after a route change, it needs to watch `$route`, but I'm not sure that `SubmissionList` will need to do even that.

So far, the one part that does seem more complicated is what to do if the URL includes an invalid filter selection. That could happen if the user copies and pastes a URL incorrectly. Once we support purging submissions, it could also happen if all the submissions for a particular submitter are purged (such that the submitter is no longer one of the filter's options). It seems potentially fine for invalid filter selections to be included in the Backend request, but I want to make sure that Frontend components won't break (specifically those in `SubmissionFilters`).

Another thing I have been wondering about is whether to include the list of selected fields in the URL. I wasn't planning to for block 14, but I could see that being useful behavior in some cases. On the other hand, the list of selected fields could be quite long. On the other other hand, maybe we could gzip and encode it or something?

matthew-white:
This PR adds a conflict filter to `EntityList`, completing the Frontend portion of getodk/central#503. The hardest part was connecting the filter to a query parameter. That's something that we've done in `SubmissionList` and have wanted to do in `AuditList` (#525), so I added a generic composable for syncing a ref with one or more query parameters. I wanted to verify that the composable worked in the `AuditList` case as well, so this PR also closes #525.

`SubmissionList` now uses the new composable, and related tests continue to pass. I've added new tests of entity components and of the composable. I also tried a couple of things locally.

The main question was how to connect the conflict filter to a query parameter. The filter logic in `SubmissionList` was a little complicated, and my hope was that a composable would simplify component code and reduce the barriers to future dynamic query parameters.

Relatedly, the data flow in `SubmissionList` felt a little suspect. `SubmissionList` defined a settable computed property for each filter. Here's an example of how things worked before:

- Navigate to .../submissions?start=2023-10-15&end=2023-10-16&reviewState=rejected&reviewState=hasIssues.
  - The `submissionDateRange` computed property is an array of `DateTime` objects (call this array "date range 1").
  - The `reviewStates` computed property is an array of review states.
- Select a different date range ("date range 2"). This will trigger a navigation, changing the `start` and `end` query parameters and the `submissionDateRange` computed property.
- `submissionDateRange` has changed. However, it is _not_ date range 2, but instead a different array with equivalent (but not equal) `DateTime` objects. That different date range is then passed back to the date range filter, which reacts to it.
- Furthermore, even though it was the date range filter that changed, `reviewStates` is also changed. It is a different array with the same values as before. That different array is passed to the review state filter, which reacts to it.

This worked in terms of user-facing behavior, but data ends up changing in unexpected ways, so components react at unexpected times. The data flow is not very clear.

The new composable makes the data flow clearer, avoiding settable computed properties. `SubmissionList` now creates a ref for each filter. Each ref is synced with one or more query parameters. If a query parameter changes, then the value of the ref is set, to which filter components react. Conversely, if a selection is changed in a filter component, then the value of the ref is set, updating query parameters. Even though that triggers a navigation, the value of the ref is not set a second time. In the terms of the example above, once date range 2 is selected, `submissionDateRange` is actually set to date range 2, and it remains date range 2 after the navigation. Unrelated refs also do not change after the navigation (`reviewStates` in the example above). The ref is synced with one or more query parameters, and something happens only if the value of the ref is changed, or one of the query parameters is changed to a truly different value (not just a technically different array).

That approach makes things simpler, but there is a downside for the submitter filter. The previous `submitterIds` computed property used the `submitters` resource, such that once the response for `submitters` was received, `submitterIds` would recompute. That's no longer the case. With the composable, each ref is associated with a function that converts query parameters to a value, but that function is not wrapped in a reactive effect. Even if the function references reactive data, it does not rerun when that data changes. It only reruns with one of the query parameters changes. As a result, we now need to watch the `submitters` resource, setting the `submitterIds` ref once the response is received. I don't love the watcher, but I don't see a way to wrap the conversion function in a reactive effect without introducing cases where the value of the ref is changed to a technically different, but equivalent value. I think there's also conceptual clarity that comes with limiting the ways that the value of the ref can be changed automatically. At the end of the day, it's also only the submitter filter that's like this: other filters have static options.

The composable itself involves some complexity, but I think I've been able to keep that complexity encapsulated in the composable. I think component code and the data flow is simpler.

The likeliest source of regression is the filters on the submissions page, but tests are passing.


test/components/entity/filters.spec.js
```diff
@@ -0,0 +1,165 @@
...
+    const app = await load('/projects/1/entity-lists/trees/entities?conflict=true');
+    const text = app.get('#entity-list .empty-table-message').text();
+    text.should.equal('There are no matching Entities.');
+  });
+
+  it('does not update dataset.entities after filtering', () => {
+    testData.extendedDatasets.createPast(1, { entities: 2 });
+    testData.extendedEntities
+      .createPast(1, { conflict: true })
+      .createPast(1, { conflict: false });
```
matthew-white, line 148: Reverse?

test/data/entities.js
```diff
@@ -23,9 +23,10 @@ export const extendedEntities = dataStore({
 
     uuid = faker.random.uuid(),
     label = faker.random.word(),
-    updates = 0,
-    updatedAt = null,
     version = 1,
+    updates = version - 1,
+    conflict = false,
```
matthew-white, line 28: `null` instead of `false`

test/components/entity/filters.spec.js
matthew-white, line 148: Replace booleans.

test/composables/query-ref.spec.js
```diff
@@ -0,0 +1,165 @@
...
+    result.value.should.equal('c,b');
+    await app.vm.$router.push('/?x=c&y=d');
+    result.value.should.equal('c,d');
+  });
+
+  describe('change to another query parameter', () => {
+    it('does not change the value of the ref', async () => {
+      const app = await load('/?x=a&z=b');
+      const fromQuery = ({ x }) => ({ someProp: x });
+      const toQuery = ({ someProp }) => someProp;
```
matthew-white, line 51: `=> ({ x: someProp })`

test/composables/query-ref.spec.js
```diff
@@ -0,0 +1,165 @@
...
+      const obj = result.value;
+      obj.should.eql({ someProp: 'a' });
+      await app.vm.$router.push('/?x=a&z=c');
+      result.value.should.equal(obj);
+    });
+
+    it('does not change value of ref even if fromQuery() uses an array parameter', async () => {
+      const app = await load('/?x=a&x=b&z=c');
+      const fromQuery = ({ x }) => ({ first: x[0], second: x[1] });
+      const toQuery = ({ first, second }) => [first, second];
```
matthew-white, line 64: `=> ({ x: [first, second] })`

test/composables/query-ref.spec.js
```diff
@@ -0,0 +1,165 @@
...
+      const app = await load('/?x=a');
+      const result = withSetup(() => useQueryRef(options.exclamation), {
+        container: app.vm.$container
+      });
+      const fake = sinon.fake();
+      // Using a beforeEach hook rather than comparing app.vm.$route before and
+      // after result.value is set because here, even if a navigation is
+      // triggered, it won't be confirmed. That's because it would be a
+      // navigation to a duplicate route.
+      app.vm.$router.beforeEach(fake);
```
matthew-white, line 145: Does the navigation guard need to return a boolean?

matthew-white (commented):

ktuite (approved):

matthew-white:
I made a number of small changes after code review, but not anything that I think requires additional review.

- One small but important change: the ref that the new composable returns is now shallow. A shallow ref is all that's needed, and I think a shallow ref is actually necessary to support the data flow laid out in the PR description. Otherwise, in cases where the value of the ref is an object, the value would be a reactive proxy of the value it was set to. I want the two values to be exactly the same.
- An important edge case: if the value of the ref changes, but that doesn't change query parameters for some reason, then a navigation should not be triggered. That comes up for the `submitterIds` ref, since it is set initially and then once again once the response from &hellip;/submitters is received.
- I added more tests, including unit tests of the `useAudit()` composable.
- More comments and other small changes

PR #875, opened 2023-10-18
Added conflict count on home page

sadiqkhoja:
Closes getodk/central#500

Tests and manual verification

Followed existing pattern

None

No

### DIFF
--- a/src/components/project/home-block.vue
+++ b/src/components/project/home-block.vue
@@ -43,7 +43,7 @@ except according to the terms contained in the LICENSE file.
       <project-dataset-row v-for="(dataset, index) of visibleDataset" :key="dataset.name" :dataset="dataset" :project="project" :show-icon="index === 0"/>
       <tr v-if="showDatasetExpander" class="project-dataset-row transparent-bg">
         <td class="col-icon"></td>
-        <td colspan="6" class="expand-button-container">
+        <td colspan="2" class="expand-button-container">
           <a href="#" role="button" class="expand-button" @click.prevent="toggleDatasetExpanded">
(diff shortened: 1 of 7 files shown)


src/components/project/home-block.vue
```diff
@@ -230,12 +246,41 @@ export default {
...
 
-    .project-dataset-row .col-icon{
-      border-right-color: #b9005c;
+    .project-dataset-row {
+      .conflicts-count {
+        text-align: right;
+        padding-right: 8px;
+
+        [class*='icon'] {
+          margin-left: 4px;
```
matthew-white, line 255: This looks to be 5px for other icons in the table.

src/components/project/home-block.vue
```diff
@@ -230,12 +246,41 @@ export default {
       }
     }
 
-    .project-dataset-row .col-icon{
-      border-right-color: #b9005c;
+    .project-dataset-row {
+      .conflicts-count {
+        text-align: right;
+        padding-right: 8px;
```
matthew-white, line 252: I think we could use `$padding-right-table-data` here. (Though isn't that the default?)

test/components/project/home-block.spec.js
matthew-white, line 1: How about adding a test that the conflict count can change depending on how the entity lists are sorted? I feel like it'd be nice to verify that the `sortFunc` prop is being used in `hiddenConflicts`.

matthew-white (commented):
Just smaller things, it's looking good!

matthew-white (approved):

PR #876, opened 2023-10-19
Enable liveReload for development

sadiqkhoja:
Accept this only if it works for everyone and their workflows

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
-vue: vue-cli-service build --mode development --watch
+vue: vue-cli-service serve --port 8080
 nginx: nginx -c "$PWD/main.nginx.conf" -p "$PWD"
--- a/main.nginx.conf
+++ b/main.nginx.conf
@@ -95,7 +95,8 @@ http {
     }
 
     location / {
-      root ./dist;
+      proxy_pass http://localhost:8080;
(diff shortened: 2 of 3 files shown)


matthew-white:
What do you think about including this as an additional command rather than replacing the current behavior? I'm thinking that `npm run dev` could continue to build and watch like it does today, then something like `npm run dev:live` could run a dev server and use live reload. Just for myself, my personal preference isn't to use live reload. One thing that I like to do sometimes during local development is to compare two versions side-by-side, to check for differences in styles or behavior. I don't think that'd be possible with live reload. That said, I can also see what's nice about live reload, and I wouldn't want to block others from using it.

One thing that I'm unsure about is how we would make this an option from the nginx side. Is it possible to add some sort of switch in the nginx config? Otherwise, could we use something like `envsubst` to pass in either `root ./dist;` or `proxy_pass` to the config?

I'm making progress on moving Frontend from Vue CLI to Vite (#671). I think things in this area will work a little differently under Vite, so this PR will need to change. I'm thinking that we should close this PR, then open a new one once I'm done with Vite. What do you think?

sadiqkhoja:
Agree, we can take this up again once Vue CLI is replaced with vite.

PR #885, opened 2023-10-31
Do not modify entity test data in place

matthew-white:
There are issues involving how entity test data is updated. @sadiqkhoja noticed that things break after `testData.extendedEntities.update()` is called, because the `entity` reference that the `entityVersions` store holds becomes out-of-date. That's because `update()` doesn't update in-place, but rather replaces the object in the `testData` store. Given that, it doesn't work for `entityVersions` to store an `entity` reference, so I'm having it store the entity UUID instead. It can use the UUID to compare entities, which is one way it used the `entity` reference before.

Beyond that, I had the thought that what `update()` does is indeed what we want in general. If a test entity is already in use in Frontend code that we're testing, then we can't let `testData` modify it in-place, as that would affect the data in Frontend. By replacing the object in the `testData` store rather than modifying it in-place, we create more of a separation between `testData` and the Frontend code we're testing. There were a few places where `testData.extendedEntities` and `testData.extendedEntityVersions` would modify in-place rather than replacing, and I've exchanged those for `update()`.

Since `update()` expects an index, making that change required various kinds of iteration over `testData` stores. I decided it was a good time to do something I've wanted to for a while, which is to make `testData` stores iterable. `testData` stores are supposed to be array-like in many ways, but they lack many array methods. I added a few methods related to iterating, finding, and filtering. Note that these are all based on creation order, which is how most of the methods work. For example, `get()`, `last()`, and `splice()` all use creation-order indices. To get an array sorted in the way that Backend will return it, `sorted()` is still the best way. In other words, `filter()` may return objects in a different order than `sorted().filter()`.

These new methods will work for both `testData` stores and `testData` views. I tried to make them performant by iterating over one object at a time rather than creating a new array of objects. I had in mind some tests that create many `testData` objects.

### DIFF
--- a/test/data/entities.js
+++ b/test/data/entities.js
@@ -36,20 +36,31 @@ const entityVersions = dataStore({
     conflictingProperties = undefined,
     creator = extendedUsers.first(),
 
-    // Internal option for the `entities` store
+    // Internal option for the `entities` store. This is an entity that is in
+    // the process of being created.
     _entity = undefined
   }) => {
     if (entities.size === 0 && _entity == null)
       throw new Error('there is no entity for which to create a new version');
-    const entity = _entity ?? (uuid != null
-      ? entities.sorted().find(e => e.uuid === uuid)
-      : entities.last());
-    if (entity == null) throw new Error('entity not found');
+    const entityIndex = _entity != null
+      ? null
+      : (uuid != null
+        ? entities.findIndex(entity => entity.uuid === uuid)
+        : entities.size - 1);
+    if (entityIndex === -1) throw new Error('entity not found');
+    const entity = _entity ?? entities.get(entityIndex);
 
-    const versions = entityVersions.sorted()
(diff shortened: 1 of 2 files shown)


test/data/entities.js
```diff
@@ -269,9 +279,9 @@ extendedEntities.resolve = (index) => {
...
 
-  for (const version of entityVersions.sorted()) {
-    if (version.entity === entity && version.conflict != null)
-      version.resolved = true;
+  for (const [i, version] of entityVersions.entries()) {
+    if (version.uuid === entity.uuid && version.conflict != null)
+      entityVersions.update(i, { resolved: true });
   }
 
   // Following should be replaced with entity.update(index) once entityVersions is linked via UUID
```
sadiqkhoja, line 287: now we can use `entities.update(index)` here, right?

sadiqkhoja (approved):

PR #889, opened 2023-11-03
Add ability to link to (scroll to) specific entity versions in feed

matthew-white:
This PR makes it possible to link to a specific element of a page using the route hash. You can link to the element from the same page or a different page. This functionality is needed for getodk/central#529.

Here's an example of the new functionality on the entity detail page. As the hash changes, the page scrolls down to a middle entry in the feed, then up to the top entry, then back down to the bottom entry. A blue border is shown around the associated entry. In the case of the bottom entry, the entry does not appear at the top of the screen because the entry is located at the very bottom of the page. However, you can still see the blue border. In other cases, there are 10px between the top of the screen and the associated entry.

https://github.com/getodk/central-frontend/assets/5970131/0eef6f30-2a95-4df3-abbe-db8d05f48810

I tried things locally and also wrote new tests. There's one thing that I wasn't able to test, which I'll make a note of.

I was able to use the scroll behavior functionality provided by Vue Router: https://router.vuejs.org/guide/advanced/scroll-behavior.html. I'll leave comments about a few choices I made.

One change I made was to modify the structure of the feed, grouping related entries in a different way from before. However, I think that if there are issues there, we'll spot them during regression testing. Other than that, there aren't a lot of changes to existing code.

### DIFF
--- a/src/components/feed-entry.vue
+++ b/src/components/feed-entry.vue
@@ -36,9 +36,10 @@ defineProps({
 <style lang="scss">
 @import '../assets/scss/mixins';
 
+$margin-bottom: 20px;
 .feed-entry {
   box-shadow: 0 7px 18px rgba(0, 0, 0, 0.05);
-  margin-bottom: 20px;
+  margin-bottom: $margin-bottom;
 }
 
 .feed-entry-heading, .feed-entry-body .markdown-view { padding: 10px 15px; }
@@ -82,5 +83,14 @@ defineProps({
   background-color: #f9f9f9;
 
   .markdown-view > p:last-child { margin: 0 0 0px; }
+}
+
+// Container of multiple FeedEntry components
+.feed-entry-group {
+  // Move the bottom margin to the container, but keep 1px between FeedEntry
+  // components.
+  margin-bottom: $margin-bottom;
+  .feed-entry { margin-bottom: 0; }
+  .feed-entry + .feed-entry { margin-top: 1px; }
 }
 </style>
--- a/src/util/router.js
+++ b/src/util/router.js
@@ -79,6 +79,49 @@ export const forceReplace = ({ router, unsavedChanges }, location) => {
   return router.replace(location);
 };
 
+// Returns a scrollBehavior function to pass to createRouter().
+export const createScrollBehavior = (animation = 'smooth') => {
+  // Function to cancel a previous attempt to scroll
+  let cancel;
+  const scrollTo = (scrollId) => {
+    const el = document.querySelector(`[data-scroll-id="${scrollId}"]`);
+    return el != null ? { el, top: 10, behavior: animation } : null;
+  };
+  return (to, from, savedPosition) => {
+    if (cancel != null) cancel();
+    if (to.hash === '') return to.path === from.path ? savedPosition : null;
+    const scrollId = to.hash.replace('#', '');
+    // Certain characters would lead to an invalid CSS selector above, so we
+    // limit the set of accepted characters.
+    if (!/^[\w-]+$/.test(scrollId)) return null;
+    // Try to scroll right away. If the element doesn't exist, then we need to
+    // wait for it. In that case, we will attempt to scroll on an interval,
+    // giving up after 3 minutes or so.
+    return scrollTo(scrollId) ?? new Promise(resolve => {
+      let tries = 0;
+      const intervalId = setInterval(
+        () => {
+          const result = scrollTo(scrollId);
+          if (result != null) {
+            clearInterval(intervalId);
+            cancel = null;
+            resolve(result);
+          } else {
+            tries += 1;
+            if (tries === 720) cancel();
+          }
+        },
+        250
+      );
+      cancel = () => {
+        clearInterval(intervalId);
+        cancel = null;
+        resolve(null);
+      };
+    });
+  };
+};
+
 
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -87,7 +130,8 @@ export const forceReplace = ({ router, unsavedChanges }, location) => {
 // Returns the resources that would be preserved after navigating from the
 // `from` route to the `to` route.
 export const preservedData = (to, from, requestData) => {
-  if (from === START_LOCATION) return requestData.resources;
+  if (from === START_LOCATION || to.path === from.path)
+    return requestData.resources;
   const results = [];
   for (const f of to.meta.preserveData) {
     const result = f(to, from);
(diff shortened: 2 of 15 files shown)


test/components/entity/activity.spec.js
```diff
@@ -64,11 +97,41 @@ describe('EntityActivity', () => {
...
+      component.findAll('.feed-entry-group').length.should.equal(1);
+      const entries = component.findAllComponents(EntityFeedEntry);
       entries.length.should.equal(2);
       entries[0].props().entry.action.should.equal('entity.create');
       entries[1].props().entry.action.should.equal('submission.create');
       entries[1].props().submission.instanceId.should.equal('some-uuid');
     });
   });
+
+  it('passes the correct entityVersion prop', () => {
```
matthew-white, line 110: This is mostly a test of existing behavior that I thought would be useful to add. We're using `entityVersion` in more ways (now in `scrollData()`), so I wanted to verify that it's set correctly.

test/components/entity/basic-details.spec.js
```diff
@@ -152,7 +152,7 @@ describe('EntityBasicDetails', () => {
           return testData.standardEntities.last();
         })
         .respondWithData(() => testData.extendedAudits.sorted())
-        .respondWithData(() => [])
+        .respondWithData(() => testData.extendedEntityVersions.sorted())
```
matthew-white, line 155: This should have been changed in #880. For #880, I searched for the word "diff", but since this was just an empty array, I didn't notice it. I caught it now because it was causing a problem: if the entity versions aren't returned, `scrollData()` isn't able to access `entityVersion` and throws an error.

test/components/entity/show.spec.js
```diff
@@ -90,7 +90,7 @@ describe('EntityShow', () => {
           return testData.standardEntities.last();
         })
         .respondWithData(() => testData.extendedAudits.sorted())
-        .respondWithData(() => []);
+        .respondWithData(() => testData.extendedEntityVersions.sorted());
```
matthew-white, line 93: Same as above: this was diffs before and should have been changed as part of #880.

src/util/router.js
```diff
@@ -87,7 +130,8 @@ export const forceReplace = ({ router, unsavedChanges }, location) => {
 // Returns the resources that would be preserved after navigating from the
 // `from` route to the `to` route.
 export const preservedData = (to, from, requestData) => {
-  if (from === START_LOCATION) return requestData.resources;
+  if (from === START_LOCATION || to.path === from.path)
```
matthew-white, line 133: This change is tested in test/router.spec.js. It wasn't needed before despite our use of filter query parameters because we preserve data after navigation between tabs — even if the navigation is from the same tab to itself.

src/util/router.js
```diff
@@ -79,6 +79,49 @@ export const forceReplace = ({ router, unsavedChanges }, location) => {
...
+export const createScrollBehavior = (animation = 'smooth') => {
+  // Function to cancel a previous attempt to scroll
+  let cancel;
+  const scrollTo = (scrollId) => {
+    const el = document.querySelector(`[data-scroll-id="${scrollId}"]`);
+    return el != null ? { el, top: 10, behavior: animation } : null;
+  };
+  return (to, from, savedPosition) => {
+    if (cancel != null) cancel();
+    if (to.hash === '') return to.path === from.path ? savedPosition : null;
```
matthew-white, line 92: I wasn't able to test `savedPosition` because we use memory history mode in history, and `savedPosition` is based on `popstate` navigation. However, I verified locally that `savedPosition` is used.

src/components/feed-entry.vue
```diff
@@ -83,4 +84,13 @@ defineProps({
 
   .markdown-view > p:last-child { margin: 0 0 0px; }
 }
+
+// Container of multiple FeedEntry components
+.feed-entry-group {
```
matthew-white, line 89: I'm not sure that the `FeedEntry` component is exactly the right place for these styles. Maybe one day `.feed-entry-group` becomes its own component.

src/util/router.js
```diff
@@ -79,6 +79,49 @@ export const forceReplace = ({ router, unsavedChanges }, location) => {
...
+    const scrollId = to.hash.replace('#', '');
+    // Certain characters would lead to an invalid CSS selector above, so we
+    // limit the set of accepted characters.
+    if (!/^[\w-]+$/.test(scrollId)) return null;
+    // Try to scroll right away. If the element doesn't exist, then we need to
+    // wait for it. In that case, we will attempt to scroll on an interval,
+    // giving up after 3 minutes or so.
+    return scrollTo(scrollId) ?? new Promise(resolve => {
+      let tries = 0;
+      const intervalId = setInterval(
```
sadiqkhoja, line 102: can we do something to avoid `interval`?

An idea: raise a custom event on the `<body>` when Entity feed is `mounted` and listen to that event here.

src/util/router.js
sadiqkhoja, line 92: Shouldn't we return `savedPosition` even when `to.path` and `from.path` are not equal?

Did you try browser back/forward buttons, are they working correctly?

matthew-white: > Did you try browser back/forward buttons, are they working correctly?

Yes, they're working! It's fun to see. `savedPosition` also seems to be working.

> Shouldn't we return `savedPosition` even when `to.path` and `from.path` are not equal?

My thinking was that if you navigate to a different page, it could be that `savedPosition` no longer exists on the page, if data has been cleared and needs to be reloaded. Or even worse, it could be that `savedPosition` exists, but it's a different element and basically just a random place on the page. It could be a different element while data is loading, or even after, depending on how the data has changed. Maybe this isn't super likely though and is outweighed by the convenience of being able to return to `savedPosition` when it does still exist. What do you think?

src/util/router.js
matthew-white: Good idea, let me see if I can avoid it somehow. That may simplify testing as well.

src/util/router.js
sadiqkhoja: If I understood following correctly then browser should remember scroll position on back/forward buttons:

> The third argument, savedPosition, is only available if this is a popstate navigation (triggered by the browser's back/forward buttons).

I checked out this branch locally, and here is the result. Back button keeps the scroll position from the current page (it should be scroll position of previous page - somewhere in the middle) and forward button is not remembering my scroll position (it should be at the bottom).

https://github.com/getodk/central-frontend/assets/447837/af690a17-8e02-4f03-a475-478345469523

matthew-white: Hmm interesting! Maybe I only tried the back button with `savedPosition`. I'd have to look further into it to figure out what's driving the difference. I'd be happy to file an issue about that, but it's probably not something I'd prioritize until we finish the rest of the entity stories.

ktuite (approved):

PR #891, opened 2023-11-17
Entity version tag on update events in entity feed

ktuite:
Closes https://github.com/getodk/central/issues/533

Adds version tags to update events in the entity feed: 

Screenshot 2023-11-16 at 4 45 36 PM

Also made the tags into links that go to `#v<version>` the anchor tag for that version. Not sure if this is useful or distracting. 

Screenshot 2023-11-16 at 4 46 25 PM

I didn't add the version to the first entity create event. I could, I'm just not sure it flows as well after "Created Entity <label> in <list name> Entity List" .

tests and trying it

### DIFF
--- a/src/components/entity/feed-entry.vue
+++ b/src/components/entity/feed-entry.vue
@@ -62,6 +62,7 @@ except according to the terms contained in the LICENSE file.
       </template>
       <template v-else-if="entry.action === 'entity.update.version'">
         <span class="icon-pencil"></span>
+        <span class="title">
         <template v-if="entry.details.submissionCreate != null">
           <i18n-t v-if="submission != null"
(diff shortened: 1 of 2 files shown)


src/components/entity/feed-entry.vue
```diff
@@ -62,6 +62,7 @@ except according to the terms contained in the LICENSE file.
       </template>
       <template v-else-if="entry.action === 'entity.update.version'">
         <span class="icon-pencil"></span>
+        <span class="title">
```
ktuite, line 65: To separate the event title from the version, which was also in the `.feed-entry-title`  div, I wound up wrapping it in another span. But only the update event title!

matthew-white (approved):

PR #899, opened 2023-11-29
Upgrade to node 20

lognaturel:
Ran locally, clicked around a bit.

Matches https://github.com/getodk/central-backend/pull/1052

No

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "test": "./test/run.sh"
   },
   "volta": {
-    "node": "18.17.0"
+    "node": "20.10.0"
   },
   "gitHooks": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -64,7 +64,7 @@
     "vue-cli-plugin-i18n": "~2"
   },
   "engines": {
-    "node": "18",
-    "npm": "9"
+    "node": "20",
+    "npm": "10"
```
matthew-white, line 68: I'd expect to see these change in package-lock.json as well.

lognaturel: Thank you!

Two observations:
- It's hard for me to remember to install dependencies in frontend when switching from backend where the `make` command takes care of it
- I'm in the habit of using `npm ci` but of course that doesn't update engines

Issue #521, opened 2021-12-17
Refresh submission keys in more cases


matthew-white:
Filing an issue based on this forum topic: https://forum.getodk.org/t/draft-submissions-not-accessible-when-project-is-encrypted-missing-passphrase-prompt-to-decrypt/36133

Right now, when submissions are refreshed (in either the Submissions tab or the Testing tab), we do not refresh the submission keys. That helps reduce the number of requests and usually works fine, with the thought that it is rare for there to be a change in the submission keys. However, in some cases, this leads to the "Decrypt and Download" modal not being shown: the user is not prompted for the passphrase.

The forum topic describe this use of the Testing tab:

- Create a form draft for an encrypted project.
- Navigate to the Testing tab.
- Create a submission.
- Refresh the submissions.
- Download the submissions: the modal is not shown.

I also remember @lognaturel encountering an issue like this in the past. I think the issue then went something like this:

- Create one or more submissions for a published form in a project that is not encrypted.
- View those submissions in the Submissions tab.
- In a different browser tab, encrypt the project.
- Create additional submissions using the new form version. These will be encrypted.
- Refresh the submissions in the first tab.
- Download the submissions in the first tab: the modal is not shown.

I think we decided that this second sequence was uncommon enough that we didn't need to make a change. However, the first sequence sounds like a more common workflow that I think we should address. (Maybe we should refresh the submission keys whenever we refresh the submissions. Or maybe we could refresh the submission keys only when there were previously no submissions: that wouldn't solve the second sequence, but it would solve the first.)

Either way, there is some complexity here because of how the components are set up: `SubmissionList` is the component shared between the two tabs and sends most of the requests, but it is the two separate tab components (`FormSubmissions` and `FormDraftTesting`) that send the requests for the submission keys.

ktuite:
QA observed related behavior in the "Analyze via OData" button on the submissions table (and entities table, coming in 2023.2).

Steps to reproduce
1. start with an encrypted project
2. add a new form
3. look at submissions for that form when there are 0 submissions - button is disabled
4. press "new" and make a new submission in a new tab
5. press "refresh" on submission page - new submission appears AND button becomes enabled
6. reloading the entire page with 1 or more submissions - button is disabled again and stays disabled

[...]

PR #910, opened 2023-12-11
Refresh keys when refreshing submissions

ktuite:
Closes https://github.com/getodk/central-frontend/issues/521

When submissions are refreshed, if we might need to refetch an encryption key (if the form is encrypted and we have no keys at the moment), then the submission list component will emit an event to the parent component, which is already responsible for the initial key request, to refetch the key.

Tests

Keys are loaded in two places, maybe that's not great.

Should fix an issue, shouldn't really be any risks. Keys (a small chunk of data) are fetched more often now, but in a way that seems necessary and useful.

### DIFF
--- a/src/components/submission/list.vue
+++ b/src/components/submission/list.vue
@@ -276,6 +276,13 @@ export default {
       })
         .finally(() => { this.refreshing = false; })
         .catch(noop);
+
+      // refresh keys when refreshing submissions (keys initially loaded in form/submissions)
+      if (refresh) {
+        this.keys.request({
+          url: apiPaths.submissionKeys(this.projectId, this.xmlFormId)
+        }).catch(noop);
+      }
     },
     fetchData() {
       this.fields.request({
--- a/src/components/submission/list.vue
+++ b/src/components/submission/list.vue
@@ -111,6 +111,7 @@ export default {
       default: (loaded) => (loaded < 1000 ? 250 : 1000)
     }
   },
+  emits: ['fetch-keys'],
   setup(props) {
     const { form, keys, resourceView } = useRequestData();
     const formVersion = props.draft
@@ -277,12 +278,9 @@ export default {
         .finally(() => { this.refreshing = false; })
         .catch(noop);
 
-      // refresh keys when refreshing submissions (keys initially loaded in form/submissions)
-      if (refresh) {
-        this.keys.request({
-          url: apiPaths.submissionKeys(this.projectId, this.xmlFormId)
-        }).catch(noop);
-      }
+      // emit event to parent component to re-fetch keys if needed
+      if (refresh && this.formVersion.keyId != null && this.keys.length === 0)
+        this.$emit('fetch-keys');
     },
     fetchData() {
       this.fields.request({
--- a/test/components/submission/list.spec.js
+++ b/test/components/submission/list.spec.js
@@ -1,6 +1,5 @@
 import Spinner from '../../../src/components/spinner.vue';
 import SubmissionDataRow from '../../../src/components/submission/data-row.vue';
-import SubmissionDownload from '../../../src/components/submission/download.vue';
 import SubmissionList from '../../../src/components/submission/list.vue';
 import SubmissionMetadataRow from '../../../src/components/submission/metadata-row.vue';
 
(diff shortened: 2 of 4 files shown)


src/components/submission/list.vue
```diff
@@ -276,6 +276,13 @@ export default {
       })
         .finally(() => { this.refreshing = false; })
         .catch(noop);
+
+      // refresh keys when refreshing submissions (keys initially loaded in form/submissions)
+      if (refresh) {
```
matthew-white, line 281: If we already have a key, then I think we can skip the request. Currently, it's only possible for there to be a single key, because an encrypted project can't be decrypted: `this.keys` is either an empty array or an array with one element.

```js
if (refresh && this.formVersion.keyId != null && this.keys.length === 0) {
```

src/components/submission/list.vue
```diff
@@ -276,6 +276,13 @@ export default {
...
+        this.keys.request({
+          url: apiPaths.submissionKeys(this.projectId, this.xmlFormId)
```
matthew-white, line 283: On the draft testing page, we need to fetch the keys for the draft, not the form. `apiPaths.submissionKeys()` takes an optional third argument for whether the form version is a draft. One place where you can see that is that `FormSubmissions` and `FormDraftTesting` send different requests for `keys`.

Actually, since those two components are already set up to send requests for `keys`, one option is for `SubmissionList` to emit an event here (e.g., `fetch-keys`) to indicate to the parent component that it should re-request `keys`. `FormSubmissions` and `FormDraftTesting` would have to listen for the event. The two components already know how to request `keys` in different ways, so it'd just be a matter of setting up the event. `ProjectShow` and `FormShow` use a similar pattern. If you take that approach though, you won't be able to use `loadSubmissionList()` to test this functionality, since it only mounts `SubmissionList`, not `FormSubmissions` or `FormDraftTesting`. You could use `load()` instead though.

ktuite: I tried this other approach, of emitting an event and having the parent component deal with it, and i like it. I'm not that sure how to test it. 

I added one test that emulates the "quirk" of the backend returning `[]` for keys when there are 0 submissions (but otherwise a key should probably be returned).

test/components/submission/list.spec.js
```diff
@@ -374,6 +374,29 @@ describe('SubmissionList', () => {
...
       });
 
+      describe('refreshing keys', () => {
+        it('sends request for encryption keys on submission refresh', () => {
+          testData.extendedProjects.createPast(1, {
+            key: testData.standardKeys.createPast(1).last(),
+            forms: 1
+          });
+          testData.extendedForms.createPast(1, {
+            submissions: 1
```
ktuite, line 384: i meant for this to be 0

test/components/submission/list.spec.js
```diff
@@ -374,6 +374,29 @@ describe('SubmissionList', () => {
...
+            keys: () => [] // if there are 0 submissions, backend returns empty key array
+          })
+            .complete()
+            .request(component =>
+              component.get('#submission-list-refresh-button').trigger('click'))
+            .beforeAnyResponse(() => {
+              testData.extendedSubmissions.createPast(1, { status: 'notDecrypted' });
+            })
+            .respondWithData(() => testData.submissionOData(1, 0))
+            .respondWithData(() => testData.standardKeys.sorted());
```
ktuite, line 396: Ideas for additional assertions in this test
1. open the modal and check that the password input is shown
2. use `testRequests` to make sure the requests are to the right form/draft key endpoints (can specify null for first element if not necessary for the test)

matthew-white (approved):

ktuite (commented):
Code review with @matthew-white

src/components/submission/list.vue
matthew-white: I was thinking more about this, and I think this comment wasn't entirely accurate. `keys` can only have one managed key, but I believe it could also have non-managed keys if form encryption was used before the project was encrypted. You can see in `SubmissionDownload` that `keys` is filtered for the managed key.

That said, I don't think anything here needs to change: form encryption should be rare outside of project encryption, and it seems like an edge case that a user would move from form-level encryption to project-level encryption (while also meeting other conditions needed for this whole issue to appear). The worst case is that `keys` isn't re-fetched. I wanted to make a note of this possibility though.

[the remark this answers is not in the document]
matthew-white: > If there isn't encryption, then we can know there isn't a managed key.

Actually, if the project is encrypted in another tab, then it could be the case that `this.formVersion.keyId == null` even while there are encryption `keys` to fetch. That case is described in the issue. However, I still think the optimization here is a good idea. The issue suggested that we don't need to address that case.

PR #913, opened 2023-12-12
Make dataset resource reactive so that entity count is updated

matthew-white:
We use `dataset.entities` to show the entity count. On the entity Data page, that allows us to show the entity count in the download button even before the OData response has been received. I don't think we show `dataset.entities` outside the Data page, but we could in the future. `odataEntities` is a resource local to the Data page, but `dataset` is available on all dataset pages.

The OData count is more likely to be up-to-date than `dataset.entities`: the dataset is requested before the OData and may have been requested some time ago. If the two counts differ, then `dataset.entities` is updated to match the OData count:

https://github.com/getodk/central-frontend/blob/294323e7f5a239fc0043a5bb001b205db41539ae/src/components/entity/list.vue#L99-L103

We do something similar with `form.submissions` and the OData count on the Submissions page, as well as with `formDraft.get().submissions` and the OData count on the Draft Testing page.

However, while looking into getodk/central#560, it seems clear that this mechanism isn't fully working. Backend can return an incorrect count for `dataset.entities` (getodk/central-backend#1062), yet even then, `dataset.entities` doesn't seem to be updated with the correct OData count.

Actually, it turns out that `dataset.entities` is updated, but that update isn't reflected in the DOM, because `dataset` isn't reactive. For performance reasons, resources have to opt into reactivity. I made `dataset` reactive, and after that, the count started being updated in the DOM.

I tried it locally and wrote a new test.

Anytime something is made reactive, it could theoretically have performance implications. However, I can't think of anywhere where that would be an issue for `dataset`. I also made `dataset` shallow reactive rather than deeply reactive, which should help with performance.

### DIFF
--- a/test/components/dataset/show.spec.js
+++ b/test/components/dataset/show.spec.js
@@ -16,11 +16,13 @@ describe('DatasetShow', () => {
     ]);
   });
 
-  it('shows the correct title', () =>
-    load('/projects/1/entity-lists/trees').then(app => {
+  it('shows the correct title', () => {
+    testData.extendedDatasets.createPast(1);
+    return load('/projects/1/entity-lists/trees').then(app => {
       const title = app.get('#page-head-title');
       title.text().should.be.equal('trees');
-    }));
+    });
+  });
 
   it('re-renders the router view after a route change', () => {
     testData.extendedDatasets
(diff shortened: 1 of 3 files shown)


test/components/dataset/show.spec.js
matthew-white, line 1: I changed two tests in this file. Previously, those tests didn't create a test dataset. I'm a little surprised that they passed! However, once I started using `shallowReactive()`, the tests started logging warnings about calling `shallowReactive()` on `undefined`. (Since there was no dataset, the response data was `undefined`.)

ktuite (approved):

PR #920, opened 2023-12-15
Do not show conflicting property explanation in tooltip of property name

matthew-white:
In the diff shown on the entity detail page, if a property name is long, then it is truncated with a tooltip. However, right now, if the property is a conflicting property, then the tooltip includes the text of the `.sr-only` element next to the property name. The tooltip should only show the property name: the text of the `.sr-only` element is shown in a different tooltip.

The fix involved moving the `v-tooltip.text` from the `<td>` element to a new `<span>` element wrapping the property name. With the introduction of the `<span>`, the selector for the tooltip element had to change in an existing test. Beyond that, I think it's hard to write a meaningful test for this fix, as any test of a tooltip already has to select the specific element that has the tooltip. Here, that element is the new `<span>`, which we know doesn't contain an `.sr-only` element.

I also viewed this change locally and was able to verify it that way.

### DIFF
--- a/src/components/entity/diff/row.vue
+++ b/src/components/entity/diff/row.vue
@@ -11,9 +11,9 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <tr class="entity-diff-row">
-    <td :class="{ property: name !== 'label', conflicting }" v-tooltip.text>
-      {{ name === 'label' ? $t('entity.label') : name }}
-      <span v-if="conflicting" class="sr-only">{{ $t('conflictingProp') }}</span>
+    <td :class="{ property: name !== 'label', conflicting }">
+      <span v-tooltip.text>{{ name === 'label' ? $t('entity.label') : name }}</span>
+      <span v-if="conflicting" class="sr-only">&nbsp;{{ $t('conflictingProp') }}</span>
     </td>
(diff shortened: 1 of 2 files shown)


src/components/entity/diff/row.vue
```diff
@@ -11,9 +11,9 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <tr class="entity-diff-row">
-    <td :class="{ property: name !== 'label', conflicting }" v-tooltip.text>
-      {{ name === 'label' ? $t('entity.label') : name }}
-      <span v-if="conflicting" class="sr-only">{{ $t('conflictingProp') }}</span>
+    <td :class="{ property: name !== 'label', conflicting }">
+      <span v-tooltip.text>{{ name === 'label' ? $t('entity.label') : name }}</span>
+      <span v-if="conflicting" class="sr-only">&nbsp;{{ $t('conflictingProp') }}</span>
```
matthew-white, line 16: With the introduction of the `<span>`, there was no whitespace between the property name and the text of the `.sr-only` element. I think whitespace is needed for screen readers, which is why I added an `&nbsp;`.

ktuite (approved):

PR #921, opened 2023-12-15
Show "Link Entity List" button after publishing new entity list

### LINKED ISSUE #916: Link to entity list button sometimes not shown on form attachments
See this [comment](https://github.com/getodk/central/issues/554#issuecomment-1846935350) for a description of the issue.

The form attachment page loads a list of datasets and decides whether to show a "Link this to existing Entity List" button, but in the case where a form was JUST published and just created a new dataset, the link button sometimes doesn't appear.

I thought I was seeing it in the 0->1 dataset count case, but I couldn't reproduce it. I can consistently recreate this error in the n->n+1 dataset case, though.

matthew-white:
Closes #916. I think the issue is that the `datasets` resource needs to be cleared/reset after the form draft is published. That's because publishing a draft can have the effect of publishing a new entity list. In `FormDraftStatus`, the `afterPublish()` method re-requests, resets, or updates many resources, but it doesn't do so for `datasets`. That means that if `FormAttachmentList` requested `datasets` before the draft was published, it won't do so again after the draft is published (unless the page is refreshed or the user navigates away from the form).

I wrote two new tests about this scenario (under "linking after publishing new dataset"). The first test involves publishing the project's first entity list, and it passed without any changes. The second test involves publishing a second entity list for the project, and it only passed after I updated `afterPublish()`.

While looking at this issue, it didn't always feel clear when or where resources were re-requested, reset, or updated. There's probably room for improvement there (though I don't have any specific ideas at the moment). However, given that `afterPublish()` is already modifying a lot of resources, I think it makes sense that it should also reset `datasets`.

### DIFF
--- a/test/components/form-attachment/list.spec.js
+++ b/test/components/form-attachment/list.spec.js
@@ -1037,15 +1037,31 @@ describe('FormAttachmentList', () => {
   });
 
   describe('dataset linking', () => {
-    beforeEach(() => {
+    it('sends a request for datasets if the project has one', () => {
+      testData.extendedProjects.createPast(1, { forms: 2, datasets: 1 });
+      testData.extendedDatasets.createPast(1);
       testData.extendedForms.createPast(1, { draft: true });
+      testData.standardFormAttachments.createPast(1);
+      return load('/projects/1/forms/f/draft/attachments')
+        .respondWithData(() => testData.extendedDatasets.sorted())
+        .testRequests([
+          null, // project
+          null, // form
+          null, // formDraft
+          null, // attachments
+          { url: '/v1/projects/1/datasets' }
+        ]);
     });
 
-    it('autolinks dataset', async () => {
+    it('renders correctly for an attachment linked to a dataset', async () => {
+      testData.extendedProjects.createPast(1, { forms: 2, datasets: 1 });
+      testData.extendedDatasets.createPast(1, { name: 'shovels' });
+      testData.extendedForms.createPast(1, { draft: true });
       testData.standardFormAttachments.createPast(1, { type: 'file', name: 'shovels.csv', datasetExists: true });
       const component = await load('/projects/1/forms/f/draft/attachments', {
         root: false
-      });
+      })
+        .respondWithData(() => testData.extendedDatasets.sorted());
       component.get('td.form-attachment-list-uploaded .dataset-label').text().should.equal('Linked to Entity List shovels');
       component.get('td.form-attachment-list-action').text().should.equal('Upload a file to override.');
       const a = component.get('td.form-attachment-list-name a');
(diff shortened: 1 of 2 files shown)


test/components/form-attachment/list.spec.js
```diff
@@ -1037,15 +1037,31 @@ describe('FormAttachmentList', () => {
   });
 
   describe('dataset linking', () => {
-    beforeEach(() => {
+    it('sends a request for datasets if the project has one', () => {
+      testData.extendedProjects.createPast(1, { forms: 2, datasets: 1 });
+      testData.extendedDatasets.createPast(1);
       testData.extendedForms.createPast(1, { draft: true });
```
matthew-white, line 1043: There used to be a `beforeEach()` hook that created a form draft. However, I think it's better to wait to create the form draft until the project is created. Otherwise, creating the form draft will create a separate project.

test/components/form-attachment/list.spec.js
```diff
@@ -1055,11 +1071,8 @@ describe('FormAttachmentList', () => {
 
     describe('Datasets preview hint', () => {
       beforeEach(() => {
-        testData.extendedProjects.createPast(1, {
-          name: 'My Project Name',
-          forms: 1,
-          datasets: 1
-        });
+        testData.extendedProjects.createPast(1, { forms: 2, datasets: 1 });
```
matthew-white, line 1074: I changed `forms` from 1 to 2 with the thought that if there's already an entity list, then there should also be another form in addition to the draft form. I don't think this change has practical ramifications, but I think it's slightly clearer.

I also removed the project name, which isn't used in the test.

test/components/form-attachment/list.spec.js
```diff
@@ -1037,15 +1037,31 @@ describe('FormAttachmentList', () => {
...
+    it('renders correctly for an attachment linked to a dataset', async () => {
+      testData.extendedProjects.createPast(1, { forms: 2, datasets: 1 });
+      testData.extendedDatasets.createPast(1, { name: 'shovels' });
+      testData.extendedForms.createPast(1, { draft: true });
       testData.standardFormAttachments.createPast(1, { type: 'file', name: 'shovels.csv', datasetExists: true });
       const component = await load('/projects/1/forms/f/draft/attachments', {
         root: false
-      });
+      })
+        .respondWithData(() => testData.extendedDatasets.sorted());
```
matthew-white, line 1064: This test makes assertions about an attachment that is linked to an entity list. However, for an attachment to be linked to an entity list, there has to be an entity list, so I created one in `testData` and returned it to `load()`. Those changes don't affect whether the test passes or not, but the test now more closely matches how the page works in practice.

test/components/form-attachment/list.spec.js
```diff
@@ -1037,15 +1037,31 @@ describe('FormAttachmentList', () => {
   });
 
   describe('dataset linking', () => {
-    beforeEach(() => {
+    it('sends a request for datasets if the project has one', () => {
```
matthew-white, line 1040: This is a test of existing behavior, but I thought it was useful to add. The reverse case — that if the project doesn't have a dataset, then it doesn't send a request — is implicitly tested by all the tests above.

ktuite (approved):

PR #928, opened 2023-12-21
Truncate long display name in navbar

### LINKED ISSUE #348: Account for long user display name in navbar
Right now if a user's display name is too long, the dropdowns in the navbar appear below the navbar:

Screen Shot 2020-07-19 at 8 55 49 AM

Instead, we should truncate the name if it is greater than a certain width, specifying `text-overflow: ellipsis`. We should add a tooltip like we do elsewhere so that the full name can still be seen.

We do something similar in the `FormVersionSummaryItem` component:

Screen Shot 2020-07-22 at 11 29 22 AM

matthew-white:
Closes #348.

If the user's display name is less than 275px, then it should look exactly like it does today. If it is 275px or more, then it should be truncated:

I tried it out locally.

275px is a bit long, but there are places in Central where we allow text to go 250px before truncation. There's room in the navbar, and I think it's nice to avoid truncation in this case, so I opted for a little more than 250px. Especially if we ever add more to the navbar, I think we could also consider removing the name from the navbar and using generic text like "My Account".

In terms of the code, the hardest part was keeping the icon on the left, the text, and the caret on the right aligned. I was initially running into issues there, but `vertical-align: top;` did the trick. I also briefly tried a flexbox strategy, but I found myself having to do some manual alignment.

### DIFF
--- a/src/components/navbar/actions.vue
+++ b/src/components/navbar/actions.vue
@@ -19,7 +19,7 @@ except according to the terms contained in the LICENSE file.
     <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button"
       aria-haspopup="true" aria-expanded="false">
       <span class="icon-user-circle-o"></span>
-      <span>{{ currentUser.displayName }}</span>
+      <span v-tooltip.text>{{ currentUser.displayName }}</span>
       <span class="caret"></span>
     </a>
     <ul class="dropdown-menu">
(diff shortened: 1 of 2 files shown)


src/components/navbar/actions.vue
```diff
@@ -71,6 +71,19 @@ export default {
 };
 </script>
 
+<style lang="scss">
+@import '../../assets/scss/mixins';
+
+#navbar-actions {
+  .dropdown-toggle .icon-user-circle-o + span {
+    @include text-overflow-ellipsis;
+    display: inline-block;
```
matthew-white, line 80: This works because the parent `<a>` has `display: block;`.

ktuite (approved):

PR #932, opened 2024-01-19
Remove use of submissionCreate from entity audit event

ktuite:
Frontend part of issue https://github.com/getodk/central-backend/issues/1036 and backend PR https://github.com/getodk/central-backend/pull/1072.

It used to be that entity events that were triggered by a Submission (e.g. entity create OR update) would return
- `sourceEvent` (the actual event that caused the entity update, whether it was a submission approval, or submission creation, etc.)
- `submission` (the Submission object, but only if a Submission was part of the event AND the Submission was not deleted)
- a `submissionCreate` object for the Submission (if a Submission was part of the event) that had enough historical information including the creation time and instance ID to describe the Submission even if it was deleted. 

The big change for https://github.com/getodk/central-backend/pull/1072 was to remove `submissionCreate` and always return some kind of Submission object, either the full details, or the instance ID, submitter, and timestamp. 

Flavors of frontend changes: 
- Switch from looking for `submissionCreate` to know if the event was related to a submission to just looking for `submission`. 
- Switch from looking for `submission` to know the submission was not deleted to looking for `submission.currentVersion`
- `submission` and `event` are contained within an entity audit event's `details` nested under `source` and all entity events have a source, though if its an API event, the source is just `{}`

Basic details of an Entity shows the source submission, and because this isn't returned with the Entity itself (though maybe it should be), the basic details component waits for the audit log and checks the entity's creation event for a source submission. 

Tests.

I was thinking about refactoring tests to have more of them use `extendedEntities.createSourceSubmission` but existing tests have such specific setups I was finding it hard to refactor.

We've long talked about removing this event detail and using different information instead.

Could change how entity events appear on frontend, but it shouldn't. 

### DIFF
--- a/src/components/entity/feed-entry.vue
+++ b/src/components/entity/feed-entry.vue
@@ -63,7 +63,7 @@ except according to the terms contained in the LICENSE file.
       <template v-else-if="entry.action === 'entity.update.version'">
         <span class="icon-pencil"></span>
         <span class="title">
-        <template v-if="entry.details.source?.submission != null">
+        <template v-if="submission != null">
           <i18n-t v-if="submission.currentVersion != null"
             keypath="title.entity.update_version.submission.notDeleted">
             <template #instanceName>
@@ -148,16 +148,11 @@ const creatingSubmissionPath = computed(() => submissionPath(
 ));
 const { t } = useI18n();
 
-// This function pulls out the submission instance ID when the event
-// is about a submission (creation, approval) and the ID is directly
-// in the event details.
 const deletedSubmission = computed(() => {
   const id = props.submission.instanceId;
   return t('title.submission.create.deleted.deletedSubmission', { id });
 });
 
-// This function pulls out the submission instance ID in events about
-// entities
 const deletedSubmissionEntityEvent = computed(() => {
   const id = props.submission.instanceId;
   return t('title.entity.update_version.submission.deleted.deletedSubmission', { id });
--- a/src/components/entity/feed-entry.vue
+++ b/src/components/entity/feed-entry.vue
@@ -15,20 +15,20 @@ except according to the terms contained in the LICENSE file.
     <template #title>
       <template v-if="entry.action === 'submission.create'">
         <span class="icon-cloud-upload"></span>
-        <i18n-t v-if="submission != null"
+        <i18n-t v-if="submission.currentVersion != null"
           keypath="title.submission.create.notDeleted">
           <template #instanceName>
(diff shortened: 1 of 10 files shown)


test/data/entities.js
```diff
@@ -258,22 +258,25 @@ export const entityOData = (top = 250, skip = 0) => {
...
+    };
 
   const auditOptions = {
     actor: submission.submitter,
     actee: formVersion,
     details: { instanceId: submission.instanceId }
   };
-  const submissionCreate = extendedAudits
+  extendedAudits
     .createPast(1, {
```
matthew-white, line 280: There's a `.last()` that's chained at the end below that could be removed.

src/components/entity/feed-entry.vue
```diff
@@ -63,8 +63,8 @@ except according to the terms contained in the LICENSE file.
       <template v-else-if="entry.action === 'entity.update.version'">
         <span class="icon-pencil"></span>
         <span class="title">
-        <template v-if="entry.details.submissionCreate != null">
-          <i18n-t v-if="submission != null"
+        <template v-if="entry.details.submission != null">
```
matthew-white, line 66:
```suggestion
        <template v-if="submission != null">
```

I think it's a little clearer just to use `submission`, since `submission` is also used on the next line.

src/components/entity/feed-entry.vue
```diff
@@ -152,14 +152,14 @@ const { t } = useI18n();
 // is about a submission (creation, approval) and the ID is directly
 // in the event details.
```
matthew-white, line 153: This comment looks possibly out-of-date (the mention of event details).

test/components/entity/feed-entry.spec.js
```diff
@@ -163,10 +170,8 @@ describe('EntityFeedEntry', () => {
         entity: { uuid: 'e' }
       };
       if (options.submission === true) {
-        testData.extendedSubmissions.createPast(1);
-        details.submissionCreate = testData.extendedAudits
-          .createPast(1, { action: 'submission.create' })
-          .last();
+        const submission = testData.extendedSubmissions.createPast(1);
```
matthew-white, line 173:
```suggestion
        const submission = testData.extendedSubmissions.createPast(1).last();
```

`createPast()` returns the test data store for chaining. I guess in the case of `entity.create`, the component is just looking at the presence of the `submission` property and not its value, which is why this wasn't an issue for tests.

matthew-white (approved):
Looks great!

PR #967, opened 2024-04-12
Hide problem alerts after successful new dataset/property requests

ktuite:
Closes https://github.com/getodk/central/issues/624 and closes https://github.com/getodk/central/issues/626

Also add custom error message for property name conflict similar to message for dataset name conflict in PR https://github.com/getodk/central-frontend/pull/966 for issue https://github.com/getodk/central/issues/625. Doing so closes getodk/central#634.

### DIFF
--- a/src/components/dataset/overview/new-property.vue
+++ b/src/components/dataset/overview/new-property.vue
@@ -38,6 +38,8 @@ except according to the terms contained in the LICENSE file.
 
 <script setup>
 import { inject, ref, watch } from 'vue';
+import { equals } from 'ramda';
+import { useI18n } from 'vue-i18n';
 
 import Modal from '../../modal.vue';
 import FormGroup from '../../form-group.vue';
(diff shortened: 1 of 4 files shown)


src/components/dataset/overview/new-property.vue
```diff
@@ -91,7 +100,10 @@ const submit = () => {
       "To add an Entity property, choose a unique property name below.",
       "You can also add new properties by uploading a Form that references them, in which case the properties are created when the Form is published."
     ],
-    "newPropertyName": "New property name"
+    "newPropertyName": "New property name",
+    "problem": {
+      "409_3": "A property already exists in this Entity List with the name of “{propertyName}”."
```
matthew-white, line 105: I usually just use quotes for strings that might contain a space. Since `propertyName` will be a single word, what do you think about removing the quotes here? No strong feelings!

matthew-white (approved):

src/components/dataset/overview/new-property.vue
ktuite: I think the quotes isolate the name better than without them, since there is no other formatting.

PR #968, opened 2024-04-12
Wrap long dataset name in success modal

ktuite:
Closes https://github.com/getodk/central/issues/630

Screenshot 2024-04-12 at 3 27 42 PM
Screenshot 2024-04-12 at 3 29 27 PM

I feel kind of unsure about these CSS changes. I did add `overflow-wrap: break-word;` to the `text-block` mixin, and then used that mixin. 

I also tried using a flexbox, but ran into an issue where the text flex item was hanging off the edge. Wasn't sure how to fix it, so that's where the 80% width came from... 


src/components/dataset/new.vue
```diff
@@ -130,8 +132,19 @@ const hideOrComplete = () => {
...
 
 #dataset-new-success {
+  display: flex;
+  align-items: center;
+  margin-bottom: 7px;
+
+  > div {
+    @include text-block;
+    width: 80%;
+    margin-left: 3px;
```
matthew-white, line 145: It looks like the `.icon-check-circle` already has `margin-right`. Could the two margins be combined in a single property?

src/components/dataset/new.vue
```diff
@@ -130,8 +132,19 @@ const hideOrComplete = () => {
 
 <style lang="scss">
 @import '../../assets/scss/variables';
+@import '../../assets/scss/mixins';
 
 #dataset-new-success {
+  display: flex;
+  align-items: center;
```
matthew-white, line 139: If this is `align-items: center;` here, then I'm guessing that the `.icon-check-circle` doesn't need `vertical-align: middle;`.

src/components/dataset/new.vue
```diff
@@ -130,8 +132,19 @@ const hideOrComplete = () => {
...
+  margin-bottom: 7px;
+
+  > div {
+    @include text-block;
```
matthew-white, line 143: Could this `<div>` be a `<p>` instead? Then `text-block` would be applied automatically.

matthew-white:
> I also tried using a flexbox, but ran into an issue where the text flex item was hanging off the edge. Wasn't sure how to fix it, so that's where the 80% width came from...

Hmm that's interesting that it hangs off the edge. Two ideas:

- Specify `flex-shrink: 0;` for the `.icon-check-circle` and `width: 100%;` for the `<div>`
- Specify `flex-shrink: 0;` for the `.icon-check-circle`. Don't specify any width for the `<div>`, but do specify `flex-grow: 1`;.

If there's a way to use the full width avaialble, that sounds nice, especially for languages where the text might be longer. But 80% also sounds completely fine if that's what gets things working.

I'll also a few more CSS thoughts below.

matthew-white (approved):

matthew-white:
I'm thinking that there will be a similar issue in `FieldKeyNew`. If it's not too hard, do you think that could be fixed at the same time? We could also file an issue about it and circle back to it later.

ktuite:
I think the issue is with `overflow-wrap: break-word`... when I tried `overflow-wrap: anywhere`, it could fit the text in the box without hanging over. 

Was looking at [this](https://blog.logrocket.com/guide-css-word-wrap-overflow-wrap-word-break/#word-wrap-overflow-wrap-css), maybe it is relevant?
> The difference between anywhere and break-word is apparent when calculating the min-content intrinsic sizes. With break-word, the browser doesn’t consider the soft wrap opportunities introduced by the word break when calculating min-content intrinsic sizes, but it does with anywhere. For more about min-content intrinsic sizes, check out our guide [here](https://blog.logrocket.com/understanding-min-content-max-content-fit-content-css/).

I ended up going with the `width: 80%` change because even with the above clues, I wasn't able to get past the text hanging out of its box.

PR #973, opened 2024-04-21
Update translations before v2024.1 regression testing

matthew-white:
This PR makes progress on getodk/central#569. I pulled the latest strings from Transifex, then ran destructure.js.


src/components/field-key/revoke.vue
```diff
@@ -93,64 +93,56 @@ export default {
     "title": "Odebrat přístup uživatele",
     "introduction": [
       "Opravdu chcete odebrat přístup uživateli aplikace {displayName}?",
-      "Stávající příspěvky od tohoto uživatele zůstanou, ale kdokoli se na tohoto uživatele obrátí, bude muset vytvořit nové, aby pokračoval ve stahování formulářů nebo odesílání příspěvků.",
-      "Tuto akci nelze vrátit zpět."
```
matthew-white, line 97: The source message is "This action cannot be undone.", and it was moved out of this component and into en.json5 for use in other components. Because it was in an array before, I decided not to use `@transifexKey`, so it will need to be retranslated. Once it is, it will appear in cs.json.

src/components/form/new.vue
```diff
@@ -357,7 +357,6 @@ export default {
       "chooseOne": "vyberte jeden"
     },
     "action": {
-      "upload": "Nahrát",
```
matthew-white, line 360: The source message is "Upload", which was moved to en.json5. I used `@transifexKey`, so this message should appear in cs.json without retranslation.

src/components/project/enable-encryption.vue
```diff
@@ -343,11 +343,7 @@ export default {
             [
               "Formuláře nakonfigurované pomocí ručně {submission} klíčů budou tyto klíče nadále používat a musí být ručně dešifrovány.",
               "Chcete-li použít automatický proces centrálního šifrování na těchto formulářích, odeberte konfiguraci {base64RsaPublicKey}."
-            ],
-            "Již nebudete mít možnost zobrazit online náhled odeslaných údajů.",
-            "Již se nebudete moci připojit k datům přes OData.",
```
matthew-white, line 348: I think these two messages were changed in #970. Because they're no longer translated, the messages after them can also no longer be included.

src/locales/cs.json
```diff
@@ -297,9 +298,6 @@
     },
     "total": "Celkem",
     "totalSubmissions": "Celkem odeslání",
-    "punctuations": {
-      "comma": ","
```
matthew-white, line 301: This message was moved elsewhere in the file using `@transifexKey`.

lognaturel (approved):
I quickly scrolled through to make sure all changes appear translation-related and they do.

PR #978, opened 2024-04-24
Do not show warning if CSV delimiter is not a comma

matthew-white:
This PR changes what happens if something other than a comma is used as the CSV delimiter. A non-comma delimiter will continue to be shown in `EntityHeaderErrors` if the header row is invalid. However, it will not be returned in the warnings from `parseCSV()` or shown in `EntityUploadWarnings`.

I updated tests to match the new behavior.


lognaturel (approved):
The removals look good. Is there still a test that verifies the message is shown in the expected case? I'm not quickly finding it. Maybe it's not that important?

matthew-white:
Ah yes, that's a test that was missing. There are still a number of those, and I've filed the issue #979 about that. But this is probably as good a time as any to add that missing test. I've just pushed a commit that tests that the `EntityUploadHeaderErrors` component use the delimiter from the CSV file.

PR #980, opened 2024-04-25
Ensure that .modal-actions is visible after entities CSV is selected

matthew-white:
Closes getodk/central#639.

I tried it locally. I don't think we need to write a test for this behavior.


lognaturel (approved):
This is the approach we discussed. Agreed it doesn't need a test!

PR #981, opened 2024-04-25
Prevent EntityMetadataRow actions from wrapping

matthew-white:
Closes getodk/central#620.

I tried it locally.

I increased the min width of the "Last Updated / Actions" column so that there is always enough room for the actions.

I also tried keeping the min width like it was, but allowing the actions to spill into the "Created at" column to the left. However, that didn't look better to me and seemed potentially confusing.

### DIFF
--- a/src/components/entity/metadata-row.vue
+++ b/src/components/entity/metadata-row.vue
@@ -107,6 +107,10 @@ const { entityPath } = useRoutes();
   .action-cell {
     padding-top: 4px;
     padding-bottom: 4px;
+
+    // Ensure that the column is wide enough that the .btn-group does not wrap.
+    min-width: 170px;
+    &:lang(id) { min-width: 215px; }
   }
 
   .col-content {


src/components/entity/metadata-row.vue
```diff
@@ -107,6 +107,10 @@ const { entityPath } = useRoutes();
   .action-cell {
     padding-top: 4px;
     padding-bottom: 4px;
+
+    // Ensure that the column is wide enough that the .btn-group does not wrap.
+    min-width: 170px;
+    &:lang(id) { min-width: 215px; }
```
matthew-white, line 113: I chose these numbers as follows:

- The min width required depends on the locale, since the text of the "More" button is translated.
- If the locale is English, then the width of a totally empty column ends up being 160px due to the width of the header.
- If an entity was updated more than a week ago, then the full date/time is shown. The column ends up being 215px wide.
- In English, the column needs to be 163px wide in order for the actions not to wrap.
- I thought about just always setting the min width to 215px. But if the column is fully empty, I think it's nice for it to be more narrow.
- If the locale is Indonesian, then the column needs to be wider to prevent wrapping, due to the text: it needs to be 207px.
- Other locales don't currently have an issue with actions wrapping. In many cases, the header is simply long enough that there is no wrapping.

lognaturel (approved):

PR #993, opened 2024-05-09
Make small changes to ESLint config

matthew-white:
I noticed that the default ESLint config from Vite has a few changes from ours (related: #671). This PR pulls in some of those changes now.

Linting continues to pass.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,15 +1,12 @@
 module.exports = {
   root: true,
   parserOptions: {
-    ecmaVersion: 2020
+    ecmaVersion: 'latest'
   },
   extends: [
     'plugin:vue/vue3-recommended',
     '@vue/airbnb'
   ],
-  env: {
-    node: true
-  },
   globals: {
     $: 'readonly',
     alert: 'readonly',


.eslintrc.js
```diff
@@ -1,15 +1,12 @@
...
-    ecmaVersion: 2020
+    ecmaVersion: 'latest'
   },
   extends: [
     'plugin:vue/vue3-recommended',
     '@vue/airbnb'
   ],
-  env: {
-    node: true
-  },
```
matthew-white, line 12: `env` tells ESLint which variables are global variables so it doesn't flag them. However, Frontend isn't a Node project, so specifying `node` here doesn't make sense. The actual global variables that we use are listed below under `globals`.

.eslintrc.js
```diff
@@ -1,15 +1,12 @@
 module.exports = {
   root: true,
   parserOptions: {
-    ecmaVersion: 2020
+    ecmaVersion: 'latest'
```
matthew-white, line 4: We target the latest two browser versions, so I think `latest` makes sense. We certainly aren't targeting 2020.

ktuite (approved):

PR #995, opened 2024-05-09
Make small changes to <meta> in index.html

matthew-white:
This PR makes a couple of small changes to index.html to more closely match the index.html that ships with Vite (related: #671).

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -13,8 +13,7 @@ except according to the terms contained in the LICENSE file.
 <html lang="en">
   <head>
     <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -13,8 +13,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
matthew-white, line 16: It looks like this was used with IE, but IE is retired. The index.html that ships with Vite doesn't include this tag. I've seen recommendations not to include this tag anymore.

ktuite (approved):

PR #1000, opened 2024-05-21
Fix css for FieldKeyQrPanel shown outside FieldKeyNew

ktuite:
Closes https://github.com/getodk/central/issues/663

Fixes some CSS issues that came from supporting a longer app user name and caused weird padding to appear behind the app user QR popover. 

Looks better now!

Screenshot 2024-05-21 at 12 15 49 PM

Looks ok here, too.
Screenshot 2024-05-21 at 12 15 34 PM

visually inspecting it.

The element for saying the new app user/QR code was successfully created is now a sibling of the element with the actual QR code in it, instead of the parent. Changed the CSS a bit to support the QR code element being shown in two places a) the success modal, and b) the normal re-viewing of a QR code in the popover.

Fixes a bug.

No.

### DIFF
--- a/src/components/field-key/new.vue
+++ b/src/components/field-key/new.vue
@@ -182,11 +182,14 @@ export default {
     width: 80%;
     margin-bottom: 0px;
   }
+
+  + .field-key-qr-panel {
+    margin: 15px auto 30px;
+  }
 }
 
 .field-key-qr-panel {
   box-shadow: $box-shadow-popover;
-  margin: 15px auto 30px;
 }
 </style>
 
--- a/src/components/field-key/new.vue
+++ b/src/components/field-key/new.vue
@@ -185,12 +185,10 @@ export default {
 
   + .field-key-qr-panel {
     margin: 15px auto 30px;
+    box-shadow: $box-shadow-popover;
   }
 }
 
-.field-key-qr-panel {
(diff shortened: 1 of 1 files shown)


src/components/field-key/new.vue
```diff
@@ -182,11 +182,14 @@ export default {
...
     margin-bottom: 0px;
   }
+
+  + .field-key-qr-panel {
+    margin: 15px auto 30px;
+  }
 }
 
 .field-key-qr-panel {
   box-shadow: $box-shadow-popover;
```
matthew-white, line 192: I think this box shadow should be moved above as well. Otherwise, it will be applied to the `FieldKeyQrPanel` shown in the popover. When I look at the [component](https://github.com/getodk/central-frontend/blob/v2023.5.0/src/components/field-key/new.vue) in v2023.5, I see that the box shadow is nested under `#field-key-new-success` and not applied to the popover. The styles here aren't automatically scoped to the `FieldKeyNew` component, so specifying a style on `.field-key-qr-panel` outside of `#field-key-new-success` will apply to all `.field-key-qr-panel` elements.

matthew-white (approved):
Looks good! When you merge the PR into `master`, would you mind using a slightly longer commit message? Maybe something like "Fix css for FieldKeyQrPanel shown outside FieldKeyNew".

PR #1001, opened 2024-05-28
Add new Web Form preview button under feature flag

sadiqkhoja:
Closes # TBA

https://github.com/getodk/central-frontend/assets/447837/0114b2ba-6451-44b7-bc5c-2b0e45a773cf

Manually verified it. Will add component tests soon.

I have also verified that initial bundle size doesn't increase with this PR, new assets are only loaded when new Web Form preview page is opened.

We discussed other options like printing URL of the new Web Forms preview in the console, but that would pollute the console log. Current approach just prints the instruction once at the start of the application about how to access new Web Forms preview. 

I have added a new `standalone` meta property to the routes, by default its value is false for all the routes and true for FormPreview. When this property is true, I don't render navigation bar and application's wrapper. Other option was to add new Odk-Web-Forms component in a separate Vue instance, which would be devoid of all the functionality that we already have like auth, request, store, etc. and probably separate build configurations.

I have create FeatureFlags composable that can be extended to include more alpha/beta features in the future. I am open to not go that route and rename this composable to just reflect new Web Forms feature.

Maybe not. We will probably announce this in our Insiders call

### DIFF
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -13,7 +13,6 @@ except according to the terms contained in the LICENSE file.
 @import './mixins';
 
 html {
-  background-color: $color-accent-secondary;
   min-height: 100%;
 }
 
--- a/src/components/app.vue
+++ b/src/components/app.vue
@@ -10,18 +10,21 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <div>
+  <div :class="features">
     <!-- If the user's session is restored during the initial navigation, that
     will affect how the navbar is rendered. -->
-    <navbar v-show="routerReady"/>
+    <navbar v-if="!$route.meta.standalone" v-show="routerReady"/>
     <alert id="app-alert"/>
     <feedback-button v-if="showsFeedbackButton"/>
     <!-- Specifying .capture so that an alert is not hidden immediately if it
     was shown after the click. -->
     <!-- eslint-disable-next-line vuejs-accessibility/click-events-have-key-events -->
-    <div class="container-fluid" @click.capture="hideAlertAfterClick">
+    <div v-if="!$route.meta.standalone" class="container-fluid" @click.capture="hideAlertAfterClick">
       <router-view/>
     </div>
+    <template v-if="$route.meta.standalone">
+      <router-view/>
+    </template>
     <div id="tooltips"></div>
   </div>
 </template>
@@ -36,6 +39,7 @@ import Navbar from './navbar.vue';
 
 import useCallWait from '../composables/call-wait';
 import useDisabled from '../composables/disabled';
+import useFeatureFlags from '../composables/feature-flags';
 import { useRequestData } from '../request-data';
 import { useSessions } from '../util/session';
 import { loadAsync } from '../util/load-async';
@@ -47,10 +51,11 @@ export default {
   setup() {
     const { visiblyLoggedIn } = useSessions();
     useDisabled();
+    const { features } = useFeatureFlags();
 
     const { centralVersion } = useRequestData();
     const { callWait } = useCallWait();
-    return { visiblyLoggedIn, centralVersion, callWait };
+    return { visiblyLoggedIn, centralVersion, callWait, features };
   },
   computed: {
     routerReady() {
--- a/src/components/form-version/standard-buttons.vue
+++ b/src/components/form-version/standard-buttons.vue
@@ -24,7 +24,6 @@ except according to the terms contained in the LICENSE file.
 </template>
 
 <script>
-import { ref } from 'vue';
 import EnketoPreview from '../enketo/preview.vue';
 import FormVersionDefDropdown from './def-dropdown.vue';
 import useRoutes from '../../composables/routes';
--- a/src/components/form-version/standard-buttons.vue
+++ b/src/components/form-version/standard-buttons.vue
@@ -14,14 +14,20 @@ except according to the terms contained in the LICENSE file.
 <template>
   <span class="form-version-standard-buttons">
     <enketo-preview :form-version="version"/>
+    <RouterLink class="btn btn-default btn-web-form" :to="formPath('preview')" target="_blank">
+      <span class="icon-eye"></span>
+      New Preview
+    </RouterLink>
     <form-version-def-dropdown :version="version"
       @view-xml="$emit('view-xml')"/>
   </span>
 </template>
 
 <script>
+import { ref } from 'vue';
 import EnketoPreview from '../enketo/preview.vue';
 import FormVersionDefDropdown from './def-dropdown.vue';
+import useRoutes from '../../composables/routes';
 
 export default {
   name: 'FormVersionStandardButtons',
@@ -32,12 +38,36 @@ export default {
       required: true
     }
   },
-  emits: ['view-xml']
+  emits: ['view-xml'],
+  setup() {
+    const { formPath } = useRoutes();
+    return { formPath };
+  }
 };
 </script>
 
 <style lang="scss">
-.form-version-standard-buttons .enketo-preview {
-  margin-right: 5px;
+.form-version-standard-buttons {
+  .enketo-preview {
+    margin-right: 5px;
+  }
+  .btn-web-form {
+    display: none;
+  }
 }
+
+// `new-web-forms` class is added to the root of App component when the
+// new web form feature is enabled.
+.new-web-forms {
+  .form-version-standard-buttons {
+    .enketo-preview {
+      display: none;
+    }
+    .btn-web-form {
+      margin-right: 5px;
+      display: inline-block;
+    }
+  }
+}
+
 </style>
--- a/src/components/form/preview.vue
+++ b/src/components/form/preview.vue
@@ -1,0 +1,79 @@
+<template>
+      <template v-if="formVersionXml.dataExists">
+        <OdkWebForm :form-xml="formVersionXml.data" @submit="handleSubmit"/>
+      </template>
+
+      <modal id="owf-submission-preview" :state="previewState" hideable backdrop
+          @hide="previewState = false">
+          <template #title>ODK Web Forms Preview</template>
+          <template #body>
+            This is the preview of new ODK Web Forms, currently you can only view your forms with it.
+          </template>
+      </modal>
+</template>
+
+<script setup>
+import { ref, createApp, getCurrentInstance } from 'vue';
+import { OdkWebForm, webFormsPlugin } from '@getodk/web-forms';
+import useForm from '../../request-data/form';
+import { apiPaths } from '../../util/request';
+import { noop } from '../../util/util';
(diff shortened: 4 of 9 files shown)


src/assets/scss/app.scss
```diff
@@ -13,7 +13,6 @@ except according to the terms contained in the LICENSE file.
 @import './mixins';
 
 html {
-  background-color: $color-accent-secondary;
```
sadiqkhoja, line 16: make it conditional

src/components/form-version/standard-buttons.vue
```diff
@@ -14,6 +14,10 @@ except according to the terms contained in the LICENSE file.
 <template>
   <span class="form-version-standard-buttons">
     <enketo-preview :form-version="version"/>
+    <RouterLink class="btn btn-default btn-web-form" :to="formPath('preview')" target="_blank">
+      <span class="icon-eye"></span>
+      New Preview
```
sadiqkhoja, line 19: use i18n

src/components/form-version/standard-buttons.vue
```diff
@@ -14,6 +14,10 @@ except according to the terms contained in the LICENSE file.
 <template>
   <span class="form-version-standard-buttons">
     <enketo-preview :form-version="version"/>
+    <RouterLink class="btn btn-default btn-web-form" :to="formPath('preview')" target="_blank">
```
sadiqkhoja, line 17: replace it with anchor link

src/components/form/preview.vue
```diff
@@ -0,0 +1,80 @@
...
+        <OdkWebForm :form-xml="formVersionXml.data" @submit="handleSubmit"/>
+      </template>
+
+      <modal id="owf-submission-preview" :state="previewState" hideable backdrop
+          @hide="previewState = false">
+          <template #title>ODK Web Forms Preview</template>
+          <template #body>
+            This is the preview of new ODK Web Forms, currently you can only view your forms with it.
+          </template>
+      </modal>
```
sadiqkhoja, line 12: use i18n

src/components/form/preview.vue
```diff
@@ -0,0 +1,80 @@
...
+    : apiPaths.formDraftDef(projectId, xmlFormId, extension);
+};
+
+const fetchForm = () => {
+  const url = apiPaths.form(props.projectId, props.xmlFormId);
+  form.request({ url, extended: true })
+    .then(() => {
+      formVersionXml.request({ url: defPath('.xml') });
+    })
+    .catch(noop);
```
sadiqkhoja, line 63: verify error handling.

src/routes.js
```diff
@@ -514,6 +514,23 @@ const routes = [
...
     }
   }),
+  asyncRoute({
+    path: '/projects/:projectId([1-9]\\d*)/forms/:xmlFormId/preview',
+    component: 'FormPreview',
+    props: true,
+    loading: 'page',
+    meta: {
+      standalone: true,
+      validateData: {
```
sadiqkhoja, line 524: remove this.

src/routes.js
```diff
@@ -514,6 +514,23 @@ const routes = [
...
+        project: () => project.permits([
+          'form.read',
+          'form.update',
+          'form.delete'
+        ])
+      },
+      title: () => ['Form Preview', form.nameOrId]
```
sadiqkhoja, line 531: use i18n

src/composables/feature-flags.js
```diff
@@ -0,0 +1,60 @@
...
+
+  const keydownEventHandler = (e) => updateCheatKeys(e, true);
+  const keyupEventHandler = (e) => updateCheatKeys(e, false);
+
+  onMounted(() => {
+    // eslint-disable-next-line no-console
+    console.log(
+      '%c ODK Central Alpha Features: \n\n%c- Press and hold the %cW and F %ckeyboard keys on a screen with a form preview button to access the new %cWeb Forms%c preview.\n\n',
+      'background-color: #009ecc; font-size: 18px; color: white',
+      ';',
```
sadiqkhoja, line 42: remove the semicolon

src/composables/feature-flags.js
```diff
@@ -0,0 +1,60 @@
...
+      'font-weight: bold; font-size: 14px;',
+      '',
+      'font-weight: bold; font-size: 14px;',
+      '',
+    );
+    document.addEventListener('keydown', keydownEventHandler);
```
sadiqkhoja, line 48: use eventListner

src/composables/feature-flags.js
```diff
@@ -0,0 +1,60 @@
...
+    cheatKeys.f = false;
+    features.value['new-web-forms'] = false;
+  }
+
+  const keydownEventHandler = (e) => updateCheatKeys(e, true);
+  const keyupEventHandler = (e) => updateCheatKeys(e, false);
+
+  onMounted(() => {
+    // eslint-disable-next-line no-console
+    console.log(
```
sadiqkhoja, line 39: don't print this in test env

src/components/form/preview.vue
```diff
@@ -0,0 +1,107 @@
...
+
+const { form, formVersionXml } = useForm();
+
+const previewState = ref(false);
+
+const defPath = (extension) => {
+  const { projectId, xmlFormId, publishedAt } = form;
+
+  return publishedAt != null
+    ? apiPaths.formVersionDef(projectId, xmlFormId, form.version, extension)
```
matthew-white, line 61: If it helps avoid a request for `form`, you could add a new `apiPaths` method for the XML of the published form. That way, you wouldn't need to specify `form.version`.

src/components/form/preview.vue
matthew-white, line 1: Needs a file header.

src/composables/feature-flags.js
matthew-white, line 1: Needs a file header.

test/components/form/preview.spec.js
```diff
@@ -0,0 +1,56 @@
...
+  beforeEach(() => {
+    mockLogin();
+    // eslint-disable-next-line no-console
+    console.log = () => {};
+  });
+
+  afterEach(() => {
+    // eslint-disable-next-line no-console
+    console.log = originalConsoleLog;
+  });
```
matthew-white, line 19: Is mocking `console.log()` needed now that `useFeatureFlags()` is checking `process.env.NODE_ENV`?

test/components/form/preview.spec.js
```diff
@@ -0,0 +1,56 @@
...
+      hide: '.btn-primary'
+    });
+  });
+
+  it('does not show navbar', async () => {
+    testData.extendedForms.createPast(1, { xmlFormId: 'a' });
+
+    const app = await load('/projects/1/forms/a/preview').complete();
+
+    should(app.findComponent({ name: 'Navbar' })).not.be.null();
```
matthew-white, line 54: I'm pretty sure this will always not be `null`. I think it should be:

```js
app.findComponent({ name: 'Navbar' }).exists().should.be.false();
```

test/components/form/preview.spec.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  it('renders new Web Form', async () => {
+    testData.extendedForms.createPast(1, { xmlFormId: 'a' });
+
+    const app = await load('/projects/1/forms/a/preview').complete();
+
+    const webForm = app.findComponent('.odk-form');
+
+    webForm.should.not.be.null();
```
matthew-white, line 36:
```suggestion
    webForm.exists().should.be.false();
```

matthew-white (commented):
This is looking really good! My main question is about how to preview the draft of a published form. I also added some other small comments (I might add a few more later).

src/components/form/preview.vue
```diff
@@ -0,0 +1,107 @@
...
+    : apiPaths.formDraftDef(projectId, xmlFormId, extension);
```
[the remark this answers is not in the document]
lognaturel: I agree it should be possible to preview a published form using WF!

test/components/form/preview.spec.js
matthew-white: Ah right, it should be `.true()` instead of `.false()`. 😅

src/components/app.vue
```diff
@@ -36,21 +41,30 @@ import Navbar from './navbar.vue';
...
 export default {
   name: 'App',
   components: { Alert, Navbar, FeedbackButton: defineAsyncComponent(loadAsync('FeedbackButton')) },
+  directives: {
+    documentColor: {
+      mounted: () => {
+        document.documentElement.style.backgroundColor = 'var(--color-accent-secondary)';
+      }
+    }
+  },
```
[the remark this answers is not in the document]
matthew-white: Would you mind adding a quick test of this change as well? Maybe in src/router.js (in a `describe()` block about `standalone`), something like:

```js
it('adds a background color if standalone is false', async () => {
  const app = await load('/login').restoreSession(false);
  document.documentElement.style.backgroundColor.should.equal('var(--color-accent-secondary)');
});
```

matthew-white (approved):
LGTM 🚀

src/components/form/preview.vue
```diff
@@ -0,0 +1,112 @@
...
+});
+
+const { form, formVersionXml } = useForm();
+
+const previewModal = modalData();
+
+const fetchForm = () => {
+  form.request({ url: apiPaths.form(props.projectId, props.xmlFormId), extended: true })
+    .then(() => formVersionXml.request({ url: apiPaths.formXml(props.projectId, props.xmlFormId, props.draft) }))
+    .catch(noop);
```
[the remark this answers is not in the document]
sadiqkhoja: I have added `alert: false` for Form request so that we don't show error alert for it and made the use of Form name in the title conditional in routes.js

sadiqkhoja:
@matthew-white thanks for reviewing this, can please double check my last commit. It fixes a subtle bug where "New Preview" button is not clickable when focus is on any element on the page. Steps to reproduce:

- Press tab until your focus is on any button or link on the page such as "User" in the top navbar
- Press W+F
- Click the "New Preview" button

src/composables/feature-flags.js
```diff
@@ -38,10 +38,12 @@ export default function useFeatureFlags() {
     }
   }
 
-  function reset() {
-    cheatKeys.w = false;
-    cheatKeys.f = false;
-    features.value['new-web-forms'] = false;
+  function reset(event) {
+    if (event.target.classList.contains('btn-web-form')) {
```
matthew-white, line 42: What happens if you click on the icon in the button? I think the icon would be the target, so this will evaluate to `false` even though we do want to reset. How about:

```js
if (event.target.closest('.btn-web-form') != null) {
```

matthew-white (approved):
Ooh interesting, good catch!

src/composables/feature-flags.js
sadiqkhoja: When I click on the icon, the button gets the focus because `span` by default is not focusable.

matthew-white: Ah good point.

PR #1003, opened 2024-07-05
Show links to FormPreview in FormList

matthew-white:
As of #1001, pressing the W+F keys shows the "New Preview" button (which links to Web Forms) instead of the default Preview button (which links to Enketo). However, pressing W+F has no effect in the project overview: even though there is a Preview button for each form, "New Preview" is not shown. This PR changes that so that W+F has an effect wherever the Preview button is shown, including the project overview.

Trying it locally.

The underlying issue is that the "New Preview" logic appears in `FormVersionStandardButtons`, but `FormRow` renders `EnketoPreview` without `FormVersionStandardButtons`. `FormVersionStandardButtons` and `FormRow` are the only two components that render `EnketoPreview`. My change involves moving the "New Preview" logic from `FormVersionStandardButtons` to `EnketoPreview`. In other words, `EnketoPreview` will manage the form preview button for both Enketo and Web Forms. I'd be happy to rename the `EnketoPreview` component if that'd make things clearer.

### DIFF
--- a/src/components/enketo/preview.vue
+++ b/src/components/enketo/preview.vue
@@ -11,16 +11,22 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <a v-if="disabledDescription == null" class="enketo-preview btn btn-default"
-    :href="href" target="_blank">
+    :href="enketoPath" target="_blank">
     <span class="icon-eye"></span>{{ $t('action.showPreview') }}
   </a>
   <button v-else type="button" class="enketo-preview btn btn-default" aria-disabled="true"
     v-tooltip.aria-describedby="disabledDescription">
     <span class="icon-eye"></span>{{ $t('action.showPreview') }}
   </button>
+
+  <router-link class="btn btn-default btn-web-form" :to="webFormsPath"
+    target="_blank">
+    <span class="icon-eye"></span>{{ $t('action.newPreview') }}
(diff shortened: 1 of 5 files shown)


src/components/enketo/preview.vue
```diff
@@ -40,16 +50,41 @@ export default {
...
+    enketoPath() {
       // enketoId probably doesn't need to be encoded, but there is also little
       // harm.
       const encodedId = encodeURIComponent(this.formVersion.enketoId);
       return `${enketoBasePath}/preview/${encodedId}`;
+    },
+    webFormsPath() {
+      return this.formPath(
+        this.formVersion.projectId,
+        this.formVersion.xmlFormId,
```
matthew-white, line 62: On form pages, `formPath()` can determine the `xmlFormId` from the route. However, it can't do that on the project overview: it needs to be explicitly passed the `xmlFormId`.

matthew-white:
It looks like there's currently a partial outage with CircleCI, but testing and linting pass for me locally.

sadiqkhoja (approved):
LGTM.

What do you think of adding following in FormRow, so that width of the columns doesn't change on pressing W+F:

```
.actions:has(.enketo-preview){
    width: 125px;
  }
```

sadiqkhoja (approved):

PR #1007, opened 2024-07-11
Add translations for traditional Chinese

matthew-white:
This PR makes progress on getodk/central#680, adding translations for traditional Chinese. The only thing that should change in this PR is translations, not any logic.

We've modified the structure of the Transifex JSON since the last release (by adding/removing a string). Due to our process, that means that we also need to update translations for other languages. I've done that in a separate commit in this PR.


src/components/form/head.vue
```diff
@@ -264,7 +264,6 @@ $tab-li-margin-top: 5px;
       "tabTitle": "Tyto funkce budou k dispozici, jakmile zveřejníte svůj koncept formuláře"
     },
     "draftNav": {
-      "title": "Koncept",
```
matthew-white, line 267: This translation still exists, it was just moved to the locale files in src/locales/ (as part of #1001).

matthew-white:
@ktuite, I think this PR just needs a very quick check that it only updates translations.

ktuite (approved):

PR #1008, opened 2024-07-11
Don't italicize Japanese or Chinese

matthew-white:
For Japanese, we use bold instead of italics for `<em>` elements. [Apparently](https://aqworks.com/en/blog/2016/09/20/perfect-japanese-typography/), italics aren't used in Japanese. That was something I [mentioned](https://github.com/getodk/central-frontend/pull/468#issuecomment-847345225) when I added Japanese in the first place. However, as I'm adding Chinese just now, I'm realizing that there are elements outside `<em>` where we italicize using CSS. This PR makes a similar change as for `<em>` elements, making those elements bold instead of italic for Japanese.

In #1006, I made Chinese bold instead of italic, similar to Japanese. I've [read](https://drwhispers.com/2021/11/16/italics-in-chinese/) that italics also aren't used in Chinese. I'm not sure that bold is the very best option (the article suggests underdots in some cases), but it sounds like it's an improvement over italic. The article makes it clear that multiple options are better than italic. If we receive feedback from the Chinese translator or from users, I'd be happy to do something different.

I spot-checked a few of these changes.

I added a mixin for this, which I think is the most ergonomic.

### DIFF
--- a/src/components/dataset/summary/row.vue
+++ b/src/components/dataset/summary/row.vue
@@ -91,7 +91,6 @@ export default {
 </script>
 
 <style lang="scss">
-@import '../../../assets/scss/_variables.scss';
 @import '../../../assets/scss/mixins';
 
 .dataset-summary-row {
@@ -144,9 +143,7 @@ export default {
       hyphens: auto;
       overflow-wrap: break-word;
 
-      .no-properties {
-        font-style: italic;
-      }
+      .no-properties { @include italic; }
     }
 }
 
--- a/src/components/entity/conflict-table.vue
+++ b/src/components/entity/conflict-table.vue
@@ -247,7 +247,9 @@ defineExpose({ resize });
 
   .unchanged {
     color: #888;
+
     font-style: italic;
+    &:lang(ja), &:lang(zh) { font-style: normal; }
   }
 
   [class^="icon-"] {
--- a/src/components/multiselect.vue
+++ b/src/components/multiselect.vue
@@ -440,6 +440,12 @@ const emptyMessage = computed(() => (searchValue.value === ''
         color: #666;
         font-style: italic;
       }
+      &:lang(ja), &:lang(zh) {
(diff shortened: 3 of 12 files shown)


matthew-white:
I'll go ahead and add this PR to the project, since it's an existing issue with Japanese.

src/components/multiselect.vue
```diff
@@ -440,6 +440,12 @@ const emptyMessage = computed(() => (searchValue.value === ''
         color: #666;
         font-style: italic;
       }
+      &:lang(ja), &:lang(zh) {
+        &::placeholder {
```
matthew-white, line 444: Because of the use of `::placeholder` above, we can't use the mixin here. It looks like the selector has to come before `::placeholder`.

src/components/entity/conflict-table.vue
```diff
@@ -247,7 +247,9 @@ defineExpose({ resize });
 
   .unchanged {
     color: #888;
+
     font-style: italic;
+    &:lang(ja), &:lang(zh) { font-style: normal; }
```
matthew-white, line 252: This is one place where I felt unsure about using bold. In the entity conflict summary table, we italicize values that haven't changed. I'm not sure it makes sense to make them bold in Japanese and Chinese: I'm not sure it conveys the same thing. (If anything, maybe we should bold the values that *have* changed.) Instead, here I'm just removing the italics. The values will still look a little different from values that have changed, because they're also gray.

src/components/dataset/summary/row.vue
```diff
@@ -91,7 +91,6 @@ export default {
 </script>
 
 <style lang="scss">
-@import '../../../assets/scss/_variables.scss';
 @import '../../../assets/scss/mixins';
```
matthew-white, line 94: `mixins` automatically imports `variables`.

ktuite (approved):

PR #1009, opened 2024-07-17
Change Overview tab to Forms and add badge

matthew-white:
This PR makes progress on getodk/central#671, adding a badge to one more nav tab. We will rename the project "Overview" tab to "Forms", then show the form count in a badge on the tab.

Trying it out some locally, new tests.

One thing I wanted to make sure was the case was that the form count in the badge matches the number of forms in the forms table. I thought that might not be the case, because the `FormTable` component had logic to filter the form list. Specifically, the component wouldn't show draft forms to Data Collector users and would filter them out. However, it looks like that logic is no longer needed in Frontend. Backend seems to be doing that filtering now, which means that the form count in the badge should match the table without additional effort. Actually, it does look like there is a related Backend issue in this area (getodk/central#687), but we can fix that in Backend, and in general, I don't think we need to do anything extra in Frontend. We just show the `forms` property from the project response in the badge, and we also update that property if for some reason it differs from the form list.

Another thing this PR does is update the `forms` property when an action can be expected to have changed it. When a form is created, the property is incremented, and when a form is deleted, it is decremented.

### DIFF
--- a/src/components/form-draft/abandon.vue
+++ b/src/components/form-draft/abandon.vue
@@ -60,9 +60,9 @@ export default {
   setup() {
     // The component assumes that this data will exist when the component is
     // created.
-    const { form } = useRequestData();
+    const { project, form } = useRequestData();
     const { request, awaitingResponse } = useRequest();
-    return { form, request, awaitingResponse };
+    return { project, form, request, awaitingResponse };
   },
   computed: {
     title() {
@@ -80,9 +80,10 @@ export default {
           : apiPaths.form(this.form.projectId, this.form.xmlFormId)
       })
         .then(() => {
-          // project.forms and project.lastSubmission may now be out-of-date. If
-          // the user navigates to ProjectOverview, project.forms should be
-          // updated. project.lastSubmission is not used within ProjectShow.
+          if (this.form.publishedAt == null) this.project.forms -= 1;
+          // project.lastSubmission may now be out-of-date. However,
+          // project.lastSubmission is not used within ProjectShow.
+
           this.$emit('success');
         })
         .catch(noop);
--- a/src/components/form-draft/abandon.vue
+++ b/src/components/form-draft/abandon.vue
@@ -58,8 +58,6 @@ export default {
   },
   emits: ['hide', 'success'],
   setup() {
-    // The component assumes that this data will exist when the component is
-    // created.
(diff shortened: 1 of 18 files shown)


src/components/form-draft/abandon.vue
```diff
@@ -80,9 +80,10 @@ export default {
           : apiPaths.form(this.form.projectId, this.form.xmlFormId)
       })
         .then(() => {
-          // project.forms and project.lastSubmission may now be out-of-date. If
-          // the user navigates to ProjectOverview, project.forms should be
-          // updated. project.lastSubmission is not used within ProjectShow.
+          if (this.form.publishedAt == null) this.project.forms -= 1;
+          // project.lastSubmission may now be out-of-date. However,
+          // project.lastSubmission is not used within ProjectShow.
```
matthew-white, line 85: Remove this comment. A draft form's submissions shouldn't be included in `project.lastSubmission`.

src/components/form-draft/abandon.vue
```diff
@@ -80,9 +80,10 @@ export default {
           : apiPaths.form(this.form.projectId, this.form.xmlFormId)
       })
         .then(() => {
-          // project.forms and project.lastSubmission may now be out-of-date. If
-          // the user navigates to ProjectOverview, project.forms should be
-          // updated. project.lastSubmission is not used within ProjectShow.
+          if (this.form.publishedAt == null) this.project.forms -= 1;
```
matthew-white, line 83: Do the same thing in `FormDelete`.

test/components/form/new.spec.js
```diff
@@ -283,12 +283,16 @@ describe('FormNew', () => {
...
 
-    it('renders the correct number of rows in the forms table', () =>
+    it('increments the form count', () =>
       createForm()
         .complete()
-        .load('/projects/1', { project: false })
+        // Navigate to a page that does not request the form list. Once the form
+        // list is received, the form count will be updated to match it.
+        .load('/projects/1/settings', { project: false })
         .afterResponses(app => {
```
matthew-white, line 292: Navigate to the forms page, but use `beforeEachResponse()`.

matthew-white (commented):

ktuite (approved):

PR #1010, opened 2024-07-19
Upgrade web forms v0.2

sadiqkhoja:
- Upgraded `getodk/web-forms` to v0.2 which brings in a appearances for Select questions.
- Added a CSS fix for the stopping bootstrap messing up with the styles of new web form component. We could have added this fix in the code of `web-forms` but it was decided that for now it is best to add in Central and update the docs of `web-forms` to let other host applications to do that same.

Manually tested it to ensure various select appearances are rendered correctly.

Second commit in this PR reverts all the styling added by bootstrap and other css files for new Web Forms component. The approach felt cleanest to me. Other options were to dynamically removed the `<style>` tags when rending new Web Forms or when `meta.standalone` is true - that would also removed styles for dialog box that we show on "Send" button.

None

Not sure we want to update the documentation for this feature or just announce it on the Forum.

### DIFF
--- a/src/assets/css/namespaces.css
+++ b/src/assets/css/namespaces.css
@@ -1,0 +1,1 @@
+@namespace svg "http://www.w3.org/2000/svg";
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -899,3 +899,11 @@ becomes more complicated.
   transition: opacity 0.6s 0.15s;
 }
 [data-mark-rows-deleted="false"] { display: none; }
+
+
+////////////////////////////////////////////////////////////////////////////////
+// Don't let styles defined in Central bleed into new Web Forms Component
+
+.odk-form :not(svg|*) {
+  all: revert;
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@floating-ui/dom": "^1.1.0",
-        "@getodk/web-forms": "^0.1.1",
+        "@getodk/web-forms": "^0.2.0",
         "axios": "^1.6.2",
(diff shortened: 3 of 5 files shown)


src/assets/css/namespaces.css
```diff
@@ -0,0 +1 @@
+@namespace svg "http://www.w3.org/2000/svg";
```
matthew-white, line 1: Why is this needed, out of curiosity? For `:not(svg)` to work below?

src/assets/scss/app.scss
```diff
@@ -899,3 +899,11 @@ becomes more complicated.
   transition: opacity 0.6s 0.15s;
 }
 [data-mark-rows-deleted="false"] { display: none; }
+
+
+////////////////////////////////////////////////////////////////////////////////
+// Don't let styles defined in Central bleed into new Web Forms Component
+
+.odk-form :not(svg|*) {
```
matthew-white, line 907: I feel like it'd be better to move this to `FormPreview`, since that's where all the other Web Form setup lives. `FormPreview` already houses some other Web Form-related CSS.

src/assets/css/namespaces.css
sadiqkhoja: That's right, with `:not(svg|*)` I am excluding all elements of svg namespace. Without this exclusion, svg icons stop working.

src/assets/scss/app.scss
sadiqkhoja: I tried that but doesn't work because styles defined in the SFC are loaded after style coming from `web-form` component, in that case this rule reverts the `web-form` styles as well.

matthew-white: Ah very interesting, that makes sense. :+1:

PR #1011, opened 2024-07-23
Remove references to old presenter objects

matthew-white:
We used to use presenter objects, defining a class for each Backend resource (e.g., `Project`). We removed those when we introduced `requestData`, but they're still referenced in a couple of places:

- The contribution guide
- A group of i18n messages is named `presenter`.

This PR removes these references in order to avoid confusion.

### DIFF
--- a/src/locales/es.json
+++ b/src/locales/es.json
@@ -319,8 +319,8 @@
       }
     }
   },
-  "presenter": {
-    "Project": {
+  "requestData": {
+    "project": {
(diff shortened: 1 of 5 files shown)


src/locales/es.json
matthew-white, line 1: Normally, it wouldn't be necessary to modify this file. It is autogenerated by destructure.js and will be overwritten the next time we update translations. However, it's necessary to modify it here because these translations are used in a test.

ktuite (approved):

PR #1032, opened 2024-10-02
Companion PR to getodk/central-backend#1205

brontolosone:
Related: getodk/central-backend#1205

Companion to getodk/central-backend#1205, related to the authz verb renames therein.

Companion PR, which means that it makes no sense without its protagonist — yet if the protagonist gets merged, then this PR also needs to be merged, otherwise the frontend will break on the changed backend.

### DIFF
--- a/src/routes.js
+++ b/src/routes.js
@@ -280,7 +280,7 @@ const routes = [
         loading: 'tab',
         meta: {
           validateData: {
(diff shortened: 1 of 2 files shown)


matthew-white:
Closing, as we've closed getodk/central-backend#1205.

PR #1035, opened 2024-10-15
Procfile: run nginx via docker

### LINKED ISSUE #1033: npm run dev fails for -e arg
Seen on: 6fd5f284811872f445e757a8b5c4496acbcf18a7
Nginx: `nginx/1.18.0 (Ubuntu)`

# Observed

```sh
$ npm run dev

> odk-central-frontend@0.1.0 dev
> nf start

11:39:19 nginx.1   |  nginx: invalid option: "e"
[DONE] Killing all processes with signal  SIGINT
11:39:19 nginx.1   Exited with exit code null
```

# Expected

`npm run dev` should start nginx.

# Possible fix

Remove ` -e stderr` from nginx invocation in `Procfile`.

# Notes

Introduced in https://github.com/getodk/central-frontend/pull/1017.

alxndrsn:
This should help to standardise the nginx version running on different dev machines.

Closes #1033


nginx-conf/nginx.conf
```diff
@@ -11,15 +11,15 @@
...
 
 events {
 }
 http {
-  access_log ./.nginx/nginx-access.log;
-  client_body_temp_path ./.nginx/client_body_temp;
-  proxy_temp_path ./.nginx/proxy_temp 1 2;
+  access_log /dev/stdout;
+  client_body_temp_path /tmp/client_body_temp;
+  proxy_temp_path /tmp/proxy_temp 1 2;
```
alxndrsn, line 22: is this necessary?

alxndrsn:
> Reverting [4527b7b](https://github.com/getodk/central-frontend/commit/4527b7bb21236fc492f0544956e275d8261a5a90), which was a mostly cosmetic change (and for that reason a separate commit in PR #1017), should fix your woes, right?

That would work for now :+1:

I think a dockerised approach will be more robust long-term.  Getting local nginx working on non-OSX was work (https://github.com/getodk/central-frontend/pull/621); avoiding future problems would be really handy.

> Yet if there has to be a container, could we have it as an option via the Makefile?

What's the argument for allowing dev environments to easily deviate?

alxndrsn:
> as found in Ubuntu 20.04 LTS (from 2020, current version is 1.27, note that support for 20.04 ends in six months…)

I'm on 22.04 which is supported until 2027: https://packages.ubuntu.com/search?keywords=nginx&searchon=names&suite=jammy&section=all

matthew-white:
I agree that if we have a Docker option, we should also have a non-Docker option. Historically for Central, we've made it possible to run things locally without Docker. Not everyone uses a lot of Docker, and I like the idea of developers being able to get started on Central without digging into Docker. That said, I also know that Docker can be very convenient. `make run-docker-postgres` has been super useful in `central-backend`, and getodk/central#525 sounds like it could also make things easier.

> I think a dockerised approach will be more robust long-term.  Getting local nginx working on non-OSX was work (https://github.com/getodk/central-frontend/pull/621); avoiding future problems would be really handy.

I definitely want it to be the case that someone starting on Central is able to get nginx up and running with minimal effort ("install nginx with version X or above"). That said, if reverting 4527b7b is enough to get nginx running on our variously configured machines, maybe that shows that our nginx config is already battle-hardened.

---

On a separate note, I really like the idea of moving the nginx config files to their own directory and out of the repo's root directory. I've had to add some files to the root directory, so it's been feeling a little cluttered lately.

matthew-white:
> The rationale for this PR was the breakage that you experienced on nginx 1.18 as found in Ubuntu 20.04 LTS (from 2020, current version is 1.27, note that support for 20.04 ends in six months…). Fair enough! Let's aim to be broadly compatible and minimally invasive.

That sounds right to me that if 1.18 is LTS, we should try to support it. One option is that we revert 4527b7b now, but plan to add it back once `-e` is standard among versions of nginx that are LTS (in six months or whenever).

alxndrsn:
> One option is that we revert [4527b7b](https://github.com/getodk/central-frontend/commit/4527b7bb21236fc492f0544956e275d8261a5a90) now

I'll open a PR for this.

PR #1037, opened 2024-10-15
Fix dev:oidc npm script

### LINKED ISSUE #1034: `npm run dev:oidc` broken
It looks like `npm run dev:oidc` no longer serves frontend with OIDC enabled:

Screenshot_2024-10-15_12-19-35

Possibly affected by 9236c61b0ce6035fbaf53f257e5f228b52874a3a

alxndrsn:
* generate public/client-config.js at foreman startup
* remove nginx's special handling for client-config.js

Closes #1034

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ npm-debug.log*
 !/.nginx/.gitkeep
 
 /.eslintcache
+/public/client-config.json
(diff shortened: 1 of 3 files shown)


matthew-white:
I'm interested in what you think of #1038, but this approach also looks reasonable to me. It'd be nice to remove the addition of Docker though, since that's being discussed at #1035.

> remove nginx's special handling for client-config.js

We have special logic around this file in prod as well, so my feeling is that it's OK to have some special handling in this local nginx config. But I also don't have any attachment to it: I'd be OK generating client-config.json locally, as this PR does, if you think that's the best approach.

alxndrsn:
#1034 not recreatable, so this change is not required.

PR #1038, opened 2024-10-15
Serve local client-config.json if it exists

### LINKED ISSUE #1034: `npm run dev:oidc` broken

matthew-white:
Now that we request client-config.json, I've been wondering whether it'd be possible to remove `npm run dev:oidc`. That command sets a single option in the config (`oidcEnabled`), but there are several other options in the config that a developer might want to set. The idea of this PR is that a developer can create a local client-config.json to match whatever they're working on, and those options won't show up in version control. Modifying src/config.js would have much the same effect, but it has the downside that it shows up in version control.

Related: #1034

### DIFF
--- a/main.nginx.conf
+++ b/main.nginx.conf
@@ -95,7 +95,10 @@ http {
     }
 
     location = /client-config.json {
-      include ./common-headers.nginx.conf;
+      root .;
+      try_files $uri @empty_json;
+    }
+    location @empty_json {
       return 200 "{}";
     }
 
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ npm-debug.log*
 !/.nginx/.gitkeep
 
 /.eslintcache
+
+/client-config.json
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
   "private": true,
   "scripts": {
     "dev": "nf start",
-    "dev:oidc": "VUE_APP_OIDC_ENABLED=true npm run dev",
     "build": "vue-cli-service build",
     "lint": "eslint --max-warnings 0 --cache --ext .js,.vue src/ bin/ test/ *.js",
     "lint:fix": "eslint --max-warnings 0 --fix --cache --ext .js,.vue src/ bin/ test/ *.js",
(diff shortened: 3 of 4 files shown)


main.nginx.conf
```diff
@@ -95,7 +95,10 @@ http {
     }
 
     location = /client-config.json {
-      include ./common-headers.nginx.conf;
```
matthew-white, line 98: We don't call `add_header` in this `location`, so I don't think we need to include this file. Locally, I can see that the expected headers are coming through.

alxndrsn: Should this change be a separate PR?  It seems to be unrelated to the aim of this PR.

alxndrsn: I would keep this line, as it doesn't do any harm, and keeps this file closer to central's `files/nginx/odk.conf.template`, and therefore easier to compare.

We could go further in mirroring central's `files/nginx/odk.conf.template` by adding:

```nginx
add_header Cache-Control no-cache;
```

alxndrsn:
I like this approach because it's more consistent with how production behaves.

matthew-white:
I still think it's a nice idea to serve a custom client-config.json locally, as described in the PR description above. However, in practice, I find it easy enough just to modify src/config.js. It's uncommon that I even have to do that: for the vast majority of Frontend work, the default config works well.

Changes have been made to Frontend that mean that this PR is now out-of-date:

- `npm run dev:oidc` has been removed.
- So has `VUE_APP_OIDC_ENABLED`.

If you want to develop Frontend locally without using the `central`/ops repo, and you're working on OIDC functionality, the best bet is to set `oidcEnabled` to `true` in src/config.js.

I'll go ahead and close this PR for now, but we can always come back to this idea. I've mentioned the idea of a local client-config.json at https://github.com/getodk/central/issues/1601#issuecomment-4437015967.

PR #1040, opened 2024-10-16
New actions related to entity form migration

ktuite:
Frontend of https://github.com/getodk/central-backend/pull/1210 (issue https://github.com/getodk/central/issues/692)

Example 1:
Screenshot 2024-10-16 at 1 56 38 PM

Example 2:
Screenshot 2024-10-16 at 1 56 48 PM

### DIFF
--- a/src/locales/en.json5
+++ b/src/locales/en.json5
@@ -160,7 +160,7 @@
       "upgrade": {
         "process_form": "Process Form",
         "process_form_draft": "Process Form Draft",
-        "process_form_entities_version": "Flag Entity Form for Upgrade"
+        "process_form_entities_version": "Flag Form for Upgrade"
       },
       // This is shown in the log of actions performed on the server. It is a
       // type of action that can be taken on or for a Web User.
--- a/src/locales/en.json5
+++ b/src/locales/en.json5
@@ -126,6 +126,7 @@
         "create": "@:audit.action.default.create",
         "update": "@:audit.action.default.update",
         "update_draft_set": "Create or Update Draft",
+        "update_draft_replace": "Replace Draft",
         "update_publish": "Publish Draft",
         "update_draft_delete": "Abandon Draft",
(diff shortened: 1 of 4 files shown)


src/locales/en.json5
```diff
@@ -158,7 +159,8 @@
       // type of action that the server performs when it is upgraded.
       "upgrade": {
         "process_form": "Process Form",
-        "process_form_draft": "Process Form Draft"
+        "process_form_draft": "Process Form Draft",
+        "process_form_entities_version": "Flag Entity Form for Upgrade"
```
matthew-white, line 163: My initial thought here was that maybe it'd be fine to present this action in the same way as the other, older form upgrade: "Process Form". It'd be one less thing to translate. Then again, I think "flag" is a better descriptor of what's actually happening. These audit events aren't making any immediate changes, they're just marking forms that need to be further evaluated and processed. That processing might not even change the form, e.g., if the form is already using the latest entity spec.

What do you think about something in the middle: "Flag Form for Upgrade"? I think if we keep it slightly generic, there's a chance we'll be able to reuse the string in the future. Also, the details on the event provide useful context, and users will be able to see those.

src/locales/en.json5
```diff
@@ -126,6 +126,7 @@
         "create": "@:audit.action.default.create",
         "update": "@:audit.action.default.update",
         "update_draft_set": "Create or Update Draft",
+        "update_draft_replace": "Replace Draft",
```
matthew-white, line 129: I think it'd be good to add this action to the filter in `AuditFiltersAction`. We can't add the upgrade event as easily, since Backend doesn't know how to filter for `upgrade.*` events as a category. However, since it's easy to add this event, and it will be returned as part of the general "Form Actions" category (/v1/audits?action=form), I think we should add `form.update.draft.replace` to the filter. No test needed, just adding it to the list.

matthew-white (approved):

PR #1048, opened 2024-11-03
Dockerize ODK CF to streamline development environment setup.

moise10r:

To verify that the changes work as intended, I have:

 - I Run the application locally using Docker with `docker-compose up --build`, ensuring that we still access the app through Ngnix on port 8989 .
- I also Executed `npm run dev` to confirm that the development server started without issues.
- I completed the linting process using npm run lint and confirmed that there are no linting errors.

It encapsulates the development environment within Docker, eliminating the need to install dependencies on the host machine. This setup simplifies onboarding for new developers and ensures a consistent development experience across different setups.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM node:20-alpine AS build
+WORKDIR /app
+
+COPY package.json ./
+COPY package-lock.json ./
+COPY vue.config.js ./
+
+COPY *.json ./
+
+RUN npm install
+
+COPY ./src ./src
+COPY ./public ./public 
+COPY ./bin ./bin 
+COPY ./docs ./docs
+COPY ./.tx ./.tx
+COPY ./transifex ./transifex
+
+RUN npm run build
+
+FROM nginx:alpine
+RUN mkdir -p /var/tmp/nginx/client_body_temp /var/tmp/nginx/proxy_temp
+COPY ./dev.nginx.conf /etc/nginx/nginx.conf
+COPY ./common-headers.nginx.conf /etc/nginx/conf.d/common-headers.nginx.conf
+COPY --from=build /app/dist /usr/share/nginx/html 
+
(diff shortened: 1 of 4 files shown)


lognaturel:
Thank you for this, @moise10r! @sadiqkhoja put together a dev docker-based stack at https://github.com/getodk/central/pull/525 and has been using it. We're hoping to get it merged soon. Would that meet your needs if we link to it clearly from the readme here and for the backend?

lognaturel:
I now realize I had forgotten the goals and scope of https://github.com/getodk/central/pull/525. Its focus is setting up all of the services other than Central frontend and backend to lower the bar to local development on those two code bases. It looks like what you want to do is run the frontend in a container while developing on it.

> eliminating the need to install dependencies on the host machine.

Did you have an issue with those? The dependencies needed for frontend are minimal, we use volta for version pinning, and it hasn't been a major pain point. We recently had some issues with different versions of nginx but these should now be resolved. If you did run into specific issues getting set up, it would be helpful to know what they were.

moise10r:
Thank you @lognaturel for the feedback, Actually when setting up the Central frontend repo, I was asked to install Nginx on my host machine, and I realized that this approach could lead to each developer potentially running different versions of Nginx. That’s why I was considering running the frontend in a container for development to ensure consistency. For the backend repo, I ran into a version mismatch with Node. After @sadiqkhoja  mentioned that Volta is used for version management, I thought it might be more efficient to have a single Docker Compose file for the backend as well. This would automate the database configuration, bringing it up with the necessary steps, and then starting the server with a dependency on the `database_service`.

lognaturel:
Thanks again for sending this in and sparking some good conversation around how we want to approach development. For now, we've decided we'd rather limit the amount of infrastructure to maintain and focus on doing host-based development. We can revisit if we get more feedback that this is a barrier.

Issue #1044, opened 2024-10-29
Remove navigator.locks from user preferences request


matthew-white:
As of #1024, Frontend sends requests to update user preferences from the `propagate()` method of the `UserPreferences` class. Currently, each request is wrapped in `navigator.locks.request()`, but after discussion, @brontolosone and I think that isn't strictly necessary. This issue is to update `propagate()`, removing `navigator.locks.request()`. I have an idea about how to do so, so I'll start by giving that a try, then sharing my idea with @brontolosone once it's ready.

Issue #787, opened 2023-05-09
transformResponse() cannot return reactive() object in development


matthew-white:
When it receives a response from Backend, a `requestData` resource by default simply stores the response data. However, the resource can first transform the data by specifying a `transformResponse()` function. `transformResponse()` can be used to make the data fully or partly reactive. By default, the data is not reactive.

All that works well in production. However, we use Pinia in limited ways in development, and things don't always end up working. If `transformResponse()` returns a nonreactive object, then the Pinia store will store that. If `transformResponse()` returns a `shallowReactive()` object, then Pinia will correctly store that. However, if `transformResponse()` returns a fully `reactive()` object, then getting the data from the store returns a nonreactive object (!).

I'm not quite sure why that is. I suspect it has something to do with the fact that the initial state passed to the Pinia store is `shallowReactive()`, not `reactive()`. Just a guess, but it's possible that Pinia doesn't offer a lot of support for `shallowReactive()` objects.

Again, everything works as expected in production (without Pinia). This hasn't come up before in development because most data isn't reactive, and the data that is is usually `shallowReactive()`. However, I'm now making the `entity` resource fully reactive and have been running into issues. For now, I'll turn off Pinia in development, matching what happens in production.

matthew-white:
Closing now that #1050 has been merged.

PR #1050, opened 2024-11-08
Don't use Pinia for requestData even in development

matthew-white:
This PR removes Pinia from `requestData` entirely. `requestData` was designed to use Pinia in development (never production). However, we haven't been using Pinia for a while due to #787. That means that this PR doesn't have a huge practical effect. However, it simplifies `requestData` and opens the door to some extensions to it. For example, for #1044, I want to be able to create an arbitrary number of `requestData` resources, yet I'm not sure that that's something that would play well with our Pinia setup.

I think it's helpful to remember why we're using Pinia in the first place. We don't need Pinia for reactivity, because we get that out of the box with Vue 3. Instead, the main value-add of Pinia is logging. Specifically, Pinia provides a nice interface within Vue DevTools to see what changes have been made to `requestData` and which components specifically made those changes.

When I created `requestData`, it felt important for it to offer robust logging. `requestData` was a replacement for Vuex, which had its own mechanism for logging changes. At first, I started out writing my own logging mechanism. However, once I realized that Pinia integrated with Vue DevTools, using Pinia seemed like a reasonably straightforward and feature-rich way to incorporate logging into `requestData`.

It's been a while since then, and it's now clear that Pinia's logging hasn't proven to be especially useful. I expected there to be more issues with `requestData`, but thankfully, they just haven't come up that often. Given the infrequency of issues, it's uncommon that we need to troubleshoot `requestData`. When we do, `console.log()` is usually enough to do the job.

So it doesn't seem like Pinia is particularly useful. Yet at the same time, it also comes with costs:

- The additional maintenance around one more Vue library. [#787](https://github.com/getodk/central-frontend/issues/787) is an example of that.
- The `patch()` pattern that's removed in this PR exists only for the sake of logging. It helps to group together multiple changes as a single event within Pinia logging. However, it requires explanation, especially for newcomers to the codebase.
- Pinia requires us to jump through some hoops in order for it to track which component is using which resource. The result is that the core logic underlying `requestData` is harder to follow.
- Local resources aren't strictly isolated from one other, running counter to the whole concept of _local_ resources. Two local resources with the same name would end up using the same Pinia store, which doesn't work. We were able to work around this issue, because it was never the case that different components on the same page would create local resources with the same name. However, it required some subtle management of `requestData` resources. The `AsyncRoute` component is an example of that that you can see in this PR.

Tests continue to pass.

### DIFF
--- a/src/components/async-route.vue
+++ b/src/components/async-route.vue
@@ -70,15 +70,10 @@ export default {
       // The main use of this.$attrs is to pass along event listeners to the
       // component.
       return { ...this.props, ...this.$attrs, key: this.k };
-    },
-    componentNameAndKey() {
-      return [this.componentName, this.k];
     }
   },
   watch: {
-    componentNameAndKey([newComponentName], [oldComponentName]) {
-      if (newComponentName !== oldComponentName) this.load();
-    }
+    componentName: 'load'
   },
   created() {
     this.load();
--- a/src/components/async-route.vue
+++ b/src/components/async-route.vue
@@ -12,12 +12,12 @@ except according to the terms contained in the LICENSE file.
 <template>
   <div v-if="loading === 'tab'">
     <loading :state="showsLoading"/>
-    <component :is="component" v-if="component != null" v-bind="propsAndAttrs"/>
+    <component :is="component" v-if="component != null" v-bind="bindings"/>
   </div>
   <page-body v-else-if="showsLoading">
     <loading :state="true"/>
   </page-body>
-  <component :is="component" v-else-if="component != null" v-bind="propsAndAttrs"/>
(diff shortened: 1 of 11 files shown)


src/components/async-route.vue
```diff
@@ -69,40 +69,15 @@ export default {
     propsAndAttrs() {
       // The main use of this.$attrs is to pass along event listeners to the
       // component.
-      return { ...this.props, ...this.$attrs };
+      return { ...this.props, ...this.$attrs, key: this.k };
```
matthew-white, line 72: Rename `propsAndAttrs`? Or pass `this.k` outside of `propsAndAttrs`?

matthew-white (commented):

ktuite (approved):

PR #1051, opened 2024-11-14
Don't require local resources and getters to have globally unique names

matthew-white:
This PR continues the work of #1050 in trying to simplify `requestData`. #1050 removed Pinia, making it such that different local resources are more independent of one another. Under Pinia, two local resources with the same name (created at different times) would share the same Pinia store. We've now removed these shared Pinia stores, but there is still a constraint on local resources: it's not possible for two local resources created at the same time to have the same name, even if they are created in different components. In other words, resource names must be globally unique, even for local resources.

My goal in this PR is to make local resources and getters as similar to `provide`/`inject` as possible. That's already the mechanism that they're using under the hood. `provide`/`inject` is happy to shadow an existing value with the same name. Now `requestData` will be fine if a local resource or getter shadows one from an ancestor component. It's not like we have a pressing need for shadowing, but by allowing it, we can simplify `requestData`. For example, we can consolidate all cleanup in `onUnmounted()` rather than using `onUnmounted()` for some things and `onBeforeUnmount()` for other things. I also think that allowing shadowing clarifies the mental model around local resources: local resources should be fully independent of one another.

### DIFF
--- a/src/request-data/index.js
+++ b/src/request-data/index.js
@@ -67,8 +67,9 @@ export const createRequestData = (container) => {
     return resource;
   };
   // Normally, requestData only provides access to app-wide resources. However,
-  // we allow tests to access local resources in order to make assertions about
-  // them.
+  // in testing, we sometimes need access outside components to local resources.
+  // For example, tests sometimes need to access local resources in order to
+  // make assertions about them.
   if (process.env.NODE_ENV === 'test')
     requestData.localResources = localResources;
 
@@ -78,6 +79,7 @@ export const createRequestData = (container) => {
   const getters = Object.create(null);
   const createGetter = (name, ref) => {
     provide(`requestData.getters.${name}`, ref);
+    getters[name] = ref;
     onUnmounted(() => { if (getters[name] === ref) delete getters[name]; });
     return ref;
   };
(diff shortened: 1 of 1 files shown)


src/request-data/index.js
```diff
@@ -58,61 +58,36 @@ export const createRequestData = (container) => {
...
+    requestData.localResources = localResources;
 
   // Getters allow you to create a computed ref in one component, then access it
   // from useRequestData() in a descendant component. Getters are only really
   // needed for computed refs that reference multiple resources.
-  const getters = new Map();
+  const getters = Object.create(null);
   const createGetter = (name, ref) => {
-    if (getters.has(name)) throw new Error(`the name ${name} is in use`);
-    getters.set(name, ref);
```
matthew-white, line 100:
```js
getters[name] = ref;
```

ktuite (approved):

src/request-data/index.js
matthew-white, line 1: Do we even need getters anymore? They seem rarely used. We could probably just use `provide`/`inject` instead. Let's see about removing them in a follow-up PR.

matthew-white (commented):
Notes from interactive review

PR #1057, opened 2024-11-19
Prevent submission row actions from wrapping in Indonesian

matthew-white:
We've added a new submission row action for deletion, which means that there are several buttons now. One of the buttons (the More button) contains text, so the width of the `.btn-group` is variable depending on the locale. That's an issue with Indonesian, where you can see that the row actions are wrapping:

This PR increases the width of the column for Indonesian only. Indonesian needs the extra width, but I didn't want to widen the column unnecessarily for all users.

This is how it looks after I increase the column width:

I viewed the change locally.

I also checked all locales other than Indonesian and didn't see a problem.

This came up previously for entity row actions. I'm basing this PR off the solution for that:

https://github.com/getodk/central-frontend/blob/acd5cf674f3460335438f428ed280996676f4823/src/components/entity/metadata-row.vue#L111-L113

### DIFF
--- a/src/components/submission/metadata-row.vue
+++ b/src/components/submission/metadata-row.vue
@@ -159,7 +159,11 @@ export default {
     max-width: 250px;
   }
 
-  .state-and-actions { min-width: 205px; }
+  .state-and-actions {
+    // Ensure that the column is wide enough that the .btn-group does not wrap.
+    min-width: 205px;
+    &:lang(id) { min-width: 221px; }
+  }
   .col-content {
     align-items: flex-start;
     display: flex;


src/components/submission/metadata-row.vue
```diff
@@ -159,7 +159,11 @@ export default {
     max-width: 250px;
   }
 
-  .state-and-actions { min-width: 205px; }
+  .state-and-actions {
+    // Ensure that the column is wide enough that the .btn-group does not wrap.
+    min-width: 205px;
+    &:lang(id) { min-width: 221px; }
```
matthew-white, line 165: On my machine, 208px was sufficient to prevent wrapping. However, I wanted to add a little wiggle room in case the users' fonts are a little different from mine. I also wanted the user to be able to see at least a fraction of the column content outside the buttons. In the screenshot, you can see that part of the review state icon is still visible to the left of the buttons. I didn't want to make the column overly wide though, so it was a balancing act.

ktuite (approved):

PR #1058, opened 2024-11-19
Changing QR text to 'Temporary Testing Code'

ktuite:
Follow up to https://github.com/getodk/central-frontend/pull/1053 and https://github.com/getodk/central/issues/746

### DIFF
--- a/src/components/collect-qr.vue
+++ b/src/components/collect-qr.vue
@@ -66,7 +66,7 @@ onMounted(() => {
   // Compute image size
   // blocks (based on length of encoded data) * cell size + (2 * margin)
   const width = code.getModuleCount() * props.cellSize + margin * 2;
-  const height = props.draft ? width + 20 : width;
+  const height = props.draft ? width + 16 : width;
   qrCanvas.value.width = width;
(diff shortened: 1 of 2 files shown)


matthew-white (approved):

src/components/collect-qr.vue
```diff
@@ -103,7 +103,7 @@ onMounted(() => {
     ctx.fillText(icon, width * 0.5, width * 0.5, width * 0.2);
 
     // Write text at the bottom
-    ctx.font = '20px Monaco, Menlo, Consolas, "Courier New", monospace';
+    ctx.font = '16px Monaco, Menlo, Consolas, "Courier New", monospace';
```
matthew-white, line 106: Since the 16 appears in two places, how about we make it a `const`?

matthew-white: I wanted to get the PR merged to get it on staging, so I went ahead and made this change. Hope that's OK! Would you mind taking a look at my commit?

ktuite: Looks good!

PR #1066, opened 2024-11-26
Make feedback button more visible

### LINKED ISSUE #673: Replace setChecked() with setValue()
We call `setChecked()` on Vue Test Utils (VTU) wrappers in a few places, but apparently that's been deprecated in favor of `setValue()`. I didn't make that change when migrating to Vue 3 because as far as I can tell, that change isn't documented in the VTU migration guide. But I'm noticing that `setChecked()` isn't mentioned anywhere in the VTU docs for Vue 3, and the VTU repo indicates that `setChecked()` isn't supported for Vue 3. It's somehow still working for us, but when we get the chance, I think we should swap it out for `setValue()`.

lognaturel:
Closes #673

Ran it locally.

I ended up with 250px from the bottom. That feels like a good balance of more eye catching but not too disruptive, especially in languages where the button text is longer.

@alyblenkin please take a quick look and if you're not loving it, let's adjust on a call next time we chat!

| 500px FR | 250px FR | 250px EN |
|-----------|----------|------------|
Screenshot 2024-11-25 at 4 25 37 PM |  Screenshot 2024-11-25 at 4 26 08 PM |  Screenshot 2024-11-25 at 4 26 25 PM

CSS-only changes

No.

### DIFF
--- a/src/components/feedback-button.vue
+++ b/src/components/feedback-button.vue
@@ -46,10 +46,10 @@ const surveyLink = computed(() => {
   position: fixed;
   z-index: 99999;
   right: 0;
-  bottom: 100px;
+  bottom: 250px;
   height: 4em;
   border: 2px solid;
-  border-color: #e3e4e4;
+  border-color: $color-accent-primary;
   border-top-left-radius: 5px;
   border-top-right-radius: 5px;
   text-align: center;


sadiqkhoja (approved):
Looks good. 

If we want we can make it middle of the screen with `calc(50vh - 2em)`

lognaturel:
> we can make it middle of the screen with calc(50vh - 2em)

That's a good option to keep in mind! From my playing around I feel like it's both more noticeable and less "in the way" when it's offset from center. Of course, this could be completely subjective, I'm not sure!

Let's try this on staging and locally for a couple days and adjust if it feels like we should try middle.

PR #1071, opened 2024-12-04
Add component to show submission review state

matthew-white:
Different components show a submission review state. To do so, they use the `reviewStateIcon()` function of the `useReviewState()` composable along with the `reviewState.*` messages in en.json5. However, one piece of functionality that isn't shared is the review state color: each component that shows a review state manages the color in its own way. This PR changes that by introducing a new component named `SubmissionReviewState`. It will show the icon and message associated with the review state, and it will also add the review state color.

I'm making this change with getodk/central#670 in mind. The hover card for a submission will show the submission's review state, and I didn't want to copy the review state colors to yet another component.

I updated tests. I also checked locally in the browser most of the components where I introduced `SubmissionReviewState`. If any issues are found during regression testing, we can patch those.

I think a component is exactly the right thing for the job, and I'm not sure why I didn't add one sooner. I may have been avoiding adding a component with `SubmissionMetadataRow` in mind. I've been trying to make that component as fast as possible to render, since there can be many submission rows on the page, and I've observed in the past that child components carry a performance penalty. However, in v2025.1, we'll be adding pagination to the submissions table, which should alleviate performance concerns. I've noticed before that one surprising performance bottleneck in components is Vue I18n. However, the new component doesn't have its own i18n messages — it only uses global messages from en.json5 — which should help.


matthew-white (commented):
Notes from interactive review

matthew-white:
I've pushed a commit with some changes. The biggest change is that I've removed some additional tests, similar to changes to other tests that @ktuite and I reviewed. I don't think we necessarily need another round of code review for this commit.

PR #1075, opened 2024-12-05
Feature #769: Show error on duplicate property name (case insensitive)

sadiqkhoja:
Closes getodk/central#769

Added tests and manual verification.

Uses existing approach of using `problemToAlert` to capture and translate the error.
Passing list of property names with bullet points to the `alert` method doesn't look super clean to me but does the job. I am open to ideas to improve that.

None.

https://github.com/getodk/docs/issues/1855

### DIFF
--- a/src/components/form/new.vue
+++ b/src/components/form/new.vue
@@ -224,9 +224,6 @@ export default {
               actual: details.value
             });
           }
-          if (code === 409.17) {
-            return `${this.$tcn('util.request.problem.409_17', details.duplicateProperties.length)}\n\n${details.duplicateProperties.map(p => `• ${p.provided} (${this.$t('common.existing')}: ${p.current})`).join('\n')}`;
-          }
           return null;
         }
       })
--- a/src/components/form/new.vue
+++ b/src/components/form/new.vue
@@ -224,6 +224,9 @@ export default {
               actual: details.value
             });
           }
+          if (code === 409.17) {
+            return `${this.$tcn('util.request.problem.409_17', details.duplicateProperties.length)}\n\n${details.duplicateProperties.map(p => `• ${p.provided} (${this.$t('common.existing')}: ${p.current})`).join('\n')}`;
+          }
           return null;
         }
       })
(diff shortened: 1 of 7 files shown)


src/locales/en.json5
```diff
@@ -436,7 +436,9 @@
     // {version} is a version number.
     "versionShort": "v{version}",
     // This text is shown on its own before details about a warning.
-    "warning": "Warning"
+    "warning": "Warning",
+    // Part of the error message format that lists the duplicate property name like "<PropertyName> (existing: <ExistingPropertyName>)"
+    "existing": "existing"
```
matthew-white, line 441: Some languages will use different punctuation from `(` and `)` and/or won't include a space character. Instead of translating "existing" specifically, I think we should translate the entire string, something like:

```js
"{provided} (existing: {current})"
```

If it includes the entire string, I think we should also change the message key. What do you think about `entity.existingProperty` or `entity.duplicateProperty`? Or maybe even better, you could change `util.request.problem.409_17` from a single message to a group of messages that includes this one.

By the way, I think it makes sense not to include the bullet point in the string.

src/components/form/new.vue
```diff
@@ -224,6 +224,9 @@ export default {
               actual: details.value
             });
           }
+          if (code === 409.17) {
+            return `${this.$tcn('util.request.problem.409_17', details.duplicateProperties.length)}\n\n${details.duplicateProperties.map(p => `• ${p.provided} (${this.$t('common.existing')}: ${p.current})`).join('\n')}`;
```
matthew-white, line 228: It looks like this is the same in `FormNew` and `FormPublish`. If a 409.17 will always be handled in the same way, I think we can go ahead and move that logic to a centralized place in `requestAlertMessage()`. We do something similar for 404.1 Problems currently.

matthew-white: In addition to `FormNew` and `FormPublish`, isn't this Problem a possibility for dataset/overview/new-property.vue? That could be another reason to centralize it.

src/components/submission/feed-entry.vue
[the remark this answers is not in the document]
matthew-white: Maybe a rebase would be useful? I'm now seeing that there's a merge conflict.

matthew-white (approved):

PR #1082, opened 2024-12-07
Move file for DatasetPropertyNew component

matthew-white:
I was thinking about the `DatasetPropertyNew` component while looking at #1075 and noticed that the component file wasn't where I expected it to be. Based on its name, I'd expect its filename to be dataset/property/new.vue, but instead it's dataset/overview/new-property.vue. We're planning to remove the entity list overview page, so I think it's good timing to move this component out of the overview/ directory.

Tests continue to pass.

Another way to make the component name and the filename consistent would have been to change the component name to match the filename: `DatasetOverviewNewProperty`. However, I prefer the current component name of `DatasetPropertyNew`: `*New` at the end matches other components, and we're soon going to be removing the `DatasetOverview` component.


src/components/dataset/property/new.vue
```diff
@@ -93,6 +93,7 @@ const submit = () => {
 
 <i18n lang="json5">
 {
+  // @transifexKey component.DatasetOverviewNewProperty
```
matthew-white, line 96: Mostly filenames don't really matter, but the one place where they do have an effect is in Transifex. For Transifex, the string key is based on the filename, under the assumption that the filename will match the component name. (I couldn't think of a better way for our Transifex scripts to figure out the component name.) We don't want these messages to be retranslated, which is why I'm specifying `@transifexKey` here.

ktuite (approved):

matthew-white:
> Another way to make the component name and the filename consistent would have been to change the component name to match the filename: `DatasetOverviewNewProperty`. &hellip; we're soon going to be removing the `DatasetOverview` component.

I ended up deciding not to remove that component: see #1131. Instead, there's a comment in the component file that explains that the component has the name that it does for historical reasons. However, I'd be happy for the entity property modal not to have `DatasetOverview` in its name. Since it's a legacy name, I think it's nice for it to be appear in as few places as possible.

I'm realizing now that instead of moving the file to src/components/dataset/property/new.vue, I could have moved it to src/components/dataset-property/new.vue. We do something similar for other subresources, for example, `FormAttachmentList` is at form-attachment/list.vue, not form/attachment/list.vue. However, I slightly prefer dataset/property/new.vue over dataset-property/new.vue:

- new.vue would be the only file in dataset-property/. There are other files related to entity properties, e.g., `DatasetOverview`, and those are all in dataset/. I think it's better for related files to be closer to each other.
- Form attachments are a subresource that you can request over the REST API: they're associated with their own set of endpoints. However, that's not the case for entity properties. In the database, entity properties have their own table, but in the API, they don't really have their own set of endpoints. To request the list of entity properties, you send a request for the entity list: there's no separate endpoint for entity properties. (Though there is a separate POST endpoint to _create_ an entity property.)

I don't think it really matters that much, and we can always change this later. I just wanted to leave some notes about my reasoning.

PR #1085, opened 2024-12-08
Add script to list messages that may be duplicates

matthew-white:
I've wanted to search the codebase for duplicate i18n messages for a while, but it felt slightly complicated. I was randomly thinking about it and realized that it wouldn't be that hard. This PR adds a script along those lines, listing i18n messages that look like duplicates.

When I ran the script, it flagged a number of false positives. However, it also successfully identified several duplicate messages: getodk/central#828.

The script doesn't list duplicates that it considers to be likely false positives. Maybe it'd be nice at some point to add a flag to list all duplicates that the script sees.

### DIFF
--- a/bin/transifex/list-duplicates.js
+++ b/bin/transifex/list-duplicates.js
@@ -1,0 +1,55 @@
+const { readFileSync } = require('node:fs');
+
+const { containsSubarray } = require('../util/util');
+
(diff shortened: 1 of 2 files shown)


matthew-white:
@ktuite, I've tagged you for review, but I think we can wait to review this until after the release goes out.

ktuite (approved):

PR #1087, opened 2024-12-09
Update text of tooltip for submission attachment shown in table

matthew-white:
Closes getodk/central#706.

Tests continue to pass.

### DIFF
--- a/src/locales/en.json5
+++ b/src/locales/en.json5
@@ -355,7 +355,7 @@
     },
     // This is shown when the Edit button is disabled.
     "editDisabled": "You cannot edit encrypted Submissions.",
-    "binaryLinkTitle": "File was submitted. Click to download.",
+    "binaryLinkTitle": "Click to download.",
     "encryptionMessage": "Data preview is not available due to encryption.",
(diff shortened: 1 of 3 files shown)


src/locales/en.json5
```diff
@@ -355,7 +355,7 @@
     },
     // This is shown when the Edit button is disabled.
     "editDisabled": "You cannot edit encrypted Submissions.",
-    "binaryLinkTitle": "File was submitted. Click to download.",
+    "binaryLinkTitle": "Click to download.",
```
matthew-white, line 358: I think we could also consider these options:

- Click to download file.
- Click to download Submission Attachment.
- Click to download Attachment.

I think it's sort of nice to say what's being downloaded. Though "file"/"Attachment" is still pretty ambiguous.

matthew-white:
I've updated the text as discussed in the issue. I also just commented on the issue, mentioning an [alternative approach](https://github.com/getodk/central/issues/706#issuecomment-2526604501) to consider.

lognaturel:
The alternative approach sounds great, let's do that. We can decide then whether we want to change the text or not.

Issue #774, opened 2023-04-09
Update Entity data directly


matthew-white:
See the release criteria [here](https://docs.google.com/document/d/1tOBNJaWZSJjkZjTYxV7vYBOFSkiZWKsnp48uWb_L5Co/edit#heading=h.970ohswy1cuc).

srujner:
**Tested with Success!**

dbemke:
Tested with Success!

PR #1094, opened 2024-12-10
Feature #774: Show message if Central version is too old

sadiqkhoja:
Closes getodk/central#774

### DIFF
--- a/src/components/home/news.vue
+++ b/src/components/home/news.vue
@@ -15,17 +15,29 @@ except according to the terms contained in the LICENSE file.
       <span>{{ $t('title') }}</span>
     </template>
     <template #body>
-      <iframe src="https://getodk.github.io/central/news.html" :title="$t('title')"></iframe>
+      <iframe :src="iframeSrc" :title="$t('title')"></iframe>
     </template>
   </page-section>
 </template>
 
 <script setup>
+import { computed, inject } from 'vue';
+import { useRequestData } from '../../request-data';
 import PageSection from '../page/section.vue';
 
 defineOptions({
   name: 'HomeNews'
 });
+
+const { centralVersion } = useRequestData();
+
+
+const container = inject('container');
+const { i18n: globalI18n } = container;
+const languageSubtag = computed(() => new Intl.Locale(globalI18n.locale).language);
+
+const iframeSrc = computed(() => `https://getodk.github.io/central/news.html?version=${centralVersion.data?.currentVersion}&lang=${languageSubtag.value}`);
+
 </script>
 
 <style lang="scss">
--- a/src/components/outdated-version-banner.vue
+++ b/src/components/outdated-version-banner.vue
@@ -1,0 +1,114 @@
+<!--
+Copyright 2024 ODK Central Developers
+See the NOTICE file at the top-level directory of this distribution and at
+https://github.com/getodk/central-frontend/blob/master/NOTICE.
+
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+
+<template>
+  <div v-if="showBanner" class="outdated-version-banner">
+    <iframe :src="iframeSrc" :title="$t('title')"></iframe>
+
+    <a class="btn btn-primary"
+      href="https://docs.getodk.org/central-upgrade/"
+      target="_blank" rel="noopener"
+      v-tooltip.aria-describedby="$t('instructionsToUpgradeTooltip')">
+        {{ $t('instructionsToUpgrade') }}
+    </a>
+
+    <button class="btn btn-danger"
+      type="button"
+      v-tooltip.aria-describedby="$t('dismissTooltip')"
+      @click="dismiss">
+        {{ $t('dismiss') }}
+    </button>
+  </div>
+</template>
+
+<script setup>
+import { computed, inject } from 'vue';
+import { useRequestData } from '../request-data';
+import { useSessions } from '../util/session';
+
+const { visiblyLoggedIn } = useSessions();
+
+const { centralVersion, currentUser } = useRequestData();
+
+defineOptions({
+  name: 'OutdatedVersionBanner'
+});
+
+const container = inject('container');
+const { i18n: globalI18n } = container;
+const languageSubtag = computed(() => new Intl.Locale(globalI18n.locale).language);
+
+const iframeSrc = computed(() => `https://sadiqkhoja.com/central/outdated-version.html?version=${centralVersion.data?.currentVersion}&lang=${languageSubtag.value}`);
+
+const showBanner = computed(() => {
+  // user is not logged in or doesn't have ability to set config (implying not an admin)
+  if (!currentUser.dataExists || !visiblyLoggedIn || !currentUser.can('config.set')) return false;
+  if (!centralVersion.dataExists) return false;
+
+  // User has seen the warning in the last 30 days, so don't show it again
+  // 864E5 is the number of milliseconds in a day
+  const dismissDate = currentUser.preferences?.site?.outdatedVersionWarningDismissDate;
+  if (dismissDate && centralVersion.currentDate.getTime() < (dismissDate.getTime() + (864E5 * 30))) return false;
+
+  // Difference between current year and Central version year is less than 2
+  const centralVersionYear = Number(centralVersion.currentVersion.match(/^(\d{4})/)[1]);
+  const currentYear = centralVersion.currentDate.getFullYear();
+  if (currentYear - centralVersionYear < 2) return false;
+
+  return true;
+});
+
+const dismiss = () => {
+  currentUser.preferences.site.outdatedVersionWarningDismissDate = centralVersion.currentDate;
+};
+</script>
+
+<style lang="scss">
+@import '../assets/scss/variables';
+
+.outdated-version-banner {
+  background-color: $color-danger-light;
+  border: 1px solid $color-danger;
+  border-radius: 8px;
+  padding: 0 20px 20px 20px;
+  text-align: center;
+  width: 700px;
+  margin: 20px auto;
+  box-shadow: 0 5px 10px rgba(0, 0, 0, 0.2);
+
+  iframe {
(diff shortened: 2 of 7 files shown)


src/request-data/user-preferences/normalizers.js
```diff
@@ -33,6 +33,21 @@ export class SitePreferenceNormalizer extends PreferenceNormalizer {
...
+  static outdatedVersionWarningDismissDate(val) {
+    // Frontend to Backend
+    if (val instanceof Date) {
+      return val.toISOString();
+    }
+
+    // Backend to Frontend
+    const isoDateRegex = /^\d{4}-\d{2}-\d{2}([T](\d{2}:\d{2}:\d{2}(\.\d+)?Z)?)?$/;
+    if (typeof (val) === 'string' && isoDateRegex.test(val)) {
+      return new Date(val);
```
matthew-white, line 46: Could it be possible to always return a string so that the return type is consistent? I know right now it's different for Backend vs. Frontend, but I think it'd be a little clearer if it was always the same.

src/components/outdated-version-banner.vue
```diff
@@ -0,0 +1,114 @@
...
+  }
+}
+
+</style>
+
+<i18n lang="json5">
+  {
+    "en": {
+      // This is a title for the outdated version banner shown for the screenreaders only
+      "title": "Outdate Version",
```
matthew-white, line 107:
```suggestion
      "title": "Outdated Version",
```

src/components/outdated-version-banner.vue
```diff
@@ -0,0 +1,114 @@
...
+      "instructionsToUpgrade": "Instructions to upgrade",
+      "instructionsToUpgradeTooltip": "Click here to see the instructions to upgrade the Central",
```
matthew-white, line 109:
```suggestion
      "instructionsToUpgradeTooltip": "Click here to see instructions to upgrade Central",
```

src/components/outdated-version-banner.vue
```diff
@@ -0,0 +1,114 @@
...
+
+.outdated-version-banner {
+  background-color: $color-danger-light;
+  border: 1px solid $color-danger;
+  border-radius: 8px;
+  padding: 0 20px 20px 20px;
+  text-align: center;
+  width: 700px;
+  margin: 20px auto;
+  box-shadow: 0 5px 10px rgba(0, 0, 0, 0.2);
```
matthew-white, line 86:
```suggestion
  box-shadow: $box-shadow-popover;
```

src/components/outdated-version-banner.vue
```diff
@@ -0,0 +1,114 @@
...
+
+const { centralVersion, currentUser } = useRequestData();
+
+defineOptions({
+  name: 'OutdatedVersionBanner'
+});
+
+const container = inject('container');
+const { i18n: globalI18n } = container;
+const languageSubtag = computed(() => new Intl.Locale(globalI18n.locale).language);
```
matthew-white, line 48: Could we pass along the whole `globalI18n.locale` instead of just the subtag? I think we'll use the Frontend Transifex for these translations, so it should be the same set of locales.

I want to leave open the possibility that Central supports Simplified Chinese one day (zh-Hans). Currently, we support Traditional Chinese (zh-Hant). The two have the same language subtag, which is part of why I think it'd be nice to send the whole locale.

src/components/outdated-version-banner.vue
```diff
@@ -0,0 +1,114 @@
...
+except according to the terms contained in the LICENSE file.
+-->
+
+<template>
+  <div v-if="showBanner" class="outdated-version-banner">
+    <iframe :src="iframeSrc" :title="$t('title')"></iframe>
+
+    <a class="btn btn-primary"
+      href="https://docs.getodk.org/central-upgrade/"
+      target="_blank" rel="noopener"
```
matthew-white, line 19: I think `target="_blank"` implies `rel="noopener"` in modern browsers.

src/components/home/news.vue
```diff
@@ -15,17 +15,29 @@ except according to the terms contained in the LICENSE file.
...
 });
+
+const { centralVersion } = useRequestData();
+
+
+const container = inject('container');
+const { i18n: globalI18n } = container;
+const languageSubtag = computed(() => new Intl.Locale(globalI18n.locale).language);
+
+const iframeSrc = computed(() => `https://getodk.github.io/central/news.html?version=${centralVersion.data?.currentVersion}&lang=${languageSubtag.value}`);
```
matthew-white, line 39: What happens if `!centralVersion.dataExists`? Does it become `?version=undefined`?

Maybe we don't need to pass `?version` here? I don't think we'd want to reload news.html once the response comes in for version.txt. (The response would cause the URL to change.) How about passing something static like `?outdatedVersionWarning=false`?

src/request-data/user-preferences/normalizers.js
```diff
@@ -33,6 +33,21 @@ export class SitePreferenceNormalizer extends PreferenceNormalizer {
...
   }
+
+  static outdatedVersionWarningDismissDate(val) {
+    // Frontend to Backend
+    if (val instanceof Date) {
+      return val.toISOString();
+    }
+
+    // Backend to Frontend
+    const isoDateRegex = /^\d{4}-\d{2}-\d{2}([T](\d{2}:\d{2}:\d{2}(\.\d+)?Z)?)?$/;
```
matthew-white, line 44: I mentioned `DateTime.fromISO().isValid` in the tests below, but I want to be explicit that I don't think we should use it here. The `luxon` package is kind of heavy, so I want to avoid including it in the main bundle.

I could see it being helpful one day to have a utility function like `isISODate()`.

matthew-white, line 44: I feel like there's one more `?` here than I'd expect. Isn't this saying that you could have `T` on its own without numbers after it?

test/util/http.js
```diff
@@ -415,7 +415,7 @@ class MockHttp {
   }
 
   respondWithData(callback) {
-    return this.respond(() => ({ status: 200, data: callback() }));
+    return this.respond(() => ({ status: 200, data: callback(), headers: new Map([['date', new Date()]]) }));
```
matthew-white, line 418: Instead of adding this here, could you add it in `_http()` (or maybe `respond()`)? Only if it's easy. I think it makes sense to always return the header, whether it's after `respondWithData()`, `respondWithSuccess()`, or `respondWithProblem()`.

src/components/outdated-version-banner.vue
```diff
@@ -0,0 +1,114 @@
...
+    </button>
+  </div>
+</template>
+
+<script setup>
+import { computed, inject } from 'vue';
+import { useRequestData } from '../request-data';
+import { useSessions } from '../util/session';
+
+const { visiblyLoggedIn } = useSessions();
```
matthew-white, line 38: `useSessions()` adds an event listener and does some other things, so it should only be called in `App`. Instead, you can `inject('visiblyLoggedIn')`. You could `provide` a value for it in the tests of the component.

test/components/outdated-version-banner.spec.js
```diff
@@ -0,0 +1,68 @@
...
+      description: 'version is too old and dismiss date is 30 days ago',
+      expectedResult: true, userLoggedIn: true, role: 'admin', centralVersion: '(v2022.1.2-sha)', currentDate: '2024-12-11', dismissDate: '2024-11-10T12:00:00Z'
+    },
+    {
+      description: 'version is too old but dismiss date is recent',
+      expectedResult: false, userLoggedIn: true, role: 'admin', centralVersion: '(v2022.1.2-sha)', currentDate: '2024-12-11', dismissDate: '2024-12-01T12:00:00Z'
+    },
+    {
+      description: 'version is not old',
+      expectedResult: false, userLoggedIn: true, role: 'admin', centralVersion: '(v2024.1.2-sha)', currentDate: '2024-12-11', dismissDate: '2024-12-11T12:00:00Z'
```
matthew-white, line 44:
```suggestion
      expectedResult: false, userLoggedIn: true, role: 'admin', centralVersion: '(v2024.1.2-sha)', currentDate: '2024-12-11', dismissDate: null
```

src/components/outdated-version-banner.vue
```diff
@@ -0,0 +1,114 @@
...
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+
+<template>
+  <div v-if="showBanner" class="outdated-version-banner">
+    <iframe :src="iframeSrc" :title="$t('title')"></iframe>
```
matthew-white, line 15: Is it possible to add a test of the value of the `src` attribute?

src/components/home/news.vue
```diff
@@ -15,17 +15,25 @@ except according to the terms contained in the LICENSE file.
       <span>{{ $t('title') }}</span>
     </template>
     <template #body>
-      <iframe src="https://getodk.github.io/central/news.html" :title="$t('title')"></iframe>
+      <iframe :src="iframeSrc" :title="$t('title')"></iframe>
```
matthew-white, line 18: Now that `src` is dynamic, how about adding a test of it?

src/components/outdated-version-banner.vue
```diff
@@ -0,0 +1,114 @@
...
+
+defineOptions({
+  name: 'OutdatedVersionBanner'
+});
+
+const container = inject('container');
+const { i18n: globalI18n } = container;
+const locale = computed(() => globalI18n.locale);
+
+const iframeSrc = computed(() => `https://getodk.github.io/central/outdated-version.html?version=${centralVersion.data?.currentVersion}&lang=${locale.value}`);
```
matthew-white, line 50: If the iframe is rendered, I think that means that `centralVersion.dataExists`. So I'm thinking the `?.` could be removed.

matthew-white (approved):

src/request-data/user-preferences/normalizers.js
sadiqkhoja: I have removed all `?` as [Data.toISOString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString) returns a simplified string which is always 24 or 27 character. In our case it will be always 24.

PR #1096, opened 2024-12-12
Use FormLink in DatasetOverview

matthew-white:
This PR is in response to https://github.com/getodk/central/issues/670#issuecomment-2535848397. It adds hover cards for the forms shown on the entity list overview.

I tried it out locally. I also updated tests.

`FormLink` expects the form passed to it to have a `projectId` property. However, the forms listed in the entity list don't have a `projectId`. I think they just have an `xmlFormId` and `name`. I ended up thinking that the easiest thing was to add a `projectId` to these forms when the response for the entity list is received. That way, components don't have to do extra legwork and can just pass each form to `FormLink` as-is.

I've also added some explanatory comments below.

### DIFF
--- a/src/components/dataset/overview/connection-to-forms.vue
+++ b/src/components/dataset/overview/connection-to-forms.vue
@@ -21,7 +21,9 @@ except according to the terms contained in the LICENSE file.
         <expandable-row v-for="(form) in propertiesByForm" :key="form.xmlFormId">
           <template #title>
             <div class="form-name">
-              <router-link :to="publishedFormPath(projectId, form.xmlFormId)" v-tooltip.text>{{ form.name }}</router-link>
+              <form-link :form="form"
+                :to="publishedFormPath(form.projectId, form.xmlFormId)"
+                v-tooltip.text/>
             </div>
           </template>
           <template #caption>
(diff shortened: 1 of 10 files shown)


src/components/dataset/overview/connection-to-forms.vue
```diff
@@ -39,44 +41,35 @@ except according to the terms contained in the LICENSE file.
...
-  props: {
-    properties: {
-      type: Array,
-      required: true
-    },
-    sourceForms: {
-      type: Array,
-      required: true
-    },
-    projectId: {
```
matthew-white, line 63: Now that `projectId` is added to each form, we don't need to pass it in as a prop. I went ahead and removed that prop, but then there was an issue in the tests of the component. The component now expects the forms to have a `projectId` property, but in the tests of the component, they didn't. The easiest way to solve that was to specify a dataset for `requestData`. Doing so automatically transforms the dataset object, adding `projectId`. Once I was passing the dataset to the test, I felt like it didn't make sense to also pass `properties` and `sourceForms` as props, since those are accessible from the dataset.

src/components/dataset/overview/connection-to-forms.vue
```diff
@@ -21,7 +21,9 @@ except according to the terms contained in the LICENSE file.
         <expandable-row v-for="(form) in propertiesByForm" :key="form.xmlFormId">
           <template #title>
             <div class="form-name">
-              <router-link :to="publishedFormPath(projectId, form.xmlFormId)" v-tooltip.text>{{ form.name }}</router-link>
+              <form-link :form="form"
+                :to="publishedFormPath(form.projectId, form.xmlFormId)"
```
matthew-white, line 25: We need to specify `:to` in this case. `FormLink` defaults to `primaryFormPath()`, but `primaryFormPath()` doesn't work super well in this case. That's because it expects the form object to have a `publishedAt` property.

src/components/dataset/overview/connection-to-forms.vue
```diff
@@ -39,44 +41,35 @@ except according to the terms contained in the LICENSE file.
...
-    return { publishedFormPath };
+    return { dataset, publishedFormPath };
   },
   computed: {
     totalProperties() {
-      return this.properties.length;
+      return this.dataset.properties.length;
     },
     propertiesByForm() {
-      const formMap = new Map(this.sourceForms.map(f => ([f.xmlFormId, { ...f, projectId: this.projectId, properties: [] }])));
```
matthew-white, line 77: We were adding `projectId` to the forms in `formMap`, but it actually didn't seem like we ever accessed `projectId` from `formMap`. We were used the `projectId` prop instead. Either way, forms will now have a `projectId` property automatically.

test/components/dataset/overview/dataset-properties.spec.js
```diff
@@ -1,33 +1,18 @@
-/*
-Copyright 2023 ODK Central Developers
```
matthew-white, line 2: We don't need a file header in test files.

matthew-white (commented):
Notes for async review

src/components/dataset/overview/connection-to-forms.vue
sadiqkhoja: Just as projectId is set, we could also set publishedAt to something not null, but this is also fine.

sadiqkhoja (approved):

src/components/dataset/overview/connection-to-forms.vue
matthew-white: I agree that that would have been a reasonable approach as well. :+1:

Issue #1084, opened 2024-12-08
Replace Bootstrap popover plugin with Floating UI


matthew-white:
We use popovers in just a couple of ways in Frontend:

- Showing a QR code
    - `FieldKeyList` shows `FieldKeyQrPanel` in a popover.
    - `FormDraftTesting` shows `FormDraftQrPanel` in a popover.
- `HoverCards` shows a popover.

We want to do more with popovers. However, because of our use of the Bootstrap popover plugin, there are a number of limitations:

https://github.com/getodk/central-frontend/blob/4158cd0994bb6087885638698797e9e9c939762a/src/components/popover.vue#L14-L29

These limitations mean that it can be pretty unwieldy from a code perspective to add a popover. In contrast, it's pretty straightforward to add a modal.

This issue is to replace the Bootstrap popover plugin with something else. Moving away from Bootstrap plugins is something we want to do in general: see #984. I used Floating UI to implement our tooltips, and I liked it. I think we should take a look at how hard it would be to replace the Bootstrap plugin with Floating UI.

One issue to keep in mind is getodk/central#797. That issue requires fairly granular control of the placement of the popover. I think Floating UI would give us that control.

The `Popover` component wraps content that we want to show in a popover, similar to the `Modal` component. I like that component-based approach, so if there's a way to keep `Popover` similar to how it is today (in terms of its contract/API/props/slots), I think that'd be nice. However, we'd need to make a number of changes within `Popover` itself in order to replace the Bootstrap plugin.

PR #1097, opened 2024-12-12
Position hover card above or below anchor if right and left don't work

matthew-white:
By default, the hover card is shown on the right. If it doesn't fit there, it is shown on the left. However, @sadiqkhoja noticed that it's possible that the hover card won't fit on either the right or the left. This PR tries to fix that. If the hover card won't fit on the right or the left, it will be shown above or below the anchor element.

I tried it out locally. We can also ask the QA team to test it.

I wouldn't sure I'm sure that this approach covers all cases, but I'm hoping it works well enough until we're able to complete #1084 and getodk/central#797.

### DIFF
--- a/src/components/hover-cards.vue
+++ b/src/components/hover-cards.vue
@@ -84,14 +84,25 @@ const component = shallowRef(null);
 const placement = ref(undefined);
 
 const popoverRef = ref(null);
-const positionPopover = () => {
-  const rightOfAnchor = hoverCard.anchor.getBoundingClientRect().right;
-  const componentElement = popoverRef.value.$el.children[0];
-  const componentWidth = componentElement.getBoundingClientRect().width;
-  // Subtracting 50 in order to provide some buffer.
-  const overflowsRight = rightOfAnchor + componentWidth >
-    document.documentElement.clientWidth - 50;
-  placement.value = overflowsRight ? 'left' : 'right';
+// The minimum distance in px between the edge of the hover card and the edge of
+// the viewport. This provides space for the popover arrow, as well some extra
+// buffer.
+const buffer = 20;
+const computePlacement = () => {
+  const anchorRect = hoverCard.anchor.getBoundingClientRect();
+  const popoverRect = popoverRef.value.$el.getBoundingClientRect();
+
+  // Try 'right' and 'left' first unless the anchor is so close to the bottom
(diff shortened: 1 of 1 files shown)


src/components/hover-cards.vue
```diff
@@ -84,14 +84,25 @@ const component = shallowRef(null);
...
+// the viewport. This provides space for the popover arrow, as well some extra
+// buffer.
+const buffer = 20;
+const computePlacement = () => {
+  const anchorRect = hoverCard.anchor.getBoundingClientRect();
+  const popoverRect = popoverRef.value.$el.getBoundingClientRect();
+
+  // Try 'right' and 'left' first unless the anchor is so close to the bottom
+  // that the popover won't fit.
+  if (window.scrollY + anchorRect.bottom + (popoverRect.height / 2) <
```
matthew-white, line 97: `/ 2` because the popover is vertically centered relative to the anchor element.

sadiqkhoja (approved):
This looks good. 

What do you think of wrapping the hover card heading after certain width?

matthew-white:
> What do you think of wrapping the hover card heading after certain width?

The heading (`.hover-card-title`) is supposed to be set up to truncate. It shouldn't cause the width of the hover card to increase. The only thing that should cause the width of the hover card to increase is the `<dt>` elements (if `:truncate-dt="false"`). That said, I see in the screenshot that you shared that the heading isn't truncated. It's a little hard to see what's going on, since the hover card is partially off-screen. When I try things out locally on this branch, I see that a long heading is truncated. Let me know if you see anything weird about the heading once this PR is merged! Or let me know if you have other thoughts about how to improve how the hover card truncates/wraps/widens.

matthew-white:
Never mind, I think I understand the problem now. If `:truncate-dt="false"`, and the `<dd>` elements are not truncated when their width is set to the width of the `<dt>` elements, then `width: auto` is set on the hover card. That's supposed to _decrease_ the width of the `<dd>` elements so that they don't take up more space than they need. However, it also allows `.hover-card-title` to grow arbitrarily wide. 😨 That's definitely not intentional. I'll fix that tomorrow!

src/components/hover-cards.vue
matthew-white: I added a code comment explaining this.

PR #1098, opened 2024-12-13
Adjust spacing between actions above submissions table

matthew-white:
This PR addresses the following to-do mentioned at https://github.com/getodk/central/issues/833#issuecomment-2529012617:

> Move the Refresh button next to the Download button

The refresh button will be moved for the draft testing page only. It will stay where it is on the regular submissions page.

This PR similarly groups the column selection dropdown with the refresh and download buttons on the right.

Additionally, I noticed issues with the vertical alignment of the actions above the table:

- The download button doesn't seem to be vertically aligned with the rest of the actions (at least on the submissions page and probably also the draft testing page). You can see that on staging if you decrease the width of the viewport enough such that the download button is positioned close to the rest of the actions.
- On the draft testing page, the test buttons are not vertically aligned with the refresh button.

This PR fixes these issues with vertical alignment.

I iterated locally on the spacing/positioning of actions. I viewed the before/after side-by-side.

As part of the PR, I moved the refresh button outside the form. (See the comment below.) That shouldn't have any effect on behavior, but I did do a quick check that the refresh button still worked. That's also something that tests look at.

This PR should only affect the spacing/positioning of actions. It's mostly changes to styles.

### DIFF
--- a/src/components/submission/list.vue
+++ b/src/components/submission/list.vue
@@ -33,13 +33,13 @@ except according to the terms contained in the LICENSE file.
           <submission-field-dropdown
             v-if="selectedFields != null && fields.selectable.length > 11"
             v-model="selectedFields"/>
-          <button id="submission-list-refresh-button" type="button"
-            class="btn btn-default" :aria-disabled="refreshing"
-            @click="fetchChunk(false, true)">
-            <span class="icon-refresh"></span>{{ $t('action.refresh') }}
-            <spinner :state="refreshing"/>
-          </button>
         </form>
+        <button id="submission-list-refresh-button" type="button"
+          class="btn btn-default" :aria-disabled="refreshing"
+          @click="fetchChunk(false, true)">
+          <span class="icon-refresh"></span>{{ $t('action.refresh') }}
+          <spinner :state="refreshing"/>
+        </button>
         <submission-download-button :form-version="formVersion"
           :aria-disabled="deleted" v-tooltip.aria-describedby="deleted ? $t('downloadDisabled') : null"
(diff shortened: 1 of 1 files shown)


src/components/submission/list.vue
```diff
@@ -485,27 +485,44 @@ export default {
...
   flex-wrap: wrap-reverse;
+  // This is 10px of space between actions on the row and also 10px between
+  // rows if the actions start wrapping.
+  gap: 10px;
+  margin-bottom: 25px;
+
+  .form-inline {
+    margin-bottom: 0;
+    padding-bottom: 0;
+  }
```
matthew-white, line 499: Different children of `#submission-list-actions` had different `margin-bottom` and `padding-bottom`, which may have been partly responsible for the vertical misalignment. Here, we remove the space below the `.form-inline`. Below, we remove the space between the download button. The latter spacing is now handled by the flexbox gap.

src/components/submission/list.vue
```diff
@@ -485,27 +485,44 @@ export default {
...
   margin-right: 5px;
 }
 #submission-download-button {
-  // The bottom margin is for if the download button wraps above the other
-  // actions.
-  margin-bottom: 10px;
   margin-left: auto;
 }
 
 #submission-list-test-in-browser {
```
matthew-white, line 513: The presence of this button is how I'm styling things differently on the draft testing page.

matthew-white (commented):
Notes for code review

ktuite (approved):

src/components/submission/list.vue
matthew-white: I've now added a code comment explaining this strategy.

PR #1100, opened 2024-12-13
Add devTools flag to config and use it to gate EntityBranchData

matthew-white:
[This comment](https://docs.google.com/document/d/1SXQjHFfOL4q2_5rOd0stLq3FGHnzJPJNv6VECfVkwLw/edit?disco=AAABZyBYceU) on the release criteria for v2024.2 asks whether we should prevent showing `EntityBranchData` unless a Frontend config is set. I think that's a good idea. It's not impossible that a user would randomly click "Offline update", and it'd be confusing if they did.

I tried it out locally. If the config is set to `true`, I can see the modal, and if it's set to `false`, I can't. Since this modal is only for development/testing, I don't think we need to write automated tests about it.

I named the config something generic: `devTools`. In the future, I think there will probably be other examples of things like this modal.

### DIFF
--- a/src/components/entity/show.vue
+++ b/src/components/entity/show.vue
@@ -37,19 +37,19 @@ except according to the terms contained in the LICENSE file.
       :label="entity.dataExists ? entity.currentVersion.label : ''"
       :awaiting-response="awaitingResponse" @hide="deleteModal.hide()"
       @delete="requestDelete"/>
-    <entity-branch-data v-bind="branchData" @hide="branchData.hide()"/>
+    <entity-branch-data v-if="config.devTools" v-bind="branchData"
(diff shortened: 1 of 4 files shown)


src/components/entity/show.vue
```diff
@@ -154,7 +155,8 @@ const requestDelete = () => {
     .catch(noop);
 };
 
-const branchData = modalData();
+const EntityBranchData = defineAsyncComponent(loadAsync('EntityBranchData'));
```
matthew-white, line 158: The component will only be loaded if it's rendered (if the `v-if` is `true`). So it won't ever have to be loaded for users.

sadiqkhoja (approved):

PR #1103, opened 2024-12-13
Resize hover cards only based on <dl>, not .hover-card-title

matthew-white:
This PR addresses an issue with how hover cards are currently sized. See the discussion that starts at https://github.com/getodk/central-frontend/pull/1097#pullrequestreview-2500713964.

I tried it out locally in Chrome and Firefox. I also wrote tests.

These are my go-to scenarios when trying things out locally:

- Select Swahili as the language. "Latest Submission" is a very long translation in Swahili.
- Hover over a form. The translation for "Latest Submission" should not be truncated.
- Hover over a form with a long form version. It should be truncated after the width of the `<dd>` reaches the width of the `<dt>`.
- Hover over a form with a short form version. The `<dd>` should be less wide than the `<dt>`.

The key idea here is to hide the heading and footer of the hover card when determining the ideal width of the body.

I'll ask the QA team to take a look.

### DIFF
--- a/src/components/hover-card.vue
+++ b/src/components/hover-card.vue
@@ -28,7 +28,7 @@ except according to the terms contained in the LICENSE file.
 <script setup>
 import { onMounted, ref } from 'vue';
 
-import { px } from '../util/dom';
+import { px, styleBox } from '../util/dom';
 
 defineOptions({
   name: 'HoverCard'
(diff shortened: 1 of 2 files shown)


src/components/hover-card.vue
```diff
@@ -55,34 +55,44 @@ const props = defineProps({
 
 const width = ref('');
 const el = ref(null);
-const resize = () => {
-  if (props.truncateDt) return;
+// Returns the desired width of .hover-card-body based on the width of the <dl>.
+// Returning 0 means that the default width should be used.
+const computeBodyWidth = () => {
+  if (props.truncateDt) return 0;
+  const dd = [...el.value.querySelectorAll('dd')];
```
matthew-white, line 62: I guess I don't need to get all the `<dd>` elements anymore, just the first.

src/components/hover-card.vue
```diff
@@ -55,34 +55,44 @@ const props = defineProps({
...
+  // Get the width excluding borders.
+  const currentBodyWidth = el.value.querySelector('.hover-card-body').clientWidth;
+
+  // Hide children other than .hover-card-body before setting `width: auto` so
+  // that we can compute the width of the hover card based on the width of
+  // .hover-card-body alone. For example, we don't want .hover-card-title to
+  // influence the width of the hover card.
+  const otherChildren = [...el.value.querySelectorAll(':scope > *:not(.hover-card-body)')];
+  for (const child of otherChildren) child.style.display = 'none';
+  el.value.style.width = 'auto';
```
sadiqkhoja, line 84: if we move `ref="el"` from the root to `.hover-card-body` then we can avoid this, right?

sadiqkhoja (approved):
Looks good to me.

PR #847, opened 2023-08-29
Labels for 2023.4 usage metrics

ktuite:
Closes #

Frontend part of https://github.com/getodk/central-backend/pull/964 

Screen Shot 2023-09-11 at 11 21 28 AM

#### Before submitting this PR, please make sure you have:

- [x] run `npm run test` and `npm run lint` and confirmed all checks still pass OR confirm CircleCI build passes
- [x] verified that any code or assets from external sources are properly credited in comments or that everything is internally sourced

PR #1104, opened 2024-12-13
Feature #847: Add close button to QR popover

sadiqkhoja:
Closes getodk/central#847

### Changes:

* Hide popover on click of a button that has data-closes-popover attribute
* Hide popover on escape key as well
* Added button in QR panel to hide popover (self)

Tried in chrome and added fews tests.

Because popover component adds click eventListener on the document that `useCapture`, I was not able to handle vue events directly on qr-panel component. Having another condition in `closeAfterClick` that checks if the event target is a close button, identified by `data-closes-popover` attribute, seems reasonable to me.

I have also added event listener for escape key, which is applicable on all popover. If we want we can add a prop to add it on selective popovers - like maybe we don't hide hover-cards on escape 🤷‍♂️.

I don't see any regression risks here.

No

### DIFF
--- a/src/components/popover.vue
+++ b/src/components/popover.vue
@@ -74,11 +74,13 @@ export default {
     // not show, then hide), we use event capturing here.
     document.addEventListener('click', this.hideAfterClick, true);
     window.addEventListener('resize', this.hideAfterResize);
+    document.addEventListener('keydown', this.hideAfterEsc);
   },
   beforeUnmount() {
     if (this.target != null) this.hide();
     document.removeEventListener('click', this.hideAfterClick, true);
     window.removeEventListener('resize', this.hideAfterResize);
+    document.removeEventListener('keydown', this.hideAfterEsc);
   },
   methods: {
     show() {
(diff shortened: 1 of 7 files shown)


src/components/popover.vue
```diff
@@ -74,11 +74,13 @@ export default {
     // not show, then hide), we use event capturing here.
     document.addEventListener('click', this.hideAfterClick, true);
     window.addEventListener('resize', this.hideAfterResize);
+    document.addEventListener('keydown', this.hideAfterEsc);
   },
   beforeUnmount() {
     if (this.target != null) this.hide();
     document.removeEventListener('click', this.hideAfterClick, true);
     window.removeEventListener('resize', this.hideAfterResize);
+    document.addEventListener('keydown', this.hideAfterEsc);
```
matthew-white, line 83:
```suggestion
    document.removeEventListener('keydown', this.hideAfterEsc);
```

matthew-white (approved):
This is looking good. :+1: Once we complete #1084, maybe we could see if there's a different approach we could take than the `data-closes-popover` attribute. But I think it makes a lot of sense given how the Bootstrap plugin works.

matthew-white:
> maybe we don't hide hover-cards on escape 🤷‍♂️.

That seems OK to me. We could always add that logic later if we need to.

PR #1112, opened 2024-12-16
Increase line-height in HoverCard to prevent text truncation

matthew-white:
This PR attempts to address the issue described at https://github.com/getodk/central/issues/670#issuecomment-2527556597.

I actually had difficulty reproducing the issue locally. I can see that Szymon is seeing a different font than I am. However, after trying each of the fonts listed in the CSS, I don't see text truncation:

https://github.com/getodk/central-frontend/blob/f1949fff55e56447e809c5c7d51f7d7046bb4405/src/assets/css/bootstrap.css#L281

I tried in both Chrome and Firefox.

At some point, we should maybe look into what options we have around standardizing fonts in Central. For now, I'm hoping I can solve the issue by increasing the line height. I strongly suspect that the text truncation that Szymon is seeing is due to `line-height` being set to 1 in the `HoverCard` component.

I wasn't able to reproduce the issue, so my main goal was to keep the look of the hover cards as consistent as possible while also increasing the line height. I compared hover cards side-by-side before/after making this change.

The main change I want to try is increasing the line height. After making that change, I also had to fiddle with margin/padding.

### DIFF
--- a/src/components/hover-card.vue
+++ b/src/components/hover-card.vue
@@ -109,6 +109,7 @@ $border-radius: 12px;
 .hover-card { width: 290px; }
 
 .hover-card-heading {
+  // Used to align the top of the icon with the top of the text.
   $icon-margin-top: 2px;
 
   background-color: #666;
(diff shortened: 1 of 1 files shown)


src/components/hover-card.vue
```diff
@@ -109,22 +109,23 @@ $border-radius: 12px;
...
   border-top: $border;
   border-top-left-radius: $border-radius;
   border-top-right-radius: $border-radius;
   color: #fff;
   display: flex;
-  // Align the top of the text with the top of the icon.
-  line-height: 1;
-  padding: $padding-panel-body;
-  padding-bottom: 12px;
+  line-height: 1.2;
```
matthew-white, line 121: This is still less than the default, but it's a number that we use in several places in Frontend.

ktuite (approved):

PR #1115, opened 2024-12-19
Prevent entity data from overflowing hover card

matthew-white:
Closes getodk/central#854.

I tried it out locally. I also wrote a test that fails when the fix is not in place.

It's surprising to me that this is necessary, because the `<div>` that applies the text wrapping/truncation already specifies `overflow: hidden`. However, the `<div>` itself actually isn't overflowing: its parent element, the `<dd>`, is. The `<dd>` ends up overflowing the `<dl>`. I think that's possible because we're now using a CSS grid for the `<dl>`. I've had to do something similar when I've used flexbox.

### DIFF
--- a/test/components/hover-card/entity.spec.js
+++ b/test/components/hover-card/entity.spec.js
@@ -1,0 +1,80 @@
+import DlData from '../../../src/components/dl-data.vue';
+import HoverCardEntity from '../../../src/components/hover-card/entity.vue';
+
+import useHoverCardResources from '../../../src/request-data/hover-card';
+
+import { truncatesText } from '../../../src/util/dom';
+
+import testData from '../../data';
+import { mergeMountOptions, mount } from '../../util/lifecycle';
+import { testRequestData } from '../../util/request-data';
+
+const mountComponent = (options = undefined) =>
+  mount(HoverCardEntity, mergeMountOptions(options, {
+    container: {
+      requestData: testRequestData([useHoverCardResources], {
+        dataset: testData.extendedDatasets.last(),
+        entity: testData.standardEntities.last()
+      })
+    }
+  }));
+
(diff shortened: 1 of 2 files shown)


test/components/hover-card/entity.spec.js
```diff
@@ -0,0 +1,80 @@
...
+      requestData: testRequestData([useHoverCardResources], {
+        dataset: testData.extendedDatasets.last(),
+        entity: testData.standardEntities.last()
+      })
+    }
+  }));
+
+describe('HoverCardEntity', () => {
+  describe('entity property data', () => {
+    it('shows the first 5 properties', () => {
```
matthew-white, line 24: This test and also the next test aren't super related to this PR. However, we don't currently have any tests of the entity hover card. (I plan to add more.) Since I'm adding a new test for this PR, I thought it made sense to add basic tests that the entity property data is being shown as expected.

test/components/hover-card/entity.spec.js
```diff
@@ -0,0 +1,80 @@
...
+    it('truncates a long property value', () => {
+      testData.extendedEntities.createPast(1, {
+        data: { x: 'y'.repeat(1000) }
+      });
+      const component = mountComponent({ attachTo: document.body });
+      const body = component.get('.hover-card-body');
+      body.element.clientWidth.should.equal(288);
+      const dd = body.get('.dl-data-dd');
+      // More than 50% of the width of .hover-card-body, but well short of 100%
+      dd.element.clientWidth.should.be.within(180, 250);
```
matthew-white, line 76: It's not the property name "x" that causes the `<dd>` element to take up >30px of width. It's another `<dd>`, "Last update".

lognaturel (approved):

PR #1117, opened 2024-12-19
Fixes tests that send user preference request when changing project sort order

sadiqkhoja:

All tests are passing, none of them are logging ERROR LOG

I have modified the tests to use mockHttp where user preferences were changed.

I was thinking to add exception for user-preference request in `logUnhandledRequest`, but went against it for the sake of consistency and to have visibility about this request while writing tests.

None 

No


matthew-white (approved):
Looks great, thank you for fixing this! 🎉

PR #1118, opened 2024-12-20
Add ability to show component in alert

matthew-white:
Right now, alerts can only show text. However, I've long wanted to lift that restriction in order to make it possible to show other content. I think lifting that restriction would make it easier to solve issues like getodk/central#775.

What got me thinking about this again was #1075. There, the alert lists one or more entity properties. It would've been nice if we could have shown a `<ul>`, but because the alert had to be text, that wasn't possible. Instead, we added bullet points to the text. With the change in this PR, it's now possible to show a `<ul>` in the alert in that case. That's done by passing a component to the alert.

I've updated existing tests. If this approach looks good, I'll add some new tests as well.

The approach in this PR is to pass a component to the alert. I also thought about passing HTML or Markdown. However, a component gives us the most flexibility. It also gives us an easy way to escape user content and not render user content as HTML.

### DIFF
--- a/src/alert.js
+++ b/src/alert.js
@@ -12,40 +12,51 @@ except according to the terms contained in the LICENSE file.
 import { shallowReactive } from 'vue';
 
 class AlertData {
+  #data;
+
   constructor() {
-    this._data = shallowReactive({
+    this.#data = shallowReactive({
       // The alert's "contextual" type: 'success', 'info', 'warning', or
       // 'danger'.
       type: 'danger',
-      message: null,
-      // `true` if the alert should be visible and `false` if not.
-      state: false,
-      // The time at which the alert was last set
+      // Either a string or an array with a component and props for the
+      // component
+      content: null,
+      // The time at which the alert was last shown
       at: new Date()
     });
   }
 
-  get type() { return this._data.type; }
-  get message() { return this._data.message; }
-  get state() { return this._data.state; }
-  get at() { return this._data.at; }
+  // Returns `true` if the alert should be visible and `false` if not.
+  get state() { return this.content != null; }
 
-  _set(type, message) {
-    this._data.type = type;
-    this._data.message = message;
-    this._data.state = true;
-    this._data.at = new Date();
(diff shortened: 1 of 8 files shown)


src/components/alert/40917.vue
```diff
@@ -0,0 +1,46 @@
...
+      <li v-for="p of duplicateProperties" :key="p.provided">
+        {{ $t('duplicateProperty', p) }}
+      </li>
+    </ul>
+  </div>
+</template>
+
+<script setup>
+defineOptions({
+  name: 'Alert40917'
```
matthew-white, line 25: I tried naming this `Alert409_17`, but the linter complained whenever I imported it:

```
Identifier 'Alert409_17' is not in camel case
```

matthew-white:
@sadiqkhoja, would you mind taking a quick first look at the approach in this PR? If it looks good, I can add more tests and get it ready for a full review.

sadiqkhoja:
I haven't looked at this yet, will do it tomorrow

src/components/alert/40917.vue
sadiqkhoja: other options to consider:

- `Alert409Dot17`
- AlertDuplicateProperties (semantic names)

src/alert.js
```diff
@@ -12,40 +12,51 @@ except according to the terms contained in the LICENSE file.
...
-    this._data = shallowReactive({
+    this.#data = shallowReactive({
       // The alert's "contextual" type: 'success', 'info', 'warning', or
       // 'danger'.
       type: 'danger',
-      message: null,
-      // `true` if the alert should be visible and `false` if not.
-      state: false,
-      // The time at which the alert was last set
+      // Either a string or an array with a component and props for the
```
sadiqkhoja, line 22: What do you think about having object instead of array? or add `component` and `props` getter so that alert.vue can access them without bracket notation.

sadiqkhoja (approved):
Approach looks good to me 👍

Suggested minor things as take it or leave it.

PR #1123, opened 2025-01-14
Replace Popover's use of Bootstrap plugin with Floating UI

ktuite:
Closes https://github.com/getodk/central-frontend/issues/1084 and https://github.com/getodk/central/issues/885

### DIFF
--- a/src/components/collect-qr.vue
+++ b/src/components/collect-qr.vue
@@ -140,13 +140,11 @@ useEventListener(document.body, 'click', (event) => {
     console.log(props.settings); // eslint-disable-line no-console
 });
 
-watch(() => props.settings, (newSettings) => {
+watch(() => props.settings, () => {
   // If the settings change, we should re-render the QR code.
   // This whole component could probably change again now that
   // the popover is allowed to have a reactive component within it.
-  if (newSettings != null) {
-    renderQrCode();
-  }
+  renderQrCode();
 });
 </script>
 
--- a/src/components/collect-qr.vue
+++ b/src/components/collect-qr.vue
@@ -23,7 +23,7 @@ let id = 0;
 <script setup>
 import qrcode from 'qrcode-generator';
 import pako from 'pako/lib/deflate';
-import { inject, onMounted, ref } from 'vue';
+import { inject, onMounted, ref, watch } from 'vue';
 import { useI18n } from 'vue-i18n';
 
 import useEventListener from '../composables/event-listener';
@@ -59,11 +59,7 @@ const imgHtml = ref('');
 const margin = 15;
 const textFontSize = 16;
 
-// The QR code is rendered to a canvas (possibly with additional icons and text)
-// and then the canvas data is converted to an image via toDataURL and used as the source
-// of an image tag.
-// It is assumed that props will not change and the QR image will only be rendered once.
-onMounted(() => {
+const renderQrCode = () => {
   const code = qrcode(0, props.errorCorrectionLevel);
   const json = JSON.stringify(props.settings);
   code.addData(btoa(pako.deflate(json, { to: 'string' })));
@@ -123,6 +119,14 @@ onMounted(() => {
   img.alt = t('altText');
 
   imgHtml.value = img.outerHTML;
+};
+
+// The QR code is rendered to a canvas (possibly with additional icons and text)
+// and then the canvas data is converted to an image via toDataURL and used as the source
+// of an image tag.
+// If the settings change, the QR code image is re-rendered.
+onMounted(() => {
+  renderQrCode();
 });
 
 id += 1;
@@ -134,6 +138,15 @@ const config = inject('config');
 useEventListener(document.body, 'click', (event) => {
   if (config.devTools && event.target.parentNode.classList.contains(idClass))
     console.log(props.settings); // eslint-disable-line no-console
+});
+
+watch(() => props.settings, (newSettings) => {
+  // If the settings change, we should re-render the QR code.
+  // This whole component could probably change again now that
+  // the popover is allowed to have a reactive component within it.
+  if (newSettings != null) {
+    renderQrCode();
+  }
 });
 </script>
 
--- a/src/components/field-key/list.vue
+++ b/src/components/field-key/list.vue
@@ -153,18 +153,6 @@ export default {
 
       event.preventDefault();
       this.managed = !this.managed;
-
-      if (this.popover.target != null) {
-        this.$nextTick(() => {
-          // Changing this.managed may change the height of the popover: the
-          // height of the QR code may change, and there may be a locale for
-          // which the height of the text will change.
-          this.$refs.popover.update();
-
-          document.querySelector('.popover .field-key-qr-panel .switch-code')
-            .focus();
-        });
-      }
     },
     afterCreate(fieldKey) {
       this.fetchData(true);
--- a/src/components/field-key/list.vue
+++ b/src/components/field-key/list.vue
@@ -153,6 +153,13 @@ export default {
 
       event.preventDefault();
       this.managed = !this.managed;
+
+      if (this.popover.target != null) {
+        this.$nextTick(() => {
+          document.querySelector('.popover .field-key-qr-panel .switch-code')
+            .focus();
+        });
+      }
     },
     afterCreate(fieldKey) {
       this.fetchData(true);
--- a/src/components/hover-card.vue
+++ b/src/components/hover-card.vue
@@ -91,8 +91,6 @@ const resize = () => {
   for (const sibling of siblings) sibling.style.display = '';
 
   if (newWidth > currentWidth)
-    // Persist the new width in width.value so that it is added to the `style`
-    // attribute. That's needed for the style to be copied to the popover.
     el.value.style.width = px(el.value.getBoundingClientRect().width + newWidth - currentWidth);
 };
 onMounted(resize);
--- a/src/components/hover-card.vue
+++ b/src/components/hover-card.vue
@@ -10,7 +10,7 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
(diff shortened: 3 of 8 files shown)


src/components/field-key/qr-panel.vue
```diff
@@ -29,13 +29,13 @@ except according to the terms contained in the LICENSE file.
...
-        <template v-if="managed">{{ $t('body[1].managed', fieldKey) }}</template>
+        <template v-if="showManaged">{{ $t('body[1].managed', fieldKey) }}</template>
         <template v-else>{{ $t('body[1].legacy') }}</template>
         <sentence-separator/>
-        <i18n-t v-if="managed" keypath="body[2].managed.full">
+        <i18n-t v-if="showManaged" keypath="body[2].managed.full">
           <template #switchToLegacy>
             <i18n-t tag="a" keypath="body[2].managed.switchToLegacy"
-              class="switch-code" href="#">
+              class="switch-code" href="#" @click.prevent="toggleManaged">
```
ktuite, line 38: Could emit a `toggleManaged` event and the parent component could toggle `managed` from the outside

ktuite: Could also use the provide/inject mechanism for a ref to more easily change legacy/managed here. Not reactive by default so it would have to explicitly provide a reactive ref.

src/components/field-key/list.vue
```diff
@@ -161,7 +161,7 @@ export default {
           // which the height of the text will change.
           this.$refs.popover.update();
 
-          document.querySelector('.popover .field-key-qr-panel .switch-code')
+          document.querySelector('.popover-content-source .field-key-qr-panel .switch-code')
             .focus();
```
ktuite, line 165: This `.focus` could probably go away because it will be the same link

test/components/field-key/new.spec.js
```diff
@@ -119,7 +120,7 @@ describe('FieldKeyNew', () => {
             .afterResponse(async (app) => {
               await app.get('.field-key-row-popover-link').trigger('click');
               await app.vm.$nextTick();
-              const panel = document.querySelector('.popover .field-key-qr-panel');
+              const panel = document.querySelector('.popover-content-source .field-key-qr-panel');
```
ktuite, line 123: Might be within `app` when bootstrap is no longer involved. `app.find` instead of `document.querySelector`

test/components/field-key/row.spec.js
```diff
@@ -52,59 +52,61 @@ describe('FieldKeyRow', () => {
     it('toggles the popover', async () => {
       const app = await load('/projects/1/app-users', { attachTo: document.body });
       await app.get('.field-key-row-popover-link').trigger('click');
-      should.exist(document.querySelector('.popover .field-key-qr-panel'));
+      should.exist(document.querySelector('.popover-content-source .field-key-qr-panel'));
       await app.get('.field-key-row-popover-link').trigger('click');
-      should.not.exist(document.querySelector('.popover'));
+      //should.not.exist(document.querySelector('.popover-content-source'));
+      // todo: need a different way to check whether popover is present or not
```
ktuite, line 58: could look at the `target` prop to see that it's null again

src/components/collect-qr.vue
```diff
@@ -59,11 +59,7 @@ const imgHtml = ref('');
 const margin = 15;
 const textFontSize = 16;
 
-// The QR code is rendered to a canvas (possibly with additional icons and text)
-// and then the canvas data is converted to an image via toDataURL and used as the source
-// of an image tag.
-// It is assumed that props will not change and the QR image will only be rendered once.
-onMounted(() => {
+const renderQrCode = () => {
```
ktuite, line 62: Realized that the QR code wasn't changing when the managed/legacy settings changed, oops! 

It actually doesn't change on staging or the QA server either.

src/components/hover-card.vue
```diff
@@ -53,50 +50,6 @@ const props = defineProps({
   }
 });
 
-const width = ref('');
```
ktuite, line 56: got rid of everything where hovercard is responsible for computing its own size and position

src/components/popover.vue
```diff
@@ -84,21 +82,26 @@ export default {
...
-          container: 'body',
-          content: this.$el.innerHTML,
-          html: true,
-          placement: this.placement,
-          trigger: 'manual',
-          animation: false,
-          sanitize: false
-        })
-        .popover('show'));
+      computePosition(this.target, this.$el, {
```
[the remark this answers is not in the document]
ktuite: I looked into this a little bit but it didn't fully make sense to me. Also, there's a lot of plumbing for the tooltips that is a bit different from popovers, and a lot of things for the hovercards which make use of the popovers, too. 

Maybe the `computePlacement` part of the popover code and updating styles isn't as pretty as it could be, but at least its self-contained and popovers can be used how they were used before.

test/components/field-key/row.spec.js
```diff
@@ -89,14 +89,6 @@ describe('FieldKeyRow', () => {
         panel.classList.contains('legacy').should.be.true;
       });
 
-      it('focuses the link to switch back to a managed QR code', async () => {
```
ktuite, line 92: bring this test back?

test/components/hover-card.spec.js
```diff
@@ -15,96 +10,4 @@ describe('HoverCard', () => {
     const icon = component.get('.hover-card-heading span');
     icon.classes('icon-file').should.be.true;
   });
-
-  describe('truncateDt prop', () => {
```
ktuite, line 19: check these tests once the resizing is brought back

test/components/hover-card.spec.js
```diff
@@ -15,96 +10,4 @@ describe('HoverCard', () => {
...
-    const Parent = {
-      template: `<popover>
```
ktuite, line 21: This doesnt need to be a popover - could be nothing or a div instead.

test/components/popover.spec.js
```diff
@@ -101,29 +99,13 @@ describe('Popover', () => {
...
-    // does not.
-    const popoverComponent = component.getComponent(Popover);
-    popoverComponent.get('p').text().should.equal('baz');
-    document.querySelector('.popover p').textContent.should.equal('foo');
-    popoverComponent.vm.update();
-    document.querySelector('.popover p').textContent.should.equal('baz');
-  });
-
-  it('uses the placement prop', async () => {
+  it.skip('uses the placement prop', async () => {
```
ktuite, line 102: Remove this test

ktuite (commented):
Interactive review with @matthew-white

src/components/hover-card.vue
```diff
@@ -53,50 +50,6 @@ const props = defineProps({
...
-
-  const newWidth = computeWidth(body);
-
-  el.value.style.width = '';
-  for (const sibling of siblings) sibling.style.display = '';
-
-  if (newWidth > currentWidth)
-    // Persist the new width in width.value so that it is added to the `style`
-    // attribute. That's needed for the style to be copied to the popover.
-    width.value = px(el.value.getBoundingClientRect().width + newWidth - currentWidth);
```
ktuite, line 97: `el.value.style.width = px(el.value.getBoundingClientRect().width + newWidth - currentWidth);`

width ref doesnt need to exist, doesn't need to be a style binding?

matthew-white (approved):

src/components/hover-card.vue
ktuite: I reverted this code so the resizing behavior is back. It seems to work (truncation behavior in Swahili looks good). There is some issue with the tests, though, and checking the width. Could it have to do with not having the `popover` wrapper? The popover wrapper didn't work in this test anymore because of the v-if visibility check.

src/components/collect-qr.vue
ktuite: Issue noted here: https://github.com/getodk/central/issues/885

matthew-white:
One thing that was surprising is that the penultimate commit a2c853b44c8296623bf0d4d48ba22674aad26de8 passed locally, but failed in CircleCI. I looked into it and posted about it in Slack:

> OK, not totally sure, but I think I see the issue. The width of the viewport is 780px in CircleCI, while it’s 1200px for me locally. The `<dt>` and the `<dd>` are expected to be the same width, but the `<dt>` is 416px, and there just isn’t room for the `<dd>` to equal it. I doubt the viewport width has changed in CircleCI, so I was wondering how the test managed to pass before. I think probably it’s due to the fact that the popover was previously rendered as a direct child of the `<body>`, whereas now it’s left where it is in the DOM and not moved. I think that means that it’s constrained by the width of its parent element, which is constrained by the width of the viewport; it’s constrained even though it’s absolutely positioned. I think rendering it as a child of the body circumvented that. We could render it as a child of the body using Vue teleport, which I think would be a good idea in general. But if that’s not something that we need yet for production, I don’t think we really need to do all that yet. I think the easier thing would just be to patch the test. I can just make the `<dt>` smaller than 416px so that both the `<dt>` and the `<dd>` can fit within 780px. 416px is wider than what we’d expect the `<dt>` to be in production, I just chose some large number for testing.

test/components/hover-card.spec.js
```diff
@@ -17,14 +18,17 @@ describe('HoverCard', () => {
 
   describe('truncateDt prop', () => {
     const Parent = {
-      template: `<hover-card icon="file" :truncate-dt="truncateDt">
+      template: `<a ref="target" href="#">Some link</a>
+      <popover :target="target">
```
matthew-white: One change here is that I added a popover back to the tests. I wrote on Slack:

> In the two FIX tests, the content of the `<dd>` is just `'x'`, so the width of the `<dd>` is expected to be quite small (just large enough to fit `'x'`). But because the hover card is outside a popover, it’s growing to fit the width of the viewport. When it’s positioned absolutely, it only grows to the width it needs. So I’m thinking we can put it back in a popover or put it in another container that positions it absolutely.

PR #1125, opened 2025-01-22
Submit via Webforms

brontolosone:
Closes getodk/web-forms#23 and closes getodk/central#883

Underlying strategy: redirect the Enketo URLs for form filling, previewing, public links etc to central-frontend routes. Then decide dynamically (based on a form's settings) whether to use Webforms, or Enketo after all (in an iframe).

See [Google doc, Jan 21 meeting notes](https://docs.google.com/document/d/1AI251Wg7MIa36B94Iaql9tYWID2aCfotvkLunWElHhs/edit) for loose ends.

Depends on: #1141 should be merged first.

This branch still uses web-forms#v0.6.0 that doesn't emits submission payload, hence I have kept dead code in here. I will enable that code in the related PR when a new version of web-forms is available.

There is one unanswered question about what to do when session times out while filling out the Form, that will be handle in getodk/central#910.

Another known issue to me is that if you copy the URL from the "New Submission" button and paste it in a new browser or incognito window, you will see error message instead of login page. I don't if that's important to fix. I can fix it with getodk/central#910 if required.


main.nginx.conf
```diff
@@ -68,6 +68,44 @@ http {
...
+    # Enketo — this one will not be intercepted.
+    location /enketo-passthrough {
+      proxy_pass http://localhost:8005/-;
+      proxy_redirect off;
+      proxy_set_header Host $host;
+    }
+
+    location ~* "^/-/(?<enketoPath>single|X)/(?<enketoId>[a-zA-Z0-9]{31})" {
+      # Form fill link, public
+      if ($request_method = GET) {
```
sadiqkhoja, line 84: I think this if condition is not necessary because we only send GET request to this location

src/components/web-form-renderer.vue
```diff
@@ -0,0 +1,304 @@
...
+      {{ $t('WebformFill.previewModal.body') }}
+      <div class="modal-actions">
+        <button type="button" class="btn btn-primary" @click="previewModal.hide()">
+          {{ $t('action.close') }}
+        </button>
+      </div>
+    </template>
+  </modal>
+
+    <modal v-bind="submissionModal" hideable backdrop @hide="submissionModal.hide()">
```
sadiqkhoja, line 31: this modal can be move to separate file, but I am expecting UI changes to this so I will move it along with those changes.

src/locales/en.json5
```diff
@@ -507,5 +506,11 @@
   },
   "router": {
     "unsavedChanges": "Are you sure you want to leave this page? Your changes might not be saved."
+  },
+  "formSetting": {
+    "formTech": "Web based form technology",
+    "enableWebForms": "Use WebForms instead of Enketo",
+    "enableWebFormsExplanation": "When users fill out forms via the web (not via ODK Collect), we use Enketo by default. Its successor is \"WebForms\", which is a work in progress that you can already test.",
+    "enableWebFormsExplanationLinktext": "Before enabling this, read up on WebForms and which form features it currently supports!"
```
sadiqkhoja, line 514: I am thinking to remove these translations and enter static text since they might change

src/components/navbar.vue
```diff
@@ -96,12 +96,18 @@ export default {
 
 $border-height: 3px;
 
+:root {
```
sadiqkhoja, line 99: I don't remember why this change was made, I should revert this.

sadiqkhoja:
When Enketo is enabled, "http://central-dev.localhost:8989/projects/2/forms/first/submissions/new" link should redirect or show some message

sadiqkhoja:
redirect /f/ based url to full URLs if webForms is enabled.

src/routes.js
```diff
@@ -660,6 +659,41 @@ const routes = [
...
+  asyncRoute({
+    path: '/projects/:projectId([1-9]\\d*)/forms/:xmlFormId/submissions/new',
+    component: 'FormSubmission',
+    name: 'WebFormNewSubmission',
+    props: true,
+    loading: 'page',
+    meta: {
+      standalone: true,
+      title: () => [form.nameOrId],
+    }
```
matthew-white, line 682: I'm noticing that we're not using `validateData` to check that the user is allowed to `submission.create`. I guess we don't even fetch the project on this page. Is the plan just to have Backend validate that aspect of things, returning a 403 if the user (e.g., a project viewer) tries to create a submission when they aren't allowed to?

src/routes.js
```diff
@@ -660,6 +659,41 @@ const routes = [
...
+    name: 'WebFormEditSubmission',
+    props: true,
+    loading: 'page',
+    meta: {
+      standalone: true,
+      title: () => [form.nameOrId],
+    }
+  }),
+  asyncRoute({
+    path: '/projects/:projectId([1-9]\\d*)/forms/:xmlFormId/submissions/new',
```
matthew-white, line 674: Wouldn't a path that matches this route also match `SubmissionShow`? Technically, `new` is a valid instance ID. I'm not too worried about someone using an instance ID of `new`, but I'm confused about how this is working exactly.

src/util/load-async.js
```diff
@@ -119,6 +119,18 @@ const loaders = new Map()
...
+    '../components/form/submission.vue'
+  )))
+  .set('EnketoIframe', loader(() => import(
+    /* webpackChunkName: "component-enketo-iframe" */
+    '../components/enketo-iframe.vue'
+  )))
+  .set('WebFormRendered', loader(() => import(
+    /* webpackChunkName: "component-web-form-renderer" */
+    '../components/web-form-renderer.vue'
+  )))
```
matthew-white, line 133: I don't see these two components being loaded async anywhere. `WebFormRendered` looks misspelled here, so I doubt this key is being utilized. `FormSubmission` will be loaded async, since it's a route component and `asyncRoute()` is used, but the components that `FormSubmission` uses don't also need to be async.

src/components/web-form-renderer.vue
```diff
@@ -0,0 +1,336 @@
...
+<style lang="scss">
+@import '../assets/scss/_variables.scss';
+
+:root {
+  font-size: 16px;
+}
+html, body {
+  background-color: var(--gray-200);
+  box-shadow: none;
+}
```
matthew-white, line 294: Is it possible for the user to navigate from a page that renders `WebFormRenderer` to a page that doesn't render `WebFormRenderer`? If so, I think we need to do something so that these styles aren't applied on the page that doesn't render the `WebFormRenderer`. Something like:

```scss
html, body {
  &:has(.odk-form) {
    background-color: var(--gray-200);
    box-shadow: none;
  }
}
```

src/components/web-form-renderer.vue
```diff
@@ -0,0 +1,336 @@
...
+          // This text is the title of a dialog box / modal shown when the user presses submit button on the preview of Web Forms.
+          "title": "ODK Web Forms Preview",
+          // This text is the body of a dialog box / modal shown when the user presses submit button on the preview of Web Forms.
+          "body": "You have completed the form preview using ODK Web Forms. The Submission was not sent."
+        },
+        "submissionModal": {
+          // This text is the title of a dialog box / modal shown when the user presses submit button on Web Forms.
+          "title": "ODK Web Forms Submission",
+          // This text is the body of a dialog box / modal shown when the user presses submit button on Web Forms.
+          "body": "You have completed filling out a form using ODK Web Forms."
```
matthew-white, line 311: Capitalize "form"?

src/locales/en.json5
```diff
@@ -321,8 +321,7 @@
     "session": "Session",
     "submission": "Submission",
     "submissions": "Submissions",
-    "projectRoles": "Project Roles",
-    "formPreview": "Form Preview"
```
matthew-white, line 325: Isn't this message still used for the title of the form preview page? I think I see it being used in src/routes.js.

src/components/form/submission.vue
```diff
@@ -0,0 +1,95 @@
...
+const fetchForm = () => {
+  const [enketoId, actionType, enketoPath] = props.path.split('/');
+
+  let formUrl = '';
+  if (props.projectId && props.xmlFormId) {
+    formUrl = apiPaths.form(props.projectId, props.xmlFormId);
+  } else {
+    formUrl = `/v1/enketo-ids/${enketoId}/form`;
+    if (route.query.st) {
+      formUrl += `?st=${route.query.st}`;
```
matthew-white, line 63: I was thinking about it more, and I think I'd feel a little better if we went ahead and URL-encoded the `st` query parameter (here and elsewhere). I think it makes it easier to reason about things knowing that a user isn't able to pass additional query parameters by including an encoded ampersand in the `st` query parameter. We have a `queryString()` utility function that I think would make this easy. It'd also let you get rid of the `if`, since `queryString()` knows how to handle the `undefined` case.

```suggestion
    formUrl += query({ st: route.query.st });
```

src/components/web-form-renderer.vue
```diff
@@ -0,0 +1,336 @@
...
+        "title": "Thank you for participating!",
+        "body": "You can close this window now."
+      },
+      "editSubmissionModal": {
+        "title": "Submission successful",
+        "body": "You will now be redirected."
+      },
+      "errorModal": {
+        "title": "Submission error",
+        "body": "Your data was not submitted. Error message: {errorMessage} You can close this dialog and try again. If the error keeps happening, please contact the person who asked you to fill this form or {supportEmail}."
```
matthew-white, line 332: Capitalize "form"?

src/composables/routes.js
```diff
@@ -94,6 +94,9 @@ export default memoizeForContainer(({ router, requestData }) => {
     return `/projects/${projectId}/forms/${encodedFormId}/submissions/${encodedInstanceId}`;
   };
 
+  const editSubmissionPath = (projectId, xmlFormId, instanceId) =>
```
matthew-white, line 97: Instead of adding a new function `editSubmissionPath()`, how about adding an optional `suffix` parameter to `submissionPath()`? We do something like that for `projectPath()` and `formPath()`.

src/components/submission/metadata-row.vue
```diff
@@ -131,6 +133,9 @@ export default {
         __system.attachmentsPresent !== __system.attachmentsExpected;
     },
     editPath() {
+      if (this.form.dataExits && this.form.webFormsEnabled) {
```
matthew-white, line 136: I think `dataExits` needs to be `dataExists`. But I'm also thinking that we could consider moving the check for `form.dataExists` from `SubmissionMetadataRow` to the `v-if` in `SubmissionTable`. We already check `project.dataExists` there. Instead of checking `form.dataExists` in each row, we could wait to render the table until the form response has been received. I think either way works fine though.

matthew-white, line 136: `webFormsEnabled` - should the "F" be capitalized? I see that it's not capitalized in other places.

matthew-white:
One thing I'm wondering is if we could remove the `enketoBasePath` constant. If we always direct the user to /f/... instead of /-/..., then I don't think we need `enketoBasePath` and could just remove it from the codebase (src/util/util.js).

src/components/form/settings.vue
```diff
@@ -25,6 +25,39 @@ except according to the terms contained in the LICENSE file.
...
+            <i18n-t tag="p" keypath="webFormsSetting.description">
+              <template #formName>
+                <strong>{{ form.nameOrId }}</strong>
+              </template>
+            </i18n-t>
+            <form id="dataset-settings-form">
+              <div class="radio">
+                <label>
+                  <input v-model="webformsEnabled" name="webformsEnabled" type="radio" :value="false"
+                    aria-describedby="dataset-setting-on-receipt" :disabled="form.awaitingResponse" @change="showConfirmation()">
```
matthew-white, line 47: Now that these radio buttons just toggle the modal and don't send a request, I think they don't need `:disabled="form.awaitingResponse"`. The modal will be laid over the radio buttons, so it won't be possible for the user to interact with the radio buttons while the form request is in progress.

matthew-white, line 47: I think this `aria-describedby` was copied from another component, it can be removed here.

src/components/form/submission.vue
```diff
@@ -0,0 +1,95 @@
...
+      component.value = WebFormRenderer;
+    } else {
+      component.value = EnketoIframe;
+      bindings.value = {
+        enketoId,
+        actionType,
+        enketoPath
+      };
+    }
+  });
```
matthew-white, line 83:
```suggestion
  }).catch(noop);
```

src/components/form/settings.vue
```diff
@@ -44,40 +77,152 @@ except according to the terms contained in the LICENSE file.
...
+      form.updatedAt = data.updatedAt;
+      form.webformsEnabled = data.webformsEnabled;
+    }
+  })
+    .catch(() => {
+      webformsEnabled.value = form.webformsEnabled;
+    })
+    .finally(() => {
+      confirmModalState.value = false;
+    });
```
matthew-white, line 191: I feel like if there's an error response, the modal should stay open, and we should show the alert within it. That's what we usually do for modals. What do you think about something like this?

```suggestion
    .then(hideAndReset)
    .catch(noop);
```

src/components/form/submission.vue
```diff
@@ -0,0 +1,95 @@
...
+const bindings = ref();
+
+const fetchForm = () => {
+  const [enketoId, actionType, enketoPath] = props.path.split('/');
+
+  let formUrl = '';
+  if (props.projectId && props.xmlFormId) {
+    formUrl = apiPaths.form(props.projectId, props.xmlFormId);
+  } else {
+    formUrl = `/v1/enketo-ids/${enketoId}/form`;
```
matthew-white, line 61: Is it possible for `enketoId` to be `undefined` here? I think not, but I wanted to check, because `enketoId` is an optional prop in the `EnketoIframe` component.

src/routes.js
```diff
@@ -660,6 +659,41 @@ const routes = [
...
+    path: '/f/:path(.*)',
+    component: 'FormSubmission',
+    name: 'WebFormPublicLink',
+    props: true,
+    loading: 'page',
+    meta: {
+      standalone: true,
+      requireLogin: false,
+      title: () => [form.nameOrId]
+    }
```
matthew-white, line 694: If this route is exclusively for public links, I think we should specify `restoreSession: false` in order to reduce the number of requests and to speed up Frontend's initial load time. In that case, there should always be an `st` query parameter. This is also something we could come back to in the future.

src/components/form/submission.vue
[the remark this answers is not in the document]
matthew-white: I feel less strongly about this now that I know that `enketoId` is URL-encoded. Part of what's useful about the `apiPaths` functions is that they handle URL-encoding. Maybe it'd be unexpected to pass an argument to an `apiPaths` function where the argument is already URL-encoded.

src/components/web-form-renderer.vue
matthew-white: I left a comment about this below, but I think it'd be nice to move it as part of the current PR if that's possible.

main.nginx.conf
```diff
@@ -68,6 +68,44 @@ http {
...
+    # whether to show a WebForm or an iframed Enketo.
+    # For that iframe to work, we'll need another path prefix under which we can reach
+    # Enketo — this one will not be intercepted.
+    location /enketo-passthrough {
+      proxy_pass http://localhost:8005/-;
+      proxy_redirect off;
+      proxy_set_header Host $host;
+    }
+
+    location ~* "^/-/(?<enketoPath>single|X)/(?<enketoId>[a-zA-Z0-9]{31}|[a-zA-Z0-9]{64})$" {
```
matthew-white, line 82: Just curious, why does only this `location` have `[a-zA-Z0-9]{64}`? The `location`s below all seem to just have `[a-zA-Z0-9]{31}`.

src/components/web-form-renderer.vue
```diff
@@ -0,0 +1,336 @@
...
+
+<i18n lang="json5">
+  {
+    "en": {
+      "WebformFill": {
+        "previewModal": {
+          // This text is the title of a dialog box / modal shown when the user presses submit button on the preview of Web Forms.
+          "title": "ODK Web Forms Preview",
+          // This text is the body of a dialog box / modal shown when the user presses submit button on the preview of Web Forms.
+          "body": "You have completed the form preview using ODK Web Forms. The Submission was not sent."
```
matthew-white, line 305: Capitalize "form"?

test/components/form/submission.spec.js
```diff
@@ -0,0 +1,51 @@
...
+      ]);
+  });
+
+  it('sends the correct initial requests - Enketo', () => {
+    testData.extendedForms.createPast(1, { xmlFormId: 'a' });
+    return load('/f/enketo-id/new?st=token')
+      .testRequests([
+        { url: '/v1/enketo-ids/enketo-id/form?st=token' }
+      ]);
+  });
```
matthew-white, line 27: How about adding one or more tests that showcase the requests to /v1/test/... endpoints?

src/components/web-form-renderer.vue
matthew-white: Ah never mind, my other comment was about the new modal in `FormSettings` for toggling `webformsEnabled`. I wanted to move that modal into its own component because it's more complicated, e.g., it sends a request. The modal here is a different one though and is just static text, so I think it's OK to leave it where it is for now.

src/components/form/preview.vue
```diff
@@ -69,70 +40,17 @@ const props = defineProps({
...
 
 const fetchForm = () => {
-  Promise.allSettled([
-    form.request({ url: apiPaths.form(props.projectId, props.xmlFormId), extended: true, alert: false }),
-    formVersionXml.request({ url: apiPaths.formXml(props.projectId, props.xmlFormId, props.draft) })
-  ]);
+  form.request({
+    url: apiPaths.form(props.projectId, props.xmlFormId),
+    extended: true,
+    alert: false
```
matthew-white, line 49: I know it was this way already, but I don't remember why we're specifying `alert: false`. If someone goes to preview a form that's been deleted, or if they don't have auth, then I'd expect them to see an error message. With `alert: false`, I think they won't.

Previously, we sent two requests. The second request, for `formVersionXml`, would show an alert: we didn't specify `alert: false` for that request. However, now `WebFormRenderer` is responsible for requesting `formVersionXml`, and we only render it if `form.dataExists`. That means that before, there would have been at least one request without `alert: false`, which would show an error message for a 404/401/403. However, now, there might not be such a request: it could be that only the request for `form` is sent. Given that, I think there's reason for us to consider removing `alert: false` now.

src/components/form/preview.vue
matthew-white, line 1: The i18n messages in this file are out-of-date / will no longer be used, right? If so, it'd be great if you could remove both `<i18n>` blocks below (both source messages and translations).

src/components/web-form-renderer.vue
```diff
@@ -0,0 +1,336 @@
...
+ * Web Form expects host application to provide a function that it can use to
+ * fetch attachments. Signature of the function is (url) => Response; where
+ * Response is subset of web standard  {@link Response}.
+ */
+const getAttachment = (url) => {
+  let requestUrl = apiPaths.formAttachment(
+    form.projectId,
+    form.xmlFormId,
+    form.draftToken !== null,
+    url.pathname.substring(1)
```
matthew-white, line 126: Before, this was `url.pathname.split('/').pop()`, which is different from `url.pathname.substring(1)`. Did something change in the `url` that Web Forms emits?

src/components/web-form-renderer.vue
```diff
@@ -0,0 +1,336 @@
...
+      }
+      Object.assign(primaryInstance, { instanceId: data.instanceId, uploadSuccess: true });
+      return true;
+    })
+    .catch(noop);
+};
+
+const uploadAttachment = async (attachmentIndex) => {
+  const attachmentDescriptor = attachmentUploads[attachmentIndex];
+  let url = apiPaths.submissionAttachment(form.projectId, form.xmlFormId, false, primaryInstance.instanceId, attachmentDescriptor.file.name);
```
matthew-white, line 202: The third argument to `apiPaths.submissionAttachment()` doesn't look right to me. Isn't it possible for this to be a form draft?

src/components/web-form-renderer.vue
```diff
@@ -0,0 +1,336 @@
...
+    // eslint-disable-next-line no-unused-vars
+    const { data, definition, status, violations } = payload;
+    if (status !== 'ready') {
+      // Status is not ready when Form is not valid and in that case submit button will be disabled,
+      // hence this branch should never execute.
+      return;
+    }
+
+    primaryInstance.file = data.instanceFile;
+    data.attachments.forEach((file, ix) => { attachmentUploads[ix] = { file, progress: null, uploadSuccess: null }; });
```
matthew-white, line 265: Don't we need to resize the array at some point? Alternatively, maybe we could use `attachmentUploads.push()` here. It seems to me that it's possible that `attachmentUploads` is an empty array here, in which case I don't think we should directly set an index on it like what's being done here.

matthew-white (approved):
Super excited to add submission via Web Forms to Frontend! 🚀

I've added a bunch of comments, but I'm ready to approve the PR. ✅ Most of my comments are pretty small. I also think that it'll help a lot to get a first version of this on staging so that we can start trying it out.

Central supports a number of different submission scenarios, and I didn't think through each scenario independently. For example, I didn't think through whether form drafts work in all cases (e.g., when the form has a published version vs. not), or whether single submission public links need to do something special compared to other public links. I figure that if we don't handle a particular scenario correctly, we'll likely discover that during testing.

This PR doesn't add many tests. @sadiqkhoja was telling me that he's considering writing Puppeteer tests to test the full flow, which involves Backend, Enketo, and nginx. That makes sense to me, and I think it's OK for tests to come later as long as we get those in before release. I also think it's possible that during the process of writing the Puppeteer tests, we'll decide that certain tests would work fine or even be easier as normal Frontend tests. For example, I think many tests that render ODK Web Forms would probably work fine as Frontend tests, since the Vue app is what's responsible for rendering Web Forms. But where Enketo needs to be served, I think it makes a lot of sense for most of those tests to be Puppeteer tests.

I expect there to be a number of small changes to this PR before it's merged. I don't feel like I need to re-review those. However, if there's any large change, I should probably review that before the PR is merged. @sadiqkhoja, could you let me know if you make a large change or any other change that you think I should review? I'm more than happy to take a look at anything that you think would be helpful.

Another type of change that I'd like to be sure to review is anything that touches the router. For example, I'm interested in changes made for the following comments:

> When Enketo is enabled, "http://central-dev.localhost:8989/projects/2/forms/first/submissions/new" link should redirect or show some message

> redirect /f/ based url to full URLs if webForms is enabled.

> If this route is exclusively for public links, I think we should specify `restoreSession: false` &hellip;

@sadiqkhoja, could you let me know when you make changes that touch the router? Alternatively, I think we could probably wait to make those changes in a follow-up PR that I review separately.

Very excited for this to get merged! 👏

sadiqkhoja:
Thanks @matthew-white for the thorough review as always 🙏.

I have incorporated all the feedback. I have tried to group related changes in separate commits. Following are the commits which I think you should have a look:

[2a71cc](https://github.com/getodk/central-frontend/pull/1125/commits/b2a71cc72b2e6daa36965c42ec566cc68c3ed1c7) Modified WebFormRenderer

[4b3bb27](https://github.com/getodk/central-frontend/pull/1125/commits/4b3bb27ed44559f801aba0bed61b12fd4093c1e1) Modified routes.js

[e969c24](https://github.com/getodk/central-frontend/pull/1125/commits/e969c24c8ed1d6a1d3fe6e63a8bae0c2a87e4cd3) change route if logged in user go to /f/:enketoId or /f/:enketoId/edit

src/components/web-form-renderer.vue
```diff
@@ -289,28 +236,21 @@ const handleSubmit = async (payload) => {
   font-size: 16px;
 }
 html, body {
-  background-color: var(--gray-200);
-  box-shadow: none;
+  &:has(.odk-form) {
```
matthew-white, line 239: I think we probably need to add this selector to `:root` above as well so that the font size on `:root` isn't applied on all pages.

src/routes.js
```diff
@@ -678,13 +686,16 @@ const routes = [
     loading: 'page',
     meta: {
       standalone: true,
+      validateData: {
+        project: () => project.permits(['form.read', 'submission.create'])
```
matthew-white, line 690: > `'form.read'`

What about Data Collectors? Would they ever navigate to this route? I believe they only have the `open_form.read` verb, not `form.read`.

If it's an issue for Frontend to check `form.read`, I actually feel like it's less important for Frontend to check `form.read` compared to `submission.create`. If the user doesn't have access to the form, then the request for the form will return a 403 and the user won't ever see the form or fill it out. I know we check `form.read` in a bunch of routes, but even in those routes, we generally fire the form request before checking the `form.read` verb (since we might not have the project response yet). I'm not sure that the check of `form.read` is ever doing that much.

That said, if you did want to validate this in Frontend, one thing that comes to mind is that `validateData` only validates a single resource at a time. You could add a watch effect to `FormSubmission` for more complex validation that checks `project` and `form` together, something like:

```js
watchEffect(() => {
  if (!(project.dataExists && form.dataExists)) return;
  // The user can read all forms in the project; no additional validation needed.
  if (project.permits('form.read')) return;
  if (!project.permits('open_form.read') || form.state == 'closed') router.push('/');
});
```

src/components/form/submission.vue
```diff
@@ -79,7 +85,16 @@ const fetchForm = () => {
   }).catch(noop);
 };
 
-fetchForm();
+// To facilate validateData in route guard
+if (props.projectId) {
+  fetchProject().then(() => {
+    fetchForm();
```
matthew-white, line 103: Is there a way to send these two requests at once, but hide the component until the project response is received? I feel like we're setting up a series of dependent requests, which could slow down page load:

- The project is requested
- The form is requested only once the project response is received
- The form XML is requested only once the form response is received

I'm thinking of something like:

```vue
<template>
  <loading :state="initiallyLoading"/>
  <component :is="component" v-if="dataExists && component != null" v-bind="bindings"/>
<template>

<script setup>
// ...

const { initiallyLoading, dataExists } = resourceStates(props.projectId ? [project, form] : [form]);

if (props.projectId) fetchProject();
fetchForm();
</script>
```

src/router.js
```diff
@@ -194,6 +194,7 @@ reused after a route change, but that shouldn't happen given our use of the
           // used before the navigation is confirmed, for example, to update the
           // DOM. Given that, we clear the data immediately.
           resource.data = null;
+          router.currentRoute.value.meta.standalone = false;
```
matthew-white, line 197: Why is this needed? Won't `router.currentRoute.value` change after the next line, which ultimately navigates to either / or /login? Both of those routes have `standalone: false`, so I feel like this change to the value of `meta.standalone` should happen automatically. I'm sure I'm missing some aspect of this though.

src/components/form/submission.vue
```diff
@@ -16,17 +16,18 @@ except according to the terms contained in the LICENSE file.
...
-import EnketoIframe from '../enketo-iframe.vue';
 import Loading from '../loading.vue';
 import useForm from '../../request-data/form';
 import { apiPaths, queryString } from '../../util/request';
 import { noop } from '../../util/util';
+import useProject from '../../request-data/project';
+import { loadAsync } from '../../util/load-async';
 
 const route = useRoute();
+const { project } = useProject();
```
matthew-white, line 34: `useProject()` has a little bit of a deceptive name. It creates project subresources used in `ProjectShow`, not the actual project resource. Certain resources are created upfront and are not local resources, including `project` and `form`. A resource can't be a local resource if it's referenced in src/routes.js, which these resources are (for page titles, I think).

Instead of calling `useProject()`, you can just grab `project` off `useRequestData()`.

Maybe `useProject()` should be renamed to `useProjectSubresources()`. 🤔

src/components/form/submission.vue
```diff
@@ -73,7 +78,14 @@ const fetchForm = () => {
       (problem.code === 404.1 ? t('formNotFound') : null)
   }).then(() => {
     if (form.data.webformsEnabled) {
-      // TODO: maybe change the route if it is /f/... and it is for "New / Edit Submission"
+      // change the route if it is /f/... and it is for "New / Edit Submission"
+      if (route.path.startsWith('/f/')) {
+        if (!actionType && !form.data.draftToken) {
+          router.replace(formPath(form.data.projectId, form.data.xmlFormId, 'submissions/new'));
+        } else if (actionType === 'edit' && route.query.instance_id) {
```
matthew-white, line 85: This is probably a question for later, but how would Web Forms work for editing if there's no `instance_id` query parameter? Instead of trying to render `WebFormRenderer`, maybe we should show an alert or render `NotFound` in that case.

matthew-white (approved):
I took a look at those three commits. They're looking good to me!

matthew-white:
I was thinking more about the case where a form is published and active, and someone creates a new form draft. Anyone with the link should be able to preview and submit to the draft, but I don't see how that'd be possible with the current setup. I think the form response doesn't include a `draftToken` in that case, so Frontend won't send requests to the /v1/test/... endpoints. I think that can be accounted for in a follow-up PR, but I thought I'd mention this case.

lognaturel:
> Anyone with the link should be able to preview and submit to the draft, but I don't see how that'd be possible with the current setup. 

To keep a record of parallel conversations we're having: I don't think that was a deliberate feature. I'd like to advocate for removing it (keeping the behavior in this PR) and if folks request the old behavior we can decide how to bring it back in a thoughtful way.

sadiqkhoja:
In this PR draft preview and draft submission are broken when web-form is enabled because both are pointing to `/f/:enketoId`. I have created a separate issue to fix that https://github.com/getodk/central/issues/932

src/components/enketo-iframe.vue
```diff
@@ -0,0 +1,93 @@
...
+});
+
+function handleIframeMessage(event) {
+  let eventData;
+  try { eventData = JSON.parse(event.data); } catch {}
+
+  if (event.origin === window.location.origin &&
+    eventData?.enketoEvent === 'submissionsuccess' &&
+    redirectUrl.value) {
+    router.push((new URL(redirectUrl.value)).pathname);
```
matthew-white, line 80: I just started wondering this: Do users ever specify their own `return_url` query parameter? Do they ever need to redirect outside of Frontend? I'm happy with how this is implemented currently: I like that we're keeping the user inside the Vue app and not redirecting them to arbitrary URLs. But I just wanted to check about this requirement.

test/components/form/submission.spec.js
```diff
@@ -4,60 +4,137 @@ import { mockLogin } from '../../util/session';
...
+    describe('data collector', () => {
+      beforeEach(() => {
+        mockLogin({ role: 'none' });
+        testData.extendedProjects.createPast(1, { role: 'formfill' });
+        testData.extendedForms.createPast(1, { xmlFormId: 'a', webformsEnabled: true });
+      });
+
+      it('can access new submission page', () =>
+        load('/projects/1/forms/a/submissions/new')
+          .respondWithData(() => simpleXml)
```
matthew-white, line 83: You could get `load()` to return `simpleXml` automatically if that'd help at all with these tests. Even if the XML is only returned in a subset of tests, it's now possible to tell `load()` to return a response conditionally rather than all the time (there are examples of this in test/util/http/data.js).

matthew-white (approved):
I took a look at the latest commit 74c02ee01749b7e8227460fac7bdadbc5a3c1d9b, and it's looking good to me. 🚀

src/components/enketo-iframe.vue
matthew-white: Sounds good. :+1: @sadiqkhoja, I sort of like the idea of addressing that in a follow-up PR so we can get this one merged sooner rather than later. What do you think about filing an issue for that? I'm happy to review it here though if doing it as part of this PR seems easier to you.

matthew-white: I've filed an issue to track this at getodk/central#942.

PR #1126, opened 2025-01-24
Reordering tests causes them to pass

matthew-white:


sadiqkhoja:
I have also tried re-ordering earlier, tests pass but I see following dialog box on my computer, which goes away after all test execution:

image

matthew-white:
Yeah, this would just fix the test, not solve #1047. (I think that dialog box is related to #1047.) For #1047, we'd need something additional like `preventDefault()` or calling `setFilename()`. I wouldn't want to fix the test just by reordering. Mostly I just wanted to make a note that reordering caused the test to pass, since that seems quite surprising to me.

PR #1127, opened 2025-01-24
Features/822 pagination

sadiqkhoja:
Closes getodk/central#822

Wrote some component tests and manually verified it. I am thinking to pass this to QA team even before we review/merge this PR.

We talked about what to do when a row is deleted, should we just remove the row from the UI or we fetch the next available row and update the count/page size. We decided that it is better to just remove the row.

**Transition Group**

I have used Vue's Transition Group (TG) component in our table-freeze component. I think this has simplified our code because model and view are alway in sync; and we no longer keep deleted row in the DOM hidden. 

I am not using css class provided by TG component and kept our transition logic in markRowChanged and markRowDeleted. Because TG doesn't help with row modified, so thought might as well keep both change/delete consistent. The only magic of TG is that it waits for `transitionend` event before removing row from the table, that event is raised when `transition` started by css property is completed.

**Snapshot Filters**

I am passing filters to odata endpoint to fetch the data for a single point in time so that submission/entity created or deleted after fetching first page don't effect our total count and number of pages. 

For alive records, those filters are records created before `now` and not deleted or deleted after `now`.
For deleted records, the filter is to fetch records that are deleted before `now`.

I am keeping a set of removed rows (instanceId or UUID), so I don't add those rows in resource store when page is changed.

Only weird case is where a user is looking at alive records and some other user undeletes a record, then next odata request will have different `count` and number of pages can change. But I think it's not super problematic.

No

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,7 @@
         "@vue/cli-service": "^5.0.8",
         "@vue/compiler-sfc": "~3",
         "@vue/eslint-config-airbnb": "^7.0.0",
-        "@vue/test-utils": "^2.4.1",
+        "@vue/test-utils": "^2.4.6",
         "chai": "^5.1.1",
         "chai-as-promised": "^8.0.0",
         "comment-json": "~4",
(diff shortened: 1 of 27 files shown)


matthew-white (approved):
Looks great! 🎉

PR #1129, opened 2025-01-24
Add preventDefault + hack to avoid karma throwing some of your tests …

### LINKED ISSUE #1047: Tests create a new "trees 20241231012345.csv" file in "~/Downloads"

sadiqkhoja:
Closes https://github.com/getodk/central-frontend/issues/1047 and makes all test pass which are currently failing.

All tests are passing and no file is created on the file system.

`preventDefault` fixes #1047.

**Fix for failing test:**

This is mysterious fix, I don't exactly know the root cause. The issue seems related to https://github.com/karma-runner/karma/issues/3887 because when I ran only two tests in `data-template.spec.ts`, I saw this error "Some of your tests did a full page reload!".

One suggestion over there is to add following:

```
beforeEach(async () => {
    window.onbeforeunload = () => "Oh no!";
});
```

So I added that and it worked. It didn't make sense to me so I removed the `onbeforeunload`, just kept `beforeEach(async () => {};` and it worked as well. That's what this PR contains.

Alternative solution: https://github.com/getodk/central-frontend/pull/1122

None

No

### DIFF
--- a/test/components/entity/upload/data-template.spec.js
+++ b/test/components/entity/upload/data-template.spec.js
@@ -12,6 +12,9 @@ const mountComponent = () => mount(EntityUploadDataTemplate, {
 });
 
 describe('EntityUploadDataTemplate', () => {
+  // hack: without this 'has the correct filename' fails. Possibly because of karma#3887
+  beforeEach(async () => {});
+
   it('has the correct data URL', () => {
     testData.extendedDatasets.createPast(1, {
       properties: [{ name: 'hauteur' }, { name: 'circonférence' }]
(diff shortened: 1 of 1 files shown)


matthew-white (approved):
Very strange that this works, but I'm glad that it does! I like how this looks. 🎉 Thank you for the fix and for looking into this!

PR #1139, opened 2025-02-24
Combine current form draft pages in new FormEdit component

matthew-white:
This PR makes progress on getodk/central#728. The main goal of this PR is the creation of the "Edit Form" page (rendered by the `FormEdit` component).

- The .../draft route will now render `FormEdit` as the route component, not `FormDraftStatus`.
- There are no longer separate routes for .../draft/attachments or .../draft/testing.
- `FormEdit` contains those three previous pages for form drafts: `FormDraftStatus`, `FormAttachmentList`, and `FormDraftTesting`.
- `FormEdit` also contains the "Create a new Draft" button that used to be in `FormHead`.
- The removal from `FormHead` of the draft tabs and the create draft button means that `FormHead` is no longer split in two halves.
- With this PR, project managers and sitewide administrators will always be able to navigate to the "Edit Form" page. If the form has a draft, then the page will show information about the draft. If the form doesn't have a draft, then the page will show the create draft button. Previously, the routes for a form draft were only accessible if a draft existed. This change means that the router no longer needs access to the `formDraft` and `attachments` resources. That allows us to make them local resources. I've moved them to `useForm()` so that they're together with local resources for other form subresources.

I updated existing tests and added tests of the new `FormEditCreateDraft` component. I also tried things out locally.

There's plenty left to do for getodk/central#728. However, I think it's useful to make these big overarching changes first. Once this PR is merged, routes will be correct, many existing tests will be updated, and `formDraft` and `attachments` will be local resources. From there, it'll be possible to continue iterating on the "Edit Form" page.


ktuite (approved):

matthew-white (commented):

PR #1141, opened 2025-02-26
Features C#906: Use web history instead of hash for routing

sadiqkhoja:
Closes getodk/central#906

Manual verification.

This change is prompted by the need to have cleaner and shorter public access link as part of web-forms integration. Since we control the server that serves frontend application we can easily use webhistory mode for routes. Even vue-router [recommends](https://router.vuejs.org/guide/essentials/history-mode.html#HTML5-Mode) web history mode. 

I have also added a small script in the index.html to redirect old hash based URLs to the newer ones, this will keep old bookmarked links (if any) still functional for the users.

#### How does this change affect users? Describe intentional changes to behavior and behavior that could have 
accidentally been affected by code changes. In other words, what are the regression risks?

None.

No

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -22,6 +22,13 @@ except according to the terms contained in the LICENSE file.
       <p>Please enable JavaScript to continue.</p>
       <p>Powered by <a href="https://getodk.org">ODK</a>.</p>
     </noscript>
+    <script>
+      // We used to have hash-based navigation, we have now switched to web-history-based
+      // navigation. To support, bookmarked links, we are redirecting old URL to the new one.
+      if (window.location.href.startsWith(window.location.origin + "/#")) {
+          window.location.replace(window.location.origin + window.location.hash.substring(1));
+      }
+    </script>
     <div id="app"></div>
     <!-- built files will be auto injected -->
   </body>
--- a/public/index.html
+++ b/public/index.html
@@ -22,13 +22,6 @@ except according to the terms contained in the LICENSE file.
       <p>Please enable JavaScript to continue.</p>
       <p>Powered by <a href="https://getodk.org">ODK</a>.</p>
     </noscript>
-    <script>
-      // We used to have hash-based navigation, we have now switched to web-history-based
-      // navigation. To support, bookmarked links, we are redirecting old URL to the new one.
-      if (window.location.href.startsWith(window.location.origin + "/#")) {
-          window.location.replace(window.location.origin + window.location.hash.substring(1));
-      }
-    </script>
     <div id="app"></div>
     <!-- built files will be auto injected -->
   </body>
--- a/src/router.js
+++ b/src/router.js
@@ -9,7 +9,7 @@ https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
 including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 */
-import { START_LOCATION, createRouter, createWebHashHistory } from 'vue-router';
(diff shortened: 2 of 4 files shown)


public/index.html
```diff
@@ -22,6 +22,13 @@
       <p>Please enable JavaScript to continue.</p>
       <p>Powered by <a href="https://getodk.org">ODK</a>.</p>
     </noscript>
+    <script>
+      // We used to have hash-based navigation, we have now switched to web-history-based
+      // navigation. To support, bookmarked links, we are redirecting old URL to the new one.
+      if (window.location.href.startsWith(window.location.origin + "/#")) {
+          window.location.replace(window.location.origin + window.location.hash.substring(1));
```
matthew-white, line 29: Is there an advantage in doing this here rather than somewhere else in the app? I was wondering about a couple of other possibilities:

**Idea 1**

Could we do this in src/main.js instead? Before Vue mounts the app, main.js could complete this check and redirect if need be. I kind of like the idea of keeping public/index.html as simple as possible and free of JavaScript.

**Idea 2**

Maybe we could even do this in Vue Router? Would it be preferable to avoid a true redirect?

```js
router.beforeEach(to => (to.path === '/' && to.hash.startsWith('/')
  ? to.hash
  : true));
```

sadiqkhoja: went with the idea 2

src/router.js
```diff
@@ -135,6 +135,9 @@ router.afterEach(unlessFailure(to => {
     };
   });
 
+  // We used to have hash-based navigation, we have now switched to web-history-based
+  // navigation. To support, bookmarked links, we are redirecting old URL to the new one.
+  router.beforeEach(to => (to.path === '/' && to.hash.startsWith('#/') ? to.hash.substring(1) : true));
```
matthew-white, line 140: What do you think about validating that the hash is a valid URL? Or is that something we can trust Vue Router to do? If we were to do the validation ourselves, that might look something like this:

```js
router.beforeEach(to => {
  if (to.path === '/' && to.hash.startsWith('#/')) {
    const path = to.hash.substring(1);
    try {
      new URL(path, window.location.origin);
    } catch (error) {
      // Just ignore the hash if it's not a valid URL.
      return '/';
    }
    return path;
  }
  return true;
});
```

matthew-white, line 140: I'm wondering about the order of the `router.beforeEach()` navigation guards. I'm thinking that this should maybe be the first navigation guard, even before the initial requests. The navigation guard for the initial requests decides whether to send a request to restore the session based on where the user is navigating to. But that won't be determined until after this new navigation guard that redirects based on the hash. For example, navigating to /#/account/claim shouldn't restore the session (though maybe we'll change that based on getodk/central#916). But I think right now, it would restore the session, because the navigation guard for the initial requests would think that the user is navigating to the homepage, not the Set Password page.

I think we could probably move this entire redirect section to before the navigation guard for the initial requests. Maybe it's nice to keep all the redirects grouped together in one place.

matthew-white, line 140: How about adding a test of this navigation guard? Something like:

```js
it('redirects if the hash is a path', () => {
  mockLogin();
  return load('/#/account/edit', {}, false)
    .respondFor('/account/edit')
    .afterResponses(app => {
      app.vm.$route.path.should.equal('/account/edit');
    });
});
```

sadiqkhoja: I think it is redundant. If the string after # is not a valid path then user will get 404 Not Found, I guess which is okay because we are adding this navigation guard for the bookmarks and the saved link, those will be most likely valid.

sadiqkhoja: moved

matthew-white (approved):

PR #1153, opened 2025-03-21
Changing text for convert pending submission modal

ktuite:
Closes https://github.com/getodk/central/issues/837

I found the second sentence in the original modal too long/hard to reason about with the future/past negatives so I made it simpler.

> Once this takes effect, Entities will no longer be generated when Submissions are marked Approved, including {records} we found that has neither been marked Approved nor Rejected.

Screenshot 2025-03-21 at 2 03 17 PM

### DIFF
--- a/src/components/dataset/pending-submissions.vue
+++ b/src/components/dataset/pending-submissions.vue
@@ -30,7 +30,7 @@ except according to the terms contained in the LICENSE file.
               <strong>{{ $t('dontConvert.label') }}</strong>
             </label>
             <p id="dataset-auto-convert-false" class="help-block">
-              {{ $tcn('dontConvert.description', pendingSubmissions) }}
+              {{ $t('dontConvert.description', pendingSubmissions) }}
             </p>
           </div>
           <div class="radio">
@@ -40,7 +40,7 @@ except according to the terms contained in the LICENSE file.
               <strong>{{ $t('convert.label') }}</strong>
             </label>
             <p id="dataset-auto-convert-true" class="help-block">
-              {{ $tcn('convert.description', pendingSubmissions) }}
+              {{ $t('convert.description', pendingSubmissions) }}
             </p>
           </div>
         </form>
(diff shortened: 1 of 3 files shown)


src/components/dataset/pending-submissions.vue
```diff
@@ -91,17 +91,17 @@ const convert = ref(null);
     "explanation": {
       "userAction": "You are setting Entity creation to occur when Submissions are first received by Central.",
       "implication": {
-        "full": "Once this takes effect, Entities will no longer be generated when Submissions are marked Approved, including {records} we found that has neither been marked Approved nor Rejected. | Once this takes effect, Entities will no longer be generated when Submissions are marked Approved, including {records} we found that have neither been marked Approved nor Rejected.",
-        "records": "{count} record | {count} records"
+        "full": "You currently have {records} not marked Approved nor Rejected.",
```
matthew-white, line 94: I think this still needs to be pluralized like it was before even though the message doesn't use `{count}`. In some languages, the word "marked" may have to change to match whether "record" is pluralized.

src/components/dataset/pending-submissions.vue
```diff
@@ -40,7 +40,7 @@ except according to the terms contained in the LICENSE file.
               <strong>{{ $t('convert.label') }}</strong>
             </label>
             <p id="dataset-auto-convert-true" class="help-block">
-              {{ $tcn('convert.description', pendingSubmissions) }}
+              {{ $t('convert.description', pendingSubmissions) }}
```
matthew-white, line 43: Now that it's just static text, we don't need to pass the count `pendingSubmissions` to `$t()`:

```suggestion
              {{ $t('convert.description') }}
```

src/components/dataset/pending-submissions.vue
```diff
@@ -91,17 +91,17 @@ const convert = ref(null);
...
-        "full": "Once this takes effect, Entities will no longer be generated when Submissions are marked Approved, including {records} we found that has neither been marked Approved nor Rejected. | Once this takes effect, Entities will no longer be generated when Submissions are marked Approved, including {records} we found that have neither been marked Approved nor Rejected.",
-        "records": "{count} record | {count} records"
+        "full": "You currently have {records} not marked Approved nor Rejected.",
+        "records": "{count} pending record | {count} pending records"
       }
     },
     "dontConvert": {
       "label": "I understand and this is not a problem for me.",
-      "description": "Change the setting and do nothing with the pending Submission. | Change the setting and do nothing with the pending Submissions."
+      "description": "Change the setting and do nothing with the pending Submission."
```
matthew-white, line 100: Does this need to change? I think it wouldn't read quite right with "the pending Submission" (singular) when there's more than one pending submission.

If you wanted to make it static text and not pluralizable, I think always using "Submissions" in the plural would work a little better, something like:

```suggestion
      "description": "Change the setting and do nothing with pending Submissions."
```

matthew-white:
> I found the second sentence in the original modal too long/hard to reason about with the future/past negatives so I made it simpler.

The new text looks reasonable to me. :+1: Might be good to get a double-check from Hélène on that change before translations get updated.

matthew-white (approved):

PR #1155, opened 2025-03-21
Let project viewers request dataset-diffs for infonav button

ktuite:
Frontend part of issue https://github.com/getodk/central/issues/897
Backend: https://github.com/getodk/central-backend/pull/1456

I was reading through the text on the issue, but thinking how these requests aren't really relevant for unpublished forms, so we still need to check the form settings.

> Once we change Backend, we should change Frontend so that it always sends those requests.

Trying it out, tests.

See reasoning in issue.

Project viewers will get infonav buttons for related entity lists only (not app users).

no

### DIFF
--- a/src/components/form/show.vue
+++ b/src/components/form/show.vue
@@ -99,8 +99,8 @@ const stopAppUsersEffect = watchEffect(() => {
   nextTick(() => stopAppUsersEffect());
 });
 const stopDatasetsEffect = watchEffect(() => {
-  if (!(project.dataExists && form.dataExists)) return;
-  if (project.permits(['form.update', 'dataset.list']) && form.publishedAt != null)
+  if (!(form.dataExists)) return;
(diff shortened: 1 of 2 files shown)


src/components/form/show.vue
```diff
@@ -99,8 +99,8 @@ const stopAppUsersEffect = watchEffect(() => {
   nextTick(() => stopAppUsersEffect());
 });
 const stopDatasetsEffect = watchEffect(() => {
-  if (!(project.dataExists && form.dataExists)) return;
-  if (project.permits(['form.update', 'dataset.list']) && form.publishedAt != null)
+  if (!(form.dataExists)) return;
```
matthew-white, line 102: Might be able to get rid of one set of parens?

```suggestion
  if (!form.dataExists) return;
```

matthew-white (approved):

PR #1158, opened 2025-03-24
Change info titles on submission and entity detail pages

ktuite:
Part of https://github.com/getodk/central/issues/924 

Entity and submission detail pages:

- Updated font sizes
- Add icons (entity or submission) beside first content header and change header name to include whether or not it is an "Entity" or "Submission" Details page
- Update other headers (just "Data" and "Activity")

Screenshot 2025-03-24 at 4 07 00 PM

Screenshot 2025-03-24 at 4 09 30 PM


src/components/entity/show.vue
```diff
@@ -166,6 +166,12 @@ const breadcrumbLinks = computed(() => [
 
 </script>
 
+<style lang="scss">
+  .page-section-heading {
+    font-size: 24px;
```
matthew-white, line 171: Is this style meant to be applied on all pages? If so, I think it should be added to `PageSection`. If it's only for this page, I think it needs to be scoped, e.g., add `id="entity-show"` to the root element, then make the selector `#entity-show .page-section-heading`. Right now, I think if you navigate to this page, then navigate in the same browser tab to another page that renders a `.page-section-heading`, this style will apply on that other page as well.

src/components/submission/show.vue
```diff
@@ -202,6 +202,12 @@ export default {
 };
 </script>
 
+<style lang="scss">
+  .page-section-heading {
+    font-size: 24px;
```
matthew-white, line 207: Similar to the comment above, I think this probably needs to be scoped.

matthew-white (approved):

PR #1160, opened 2025-03-28
Draft preview and draft submission requires logged in user + some refactoring and tests

sadiqkhoja:
Closes getodk/central#932

**Changes:**

- Created EnketoRedirector component, which redirects user from old enketo url to canonical url i.e. from `/f/:enketoId/new` to `/projects/:id/forms/:name/submissions/new`, similarly for `edit`, `offline` and `preview`
- Modified route paths to be specific instead of wildcard matching 
- Updated "New" submission button href to the canonical url
- Removed the dynamic `component` from the FormSubmission and added components in the template with `defineAsyncComponent`. Same change for FormPreview component
- Added several tests for EnketoIframe, EnketoRedirector, FormSubmission

Added tests, and manual verification

All these changes makes the component organization clean and easier to reason about.

Everything related to submission creation, draft submission and draft preview should be tested.


src/routes.js
```diff
@@ -805,8 +827,16 @@ const routesByName = new Map();
...
+  [
+    'NewSubmission',
+    'NewDraftSubmission',
+    'EditSubmission',
+    'FormPreview',
+    'DraftFormPreview'
+  ].forEach(redirectTo => {
+    const preserve = (_, from) => from.name === 'EnketoRedirector' && [form];
+    routesByName.get(redirectTo).meta.preserveData.push(preserve);
+  });
```
sadiqkhoja, line 839: can't use `preserveDataBetweenRoutes` because it add preserveData for all the routes provided as first argument. We want to preserveData one way not between all.

test/util/router.js
```diff
@@ -74,7 +74,7 @@ export const mockRouter = (location = undefined) => (container) => {
       app.component('RouterView', RouterViewStub);
     }
   };
-  for (const prop of ['getRoutes', 'hasRoute', 'resolve'])
+  for (const prop of ['getRoutes', 'hasRoute', 'resolve', 'push'])
```
sadiqkhoja, line 77: added to test `redirects on submissionsuccess message with return_url`, is there a better way to test that?

test/components/form/submission.spec.js
```diff
@@ -1,9 +1,20 @@
...
 
 describe('FormSubmission', () => {
+  beforeAll(() => {
+    // Mock WebFormRenderer - loading the real component creates dependency between tests because
+    // it is loaded asynchronously
+    setLoader('WebFormRenderer', async () => ({
+      default: { foo: 'bar' },
+      template: '<div class="odk-form">dummy renderer</div>'
+    }));
+  });
```
sadiqkhoja, line 16: When I was using real WebFormRenderer, the first test was sending Form XML request after `WebFormRenderer` was async-loaded but subsequent tests were sending Form XML request at the mount time. Because of that first test became dependency for the rest of the test.  Since `WebFormRenderer` is loaded via FormPreview and FormSubmissions components, it became unreliable to predict which test suite will run first.

src/components/web-form-renderer.vue
sadiqkhoja, line 1: maybe remove `withToken` function

test/util/router.js
matthew-white: I have two ideas:

1. Usually if `router.push()` needs to be called, that's a sign that a real router is needed and not `mockRouter()`. `mockRouter()` is just supposed to do basic, read-only things on a single page. So usually that means using `load()`.
2. Maybe it'd be fine to use `mockRouter()` in this case though? The test isn't actually trying to redirect and render a new page, it just wants to see how `router.push()` is called. Maybe we could set `mock.push` to a Sinon fake that we could make assertions about in the test. I'm thinking of:

```js
mock.push = sinon.fake();
```

Then in the test:

```js
router.push.calledWith('/projects/1').should.be.true;
```

src/routes.js
```diff
@@ -805,8 +827,16 @@ const routesByName = new Map();
...
   );
 
-
-
+  [
+    'NewSubmission',
+    'NewDraftSubmission',
+    'EditSubmission',
+    'FormPreview',
+    'DraftFormPreview'
```
matthew-white, line 835: At least when naming components, we tend to list the resource first, then the action or descriptor after, e.g., `EntityUpdate` or `ProjectOverview`. What do you think about doing something similar here?

Also, in the past, we've named things `FormDraft` rather than `DraftForm`. I think either one makes sense, but I think consistency is helpful.

```suggestion
    'SubmissionNew',
    'DraftSubmissionNew',
    'SubmissionEdit',
    'FormPreview',
    'FormDraftPreview'
```

test/data/form-with-attachment.js
matthew-white, line 1: Could we put XForms in their own directory just to keep them separate from the test data stores? Like test/data/xforms or test/data/xml or test/xforms?

Or could we have a single file that exports all XForms? Do we need a separate file for each XForm?

test/util/http/data.js
```diff
@@ -84,12 +83,17 @@ const responsesByComponent = {
     ]
   }),
   FormPreview: componentResponses({
-    form: () => testData.extendedForms.last(),
-    xml: () => mockResponse.of(simpleXml)
+    form: [
+      ({ url }) => /^\/v1\/projects\/\d+\/forms\/[^/]+$/.test(url),
```
matthew-white, line 87: I'm trying to figure out a way to make this part easier. Suggestions welcome! Maybe we could use a library that converts route paths to regular expressions. 🤔

test/components/enketo-iframe.spec.js
```diff
@@ -0,0 +1,59 @@
...
+  it('redirects on submissionsuccess message with return_url', async () => {
+    const wrapper = mountComponent({ enketoId, actionType: 'new', instanceId: 'test-instance' });
+    const iframe = wrapper.find('iframe');
+
+    wrapper.vm.$route.query = { return_url: 'http://localhost/projects/1' };
+
+    wrapper.vm.$router.push = sinon.fake();
+
+    iframe.element.contentWindow.eval(`
+      window.parent.postMessage(JSON.stringify({ enketoEvent: 'submissionsuccess' }), "*");
```
matthew-white, line 52: I always feel a little nervous when I see `eval()`, even though I know it's safe here. 😅 Is there another way we could call `postMessage()` though? Is `eval()` needed?

test/components/enketo-iframe.spec.js
```diff
@@ -0,0 +1,59 @@
...
+    const iframe = wrapper.find('iframe');
+    iframe.exists().should.be.true;
+    iframe.attributes('src').should.contain(`/enketo-passthrough/edit/${enketoId}?instance_id=test-instance`);
+  });
+
+  it('redirects on submissionsuccess message with return_url', async () => {
+    const wrapper = mountComponent({ enketoId, actionType: 'new', instanceId: 'test-instance' });
+    const iframe = wrapper.find('iframe');
+
+    wrapper.vm.$route.query = { return_url: 'http://localhost/projects/1' };
```
matthew-white, line 47: I'm a little surprised that Vue Router lets us mutate the `$route` object. I don't mind keeping this line the way it is, since it works, but I thought I'd throw out another idea: we could pass the query parameter to `mockRouter()`.

```js
mockRouter('/?return_url=http%3A%2F%2Flocalhost%2Fprojects%2F1')
```

What I often do is specify a default `mockRouter()` in `mountComponent()`, but allow a specific test to override it. That can be done using `mergeMountOptions()`.

matthew-white (commented):
I haven't finished reviewing everything yet, but I thought I'd go ahead and leave these comments since they're ready. It's looking good so far!

test/util/router.js
matthew-white: Now that the test is using `sinon.fake()`, does this file need to change at all? Could we remove `push` from the list?

As a related thought, I'm thinking that it'd probably also be OK for this file to always set `mock.push = sinon.fake()`, if that's something you think would be helpful. Then any future test that needs to make assertions about `push()` would be able to do so easily. We could also wait to make that change later if/when it comes up again.

test/components/form/submission.spec.js
matthew-white: I don't think we've used `setLoader()` before for real components. Mostly it's for adding or replacing async components that are only used in testing, e.g., in the tests of the `AsyncRoute` component. We don't have a hook to undo the effects of `setLoader()`, so if you mock `WebFormRenderer` here, that mock will be in effect for all tests that follow this one.

Could the `beforeAll()` hook just load the `WebFormRenderer` component? Hopefully then all tests would send the request at the same time (at mount time), since `WebFormRenderer` will have been async-loaded by then.

```js
beforeAll(loadAsync('WebFormRenderer'));
```

We already preload all route components in test/util/load-async.js, so I think it's fair to preload `WebFormRenderer`.

If we really need to use `setLoader()`, I have ideas about how we could reverse its effects so that this `beforeAll()` hook doesn't effect tests outside this file. But I'm hoping that there's a way around using `setLoader()`.

matthew-white: Another idea: Does Vue Test Utils allow you to stub async components? Could these tests stub `<web-form-renderer>`?

test/util/http/data.js
sadiqkhoja: what do you think about this function:

```
const DIGITS = 'DIGITS';
const ALPHA_NUMERIC = 'ALPHA_NUMERIC';

const apiPathToRegex = (apiPathWithPlaceholders) => {
  const DIGITS_REGEX = '\\d+';
  const ALPH_NUMERIC_REGEX = '[^/]+';

  return new RegExp(apiPathWithPlaceholders
    .replace(DIGITS, DIGITS_REGEX)
    .replace(ALPHA_NUMERIC, ALPH_NUMERIC_REGEX));
};
```

and then

```
apiPathToRegex(apiPaths.form(DIGITS, ALPHA_NUMERIC)).test(url)
```

test/data/form-with-attachment.js
sadiqkhoja: moved the Forms to `test/data/xml` directory and rename them to xml so that they play nice in IDE. Had to install raw-loader as dev-dependency so that webpack is able to load xml files.

test/components/enketo-iframe.spec.js
matthew-white: Is it possible to do something like:

```js
iframe.element.contentWindow.parent.postMessage(/*...*/);
```

Or is that restricted in some way?

sadiqkhoja: I wanted to send message from inside the iframe, but I guess this is also fine.

src/components/enketo-redirector.vue
```diff
@@ -0,0 +1,77 @@
...
+  url: `/v1/enketo-ids/${props.enketoId}/form`,
+})
+  .then(() => {
+    let target = '/';
+
+    // Edit
+    // note: we don't support editing of draft submissions
+    if (props.actionType === 'edit' && route.query.instance_id) {
+      // if route.query.instance_id is not there then it will not match any path and page not found
+      // will be displayed.
```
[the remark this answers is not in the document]
sadiqkhoja: I was assuming that `submissionPath` would return `/projects/1/forms/f/submissions//edit` but that's not the case.

I will made the change

src/composables/routes.js
```diff
@@ -88,6 +88,25 @@ export default memoizeForContainer(({ router, requestData }) => {
     ? publishedFormPath(form.projectId, form.xmlFormId)
     : formPath(form.projectId, form.xmlFormId, 'draft'));
 
+  const newSubmissionPath = (projectIdOrDraft, xmlFormId, draft) => {
```
matthew-white, line 91: This might be off-topic (I could file a follow-up issue), but could we use `newSubmissionPath()` in `ProjectFormRow` as well? Right now that component is linking to a /- path.

matthew-white (commented):
I still have a few things to look at, but I'm making progress. It's still feeling like a good approach to me. :+1:

test/components/enketo-iframe.spec.js
matthew-white: I guess injecting a `<script>` tag doesn't feel too different from `eval()` to me. But maybe at the end of the day, both are actually fine. If what you're doing feels like the right approach to you because it's inside the iframe, I'm happy for us to run with that.

test/util/http/data.js
```diff
@@ -112,28 +147,28 @@ const responsesByComponent = {
...
       () => testData.formDraftDatasetDiffs.sorted()
     ],
     datasets: [
-      ({ url }) => /^\/v1\/projects\/\d+\/datasets$/.test(url),
+      ({ url }) => matchesApiPath(apiPaths.datasets, url),
       () => testData.extendedDatasets.sorted()
     ],
     keys: [
-      ({ url }) => /^\/v1\/projects\/\d+\/forms\/[^/]+\/draft\/submissions\/keys$/.test(url),
+      ({ url }) => matchesApiPath((projectId, xmlFormId, draft) => apiPaths.submissionKeys(projectId, xmlFormId, draft), url),
```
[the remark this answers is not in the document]
sadiqkhoja: I tried to use Ramda but `partialRight` sets the right most required argument

src/composables/routes.js
sadiqkhoja: changed in `PublicLinkRow` as well

PR #1163, opened 2025-03-31
Allow user to drop form attachments anywhere on Edit Form page

matthew-white:
This PR makes progress on getodk/central#728. It makes it possible for the user to drag and drop attachments anywhere on the Edit Form page. Currently, drag and drop is limited to the `FormAttachmentList` component within the Attachments section of the page.

Existing tests continue to pass. I also tried it out some locally.

There's a little bit of awkwardness in that the `FileDropZone` component is rendered in `FormEdit`, but all the drag and drop logic is in `FormAttachmentList`, which is a few layers away. Rather than trying to emit and re-emit up through those layers, I'm using provide/inject to pass setter functions from `FormEdit` to `FormAttachmentList`. That allows `FormAttachmentList` to interact with the `FileDropZone` without getting intermediate components involved.

I also wondered whether it was possible to keep the `FileDropZone` in `FormAttachmentList`, but teleport it to `FormEdit`. I don't think that really works though, because we also need to pass content in a slot to `FileDropZone`. Specifically, that content needs to be the contents of `FormEdit`.

### DIFF
--- a/src/components/form-attachment/list.vue
+++ b/src/components/form-attachment/list.vue
@@ -10,8 +10,7 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <file-drop-zone id="form-attachment-list" :disabled="uploading"
-    :styled="false" @dragenter="dragenter" @dragleave="dragleave" @drop="drop">
+  <div id="form-attachment-list">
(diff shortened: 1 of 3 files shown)


matthew-white (commented):
Notes from interactive review

ktuite (approved):

PR #1172, opened 2025-04-03
Show tag about new or changed form attachments

matthew-white:
This PR makes progress on getodk/central#728. It implements the following release criteria:

> I should see a “{# new file(s)} {and} {# changed file(s)} since the published version” tag if any files are different, where I should only see nonzero {parts}

New tests. I also tried out some cases locally.

We determine the number of changed files by comparing blob hashes. But that means that when we upload a file, we need Backend to tell us its hash. That required a change to Backend: see getodk/central-backend#1471. That PR has additional notes about this change and the reasoning behind it.

One aspect of the release criteria is that it requires us to pluralize "# new file(s)" separately from "# changed file(s)". After we pluralize them, we add the rest of the text ("since the published version") using another i18n message, `diff.summary`. I didn't want there to be multiple versions of `diff.summary` — one for when there is only one {part} and one for when there are two — so I expanded our `formatList()` function to combine "# new file(s)" and "# changed file(s)". `formatList()` is responsible for adding the "and" between them. The browser knows how to do this automatically using `Intl.ListFormat`. Previously, `formatList()` only knew how to format lists in one way, but now it knows how to do so in two ways: it now accepts an optional `key` argument that determines the list style. In this way, we handle `formatList()` and `Intl.ListFormat` in a very similar way as `formatRange()` and `Intl.NumberFormat`.

### DIFF
--- a/src/components/form/edit/attachments.vue
+++ b/src/components/form/edit/attachments.vue
@@ -43,8 +43,7 @@ import Loading from '../../loading.vue';
 import { useI18nUtils } from '../../../util/i18n';
 import { useRequestData } from '../../../request-data';
 
-const { form, publishedAttachments, draftAttachments, resourceStates } = useRequestData();
-const { dataExists } = resourceStates([form, publishedAttachments, draftAttachments]);
+const { publishedAttachments, draftAttachments } = useRequestData();
 
 const hasAttachments = computed(() =>
   draftAttachments.dataExists && draftAttachments.size !== 0);
@@ -54,8 +53,7 @@ const hasMissing = computed(() =>
 const { t } = useI18n();
 const { tn, formatList } = useI18nUtils();
 const tag = computed(() => {
-  if (!(dataExists.value && hasAttachments.value && form.publishedAt != null))
-    return '';
+  if (!(hasAttachments.value && publishedAttachments.dataExists)) return '';
 
   let newCount = 0;
   let changedCount = 0;
@@ -94,8 +92,8 @@ const tag = computed(() => {
     "diff": {
       "newCount": "{count} new attachment | {count} new attachments",
       "changedCount": "{count} changed attachment | {count} changed attachments",
-      // {changes} describes the changes that have been made to Form
-      // attachments. For example: "1 new attachment and 2 changed attachments"
+      // {changes} describes changes that have been made to Form attachments.
+      // For example: "1 new attachment and 2 changed attachments"
       "summary": "{changes} since the published version"
     }
   }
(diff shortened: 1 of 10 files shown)


src/components/form/edit/attachments.vue
```diff
@@ -33,19 +34,47 @@ except according to the terms contained in the LICENSE file.
...
 
 const hasAttachments = computed(() =>
   draftAttachments.dataExists && draftAttachments.size !== 0);
 const hasMissing = computed(() =>
   draftAttachments.dataExists && draftAttachments.missingCount !== 0);
+
+const { t } = useI18n();
+const { tn, formatList } = useI18nUtils();
+const tag = computed(() => {
+  if (!(dataExists.value && hasAttachments.value && form.publishedAt != null))
```
matthew-white, line 57: We don't need to check `form.publishedAt != null` explicitly because there won't be `publishedAttachments` if the form isn't published.

matthew-white (commented):
Notes from interactive review

ktuite (approved):

matthew-white:
I've made a couple of small changes after code review. @ktuite, I think these changes are small enough that they don't need to be reviewed (I'm going to go ahead and merge). However, I thought I'd describe a couple of the changes.

As part of using the POST and PATCH responses from Backend (i.e., the updated attachment), the first commit of this PR called `Object.assign()` to update the reactive Frontend data for the individual attachment. However, I ended up wanting to tweak that reactivity. Previously, we had a non-reactive `Map` of reactive attachments. Instead, I think it's better to have a shallow-reactive `Map` of non-reactive attachments. Instead of updating an attachment in place, we can replace it outright in the `Map`. This approach is closer to how we manage other reactive collections, e.g., submissions. Rather than creating potentially many reactive proxies (one for each attachment), I think it's better to create a single proxy at the top level. Moving the level at which the reactivity happens also makes it easy to remove reactivity where it's not needed at all. The first commit of this PR made `publishedAttachments.data` a `Map` of reactive attachments, but `publishedAttachments.data` doesn't need to be reactive at all: even the `Map` doesn't need to be shallow-reactive. That's because individual published attachments never change: `publishedAttachments.data` only changes when it's replaced as a whole by a new response.

Another set of changes I made was to `testData.standardAttachments`. I noticed I made a couple of small logic mistakes when updating that `testData` store, so I fixed those. I also ended up questioning more of the validation that the `testData` store was doing. I decided just to remove all that validation. After all, it's pretty unusual for a `testData` store to do any validation.

PR #1174, opened 2025-04-04
Enhancement: simplify request headers

sadiqkhoja:

### DIFF
--- a/src/container.js
+++ b/src/container.js
@@ -29,6 +29,8 @@ const provide = [
   'logger'
 ];
 
+axios.defaults.headers.post['X-Requested-With'] = 'XMLHttpRequest';
+
 export default ({
   // `router` must be a function that returns an object. The function will be
   // passed a partial container. It is also possible to create a container
--- a/src/util/request.js
+++ b/src/util/request.js
@@ -68,6 +68,7 @@ const entityPath = (suffix) =>
 export const apiPaths = {
   // Backend generates session tokens that are URL-safe.
   session: (token) => `/v1/sessions/${token}`,
+  currentSession: () => '/v1/sessions/current',
   users: (query = undefined) => `/v1/users${queryString(query)}`,
   user: (id) => `/v1/users/${id}`,
   password: (id) => `/v1/users/${id}/password`,
@@ -212,17 +213,6 @@ for (const prop of ['post', 'put', 'patch']) {
 // function references `this`, you must bind `this` for it first.
 export const withHttpMethods = (f) => Object.assign(f, httpMethods);
 
-export const withAuth = (config, token) => {
-  const { headers } = config;
-  if ((headers == null || headers.Authorization == null) &&
-    config.url.startsWith('/v1/') && token != null) {
-    return {
-      ...config,
-      headers: { ...headers, Authorization: `Bearer ${token}` }
-    };
-  }
-  return config;
-};
 
 export const logAxiosError = (logger, error) => {
   if (error.response == null)
--- a/src/util/session.js
+++ b/src/util/session.js
@@ -16,10 +16,8 @@ object, as well as an associated cookie that is Secure and HttpOnly. If the user
 then opens Frontend in a new tab, Frontend will use the cookie to restore the
 session.
 
-The cookie is used in only limited ways: mostly Frontend specifies the session
-token as a bearer token. The cookie is used to restore the session. It is also
-used for non-AJAX requests, including download links and iframe forms. When the
-user logs out, the cookie is removed.
+Frontend relies on session cookie for the authentication for all types of
+requests. When the user logs out, the cookie is removed.
 
 Across tabs, Frontend allows only one user to be logged in at a time. (Otherwise
 one user would use another user's cookie.) Further, Frontend allows only one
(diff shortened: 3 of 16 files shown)


src/util/request.js
```diff
@@ -68,6 +68,7 @@ const entityPath = (suffix) =>
 export const apiPaths = {
   // Backend generates session tokens that are URL-safe.
   session: (token) => `/v1/sessions/${token}`,
+  currentSession: () => '/v1/sessions/current',
```
matthew-white, line 71: I usually only add an `apiPaths` function when there's a variable/argument. No harm in adding a constant function, but I also think it'd be OK to hard-code it where it's used.

src/util/util.js
```diff
@@ -16,3 +16,7 @@ export const noop = () => {};
 export const noargs = (f) => () => f();
 
 export const sumUnderThreshold = (list, threshold) => list.reduce((acc, i) => acc + Math.min(i, threshold), 0);
+
+export const getCsrf = () => decodeURIComponent(document.cookie.split(';')
```
matthew-white, line 20: When I `console.log(document.cookie)`, I see that the separator is `'; '` with a space after the semicolon. MDN [suggests](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#read_all_cookies_accessible_from_this_location) that the amount of white space may vary by browser:

> Note that each _key_ and _value_ may be surrounded by whitespace (space and tab characters): in fact, RFC 6265 mandates a single space after each semicolon, but some user agents may not abide by this.

Maybe we could split by `/;[ \t]*/`?

test/components/account/claim.spec.js
```diff
@@ -30,10 +30,9 @@ describe('AccountClaim', () => {
         await component.get('input').setValue('testPassword');
         return component.get('form').trigger('submit');
       })
-      .beforeEachResponse((_, { method, url, headers, data }) => {
+      .beforeEachResponse((_, { method, url, data }) => {
         method.should.equal('POST');
         url.should.equal('/v1/users/reset/verify');
-        headers.Authorization.should.equal('Bearer foo');
```
matthew-white, line 36: I think we should probably keep this assertion. This isn't a generic assertion about the `Authorization` header and how it's usually set. The `AccountClaim` component does something special and sets the `Authorization` header based on the token in the URL.

test/util/session.js
```diff
@@ -18,6 +19,10 @@ export const mockLogin = (options = undefined) => {
   );
 
   loggedIn = true;
+
+  const token = faker.random.alphaNumeric(64);
+  const csrf = faker.random.alphaNumeric(64);
+  document.cookie = `__csrf=${csrf};session=${token}`;
```
matthew-white, line 25: The session cookie is `HTTPOnly`, so it won't show up in `document.cookie`.

test/util/http/common.js
```diff
@@ -4,9 +4,8 @@ import Modal from '../../../src/components/modal.vue';
 import Spinner from '../../../src/components/spinner.vue';
 
 import { relativeUrl } from '../request';
-import { withAuth } from '../../../src/util/request';
 
-const assertRequestsMatch = (actual, expected, component) => {
+const assertRequestsMatch = (actual, expected) => {
```
matthew-white, line 8: I really like that we don't have to pass in the component anymore: it feels cleaner like this. `assertRequestsMatch()` is used in two places in this file. How about updating those calls to remove the component?

matthew-white (commented):
So nice that this is an overall subtraction of lines. 🎉

src/container.js
```diff
@@ -29,6 +29,8 @@ const provide = [
   'logger'
 ];
 
+axios.defaults.headers.post['X-Requested-With'] = 'XMLHttpRequest';
```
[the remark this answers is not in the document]
matthew-white: Makes sense to me to just do it for POST: that's the only time that Backend cares about it. Maybe it'd be helpful to add a comment about it though.

src/util/util.js
```diff
@@ -16,3 +16,8 @@ export const noop = () => {};
 export const noargs = (f) => () => f();
 
 export const sumUnderThreshold = (list, threshold) => list.reduce((acc, i) => acc + Math.min(i, threshold), 0);
+
+export const getCsrf = () => decodeURIComponent(document.cookie.split(/;[ \t]*/)
+  .map(cookie => cookie.trim())
+  .find(cookie => cookie.startsWith('__csrf'))
+  ?.split('=')[1] || '');
```
matthew-white, line 23: Since #1186 does similar cookie parsing, what do you think about adding a little utility function for that? I feel like there's enough going on in the cookie parsing that it'd be nice to have a little reusable function.

src/util/util.js
```diff
@@ -16,3 +16,8 @@ export const noop = () => {};
 export const noargs = (f) => () => f();
 
 export const sumUnderThreshold = (list, threshold) => list.reduce((acc, i) => acc + Math.min(i, threshold), 0);
+
+export const getCsrf = () => decodeURIComponent(document.cookie.split(/;[ \t]*/)
+  .map(cookie => cookie.trim())
+  .find(cookie => cookie.startsWith('__csrf'))
```
matthew-white, line 22: How about adding `=` just to make absolute sure that we're matching the `__csrf` cookie, not another cookie with the same prefix, like `__csrf_but_not_really`?

```suggestion
  .find(cookie => cookie.startsWith('__csrf='))
```

src/util/session.js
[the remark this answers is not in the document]
sadiqkhoja: I have added a clause for 404.1 and not removed 401/403 just in case we decide to change authHandler for session cookie.

Part of me also thinks that `DELETE /v1/sessions/current` should never return error, if the provided session exists and is valid then it is deleted otherwise noop but in both cases return response with deleting the cookies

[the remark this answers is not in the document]
matthew-white: That change sounds reasonable to me, though I'm a little wary of changing the API unless we need to. Is there a specific use case you had in mind?

matthew-white (approved):

PR #1156, opened 2025-03-21
Redo breadcrumbs

ktuite:
Part of https://github.com/getodk/central/issues/924

This redoes the breadcrumbs to have icons, have the names of the objects, and to feel more hierarchical. 

Project name --> form list or entity lists list (basically 2 views of the project home)
Form name --> form home
Entity list name --> entity list home
Submission ID --> uhh I had it link to the form home which is also the submission list but it could link to itself or something?
Entity ID --> also links to entities on entity list page, essentially the entity list home

Screenshot 2025-03-21 at 3 55 02 PM

#### What has been done to verify that this works as intended?

Testing it and trying it out.

#### Why is this the best possible solution? Were any other approaches considered?

The code feels easier to think about because there is a much clearer hierarchy/repetition!
It seems nice that the name of the thing goes to the page about the thing and there aren't extraneous links.

#### Before submitting this PR, please make sure you have:

- [ ] run `npm run test` and `npm run lint` and confirmed all checks still pass OR confirm CircleCI build passes
- [ ] verified that any code or assets from external sources are properly credited in comments or that everything is internally sourced

PR #1176, opened 2025-04-07
Remove trailing slash on breadcrumbs

ktuite:
Related to issue https://github.com/getodk/central/issues/924
Followup to PR https://github.com/getodk/central-frontend/pull/1156

Changes:
- Removed the trailing slash.
- Made final crumb on entity and submission show pages into plain text (no link). For the parent pages (form or entity list), the name of the form/entity list is a link because those pages have multiple tabs and the link goes to the root tab.

### DIFF
--- a/src/components/breadcrumbs.vue
+++ b/src/components/breadcrumbs.vue
@@ -13,10 +13,14 @@ except according to the terms contained in the LICENSE file.
   <div class="breadcrumbs">
     <template v-for="(link, index) in links" :key="index">
       <div class="breadcrumb-item" v-tooltip.text>
-        <router-link :to="link.path">
+        <router-link v-if="link.path != null" :to="link.path">
           <span v-if="link.icon" :class="link.icon"></span>
           {{ link.text }}
         </router-link>
+        <template v-else>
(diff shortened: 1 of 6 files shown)


src/components/breadcrumbs.vue
```diff
@@ -13,12 +13,16 @@ except according to the terms contained in the LICENSE file.
...
       <div class="breadcrumb-item" v-tooltip.text>
-        <router-link :to="link.path">
+        <router-link v-if="link.path != null" :to="link.path">
           <span v-if="link.icon" :class="link.icon"></span>
           {{ link.text }}
         </router-link>
+        <template v-else>
+          <span v-if="link.icon" :class="link.icon"></span>
+          {{ link.text }}
+        </template>
```
matthew-white, line 23: This pattern has come up before of sometimes wrapping content in a link and sometimes not. We have the `Linkable` component that you could use for it. It makes it possible to avoid duplicating the inner content:

```suggestion
        <linkable :to="link.path">
          <span v-if="link.icon" :class="link.icon"></span>
          {{ link.text }}
        </linkable>
```

matthew-white (approved):

PR #1178, opened 2025-04-11
Allow user to make another submission

sadiqkhoja:
Closes getodk/central#928

Added a test and manually verified it.

Initially I created a `clearForm` function to call the OWF callback, which was called from the hideSubmissionModal function. Current `callback` function is called immediately after all attachment upload, which makes the code clean.

Later we can pass a promise to this `callback` function to communicate to OWF that the submission is being sent to the server. See https://getodk.slack.com/archives/C01TKJ7TG7J/p1744395959590219

No

No

-->

### DIFF
--- a/src/components/web-form-renderer.vue
+++ b/src/components/web-form-renderer.vue
@@ -13,9 +13,7 @@ except according to the terms contained in the LICENSE file.
 <template>
   <loading :state="initiallyLoading"/>
   <template v-if="dataExists">
-    <!-- update newInstanceId to rerender the component -->
     <OdkWebForm
-      :key="newInstanceId"
       :form-xml="formVersionXml.data"
       :edit-instance="editInstanceOptions"
       :fetch-form-attachment="getAttachment"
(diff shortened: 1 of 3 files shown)


src/components/web-form-renderer.vue
```diff
@@ -46,7 +44,15 @@ except according to the terms contained in the LICENSE file.
...
-      <div class="modal-actions">
+      <div v-if="submissionModal.type === 'submissionModal'" class="modal-actions">
+        <button type="button" class="btn btn-primary" @click="hideModals()">
+          {{ $t('submissionModal.fillOutAgain') }}
+        </button>
+        <button type="button" class="btn btn-link" @click="closeWindow()">
+          {{ $t('action.close') }}
+        </button>
+      </div>
+      <div v-else>
```
matthew-white, line 55: Does this need `class="modal-actions"`?

src/components/web-form-renderer.vue
```diff
@@ -359,8 +359,9 @@ html, body {
         "body": "The data you entered is valid, but it was not submitted because this is a Form preview."
       },
       "submissionModal": {
-        "title": "Submission successful",
-        "body": "Your data was submitted."
+        "title": "Form successful sent!",
+        "body": "You can fill this Form out again or close if you’re done.",
+        "fillOutAgain": "Fill out again"
```
matthew-white, line 364: I usually put button text inside an `"action": {` block in the JSON so that it automatically gets a comment about how it's used.

matthew-white (approved):

PR #1181, opened 2025-04-12
Update modal styling to reflect new spacious design

ktuite:
Closes https://github.com/getodk/central/issues/914

The buttons are still on the left side of the footer. I plan to have a separate PR where we position these buttons on the right, but also check every `<div class="modal-actions">` to reorder the buttons if there is more than one button. 

For the edit entity modal, there used to be no padding/space around the outside of the editable fields, but I decided to leave the space in this time. 

Sampling of different modals with new styles:
Screenshot 2025-04-12 at 4 22 17 PM
Screenshot 2025-04-12 at 4 19 07 PM
Screenshot 2025-04-12 at 4 18 54 PM
Screenshot 2025-04-12 at 4 18 49 PM
Screenshot 2025-04-12 at 4 18 37 PM
Screenshot 2025-04-12 at 4 18 07 PM
Screenshot 2025-04-12 at 4 17 52 PM
Screenshot 2025-04-12 at 4 17 38 PM
Screenshot 2025-04-12 at 4 17 16 PM
Screenshot 2025-04-12 at 4 17 02 PM

### DIFF
--- a/src/components/entity/update.vue
+++ b/src/components/entity/update.vue
@@ -169,9 +169,7 @@ const currentVersion = computed(() =>
     overflow-y: auto;
   }
 
-  .modal-body { padding: 0; }
   table { margin-bottom: 0; }
-  .modal-actions { margin: 0; }
 
   table { table-layout: fixed; }
   thead {
--- a/src/components/modal.vue
+++ b/src/components/modal.vue
@@ -20,11 +20,13 @@ may add the `in` class to the element, and the checkScroll() method may add the
     @click="modalClick" @keydown.esc="hideIfCan" @focusout="refocus">
     <div class="modal-dialog" :class="sizeClass" role="document">
       <div class="modal-content">
-        <div class="modal-header">
+        <div class="modal-top-actions">
           <button type="button" class="close" :aria-disabled="!hideable"
             :aria-label="$t('action.close')" @click="hideIfCan">
             <span aria-hidden="true">&times;</span>
           </button>
+        </div>
+        <div class="modal-header">
           <h4 :id="titleId" class="modal-title"><slot name="title"></slot></h4>
         </div>
         <div ref="body" class="modal-body">
@@ -241,27 +243,32 @@ const titleId = `modal-title${id}`;
 
 .modal-dialog {
   margin-top: 20vh;
+  border-radius: 6px;
 
   .modal-content {
     border: none;
-    border-radius: 0;
     box-shadow: $box-shadow-panel-main;
 
-    .modal-header {
-      background-color: $color-accent-primary;
-      color: #fff;
-      padding: 10px $padding-right-modal-header 9px $padding-left-modal-header;
+    .modal-top-actions {
+      text-align: right;
+      padding-right: 16px;
+      padding-top: 16px;
 
       .close {
-        color: #fff;
-        font-weight: normal;
-        margin-top: 0;
-        opacity: 1;
+          float: none;
+          color: $color-input;
+          font-weight: normal;
+          opacity: 1;
 
-        &[aria-disabled="true"] {
-          cursor: not-allowed;
+          &[aria-disabled="true"] {
+            cursor: not-allowed;
+          }
         }
       }
+
+    .modal-header {
+      color: $color-text;
+      padding: 0px $padding-right-modal-header 20px $padding-left-modal-header;
 
       h4 {
         @include text-overflow-ellipsis;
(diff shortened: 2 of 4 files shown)


ktuite:
I don't remember how the modal resizing works well enough to know why these three tests are failing:

```
1) emits a resize event after the height of the modal changes
     Modal
     AssertionError: expected [ [ 10 ], [ 20 ] ] to deeply equal [ [ 40 ], [ 50 ] ]
    at Context.<anonymous> (dist/commons.js:323006:35)

2) emits a resize event
     Modal state prop is initially true
     AssertionError: expected [ [ 10 ] ] to deeply equal [ [ 40 ] ]
    at Context.<anonymous> (dist/commons.js:322932:37)

3) emits a resize event
     Modal after the state prop changes to false
     AssertionError: expected [ [ 10 ], [ +0 ] ] to deeply equal [ [ 40 ], [ +0 ] ]
    at Context.<anonymous> (dist/commons.js:322993:37)
```

matthew-white:
I don't totally remember either. I don't have a ton of time right now to dig into it on my own, but I'd be happy to pair on it at any point. The resizing stuff is important behavior, but since you didn't change any of the JS, I sort of doubt it's really broken. Probably numbers just need to change in tests to match the new padding, etc.

matthew-white:
> The resizing stuff is important behavior

But maybe the `resize` event specifically isn't that important? When I search the codebase for `@resize`, I don't see anything. Maybe the event isn't being used. I think I added the `resize` event at the same time I added the `mutate` event. The `mutate` event is actually used, in `EntityUpload`. I could see the `resize` event being useful one day, and I think we should try to keep it, but we probably don't need to spend a ton of time digging into it if it's not currently being used.

matthew-white:
> I plan to have a separate PR where we position these buttons on the right

Should this PR not close the issue then? Otherwise, what do you think about filing a follow-up issue for this aspect of things?

src/components/entity/update.vue
matthew-white, line 1: It looks like this component hard-codes some heights that I'm guessing need to be adjusted:

```scss
max-height: calc(70vh -
  #{/* .modal-header */ 46px + /* .modal-actions */ 50px});
```

src/components/modal.vue
matthew-white, line 1: Could you check whether `FormAttachmentPopups` still looks right after these changes? It also uses the modal classes, though maybe it'd be better for it not to. I'm guessing that we don't want all the changes here to apply to that component.

src/components/entity/update.vue
```diff
@@ -169,9 +169,7 @@ const currentVersion = computed(() =>
     overflow-y: auto;
   }
 
-  .modal-body { padding: 0; }
```
matthew-white, line 172: `EntityUpdateRow` uses `$padding-left-modal-header`. Now that the padding here isn't being zeroed out, maybe `EntityUpdateRow` doesn't need to use it anymore?

src/components/entity/update.vue
```diff
@@ -169,9 +169,7 @@ const currentVersion = computed(() =>
...
   table { margin-bottom: 0; }
-  .modal-actions { margin: 0; }
```
matthew-white, line 174: Not super important because it's not user-facing, but `EntityBranchData` zeroes things out in a similar way to `EntityUpdate`.

src/components/entity/update.vue
matthew-white, line 1: Does `EntityResolve` need to change at all? It doesn't seem to zero out padding like `EntityUpdate` does, but I do see it referencing `$padding-modal-body` in negative margin.

matthew-white (commented):
The changes in here are looking good to me. I'm liking the look of the new modals. ✨ The main thing I'm wondering about is that there are a few places where I think we may need to make additional changes.

src/components/entity/update.vue
ktuite: I've given it a bit more space so this modal has a bit of space between its bottom edge and the bottom of the screen.

ktuite: I've checked it and it tastefully makes the conflict table extend to the edge of the modal. 
Screenshot 2025-04-23 at 4 15 39 PM

src/components/entity/update.vue
ktuite: I've changed some of the padding here so the update table looks nice.
Screenshot 2025-04-23 at 4 26 51 PM

src/components/entity/update.vue
ktuite: It's looking like this now:

Screenshot 2025-04-23 at 4 22 49 PM

src/components/modal.vue
ktuite: It did look off before but I've made some changes and I looks normal now and uses some non-modal classes in some places.

matthew-white (approved):

PR #1182, opened 2025-04-14
Handle session timeout for Web Forms

sadiqkhoja:
Closes getodk/central#910

### Changes:

- Skip auto logout and session expiring alert for Enketo/Web Form submissions.
- Show modal on submit button if session has expired and ask user to login in another tab

### DIFF
--- a/src/components/web-form-renderer.vue
+++ b/src/components/web-form-renderer.vue
@@ -40,6 +40,14 @@ except according to the terms contained in the LICENSE file.
             <a href="emailto:support@getodk.org">support@getodk.org</a>
           </template>
         </i18n-t>
+        <i18n-t v-else-if="submissionModal.type === 'sessionTimeoutModal'" tag="p" keypath="sessionTimeoutModal.body">
+          <template #here>
+              <a href="/login" target="_blank">{{ $t('sessionTimeoutModal.here') }}</a>
+          </template>
+          <template #supportEmail>
+            <a href="emailto:support@getodk.org">support@getodk.org</a>
+          </template>
+        </i18n-t>
         <p v-else>
           {{ $t(submissionModal.type + '.body') }}
         </p>
@@ -269,7 +277,11 @@ const postPrimaryInstance = (file) => {
   })
     .then(({ data }) => {
       if (isProblem(data)) {
-        showModal(submissionModal, { type: 'errorModal', errorMessage: data.message });
+        if (data.code === 403.1) {
+          showModal(submissionModal, { type: 'sessionTimeoutModal' });
+        } else {
+          showModal(submissionModal, { type: 'errorModal', errorMessage: data.message });
+        }
         return false;
       }
       return data.currentVersion.instanceId;
@@ -386,6 +398,11 @@ html, body {
       "sendingDataModal": {
         "title": "Sending Submission",
         "body": "Your data is being submitted. Please don’t close this window until it’s finished."
+      },
+      "sessionTimeoutModal": {
+        "title": "Session expired",
+        "body": "Please log in {here} in a different browser tab and try again.",
+        "here": "here"
       }
     }
   }
--- a/src/components/web-form-renderer.vue
+++ b/src/components/web-form-renderer.vue
@@ -40,12 +40,9 @@ except according to the terms contained in the LICENSE file.
             <a href="emailto:support@getodk.org">support@getodk.org</a>
           </template>
         </i18n-t>
-        <i18n-t v-else-if="submissionModal.type === 'sessionTimeoutModal'" tag="p" keypath="sessionTimeoutModal.body">
+        <i18n-t v-else-if="submissionModal.type === 'sessionTimeoutModal'" tag="p" keypath="sessionTimeoutModal.body.full">
           <template #here>
-              <a href="/login" target="_blank">{{ $t('sessionTimeoutModal.here') }}</a>
-          </template>
-          <template #supportEmail>
-            <a href="emailto:support@getodk.org">support@getodk.org</a>
+              <a href="/login" target="_blank">{{ $t('sessionTimeoutModal.body.here') }}</a>
           </template>
         </i18n-t>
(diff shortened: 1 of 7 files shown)


src/components/web-form-renderer.vue
```diff
@@ -386,6 +398,11 @@ html, body {
       "sendingDataModal": {
         "title": "Sending Submission",
         "body": "Your data is being submitted. Please don’t close this window until it’s finished."
+      },
+      "sessionTimeoutModal": {
+        "title": "Session expired",
+        "body": "Please log in {here} in a different browser tab and try again.",
+        "here": "here"
```
matthew-white, line 405: Could you group these two messages in an object where the main message is named `full`? Doing so will automatically add comments for the two messages about the context in which "here" appears. You should see those comments show up in transifex/strings_en.json.

```json
"body": {
  "full": "Please log in {here} in a different browser tab and try again.",
  "here": "here"
}
```

src/components/web-form-renderer.vue
```diff
@@ -40,6 +40,14 @@ except according to the terms contained in the LICENSE file.
             <a href="emailto:support@getodk.org">support@getodk.org</a>
           </template>
         </i18n-t>
+        <i18n-t v-else-if="submissionModal.type === 'sessionTimeoutModal'" tag="p" keypath="sessionTimeoutModal.body">
+          <template #here>
+              <a href="/login" target="_blank">{{ $t('sessionTimeoutModal.here') }}</a>
+          </template>
+          <template #supportEmail>
+            <a href="emailto:support@getodk.org">support@getodk.org</a>
+          </template>
```
matthew-white, line 49: I don't see `{supportEmail}` mentioned in the `sessionTimeoutModal.body` message.

src/routes.js
```diff
@@ -161,6 +161,10 @@ The following meta fields are supported for bottom-level routes:
 
   - standalone (default: false): If standalone is `true` then application layout
     elements like navigation bar, background color, etc are not rendered.
+
+  - skipAutoLogout (default: false): If `true`, no alert will be displayed when
```
matthew-white, line 165: To make the default `false` rather than `undefined`, you'll need to update `normalizeMeta()` below. I like the idea of making it `true`/`false` like other route meta fields.

matthew-white, line 165: What do you think about moving this closer to the descriptions of the route meta fields related to login? They feel thematically related. They could all go in one section: section about login could be renamed to "Login/logout" or "Session". Or `skipAutoLogout` could go in its own new section named "Logout" that comes after the section named "Login". No preference from my end on the exact section titles, the main thing I think is that it'd be nice for these route meta fields to be close by each other.

test/components/form/submission.spec.js
```diff
@@ -184,4 +187,36 @@ describe('FormSubmission', () => {
...
+      testData.extendedForms.createPast(1, { xmlFormId: 'a', webformsEnabled: true });
+      const session = testData.sessions.createNew({ expiresAt: '1970-01-01T00:05:00Z' });
+
+      const clock = sinon.useFakeTimers();
+
+      await load('/projects/1/forms/a/submissions/new', mountOptions({
+        container: {
+          requestData: testRequestData([], {
+            session
+          }),
```
matthew-white, line 206: It feels a little awkward to call `mockLogin()`, then create a different session from the one that `mockLogin()` creates and uses. For example, `sessionExpires` in local storage will end up not matching the new session.

What do you think about modifying `mockLogin()` to make it easier to customize the session that it creates? `mockLogin()` accepts options that it will use to create the user. We could add a second parameter for options that it will use to create the session.

If we did something like that, you wouldn't need to pass `session` to `requestData`, since `mockLogin()` automatically passes the session it creates to `requestData`.

test/components/form/submission.spec.js
```diff
@@ -184,4 +187,36 @@ describe('FormSubmission', () => {
...
+  describe('session timeout', () => {
+    beforeEach(() => {
+      mockLogin();
+    });
+
+    it('does not auto log out', async () => {
+      testData.extendedForms.createPast(1, { xmlFormId: 'a', webformsEnabled: true });
+      const session = testData.sessions.createNew({ expiresAt: '1970-01-01T00:05:00Z' });
+
+      const clock = sinon.useFakeTimers();
```
matthew-white, line 200: I usually call `sinon.useFakeTimers()` before creating `testData` so that the default timestamps that `testData` uses reflect the faked time.

test/components/form/submission.spec.js
```diff
@@ -184,4 +187,36 @@ describe('FormSubmission', () => {
...
+        .complete()
+        .request(() => {
+          clock.tick('24:00:00');
+        })
+        .respondWithData(() => '(v2024.1.2-sha)')
+        .testRequests([
+          { url: '/version.txt' }
+        ])
+        .afterResponses((app) => {
+          app.vm.$container.requestData.session.dataExists.should.be.true;
```
matthew-white, line 218: Could you also assert that the route path hasn't changed? In other words, that the user wasn't redirected to /login.

src/util/session.js
```diff
@@ -177,7 +178,7 @@ const logOutBeforeSessionExpires = (container) => {
 const logOutAfterStorageChange = (container) => (event) => {
   // event.key == null if the user clears local storage in Chrome.
   if ((event.key == null || event.key === 'sessionExpires') &&
-    container.requestData.session.dataExists) {
+    container.requestData.session.dataExists && !container.router.currentRoute.value.meta.skipAutoLogout) {
```
matthew-white, line 181: How about adding a test of this line? Maybe it's not critical though.

matthew-white (approved):
Looks great! I really like the idea of using a route meta field for this.

test/components/form/submission.spec.js
sadiqkhoja: this is not needed after I moved sinon.useFakeTimers() before everything else including mockLogin

PR #1183, opened 2025-04-14
Ability to redirect to external sites

sadiqkhoja:
Closes getodk/central#942

Manual verification and added couple of tests.

NA

No

Added in https://github.com/getodk/docs/issues/1896

### DIFF
--- a/src/components/enketo-iframe.vue
+++ b/src/components/enketo-iframe.vue
@@ -18,9 +18,12 @@ except according to the terms contained in the LICENSE file.
 <script setup>
 import { computed, inject, ref } from 'vue';
 import { useRoute, useRouter } from 'vue-router';
-import useEventListener from '../composables/event-listener';
 import { queryString } from '../util/request';
+import { useRequestData } from '../request-data';
+
 import Loading from './loading.vue';
+import useEventListener from '../composables/event-listener';
+import useRoutes from '../composables/routes';
 
 defineOptions({
   name: 'EnketoIframe'
@@ -40,10 +43,12 @@ const props = defineProps({
 
 const { location } = inject('container');
 
+const { form } = useRequestData();
 const route = useRoute();
 const router = useRouter();
+const { submissionPath } = useRoutes();
 
-const redirectUrl = computed(() => route.query.return_url);
+const redirectUrl = computed(() => route.query.return_url || route.query.returnUrl);
 
 const lastSubmitted = (enketoOnceId) => {
   const iframe = document.createElement('iframe');
@@ -72,7 +77,7 @@ const setEnketoSrc = () => {
     basePath = `/#${basePath}`;
   }
   let prefix = basePath;
-  const { return_url: _, ...query } = route.query;
+  const { return_url: _, returnUrl: __, ...query } = route.query;
 
   query.parentWindowOrigin = location.origin;
 
@@ -104,7 +109,7 @@ const setEnketoSrc = () => {
 
 setEnketoSrc();
 
-function handleIframeMessage(event) {
+const handleIframeMessage = (event) => {
   if (event.origin === location.origin) {
     const { parentWindowOrigin } = route.query;
     // For the cases where this page is embedded in external iframe, pass the event data to the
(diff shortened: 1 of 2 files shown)


src/components/enketo-iframe.vue
```diff
@@ -97,8 +103,22 @@ function handleIframeMessage(event) {
...
-      router.push((new URL(redirectUrl.value)).pathname);
+    if (eventData?.enketoEvent === 'submissionsuccess') {
+      if (props.actionType === 'edit') {
+        // for edit we always redirect to Submission details page
+        router.push(submissionPath(form.projectId, form.xmlFormId, props.instanceId));
+      } else if (!props.actionType && redirectUrl.value) {
+        // for public link, we read return value from query parameter. The value could be 3rd party
+        // site as well, typically a thank you page
+        try {
+          const normalizedUrl = new URL(redirectUrl.value);
```
sadiqkhoja, line 114: if this is failed then Enketo will show its default thank you page

src/util/util.js
```diff
@@ -16,3 +16,8 @@ export const noop = () => {};
 export const noargs = (f) => () => f();
 
 export const sumUnderThreshold = (list, threshold) => list.reduce((acc, i) => acc + Math.min(i, threshold), 0);
+
+// To be able to mock it
+export const externalNavigator = {
```
sadiqkhoja, line 21: use location from the container

src/components/enketo-iframe.vue
```diff
@@ -102,11 +107,26 @@ function handleIframeMessage(event) {
...
```
matthew-white, line 118: I think I remember us talking about doing a little validation of `normalizedUrl`. Like checking that it's not something truly strange like an ftp:// or file:// URL. Does Enketo do any validation of the URL?

src/components/enketo-iframe.vue
```diff
@@ -102,11 +107,26 @@ function handleIframeMessage(event) {
...
+          if (normalizedUrl.origin === location.origin) {
+            router.push(normalizedUrl.pathname);
+          } else {
+            location.assign(normalizedUrl);
+          }
+        } catch (e) {}
```
matthew-white, line 124: How about adding a test of the `catch` case? Basically just checking that if the specified URL is invalid, the user stays on the page.

src/components/enketo-iframe.vue
```diff
@@ -55,11 +60,11 @@ const invalidProps = computed(() => {
   return false;
 });
 
-const redirectUrl = computed(() => route.query.return_url);
+const redirectUrl = computed(() => route.query.return_url || route.query.returnUrl);
```
matthew-white, line 63: Could we do a type check of these query parameters, checking that they're string? Vue Router allows query parameters to be array as well.

src/components/enketo-iframe.vue
```diff
@@ -102,11 +107,26 @@ function handleIframeMessage(event) {
     let eventData;
     try { eventData = JSON.parse(event.data); } catch {}
 
-    if (eventData?.enketoEvent === 'submissionsuccess' && redirectUrl.value) {
-      router.push((new URL(redirectUrl.value)).pathname);
+    if (eventData?.enketoEvent === 'submissionsuccess') {
+      if (props.actionType === 'edit') {
+        // for edit we always redirect to Submission details page
+        router.push(submissionPath(form.projectId, form.xmlFormId, props.instanceId));
```
matthew-white, line 113: Do we have a test of this case already (redirect to submission detail page after Enketo edit)?

test/components/enketo-iframe.spec.js
```diff
@@ -77,6 +77,33 @@ describe('EnketoIframe', () => {
...
+        }
+      }
+    });
+    const iframe = wrapper.find('iframe');
+
+    const script = document.createElement('script');
+    script.textContent = `
+      window.parent.postMessage(JSON.stringify({ enketoEvent: 'submissionsuccess' }), "*");
+    `;
+    iframe.element.contentDocument.body.appendChild(script);
```
matthew-white, line 100: This is just a few lines, but I feel like it could help to move it into a little utility function that's housed in this file. Then the test would be shorter and easy to scan quickly.

src/components/enketo-iframe.vue
sadiqkhoja: We have one for WebForms, but not for Enketo. 

Added one

PR #1185, opened 2025-04-14
getodk/central#937: add a negative test

sadiqkhoja:
Addresses https://github.com/getodk/central-frontend/pull/1179#discussion_r2042669745

### DIFF
--- a/src/util/util.js
+++ b/src/util/util.js
@@ -16,10 +16,3 @@ export const noop = () => {};
 export const noargs = (f) => () => f();
 
 export const sumUnderThreshold = (list, threshold) => list.reduce((acc, i) => acc + Math.min(i, threshold), 0);
-
-// To be able to mock it
-export const $location = {
-  get origin() {
-    return window.location.origin;
-  }
-};
--- a/src/util/util.js
+++ b/src/util/util.js
@@ -16,3 +16,10 @@ export const noop = () => {};
 export const noargs = (f) => () => f();
 
 export const sumUnderThreshold = (list, threshold) => list.reduce((acc, i) => acc + Math.min(i, threshold), 0);
+
+// To be able to mock it
+export const $location = {
+  get origin() {
+    return window.location.origin;
+  }
+};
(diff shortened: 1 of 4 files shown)


src/util/util.js
```diff
@@ -16,3 +16,10 @@ export const noop = () => {};
 export const noargs = (f) => () => f();
 
 export const sumUnderThreshold = (list, threshold) => list.reduce((acc, i) => acc + Math.min(i, threshold), 0);
+
+// To be able to mock it
+export const $location = {
```
matthew-white, line 21: Could we name this something other than `$location`? The `$` makes me think of a Vue global. Ideas:

- `mockableLocation`. Not a pretty name, but it makes it clear why it's being used over `window.location`.
- `appLocation`
- We could also add `origin` to the container in `createContainer()`, then replace it in the test. `origin` could be a getter on the container. The container is a common destination for global-y things or things that need to be mocked.

test/components/enketo-iframe.spec.js
```diff
@@ -99,4 +100,29 @@ describe('EnketoIframe', () => {
...
+
+  it('should not bubble up the message event if origin is different', async () => {
+    const wrapper = mountComponent({
+      props: { enketoId, actionType: 'new', instanceId: 'test-instance' },
+      container: {
+        router: mockRouter(`/?parentWindowOrigin=${window.location.origin}`)
+      }
+    });
+
+    sinon.stub($location, 'origin').value('https://example.com');
```
matthew-white, line 112: I had the thought that instead of mocking the location, we could mock the entire event. Then again, what you're doing here is similar to what's done elsewhere in this file and is probably closer to how things would actually happen in production. It at least triggers the event from within the iframe. But if you wanted to avoid mocking the location, I think mocking the event is also an option, so I thought I'd mention it. It'd be something like:

```js
iframe.element.dispatchEvent(new MessageEvent('message', {
  origin: 'https://example.com',
  data
}));
```

matthew-white (approved):

src/util/util.js
sadiqkhoja: I like the 3rd idea. But I have added `location` instead of `origin` in the container as there is another place where I need to mock location.assign

matthew-white (approved):

Issue #1157, opened 2025-03-24
Add more tests for Web Forms integration


sadiqkhoja:
Add test for:

- [x] Validating existing public link continue to work. 
- [x] Enketo is rendered if web-forms in not enabled
- [x] Redirection configured in nginx config is working correctly
- [x] Single Submission link allows only one submission

PR #1186, opened 2025-04-15
Enketo: Ensure only one submission can be made with "Single" public link

sadiqkhoja:
TBA

-->


src/components/enketo-iframe.vue
```diff
@@ -112,4 +144,8 @@ useEventListener(window, 'message', handleIframeMessage, false);
   height: 100vh;
   width: 100%;
 }
+
+.checkSubmittedIframe {
+  display: none;
+}
```
sadiqkhoja, line 150: remove this

src/components/enketo-iframe.vue
matthew-white, line 1: Is there a way to add a test of this behavior? Like could we set a cookie in a test, then assert that the Enketo thanks page is rendered? Or are you planning to test this in an end-to-end test (#1157)?

matthew-white (approved):

src/components/enketo-iframe.vue
```diff
@@ -47,24 +47,43 @@ const { t } = useI18n();
...
-
-  query.parentWindowOrigin = window.location.origin;
-
+const lastSubmitted = (enketoOnceId) => {
+  const iframe = document.createElement('iframe');
+  iframe.src = '/-/single/check-submitted';
+  iframe.style.display = 'none';
+
+  return new Promise((resolve) => {
+    iframe.onload = () => {
```
matthew-white, line 63: This might be done in another PR, but what do you think about showing a loading message while `enketoSrc.value == null`? Since checking the `enketoOnceId` cookie is an async process, I think there's a possibility that the page could be completely blank while the cookie iframe is loading.

matthew-white: Related idea: now that `enketoSrc.value` can be nullish, what do you think about checking for that possibility in the template? I.e., changing the `v-else` to `v-else-if`. I don't think it was possible before for `enketoSrc.value` to be nullish, at least as long as `invalidProps.value` was `false`.

```vue
<iframe v-else-if="enketoSrc != null" ... :src="enketoSrc">
```

matthew-white (approved):
I took a look at the latest commit a589510aa6cce765c2e402f6b423879d9c7e7f44, and it looks good to me.

PR #1193, opened 2025-04-20
fixes 962: remove length of enketoId

sadiqkhoja:
Closes getodk/central#962

**Background**

Enketo Express allows length of enketoId to be configurable between 4-31. That's why we don't want to hard code length of enketoId at our end.

### DIFF
--- a/main.nginx.conf
+++ b/main.nginx.conf
@@ -68,24 +68,23 @@ http {
     # whether to show a WebForm or an iframed Enketo.
     #
     # Following are the locations that serve a Form and these are redirected to the frontend:
-    # N.B: For Single submission Enketo ID (enketoOnceId) is 64 characters long otherwise it's 31 char
-    location ~ "^/-/single/(?<enketoId>[a-zA-Z0-9]{31}|[a-zA-Z0-9]{64})$" {
+    location ~ "^/-/single/(?<enketoId>[a-zA-Z0-9]+)$" {
       # Form fill link, public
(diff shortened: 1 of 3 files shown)


src/routes.js
```diff
@@ -735,15 +735,6 @@ const routes = [
       restoreSession: true,
       requireLogin: false,
       title: () => [form.nameOrId]
-    },
-    // onceEnketoId can be used by only public-links as it doesn't support offline mode
-    beforeEnter: (to) => {
-      const { enketoId, actionType } = to.params;
-
-      if (enketoId.length === 64 && actionType === 'offline') {
```
matthew-white, line 743: Could you say more about why we don't need this check anymore?

Would it be worth moving the check to `FormSubmission`? Something like:

```js
watchEffect(() => {
  if (form.dataExists && props.enketoId === form.enketoOnceId && props.actionType === 'offline')
    router.push('/');
});
```

main.nginx.conf
matthew-white, line 1: Just to mention it somewhere, we'll need to make these changes in the `central` repo as well.

src/routes.js
sadiqkhoja: I removed it as it was not adding any value, if somebody tries to access offline mode with enketoOnceId, then Enketo will throw 404 error

matthew-white (approved):

main.nginx.conf
sadiqkhoja: https://github.com/getodk/central/pull/969

PR #1190, opened 2025-04-16
Add tests of user preferences

matthew-white:
This PR closes #1049, adding tests of user preferences. The goals of this PR are to:

- Prevent someone from accidentally deleting all user preference functionality outright.
- Make it clear how to read and write user preferences in Frontend / what the contract is.
- Provide a template for future tests. Quite possibly the core functionality won't ever need to change, but as we add new specific preferences, we'll also need to write new preference normalizers. It was with that in mind that I added tests of the existing normalizers.

I didn't write tests for absolutely all behavior. I also didn't write any new tests of components that read or write user preferences. However, I did write more than the 3 or 4 tests that the comment on the issue suggests.

#### Before submitting this PR, please make sure you have:

- [x] run `npm run test` and `npm run lint` and confirmed all checks still pass OR confirm CircleCI build passes
- [x] verified that any code or assets from external sources are properly credited in comments or that everything is internally sourced

PR #1195, opened 2025-04-21
Whats new modal for form edit changes with user preferences

ktuite:
Closes https://github.com/getodk/central/issues/801

(needs tests)


src/components/whats-new.vue
matthew-white, line 1: I think there needs to be some sort of logic to prevent the modal from being shown if it's an entirely new Central install. I think that's what the issue means by:

> Users who &hellip; had account prior to update

src/components/whats-new.vue
```diff
@@ -0,0 +1,70 @@
...
+}
+
+</script>
+
+<i18n lang="json5">
+  {
+    "en": {
+      // This is the title at the top of a pop-up.
+      "title": "Form drafts have moved",
+      "body": "Create a new form and edit it on the new Edit Form tab",
```
matthew-white, line 63: Capitalize "form"?

src/components/whats-new.vue
```diff
@@ -0,0 +1,70 @@
...
+
+defineOptions({
+  name: 'WhatsNew'
+});
+
+watch(() => projects.dataExists, () => {
+  const canUpdateForm = currentUser.can('form.update') ||
+    projects.data.some(project => project.verbs.has('form.update'));
+  if (canUpdateForm && !currentUser.preferences.site.whatsNewDismissed2025_1) {
+    isVisible.value = true;
```
matthew-white, line 47: Given that the modal isn't shown until after the projects response is received, and we know that response can be a little slow, I think it'd be a good idea to account for the case where another modal is already open. That modal could be `ProjectNew` or `AnalyticsIntroduction` or (in the future) the modal for getodk/central#849.

One way to check whether another modal is open is to see whether `document.body` has the `modal-open` class. I don't love using an HTML class like that, but it would probably work for now. I've had the thought multiple times that at some point, it could be helpful to store a piece of state in the `container` object that indicates whether any modal is open.

src/components/whats-new.vue
```diff
@@ -0,0 +1,70 @@
...
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+<template>
+  <modal :state="isVisible" backdrop :hideable="true" @hide="hideModal">
+    <template #title>{{ $t('title') }}</template>
+    <template #body>
+      <div class="modal-introduction">
```
matthew-white, line 16:
```suggestion
      <p class="modal-introduction">
```

We make the `.modal-introduction` element a `<div>` when it's wrapping multiple `<p>` elements, but there's always some `<p>`.

src/components/whats-new.vue
ktuite: I'm not sure how to do this. We have the `createdAt` on the user response but I don't know how to check when an upgrade happened. We were talking about this last week of how to log this in the audit log. 

I could approximate it by checking against a date? If the user was created after ~end of April, even if they started on some older version of Central and then updated, then this message isn't too relevant to them?

matthew-white: Checking against a date sounds good to me. Something like: a user created after May 6 should not see the message. May 6 is the approximate date of release, a little on the early side so we don't miss showing the modal to anyone who should see it.

ktuite:
Wanted to capture some thoughts about forward/backward compatibility:

I'm assuming that for a LATER release, like 2025.2, whether or not we have a new "What's New" modal, we take out the code that checks for the 2025.1 "show modal about form drafts being moved" condition. 

If there's something we want to highlight in 2025.2 (e.g. entities from repeats!) that is a new condition, a user preference, different modal content. The old modal does not show and the old code is removed. Or if there's nothing new, we either keep the old what's new modal for one more release (with the same preferences) or remove the modal entirely. 

If people skip an update (or more) they may miss some info about what's changed but a) it won't be an overwhelming slew of popups for the user and b) it will be easier for us to implement. Some user preferences will be stale/irrelevant but that is ok. The alternative of trying to see how much of an update delta the server and each user has experienced and craft a tailored summary of that sounds like a lot of work.

matthew-white:
I agree that we should never have more than one What's New modal, all competing to be shown. I also think that trying to calculate an update delta doesn't sound fun.

I think it'd be OK for the modal in this PR to exist in the codebase for multiple releases, unless there's something else we want to highlight. v2025.2 will hopefully be out about a month after v2025.1, so there will probably be users upgrading directly from v2024.3 to v2025.2. I think we probably want those users to see this messaging.

matthew-white:
@ktuite, if you're working on tests, feel free to cherry-pick the changes to test/data/users.js from #1190 if those would be helpful to you. Alternatively, you can always just specify the full `preferences` object, which can include empty objects for `site` and `projects`.

test/components/whats-new.spec.js
```diff
@@ -0,0 +1,79 @@
...
+  describe('shows modal', () => {
+    it('shows modal to admin with 0 projects', async () => {
+      mockLogin({ createdAt: '2025-01-01' });
+      const app = await load('/', { root: false });
+      const baseModal = app.findComponent(WhatsNew).findComponent(Modal);
+      baseModal.exists().should.be.true;
+      baseModal.props().state.should.be.true;
+    });
+
+    it('shows modal to admin with more than 1 project', async () => {
```
matthew-white, line 18: This says "more than 1", but the test only seems to create 1.

test/components/whats-new.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      mockLogin({ createdAt: '2025-01-01' });
+      testData.extendedProjects.createPast(1);
+      const app = await load('/', { root: false });
+      const baseModal = app.findComponent(WhatsNew).findComponent(Modal);
+      baseModal.exists().should.be.true;
+      baseModal.props().state.should.be.true;
+    });
+
+    it('shows modal to project manager with 1 project', async () => {
+      mockLogin({ role: 'manager' });
```
matthew-white, line 28: The `role` here is the sitewide role.

```suggestion
      mockLogin({ role: 'none' });
```

test/components/whats-new.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      mockLogin({ createdAt: '2025-05-07' });
+      testData.extendedProjects.createPast(1);
+      const app = await load('/', { root: false });
+      const baseModal = app.findComponent(WhatsNew).findComponent(Modal);
+      baseModal.exists().should.be.true;
+      baseModal.props().state.should.be.false;
+    });
+
+    it('does not show modal to project viewer with no management role', async () => {
+      mockLogin({ role: 'viewer' });
```
matthew-white, line 48:
```suggestion
      mockLogin({ role: 'none' });
```

matthew-white (approved):

src/components/whats-new.vue
matthew-white: > If the user was created after ~end of April, even if they started on some older version of Central and then updated, then this message isn't too relevant to them?

I was thinking more about this case. Some users won't upgrade their Central deployment right away, so you're right that it's definitely conceivable that the What's New modal could be relevant to a user created after May 6 (e.g., a project manager is created under v2024.3 on May 7, then the deployment is upgraded on June 7). But I think it's OK if we miss showing the modal to a few users like that. Presumably such users also won't be as familiar with the old interface.

In any case, I don't think we really have another option beyond what you're doing, until we implement getodk/central#971. But I think checking against a date is quite reasonable.

PR #1198, opened 2025-04-24
Fixes: single submission public link

sadiqkhoja:
Addresses: https://github.com/getodk/central/issues/961#issuecomment-2823302606

### DIFF
--- a/main.nginx.conf
+++ b/main.nginx.conf
@@ -84,7 +84,7 @@ http {
       # Offline new Submissions - can be public if it has st query parameter
       return 301 "/f/$enketoId/offline$is_args$args";
     }
-    location ~ "^/-/(?<enketoId>[a-zA-Z0-9]+)$" {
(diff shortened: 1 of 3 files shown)


main.nginx.conf
```diff
@@ -84,7 +84,7 @@ http {
       # Offline new Submissions - can be public if it has st query parameter
       return 301 "/f/$enketoId/offline$is_args$args";
     }
-    location ~ "^/-/(?<enketoId>[a-zA-Z0-9]+)$" {
+    location ~ "^/-(?!thanks$)(?<enketoId>[a-zA-Z0-9]+)$" {
```
sadiqkhoja, line 87: because we removed length of `enketoId` from here so now it is matching `thanks` as well.

Updated regex to excluding 'thanks' from matching.

matthew-white (approved):

PR #1207, opened 2025-05-02
Features/970 handle submission attachments

sadiqkhoja:
Closes getodk/central#970

### DIFF
--- a/src/components/web-form-renderer.vue
+++ b/src/components/web-form-renderer.vue
@@ -359,7 +359,7 @@ const handleResult = () => {
     if (error.response && isProblem(error.response.data) && error.response.data.code === 403.1) {
       showModal({ type: 'sessionTimeoutModal' });
     } else {
-      showModal({ type: 'errorModal', errorMessage: requestAlertMessage(i18n, error, false) });
+      showModal({ type: 'errorModal', errorMessage: requestAlertMessage(i18n, error) });
     }
   }
 
--- a/src/components/web-form-renderer.vue
+++ b/src/components/web-form-renderer.vue
@@ -20,7 +20,7 @@ except according to the terms contained in the LICENSE file.
       @submit="handleSubmit"/>
   </template>
 
-  <modal id="web-form-renderer-submission-modal" v-bind="submissionModal" :hideable="submissionModal.hideable" backdrop @hide="hideModal()">
+  <modal id="web-form-renderer-submission-modal" v-bind="submissionModal" backdrop @hide="hideModal()">
     <template #title>{{ $t(submissionModal.type + '.title') }}</template>
     <template #body>
       <div class="modal-introduction">
@@ -77,7 +77,7 @@ except according to the terms contained in the LICENSE file.
 </template>
 
 <script setup>
-import { computed, createApp, getCurrentInstance, inject, onUnmounted, reactive, ref } from 'vue';
+import { computed, createApp, getCurrentInstance, inject, onUnmounted } from 'vue';
 import { useRoute, useRouter } from 'vue-router';
 /* eslint-disable-next-line import/no-unresolved -- not sure why eslint is complaining about it */
 import { OdkWebForm, webFormsPlugin, POST_SUBMIT__NEW_INSTANCE } from '@getodk/web-forms';
@@ -136,7 +136,7 @@ const isEdit = computed(() => props.actionType === 'edit');
  * Convert AxiosResponse into subset of web standard  {@link Response} that satisfies Web-Forms'
  * requirements
  */
-const transformAttachmentResponse = async (axiosResponse) => {
+const transformAttachmentResponse = (axiosResponse) => {
   const { data, status, statusText, headers } = axiosResponse;
 
   const fetchHeaders = new Headers();
@@ -202,7 +202,7 @@ const fetchSubmissionAttachment = (attachmentName) => {
     url: requestUrl,
     alert: false
   })
-    .then(transformAttachmentResponse) // TODO: prob same things as getFormAttachment fn
+    .then(transformAttachmentResponse)
     .catch(noop);
 };
 
@@ -312,14 +312,18 @@ const uploadAttachment = async (attachment, instanceId) => {
   return { name: attachment.name, result };
 };
 
-const submissionResult = reactive({});
-const submissionData = ref();
+const submissionResult = {};
+/**
+ * Holds the data received from ODK Web Forms and sent to the server when the Send or Retry
+ * buttons are pressed. This supports retry functionality, since ODK Web Forms does not provide
+ * a way to access submission data at an arbitrary point in time.
+ */
+let submissionData = {};
 let clearForm;
 
 const initializeSubmissionState = (data, clearFormCallback) => {
-  submissionData.value = data;
+  submissionData = data;
 
-  submissionResult.inProgress = false;
   submissionResult.primaryInstanceResult = {
     success: false
   };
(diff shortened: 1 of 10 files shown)


test/run.sh
```diff
@@ -5,11 +5,12 @@ output=$(mktemp)
...
 NODE_ENV=test karma start | tee "$output"
 # Search for: warnings from console.warn(), including Vue warnings; Sass
-# warnings; and warnings from Karma.
-if grep -F -e 'WARN LOG:' -e 'ERROR LOG:' -e 'Module Warning' -e 'WARN [web-server]:' -C5 "$output"; then
-	# Reset the text format in case the search results contained formatted text.
-	tput sgr0
-	echo
-	echo 'All tests passed, but there were warnings: see above.'
-	exit 1
+# warnings; and warnings from Karma. Ignore build warnings
```
matthew-white, line 8: If we were to merge #1217 and address getodk/central#1029, would that resolve all the build warnings?

matthew-white: Hmm maybe not, I think there'd still be this warning:

```
WARN LOG: 'Feature flag __VUE_PROD_HYDRATION_MISMATCH_DETAILS__ is not explicitly defined.
```

matthew-white: I think it is possible for there to be build warnings that we'd want to be informed about. At the same time, we need tests to pass despite these warnings. What do you think about having test/run.sh count the number of warnings and assert that it's less than some fixed number? (3?)

src/components/web-form-renderer.vue
```diff
@@ -305,36 +405,18 @@ const uploadAttachment = async (attachment, instanceId) => {
...
-    if (instanceId) {
-      const attachmentRequests = data.attachments.map(a => () => uploadAttachment(a));
-      const attachmentResults = await runSequentially(attachmentRequests);
-
-      // TODO: what to do if attachments upload fail - blocked, need to define requirements / UX
-      if (attachmentResults.every(r => !isProblem(r))) {
-        callback({ next: POST_SUBMIT__NEW_INSTANCE }); // tell OWF to clear out the Form
-        if (isPublicLink.value) {
-          showModal(submissionModal, { type: 'thankYouModal' });
-          formVersionXml.reset(); // hides the Form
```
matthew-white, line 329: Is this line still important? Or was it not necessary before?

matthew-white (approved):
I haven't taken a look at tests yet, but the changes to src/ look good to me. There are a number of different cases, but the logic flows nicely. 💯 @sadiqkhoja, feel free to merge at any point. I'll start taking a look at tests now.

vue.config.js
```diff
@@ -25,6 +25,12 @@ module.exports = {
     if (process.env.NODE_ENV === 'test') {
       config.resolve.alias.set('vue$', 'vue/dist/vue.esm-bundler.js');
     }
+
+    config.plugin('define').tap(definitions => {
+      // eslint-disable-next-line no-param-reassign
+      definitions[0].__VUE_PROD_HYDRATION_MISMATCH_DETAILS__ = JSON.stringify(false);
```
matthew-white, line 31: I think you mean to revert this, but I thought I'd add a comment. I actually sort of like that it's telling us about issues like getodk/central#1029. But I also think that once we move to Vitest, this warning will probably go away entirely.

test/run.sh
sadiqkhoja: build warning tolerance was added in https://github.com/getodk/central-frontend/pull/1218

test/components/web-form-renderer.spec.js
```diff
@@ -168,15 +170,94 @@ describe('WebFormRenderer', () => {
...
+  });
+
+  const fileToUpload = new File(['dummy content'], '1746140510984.jpg', { type: 'image/jpeg' });
+
+  const uploadFile = async (component) => {
+    const fileInput = component.find('.odk-form input[type="file"]');
+    const dataTransfer = new DataTransfer();
+    dataTransfer.items.add(fileToUpload);
+    fileInput.element.files = dataTransfer.files;
+    await fileInput.trigger('change');
```
matthew-white, line 191: We have a `setFiles()` utility function that can do this part.

test/components/web-form-renderer.spec.js
matthew-white, line 1: I was thinking about whether there were any more tests that could be added:

- How about adding a test of what happens if the first submission request (with the XML) fails? We can assert that there isn't a "Try again" button. It's different from if a submission attachment request fails. Do we already have a test like that?
- Testing when the `sessionTimeoutModal` is `hideable` vs. not, i.e., testing a 403.1 response to a submission attachment POST. That could make an assertion about the `hideable` prop of the `Modal` component and also assert which buttons are visible.

src/components/web-form-renderer.vue
```diff
@@ -185,22 +195,24 @@ const fetchSubmissionAttachments = () => {
...
     .catch(noop);
 };
 
 const editInstanceOptions = computed(() => {
   if (isEdit.value && submissionAttachments.dataExists) {
     return {
       resolveInstance: fetchSubmissionXml,
-      attachmentFileNames: submissionAttachments.data,
+      attachmentFileNames: submissionAttachments.data
+        .filter(a => a.exists)
```
matthew-white, line 214: How about adding a test of this filtering?

src/components/web-form-renderer.vue
```diff
@@ -256,46 +264,138 @@ const getAttachment = (url) => {
...
+    };
+    const { data } = await request(requestOptions);
+    result.success = true;
+    result.data = data;
+  } catch (error) {
+    result.success = false;
+    result.data = error;
+  }
+
+  return { name: attachment.name, result };
```
matthew-white, line 312: Does this need to return the name? The `attachmentResult` map already maps attachment name to result, so I feel like the name isn't needed here too. Not a big deal at all, might just be slightly simpler without it.

src/components/web-form-renderer.vue
```diff
@@ -256,46 +264,138 @@ const getAttachment = (url) => {
...
+};
+
+const submissionResult = reactive({});
+const submissionData = ref();
+let clearForm;
+
+const initializeSubmissionState = (data, clearFormCallback) => {
+  submissionData.value = data;
+
+  submissionResult.inProgress = false;
```
matthew-white, line 322: Does `inProgress` end up getting used? I don't see it referenced anywhere.

matthew-white (approved):

src/components/web-form-renderer.vue
sadiqkhoja: I wanted to keep this line but it is making styling of modal weird. 

Previously `thankYouModal` was hideable so it was important to hide the Form but now it is not

src/components/web-form-renderer.vue
```diff
@@ -256,46 +264,138 @@ const getAttachment = (url) => {
...
+  } catch (error) {
+    result.success = false;
+    result.data = error;
+  }
+
+  return { name: attachment.name, result };
+};
+
+const submissionResult = reactive({});
+const submissionData = ref();
```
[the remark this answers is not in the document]
sadiqkhoja: I was initialing thinking to use watcher to show appropriate modal based on the submissionData and submissionResult but then dropped that idea.

test/components/web-form-renderer.spec.js
sadiqkhoja: first one is tested by `should show error modal in case of submission failure`, I have added an assertion that "try again" button is not there

src/components/web-form-renderer.vue
sadiqkhoja: it is needed to the value in `attachmentResult` map in this line: 

```
      submissionResult.attachmentResult.set(r.name, r.result);
```

PR #1208, opened 2025-05-02
Add a slot for modal banner

ktuite:
Part of https://github.com/getodk/central/issues/944

Here's how an image COULD look as a banner in a modal. This was set up as a test (with a local image file on my machine) because we're waiting on vite to load images properly.
Screenshot 2025-05-02 at 1 19 27 PM

This PR includes this SVG gradient in the What's New modal as an example usage of the banner slot:
Screenshot 2025-05-02 at 1 02 17 PM


lognaturel (approved):
Nice and flexible, styling makes sense, I see that it's been tried with an image and I believe that it does work.

PR #1221, opened 2025-05-06
Upgarde fakerjs to the latest

sadiqkhoja:
faker v4 is dead for a long time, I can't find adequate documentation or git repo for that version either.

What forced me to do this now? `faker.random.word()` sometimes generates multiple words causing some e2e tests to fail.


matthew-white:
Good idea, just having good access to docs is reason enough to make this change. :+1:

It'd be convenient if we could merge this PR after #1222.

matthew-white (approved):
Awesome, thank you for doing this! ✨

PR #1223, opened 2025-05-06
Teleport modals

matthew-white:
The Bootstrap [docs](https://getbootstrap.com/docs/3.4/javascript/#modals) say about modals:

> Always try to place a modal's HTML code in a top-level position in your document to avoid other components affecting the modal's appearance and/or functionality.

I've wanted to do this for a while. However, it wasn't super easy to do it in Vue. However, now that we have [teleport](https://vuejs.org/guide/built-ins/teleport) in Vue 3, it's something we can do.

I haven't confirmed yet, but I'm hopeful that this change will close getodk/central#1003. That issue is happening because the modal is a descendant element of the drop zone component. But if we move the modal somewhere else in the document, that won't be the case anymore.

I'm hoping to get this in before regression testing so that if this approach doesn't work for any modal, we'll learn about it.

The main issue I ran up against was in testing. I kept running into errors and warnings. I tried stubbing `Teleport` in a few different ways, but it didn't work. I ended up running with a different approach, adding the component `TeleportIfExists`. I actually think it might have value on its own. For example, I think it works well in `SubmissionList`.

Here's how I tried stubbing `Teleport`. None of these worked:

1. What's on `master`: `teleport: true`
2. The same as what we do for `transition-group`: `{ inheritAttrs: false, template: '<slot />' }`
3. I tried this single file component:
```vue
<template>
  <slot v-bind="$attrs"></slot>
</template>

<script setup>
defineOptions({
  inheritAttrs: false
});
defineProps({
  to: {
    type: String,
    required: true
  }
});
</script>
```

Vue Test Utils [suggests](https://test-utils.vuejs.org/guide/advanced/teleport.html#Mounting-the-Component) creating the teleport target in the DOM in a `beforeEach` hook, but that seemed more cumbersome to me.

### DIFF
--- a/test/util/http/common.js
+++ b/test/util/http/common.js
@@ -99,6 +99,11 @@ export function testModalToggles({
   // shown
   respond = (series) => series
 }) {
+  if (typeof modal === 'string')
+    throw new Error('modal must be a component, not a string');
+  const getModal = (component) => (modal === Modal
+    ? component.getComponent(Modal)
+    : component.getComponent(modal).getComponent(Modal));
   return this
     // First, test that the show button actually shows the modal.
     .afterResponses(component => {
@@ -107,7 +112,7 @@ export function testModalToggles({
     .request(component => component.get(show).trigger('click'))
     .modify(respond)
     .afterResponses(async (component) => {
-      const m = component.getComponent(modal);
+      const m = getModal(component);
       m.props().state.should.be.true;
 
       // Next, test that `modal` listens for `hide` events from Modal.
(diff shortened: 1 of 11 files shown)


test/util/http/common.js
```diff
@@ -99,6 +99,11 @@ export function testModalToggles({
   // shown
   respond = (series) => series
 }) {
+  if (typeof modal === 'string')
+    throw new Error('modal must be a component, not a string');
```
matthew-white, line 103: I needed to change `modal: '#web-form-renderer-submission-modal'` to `modal: Modal` above. `.getComponent()` doesn't seem to play well with a string when working with modals.

sadiqkhoja (approved):
Looks good to me.

PR #1230, opened 2025-05-07
Fix: don't split css into multiple files

sadiqkhoja:
Closes getodk/central#1034

Locally I don't see any splashes on the home page. Deployed this change on dev server as well, there are no more splashes on Home or Submission pages

There is an open issue (https://github.com/vitejs/vite/issues/3924) in vite related to this. I think loading component dynamically causes js to load before css. I feel vite/vue should have ability to wait for css loading before adding elements in the DOM. I think we would be able to fix this issue in much clearer way in https://github.com/getodk/central-frontend/issues/1212.

This change fixes the problem in smallest possible way without having any significant impact. I'm also thinking if there is any benefit of serving component css and js file separately in our situation?

No

No

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -53,7 +53,8 @@ export default defineConfig(({ mode }) => ({
   build: {
     target: buildTarget,
     // `false` during dev for performance reasons
-    reportCompressedSize: mode === 'production'
+    reportCompressedSize: mode === 'production',
+    cssCodeSplit: false
   },
   // Not sure why this is needed in addition to build.target above and why it's
   // only an issue in development. `npm run dev` doesn't work without this.


matthew-white:
When I read about `cssCodeSplit`, it says:

> If disabled, all CSS in the entire project will be extracted into a single CSS file.

I tried it out locally, and it does look like that's what it's doing. With `cssCodeSplit: true` (the default), there are many CSS files. The largest one is this one:

```
dist/assets/index-DezQaQDP.css                              111.12 kB │ gzip:    19.63 kB
```

With `cssCodeSplit: false`, there is just one CSS file:

```
dist/assets/style-BAOwY-Kl.css                              201.03 kB │ gzip:    35.02 kB
```

It doesn't look like `cssCodeSplit` affects the size of .js files. For example, date-range-picker-*.js is the same either way. I think that means that the JS and CSS aren't bundled together or anything like that. Am I getting that right, @sadiqkhoja?

It's a little unfortunate that `cssCodeSplit: false` will increase the size of the initial bundle. At the same time, 35 kB gzipped seems pretty OK.

> I feel vite/vue should have ability to wait for css loading before adding elements in the DOM.

Agreed! It's unfortunate that there's no such option. We can keep our eye on the issue you mentioned.

> I think we would be able to fix this issue in much clearer way in #1212.

How would #1212 help with this? I thought that issue ended up being mostly just about asset filenames. Are you thinking that there may be another rollup option that would help with this?

Bottom line: With the change in this PR, we can get rid of the `<`. I also think that 35 kB is acceptable for the CSS asset. We can keep our eye on other options, but I think it makes sense to run with this.

matthew-white (approved):

sadiqkhoja:
yes, you are right. My assumption that css is merged into JS is wrong. All the CSS styles are combined into one.

With https://github.com/getodk/central-frontend/issues/1212, I am thinking we might be able to remove AsyncRoute and leverage [Vue Router's Lazy Loading](https://router.vuejs.org/guide/advanced/lazy-loading.html)

matthew-white:
> With #1212, I am thinking we might be able to remove AsyncRoute and leverage [Vue Router's Lazy Loading](https://router.vuejs.org/guide/advanced/lazy-loading.html)

Yeah, `AsyncRoute` might be less useful today than it was in the past. Previously, `AsyncRoute` guaranteed that navigation itself was synchronous. If the code called `router.push()`, the user would be navigated synchronously to the next page: they wouldn't linger on the previous page. However, at some point (with Vue 3, I think?), router navigation as a whole became asynchronous. It's still near-instananeous though: I think navigation is completed as a microtask unless a navigation guard slows it down with a request or something.

I also ran into a bunch of issues with route component reuse that caused code to be more complex (logic related to the `beforeRouteUpdate` navigation guard). That's why `AsyncRoute` has the `key` option. That could probably be separated out though: async components and component reuse are two different concerns.

I'd be open to looking into removing `AsyncRoute`, but I feel like we should do that as part of a separate issue. Would you be up for filing such an issue? For #1212, at this point I'm picturing just a small change: removing webpack magic comments that are now no-ops and just noise in the codebase.

PR #1237, opened 2025-05-09
Change webform opt in from confirmation to regular modal and add image

ktuite:
Part of https://github.com/getodk/central/issues/944 for the Webforms Enabled modal

I changed the opt in to webforms popup to be a regular modal so i could use the new banner slot. 

### DIFF
--- a/src/components/form/web-forms-settings-confirmation.vue
+++ b/src/components/form/web-forms-settings-confirmation.vue
@@ -10,10 +10,14 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <confirmation v-bind="confirm" @hide="$emit('hide')" @success="setWebformsEnabled">
+  <modal :state="state" :hideable="!form.awaitingResponse" backdrop
+  @hide="$emit('hide')">
+    <!-- TODO: for ODK Web Forms, we need to show modal with an image. Reuse "what's new" modal,
+        once it is done in getodk/central#801 -->
+    <template v-if="webformsEnabled" #title>ODK Web Forms</template>
+    <template v-else #title>Enketo</template>
+
     <template v-if="webformsEnabled" #body>
-      <!-- TODO: for ODK Web Forms, we need to show modal with an image. Reuse "what's new" modal,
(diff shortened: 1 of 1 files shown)


src/components/form/web-forms-settings-confirmation.vue
```diff
@@ -10,10 +10,14 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <confirmation v-bind="confirm" @hide="$emit('hide')" @success="setWebformsEnabled">
+  <modal :state="state" @hide="$emit('hide')" @success="setWebformsEnabled">
```
matthew-white, line 13: I think we need `hideable` and `backdrop`. Also, `@success` can be removed.

src/components/form/web-forms-settings-confirmation.vue
```diff
@@ -102,6 +106,12 @@ const setWebformsEnabled = () => {
 };
 </script>
 
+<style lang="scss">
+  #modalBanner {
```
matthew-white, line 110: I could imagine another component using an `id` named `modalBanner`. CSS isn't scoped to components: once loaded here, it will apply anywhere in the app. How about either: (1) renaming this id (e.g., to `form-web-forms-settings-confirmation-banner`), or (2) moving this style to the `Modal` component (e.g., `.modal-banner img { width: 100%; }`).

src/components/form/web-forms-settings-confirmation.vue
```diff
@@ -30,27 +34,43 @@ except according to the terms contained in the LICENSE file.
...
         </template>
       </i18n-t>
+      <div class="modal-actions">
+        <button class="btn btn-link" type="button" @click="$emit('hide')">
+          {{ $t('action.cancel') }}
+        </button>
+        <button type="button" class="btn btn-primary"
+          @click="setWebformsEnabled">
+          {{ $t('webformsConfirmation.useOdkWebForms') }}
+        </button>
```
matthew-white, line 44: I see that `confirm.value` had an `awaitingResponse` property. That makes me think that we need a `<spinner>` here.

PR #1238, opened 2025-05-09
Fixes getodk/central#1047: Update filter styles

sadiqkhoja:
Closes getodk/central#1047

image

Besides changes requested in the issue, I have made the "type" dropdown on audits pages to open on click of label as well (not 100% true, actually dropdown is now bigger and label is on a upper z-index because we can't open select on the click of label - not even using JS). Also added custom icon-arrow for it, default looked different than the other filter arrow icon.

I also took the liberty to move column selected to the left, it was not aligning well with the refresh button after matching its style with other filters.

### DIFF
--- a/src/components/audit/filters/action.vue
+++ b/src/components/audit/filters/action.vue
@@ -11,6 +11,7 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <label id="audit-filters-action" class="form-group">
+    <span class="form-label">{{ $t('field.type') }}</span>
     <select class="form-control" :value="modelValue"
       @change="$emit('update:modelValue', $event.target.value)">
       <option v-for="option of options" :key="option.value"
@@ -18,7 +19,7 @@ except according to the terms contained in the LICENSE file.
         {{ option.text }}
       </option>
     </select>
-    <span class="form-label">{{ $t('field.type') }}</span>
+    <span class="icon-angle-down"></span>
   </label>
 </template>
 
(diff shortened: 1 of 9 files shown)


src/components/audit/filters/action.vue
```diff
@@ -105,8 +106,34 @@ export default {
...
-  // Not all browsers support styling an <option> element this way.
-  font-weight: bold;
+#audit-filters-action {
+  select {
+    appearance: none;
+    cursor: pointer;
+    padding-top: 23px;
+    padding-right: 25px;
+    height: 48px;
+  }
```
matthew-white, line 116: Should some of these styles apply to `<select>` elements outside this component? I'm thinking of `ProjectSort`, for example. I wonder whether we should create a generic `<single-select>` component at some point.

src/components/audit/filters/action.vue
```diff
@@ -105,8 +106,34 @@ export default {
...
-.audit-filters-action-category {
-  // Not all browsers support styling an <option> element this way.
-  font-weight: bold;
+#audit-filters-action {
+  select {
+    appearance: none;
+    cursor: pointer;
+    padding-top: 23px;
+    padding-right: 25px;
+    height: 48px;
```
matthew-white, line 115: What do you think about making this `min-height`? Like maybe it ends up bigger due to the font that a user is using. You can use a browser setting to override a website's font size, e.g., to make it bigger.

matthew-white (approved):

matthew-white:
I went ahead and merged in order to get this on the QA server. I suggested turning a `height` into a `min-height`, but that could happen in a follow-up PR.

src/components/audit/filters/action.vue
sadiqkhoja: https://github.com/getodk/central-frontend/issues/1240

PR #1245, opened 2025-05-12
Expand entity properties by default on edit form page

ktuite:
Closes https://github.com/getodk/central/issues/1055

### DIFF
--- a/src/components/expandable-row.vue
+++ b/src/components/expandable-row.vue
@@ -40,7 +40,7 @@ defineOptions({
   name: 'ExpandableRow'
 });
 
-const expanded = ref(false);
+const expanded = ref(true);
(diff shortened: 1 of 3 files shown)


src/components/expandable-row.vue
```diff
@@ -40,7 +40,7 @@ defineOptions({
   name: 'ExpandableRow'
 });
 
-const expanded = ref(false);
+const expanded = ref(true);
```
matthew-white, line 43: `ExpandableRow` is also used in `ConnectionToForms`. Should it be expanded by default there as well?

ktuite: No, probably not. There can be a lot of datasets/properties there that would be annoying to have all expanded by default.

matthew-white (approved):

PR #1248, opened 2025-05-12
Restyle section icon in FormEditSection

matthew-white:
Closes getodk/central#1056.

I tried it out locally.

### DIFF
--- a/src/assets/scss/_mixins.scss
+++ b/src/assets/scss/_mixins.scss
@@ -1,5 +1,12 @@
 @import './variables';
 
+@mixin icon-box {
+  padding: 10px;
+  border-radius: 6px;
+  color: $color-accent-primary;
+  background-color: rgba($color-accent-primary, 0.1);  
+}
+
 
 
 ////////////////////////////////////////////////////////////////////////////////
--- a/src/components/form/edit/section.vue
+++ b/src/components/form/edit/section.vue
@@ -99,9 +99,9 @@ $heading-margin-bottom: 10px;
   // Use flexbox to center the icon.
   display: flex;
   align-items: center;
-  flex-shrink: 0;
   justify-content: center;
 
+  flex-shrink: 0;
   position: relative;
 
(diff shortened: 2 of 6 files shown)


src/assets/scss/_mixins.scss
```diff
@@ -1,5 +1,12 @@
 @import './variables';
 
+@mixin icon-box {
+  padding: 10px;
+  border-radius: 6px;
+  color: $color-accent-primary;
+  background-color: rgba($color-accent-primary, 0.1);  
+}
```
matthew-white, line 8: We use these styles in a few different places. I didn't want to copy them somewhere additional, so I moved them into a mixin. I could have made it an HTML class or a component instead, but this felt like the lightest-weight approach.

src/components/form/edit/section.vue
```diff
@@ -81,24 +81,27 @@ $heading-margin-bottom: 10px;
...
     overflow-x: hidden;
-    padding-top: 16px;
+    padding-top: 6px;
   }
 }
 
 .form-edit-section-icon-container {
+  @include icon-box;
+  font-size: 24px;
+  padding-inline: 0;
```
matthew-white, line 91: This zeroes out the horizontal padding from `icon-box`, but not the vertical padding. We want the box to be a fixed width so that the different section icons all line up. But the width of different icons vary quite a bit. In some cases, I think it works best for the icon to eat into the padding. To do so, we remove the horizontal padding and let the fixed with do all the work.

src/components/submission/basic-details.vue
```diff
@@ -114,11 +114,8 @@ export default {
 
 #submission-basic-details {
   .icon-file {
+    @include icon-box;
     font-size: 20px;
-    padding: 10px;
-    border-radius: 6px;
-    margin-right: $margin-right-icon;
```
ktuite, line 120: Leave this margin in

ktuite (approved):

PR #1249, opened 2025-05-12
Use outlined style for more buttons on Edit Form page

matthew-white:
This PR closes getodk/central#929.

I tried it out locally.

### DIFF
--- a/src/assets/scss/app.scss
+++ b/src/assets/scss/app.scss
@@ -266,7 +266,7 @@ $btn-focus-box-shadow: 0 0 0 3px rgba(0, 0, 0, 0.18);
 .btn-danger-outlined {
   background-color: #fff;
   color: $color-danger;
-  border: 1px solid $color-danger;  
+  border: 1px solid $color-danger;
 
   &:hover, &:focus, &:active:focus {
     background-color: $color-danger;
@@ -276,7 +276,7 @@ $btn-focus-box-shadow: 0 0 0 3px rgba(0, 0, 0, 0.18);
       background-color: #fff;
       color: $color-danger;
     }
-  }  
+  }
 }
 
 a, .btn-link {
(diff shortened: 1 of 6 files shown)


src/assets/scss/app.scss
```diff
@@ -241,7 +241,7 @@ $btn-focus-box-shadow: 0 0 0 3px rgba(0, 0, 0, 0.18);
   &:hover, &:focus, &:active:focus {
     background-color: $color-danger-dark;
 
-    &[aria-disabled="true"], &.disabled, fieldset[disabled] {
+    &[aria-disabled="true"], &.disabled, fieldset[disabled] & {
```
matthew-white, line 244: This isn't really related to this PR, I just noticed that this style was incorrect and was missing an `&`. It doesn't follow the same pattern as other buttons.

src/assets/scss/app.scss
```diff
@@ -255,10 +255,30 @@ $btn-focus-box-shadow: 0 0 0 3px rgba(0, 0, 0, 0.18);
   &:hover, &:focus, &:active:focus {
     color: #fff;
     background-color: $color-action-background;
-    &[aria-disabled="true"], &.disabled, fieldset[disabled] & { background-color: #ddd; }
+
+    &[aria-disabled="true"], &.disabled, fieldset[disabled] & {
+      background-color: #fff;
+      color: $color-action-background;
+    }
```
matthew-white, line 262: I think it's nice for a disabled button not to have a hover effect. There's already reduced opacity and `cursor: not-allowed`; to indicate that the button is disabled.

sadiqkhoja (approved):
Looks good to me.

src/components/enketo/preview.vue
```diff
@@ -35,7 +35,8 @@ export default {
     formVersion: {
       type: Object,
       required: true
-    }
+    },
+    outlined: Boolean
```
sadiqkhoja, line 39: I wish there was a easier/simpler way to achieve the desire result with css-class as pass-through attribute instead of having a prop here

PR #1250, opened 2025-05-13
Correctly position FormAttachmentPopups in wide viewport

matthew-white:
Closes getodk/central#964.

One thing that was telling to me was that I could only reproduce the issue when my viewport was wide. When my window was smaller and not full-screen, I saw that the popup was at the bottom-right as expected.

I tried out this change locally. With this change, I'm no longer able to reproduce the issue.

### DIFF
--- a/src/components/form-attachment/popups.vue
+++ b/src/components/form-attachment/popups.vue
@@ -157,7 +157,6 @@ export default {
 </script>
 
 <style lang="scss">
-@use 'sass:math';
 @import '../../assets/scss/variables';
 
 $z-index-backdrop: 1;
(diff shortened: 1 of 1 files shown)


src/components/form-attachment/popups.vue
```diff
@@ -173,7 +173,7 @@ $popup-width: 300px;
 #form-attachment-popups-main {
   bottom: $edge-offset;
   position: fixed;
-  right: calc($edge-offset + max(50% - #{math.div($max-width-page-body, 2)}, 0px));
```
matthew-white, line 176: The popup has `position: fixed`, and we want to position it near the right edge of the window. That's quite easy to do. The only reason it was more complicated before is that the Attachments page wasn't full-width. I think we were using `$max-width-page-body` before so that the popup would line up with the rest of the content of the page.

lognaturel (approved):
Nice and simple!

PR #1255, opened 2025-05-14
Upgrade Web Forms for multiline text

lognaturel:
Tried it locally with [ODK Summit preparation survey 2025.xlsx](https://github.com/user-attachments/files/20199963/ODK.Summit.preparation.survey.2025.xlsx)

It adds support for the `rows` attribute on text questions. No alternatives considered.

The release has one small refactor related to labels which is extensively tested and the `rows` addition. Both seem low-risk.

No.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.1.0",
       "dependencies": {
         "@floating-ui/dom": "^1.1.0",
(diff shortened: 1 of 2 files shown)


latin-panda (approved):
I've checked previous commits in case there's anything else to do, but this seems like what we need and safe to merge.

PR #1258, opened 2025-05-15
Enable buttons to preview/fill form if Web Forms is enabled

matthew-white:
Right now, `EnketoPreview` and `EnketoFill` are disabled if the form doesn't have an `enketoId`. That's true even if `webformsEnabled` is `true`, in which case `enketoId` isn't needed. This PR changes it so that `EnketoPreview` and `EnketoFill` are never disabled when `webformsEnabled` is `true`. See getodk/central#1065.

Note that because `WebFormRenderer` uses the REST API instead of OpenRosa, we also don't need to check the form state. The buttons don't need to be disabled if the form is closed or closing.

New tests. I also tried it out locally by changing a form's state to closing. I was able to preview and fill the form using Web Forms.

### DIFF
--- a/src/components/enketo/preview.vue
+++ b/src/components/enketo/preview.vue
@@ -10,16 +10,16 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <a v-if="disabledDescription == null" :class="enketoClass" :href="enketoPath"
+  <a v-if="disabledDescription == null" class="enketo-preview btn" :class="btnClass" :href="previewPath"
     target="_blank">
     <span class="icon-eye"></span>{{ $t('action.showPreview') }}
   </a>
-  <button v-else type="button" :class="enketoClass" aria-disabled="true"
+  <button v-else type="button" class="enketo-preview btn" :class="btnClass" aria-disabled="true"
     v-tooltip.aria-describedby="disabledDescription">
     <span class="icon-eye"></span>{{ $t('action.showPreview') }}
   </button>
 
-  <router-link class="btn btn-default btn-web-form" :to="webFormsPath"
+  <router-link class="btn btn-web-form" :class="btnClass" :to="webFormsPath"
     target="_blank">
     <span class="icon-eye"></span>{{ $t('action.newPreview') }}
   </router-link>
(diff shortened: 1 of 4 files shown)


src/components/enketo/preview.vue
```diff
@@ -10,16 +10,16 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <a v-if="disabledDescription == null" :class="enketoClass" :href="enketoPath"
+  <a v-if="disabledDescription == null" class="enketo-preview btn" :class="btnClass" :href="previewPath"
```
matthew-white, line 13: I introduced `enketoClass` recently in #1249, but when I did so, I thought for some reason that this link would always go to Enketo. That's not the case: it links to the canonical preview path (either Enketo or Web Forms depending on the setting). Given that, I think it's a good idea not to use names like `enketoClass` or `enketoPath`. I want to minimize references to Enketo that could equally apply to Web Forms. (On that note, we should probably rename this component to something like `FormPreviewButton`. That's for a different PR though.)

sadiqkhoja:
> Note that because WebFormRenderer uses the REST API instead of OpenRosa, we also don't need to check the form state. The buttons don't need to be disabled if the form is closed or closing.

Should we allow new submission to be created by Data Collectors when Form is closed/closing?

sadiqkhoja:
The `edit` button on Submission details page is sending backend request hence not working.

matthew-white:
> The `edit` button on Submission details page is sending backend request hence not working.

For this PR, I was just trying to fix up the `EnketoPreview` and `EnketoFill` components. The edit button doesn't use those components, but how about I update the edit button in a follow-up PR?

matthew-white:
> Should we allow new submission to be created by Data Collectors when Form is closed/closing?

Data Collectors should not be allowed to create a new submission to a form that is closed or closing, but I don't think that needs to be enforced by the `EnketoFill` component. With Enketo, I don't think any user (whether data collector or administrator) can currently create a new submission to a form. Enketo will say "This form does not exist or you no longer have access to it", because the OpenRosa formList excludes closed forms (related: getodk/central-backend#469). I think that's why `EnketoFill` concerned itself with form state at all: not for data collectors, but to prevent other users from attempting an Enketo operation that would fail.

I think we prevent Data Collectors from creating a new submission to a form that is closed or closing in the following ways:

- The `FormRow` component doesn't render `EnketoFill` in that case. I think that's the only instance of `EnketoFill` that Data Collectors encounter.
- If the user can only `open_form.read`, `FormSubmission` checks that the form isn't closed.
- I think Backend also checks the form state when it receives a submission.

sadiqkhoja (approved):

PR #1259, opened 2025-05-16
Link Edit button in SubmissionActivity to ODK Web Forms

matthew-white:
This PR addresses https://github.com/getodk/central-frontend/pull/1258#issuecomment-2885022176. If ODK Web Forms is enabled for a form, then the Edit button in the submissions table will link to the Web Form. However, the Edit button on the submission details page still links to Backend, which redirects to Enketo. The redirects actually end up working out: as far as I can tell, you can still use the Edit button on the submission details page to edit a submission in Web Forms. However, I think it makes sense to make the two buttons consistent and to cut out the unneeded request to Backend and the subsequent redirects.

I wrote tests of a new function. I also tried it out locally.

My approach is to add the function `editSubmissionPath()` to the `useRoutes()` composable. If `webformsEnabled` is `true`, the function will send the user to ODK Web Forms, staying within Frontend. If it is `false`, it will link to Backend. It's a little unusual for a function in `useRoutes()` to link to Backend, but I think it makes sense in this case. My main goal is to centralize the logic in a single place and not duplicate it between `SubmissionMetadataRow` and `SubmissionActivity`.

I considered adding a component to do this. It would be something similar to `EnketoFill` or `EnketoPreview`. However, that felt like more than what was needed in this case. Unlike `EnketoFill` or `EnketoPreview`, we never disable the Edit button.

I also thought about always linking to `FormSubmission`, then having `FormSubmission` redirect to the Backend link as needed. However, I think that just adds complexity and another layer of redirecting.

### DIFF
--- a/test/components/submission/activity.spec.js
+++ b/test/components/submission/activity.spec.js
@@ -31,7 +31,7 @@ const mountComponent = () => {
       router: mockRouter('/projects/1/submissions/s'),
       requestData: testRequestData(
         [useSubmission, useFields],
-        { project, submission, audits, comments, diffs, fields }
+        { project, form, submission, audits, comments, diffs, fields }
       )
     }
   });
--- a/src/components/submission/activity.vue
+++ b/src/components/submission/activity.vue
@@ -58,7 +58,7 @@ import PageSection from '../page/section.vue';
 import SubmissionComment from './comment.vue';
 import SubmissionFeedEntry from './feed-entry.vue';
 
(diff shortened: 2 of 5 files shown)


test/components/submission/activity.spec.js
matthew-white, line 1: I didn't add a new test of `SubmissionActivity`. It felt enough that I added tests of the composable.

sadiqkhoja (approved):

PR #1260, opened 2025-05-16
Remove FormVersionStandardButtons

matthew-white:
At https://github.com/getodk/central/issues/896#issuecomment-2767205982, I suggested using `FormVersionStandardButtons` in `FormVersionRow`. However, I realized that that component actually isn't used anywhere else anymore. In other words, `FormVersionStandardButtons` is not so "standard" anymore. Given that it's not used elsewhere, and given that the component does so little, I think it makes sense to go ahead and remove it. I think it makes the logic in `FormVersionRow` clearer that way.

I updated a test. I also checked out the Versions page locally to make sure that it still looked right.


src/components/form-version/row.vue
```diff
@@ -66,6 +65,8 @@ defineEmits(['view-xml']);
 
     > :last-child { flex-shrink: 0; }
   }
+
+  * + .form-version-def-dropdown { margin-left: 5px; }
```
matthew-white, line 69: I also could have done:

```css
.form-version-def-dropdown:not(:first-child) { margin-left: 5px; }
```

I have a slight preference for my version because it avoids a nested selector.

matthew-white:
@ktuite, I think this PR can be reviewed async.

ktuite (approved):

PR #1263, opened 2025-05-17
Add CTA to toasts

matthew-white:
This PR makes progress on getodk/central#1011 and getodk/central#1030:

- It adds a mechanism for showing a Call to Action (CTA) in a toast. This is called for in [#1011](https://github.com/getodk/central/issues/1011).
- It adds the specific CTAs listed in [#1030](https://github.com/getodk/central/issues/1030). All of them happen to be about refreshing the page, but the CTA mechanism as a whole is generic.

The CTA is not styled/positioned correctly yet. I'll do that in a follow-up PR as I continue working on #1011. I'm making this change now so that I can close out #1030 quickly. This is what the CTA looks like now:

Image

New tests. I also tried out one of the CTAs locally.

My goal was to make it as simple as possible as possible to add a CTA. I think that's going to be a relatively common pattern. For example, I don't want us to have to write a component/HTML whenever we want to add a CTA. I wanted something generic/reusable.

We also need something that will work outside of component code. For example, src/util/session.js needs to add a CTA.

I'll also add comments explaining individual lines.

### DIFF
--- a/test/components/app.spec.js
+++ b/test/components/app.spec.js
@@ -44,7 +44,10 @@ describe('App', () => {
     describe('after a version change', () => {
       it('shows an alert', () => {
         const clock = sinon.useFakeTimers(Date.now());
-        return load('/')
+        const reload = sinon.fake();
+        return load('/', {
+          container: { location: { reload } }
+        })
           .complete()
           .request(() => {
             clock.tick(15000);
@@ -64,11 +67,13 @@ describe('App', () => {
             clock.tick(60000);
           })
           .respondWithData(() => '(v2024.1.3-sha)')
-          .afterResponse(app => {
-            clock.tick(0);
+          .afterResponse(async (app) => {
+            await clock.tickAsync(0);
             app.should.alert('info', (message) => {
               message.should.startWith('The server has been updated.');
             });
+            await app.get('.alert-cta').trigger('click');
+            reload.called.should.be.true;
           });
       });
 
--- a/test/components/async-route.spec.js
+++ b/test/components/async-route.spec.js
@@ -1,3 +1,5 @@
+import sinon from 'sinon';
+
 import AsyncRoute from '../../src/components/async-route.vue';
 import Loading from '../../src/components/loading.vue';
 import PageBody from '../../src/components/page/body.vue';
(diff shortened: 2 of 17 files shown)


test/util/util.js
```diff
@@ -2,6 +2,13 @@ import sinon from 'sinon';
 
 export const mockLogger = () => ({ log: sinon.fake(), error: sinon.fake() });
 
+export const mockLocation = (mockedProperties) => new Proxy({}, {
+  get: (_, name) =>
+    (Object.prototype.hasOwnProperty.call(mockedProperties, name)
+      ? mockedProperties[name]
+      : window.location[name])
+});
```
matthew-white, line 10: For the tests of the CTAs, I want to mock `window.location.reload()`. `window.location` doesn't make that easy for us: `sinon.replace()` isn't enough. That's why I'm using this proxy. It allows us to mock specific properties like `reload` while returning actual values from `window.location` for other properties.

test/components/async-route.spec.js
```diff
@@ -75,21 +80,33 @@ describe('AsyncRoute', () => {
...
+      const reload = sinon.fake();
+      const asyncRoute = mountComponent({
+        props: { componentName: 'MyComponent' },
+        container: {
+          location: { reload }
+        }
+      });
       await wait();
       asyncRoute.should.alert('danger');
+      asyncRoute.vm.$container.alert.ctaHandler();
```
matthew-white, line 92: Here we call `alert.ctaHandler()` directly. We can't click `.alert-cta` because the alert isn't rendered in the DOM. This test just mounts the `AsyncRoute` component, not the whole. We independently test `alert.ctaHandler()` and `.alert-cta` elsewhere, so I think it's OK to call `alert.ctaHandler()` directly here.

test/components/app.spec.js
```diff
@@ -64,11 +67,13 @@ describe('App', () => {
             clock.tick(60000);
           })
           .respondWithData(() => '(v2024.1.3-sha)')
-          .afterResponse(app => {
-            clock.tick(0);
+          .afterResponse(async (app) => {
+            await clock.tickAsync(0);
```
matthew-white, line 71: `tick()` needs to become `tickAsync()` in order to give the DOM a chance to update. We need `.alert-cta` to be rendered before we click it.

matthew-white:
@ktuite, I'm thinking it might be easiest to review this one interactively.

ktuite (approved):

Issue #1256, opened 2025-05-14
Intermittent test failure: future date


matthew-white:
I've seen this error from `fakePastDate()` a few times. I'm not sure if it's always from the same test or from a variety of tests.

```
1) should load correct page on clicking next twice
     SubmissionList pagination
     Error: future date
    at fakePastDate (dist/commons.js:224744:25)
    at factory (dist/commons.js:218689:70)
    at Store._create (dist/commons.js:216776:25)
    at Store.createPast (dist/commons.js:216766:12)
    at createSubmissions (dist/commons.js:210942:69)
    at Context.<anonymous> (dist/commons.js:211855:7)
```

matthew-white:
```
1) renders the edit button correctly
     EntityMetadataRow
     Error: future date
    at fakePastDate (dist/commons.js:224930:25)
    at factory (dist/commons.js:217324:72)
    at Store._create (dist/commons.js:216962:25)
    at Store.createPast (dist/commons.js:216952:12)
    at dist/commons.js:217420:37
    at factory (dist/commons.js:217433:7)
    at Store._create (dist/commons.js:216962:25)
    at Store.createPast (dist/commons.js:216952:12)
    at View.createPast (dist/commons.js:217049:17)
    at Context.<anonymous> (dist/commons.js:190917:68)
```

matthew-white:
This issue seems to be resolved. I was seeing it somewhat frequently before, but it hasn't shown up in the past 25 runs of CircleCI. Maybe I solved it somehow in #1265? I'll go ahead and close the issue, but we can reopen it if it shows up again.

PR #1265, opened 2025-05-18
Make improvements to fakePastDate()

matthew-white:
The main goal of this PR is to make it easier to figure out what's going on with #1256. I'm not sure where those failures are coming from, but I've updated the error message to include more information. I'm hoping that extra information will give us some insight into #1256.

### DIFF
--- a/test/util/date-time.js
+++ b/test/util/date-time.js
@@ -40,12 +40,16 @@ afterEach(() => {
 // Returns an ISO string for a date in the past that is no earlier than the
 // specified dates.
 export const fakePastDate = (dateStrings) => {
-  const dateTimes = dateStrings
+  const parsed = dateStrings
     .filter(s => s != null)
-    .map(s => DateTime.fromISO(s));
-  if (dateTimes.length === 0) return faker.date.past().toISOString();
-  const from = DateTime.max(...dateTimes);
-  const now = DateTime.local();
-  if (from > now) throw new Error('future date');
-  return faker.date.between({ from: from.toISO(), to: now.toISO() }).toISOString();
+    .map(s => Date.parse(s));
+  if (parsed.length === 0) return faker.date.past().toISOString();
+  const from = Math.max(...parsed);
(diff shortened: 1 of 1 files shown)


test/util/date-time.js
```diff
@@ -40,12 +40,16 @@ afterEach(() => {
...
-    .map(s => DateTime.fromISO(s));
-  if (dateTimes.length === 0) return faker.date.past().toISOString();
-  const from = DateTime.max(...dateTimes);
-  const now = DateTime.local();
-  if (from > now) throw new Error('future date');
-  return faker.date.between({ from: from.toISO(), to: now.toISO() }).toISOString();
+    .map(s => Date.parse(s));
+  if (parsed.length === 0) return faker.date.past().toISOString();
+  const from = Math.max(...parsed);
+  const to = Date.now() - 1000;
```
matthew-white, line 48: `- 1000` is a change from the current behavior, since `now` was used before, and that went up to the current millisecond. I noticed that `faker.date.past()` does `- 1000`, and I thought it made sense to try to match it: https://github.com/faker-js/faker/blob/549d71cf3361cca787d685a1c053ef8135e13641/src/modules/date/index.ts#L88

test/util/date-time.js
```diff
@@ -40,12 +40,16 @@ afterEach(() => {
 // Returns an ISO string for a date in the past that is no earlier than the
 // specified dates.
 export const fakePastDate = (dateStrings) => {
-  const dateTimes = dateStrings
+  const parsed = dateStrings
     .filter(s => s != null)
-    .map(s => DateTime.fromISO(s));
```
matthew-white, line 45: I like `DateTime`/Luxon in general, but I've removed it from this function. `DateTime` isn't needed: `Date.parse()` is enough here. I sort of think that `DateTime` just added an extra layer rather than simplifying the code.

sadiqkhoja (approved):

PR #1270, opened 2025-05-20
Fix: modals without actions should have bottom padding

sadiqkhoja:
image

### DIFF
--- a/test/components/modal.spec.js
+++ b/test/components/modal.spec.js
@@ -71,7 +71,7 @@ describe('Modal', () => {
         },
         attachTo: document.body
       });
-      modal.emitted().resize.should.eql([[10]]);
+      modal.emitted().resize.should.eql([[22]]);
     });
   });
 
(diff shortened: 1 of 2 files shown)


test/components/modal.spec.js
matthew-white, line 1: These changes look good, but FYI Kathleen recently skipped these tests. Related: getodk/central#1050

matthew-white (approved):

test/components/modal.spec.js
ktuite: I skipped and then unskipped them when I realized I could reset them to the old values. I still don't know what's going on with them, though. E.g. why are the new numbers in this PR the magic numbers that make the tests pass?

ktuite (approved):

test/components/modal.spec.js
matthew-white: It has to do with the height of the `.modal-body`, which includes padding.

PR #1271, opened 2025-05-20
Remove unused components

matthew-white:
This PR removes two components that are no longer used anywhere: `FloatRow` and `ChecklistStep`. These are essentially orphaned code now.

- `FloatRow`. I had wanted to refactor this away anyway because we tend to use flexbox now instead of floats. But its last use was removed, so we can just go ahead and remove it without needing to refactor anything else.
- `ChecklistStep`. Its uses were removed in v2025.1 as part of [getodk/central#865](https://github.com/getodk/central/issues/865) and [getodk/central#728](https://github.com/getodk/central/issues/728).

If we ever need these components again for some reason, we can add them back. I doubt we will need them again though.


matthew-white:
@ktuite, I think this one can be reviewed async.

ktuite (approved):

PR #1273, opened 2025-05-28
fix: handle all file types for attachments

latin-panda:
When working on [Select with images](https://github.com/getodk/web-forms/pull/417) I noticed that PNG images weren't rendering. 

The All Types form renders images and GeoJSON. Please suggest any other test I should make

Screenshot 2025-05-28 at 3 23 05 PM

This fix treats attachments as a blob, since it can be any file (audio, video, image, text, etc.)

None

No

### DIFF
--- a/src/components/web-form-renderer.vue
+++ b/src/components/web-form-renderer.vue
@@ -151,7 +151,7 @@ const transformAttachmentResponse = (axiosResponse) => {
   }
 
   let body;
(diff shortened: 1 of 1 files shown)


latin-panda:
@sadiqkhoja can you please look at this PR? It's tiny

sadiqkhoja (approved):

PR #1274, opened 2025-05-29
Bind vite dev server to 127.0.0.1

yoursdearboy:
This PR introduces extra config option to make Vite dev server accessible using a devcontainer and VS Code port forwarding feature. It just changes server address from `localhost` to `127.0.0.1`, so shouldn't break anything.

For more info see [Vite docs](https://github.com/sapphi-red/vite/blob/main/docs/guide/troubleshooting.md#dev-containers--vs-code-port-forwarding).

Run dev server using `npm run dev` in VS Code and opened it through VS Code port forwarding.

It's a well-known workaround.

Doesn't affect users.

No.


sadiqkhoja:
I also noticed the same issue while running e2e tests. I have been adding `host: true` but I think specifying `127.0.0.1` is better approach.

sadiqkhoja (approved):

PR #1280, opened 2025-06-04
Test translations

alxndrsn:
I had this lying around; might be of interest.

Current checks:

* missing translations (present in `en`, but missing in another language)
* extra translations (present in another language, but missing in `en`)
* missing variable (present in `en`, but missing in another language)
* extra variable (present in another language, but missing in `en`)

Probably transifex already does a similar report?

### DIFF
--- a/.github/workflows/test-translations.yml
+++ b/.github/workflows/test-translations.yml
@@ -1,0 +1,17 @@
+name: Translations
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  test-translations:
+    timeout-minutes: 2
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
+      with:
+        node-version: 22.12.0
+        cache: 'npm'
+    - ./bin/test-translations.js
--- a/.github/workflows/test-translations.yml
+++ b/.github/workflows/test-translations.yml
@@ -14,4 +14,4 @@ jobs:
       with:
         node-version: 22.12.0
         cache: 'npm'
-    - ./bin/test-translations.js
+    - run: ./bin/test-translations.js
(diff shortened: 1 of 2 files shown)


matthew-white:
Thanks, @alxndrsn! We do most of these checks already, but let me add more detail. If we were going to do additional checks, I'd probably want to add them to bin/util/transifex.js, where we do the rest of our translation validation. I'd be happy to meet to discuss any of this if you're interested in working in this area.

> missing translations (present in `en`, but missing in another language)

This is a normal situation for better or for worse. Some languages get translated more quickly than others and are more maintained in general. For each language, Transifex tells us how many translations are missing / what percentage of strings are translated.

> extra translations (present in another language, but missing in `en`)

This can happen in the normal course of things, up until we pull translations from Transifex. We do that before regression testing, then before release. For example, if I remove a source message in a PR, the corresponding translation will remain in the codebase until destructure.js is run. There are more details about that workflow in CONTRIBUTING.md if you're interested.

That said, when we do pull translations from Transifex before regression testing and before release, we might be assuming that the translation files have the same JSON structure as the source JSON (strings_en.json). I think #858 tells us that bin/util/transifex.js could do a better job of showing a nice error message when the structure differs. Transifex should ensure that the structure matches, but I think it'd be nice for bin/util/transifex.js to check that as well. I think that'll be done as part of #858.

<blockquote>

* missing variable (present in `en`, but missing in another language)
* extra variable (present in another language, but missing in `en`)
</blockquote>

Transifex definitely checks for missing variables. I don't remember whether it checks for extra variables. Transifex has a concept of [translation checks](https://help.transifex.com/en/articles/6241794-setting-translation-checks), which we definitely make use of.

We also check this in bin/util/transifex.js:

https://github.com/getodk/central-frontend/blob/7ca500f1f01e128408b86e65e281fad7efe35707/bin/util/transifex.js#L971-L973

PR #1284, opened 2025-06-05
Added search-textbox component

sadiqkhoja:
Closes getodk/central#1023

Added tests.

Uses existing pattern.

No

No

### DIFF
--- a/src/components/entity/list.vue
+++ b/src/components/entity/list.vue
@@ -16,7 +16,7 @@ except according to the terms contained in the LICENSE file.
         <div class="form-group">
           <span class="icon-filter"></span>
         </div>
-        <search-textbox v-model="searchTerm" :label="$t('common.search')" :no-label="true" :disabled="deleted"/>
+        <search-textbox v-model="searchTerm" :label="$t('common.search')" :hide-label="true" :disabled="deleted"/>
         <entity-filters v-model:conflict="conflict" :disabled="deleted"
         :disabled-message="deleted ? $t('filterDisabledMessage') : null"/>
       </form>
@@ -138,7 +138,6 @@ export default {
       })
     });
 
-    // works in conjunction with searchTextbox
     const searchTerm = useQueryRef({
       fromQuery: (query) => (typeof query.search === 'string' ? query.search : null),
       toQuery: (value) => ({
@@ -177,8 +176,6 @@ export default {
       pagination: { page: 0, size: this.pageSizeOptions[0], count: 0 },
       now: new Date().toISOString(),
       snapshotFilter: '',
-
-      searchTextbox: this.searchTerm ?? ''
     };
   },
   computed: {
--- a/src/components/entity/list.vue
+++ b/src/components/entity/list.vue
@@ -16,14 +16,7 @@ except according to the terms contained in the LICENSE file.
         <div class="form-group">
           <span class="icon-filter"></span>
         </div>
-        <div class="form-group">
-          <input v-model="searchTextbox" class="form-control search-textbox" :placeholder="$t('common.search')"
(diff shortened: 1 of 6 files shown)


src/components/search-textbox.vue
```diff
@@ -0,0 +1,58 @@
...
+
+
+const props = defineProps({
+  modelValue: String,
+  disabled: Boolean,
+  /**
+   * Text for label and placeholder
+   */
+  label: String,
+  noLabel: Boolean
```
matthew-white, line 36: I see that the `SearchTextbox` in `EntityList` specifies props for both `label` and `noLabel`. That makes me wonder about the following:

1. Is the `label` prop always expected? I think so, because it will used in the `placeholder` and `aria-label`. If that's right, how about specifying `required: true` for the prop?
2. If the `label` prop is always expected, I think the name `noLabel` is slightly confusing. What do you think about a name like `hideLabel` or `hiddenLabel`? It seems like that prop has to do with whether a `.form-label` element is rendered, not whether there's a label at all.

src/components/project/user/list.vue
```diff
@@ -152,6 +142,15 @@ export default {
         : this.$t('common.noResults');
     }
   },
+  watch: {
+    searchTerm(searchTerm) {
+      if (!searchTerm) {
+        this.clearSearch();
+      } else {
+        this.search(true);
```
matthew-white, line 150: I don't think `this.search()` takes an argument.

```suggestion
        this.search();
```

src/components/project/user/list.vue
matthew-white, line 1: Are the following styles still relevant? Should they be removed? If they are useful, maybe they could be moved into the `SearchTextbox` component?

https://github.com/getodk/central-frontend/blob/4ef4186b5517e70e9ad04d6809ab8cb8aa92b0a5/src/components/project/user/list.vue#L231-L235

src/components/entity/list.vue
```diff
@@ -16,14 +16,7 @@ except according to the terms contained in the LICENSE file.
         <div class="form-group">
           <span class="icon-filter"></span>
         </div>
-        <div class="form-group">
-          <input v-model="searchTextbox" class="form-control search-textbox" :placeholder="$t('common.search')"
```
matthew-white, line 20: Are we no longer using the `searchTextbox` data property? If so, can it be removed from `data()`?

test/components/project/user/list.spec.js
matthew-white, line 1: Are there any tests in this file related to the search bar that you think we could get rid of now? If there's functionality that's already covered by the tests of `SearchTextbox`, I feel like it'd be nice not to also test it here.

matthew-white (approved):

src/components/search-textbox.vue
```diff
@@ -0,0 +1,58 @@
...
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+
+<template>
+  <label class="form-group">
+    <input v-model="searchTextbox" class="form-control search-textbox" :placeholder="label"
+      :aria-label="label" :aria-disabled="disabled" autocomplete="off" @keydown.enter="setSearchTerm" @focusout="revert">
+    <button v-show="searchTextbox" type="button" class="close"
+      :aria-label="$t('action.clearSearch')" @click="clearSearch">
```
matthew-white, line 18: I'm realizing that `action.clearSearch` doesn't exist in en.json5. It's in `ProjectUserList`. How about removing it from `ProjectUserList` and adding it to this component? You can use `@transifexKey` to prevent it from changing in Transifex and being retranslated.

PR #1286, opened 2025-06-06
Remove loading component from web-form-renderer and enketo-iframe components

sadiqkhoja:
Closes getodk/central#1120

The loading message changes size and font because of WF css bleeding, which might give an impression of application glitch or slow. That can be fixed by adding following style in loading component but that doesn't seem best solution to the problem. Ideally WF should not modify the styles of host application (Central):

```
  font-size: 14px;
  font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
  line-height: 1.42857143;
```

https://github.com/user-attachments/assets/b7825497-2f9d-4e81-846d-cf995ee148db

### DIFF
--- a/src/components/form/preview.vue
+++ b/src/components/form/preview.vue
@@ -22,7 +22,7 @@ except according to the terms contained in the LICENSE file.
 </template>
 
 <script setup>
-import { defineAsyncComponent, ref } from 'vue';
+import { defineAsyncComponent, ref, watch } from 'vue';
 import { useRequestData } from '../../request-data';
 import { apiPaths } from '../../util/request';
 import { noop } from '../../util/util';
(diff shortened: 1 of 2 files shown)


src/components/form/preview.vue
matthew-white, line 1: How about the case where the request for `form` results in an error? Then `WebFormRenderer` and `EnketoIframe` won't ever be rendered and won't ever emit `loaded`. In that case, even though there was an error and nothing more will happen, the page will show the loading message forever. Maybe it could be something like:

```js
watch(() => form.initiallyLoading, (value) => {
  if (!value) loadingState.value = form.dataExists;
});
```

matthew-white: I think a variation on this comment also applies to `FormSubmission`.

matthew-white (approved):

PR #1290, opened 2025-06-09
Add ability to toggle ownerOnly

matthew-white:
This PR implements the Frontend part of getodk/central#952.

I tried it out locally. I also wrote new tests.

Mostly I tried to follow existing patterns. There were a couple of places where I did things differently; I'll add comments about those.

The change is mostly additive. There's a new component. Given that, I think the risk of regression is low.

I've updated the issue in the docs repo for v2025.2.

### DIFF
--- a/src/components/dataset/owner-only.vue
+++ b/src/components/dataset/owner-only.vue
@@ -62,7 +62,8 @@ except according to the terms contained in the LICENSE file.
 </template>
 
 <script setup>
-import { ref, watch } from 'vue';
+import { inject, ref, watch } from 'vue';
+import { useI18n } from 'vue-i18n';
 
 import Modal from '../modal.vue';
 import Spinner from '../spinner.vue';
@@ -75,6 +76,9 @@ import { useRequestData } from '../../request-data';
 defineOptions({
   name: 'DatasetOwnerOnly'
 });
+
+const { t } = useI18n();
+const { alert } = inject('container');
 
 // The component assumes that this data will exist when the component is
 // created.
@@ -101,7 +105,12 @@ const confirm = () => {
     data: { ownerOnly: ownerOnly.value },
     patch: ({ data }) => { dataset.ownerOnly = data.ownerOnly; }
   })
-    .then(() => { confirmationModal.hide(); })
+    .then(() => {
+      confirmationModal.hide();
+      alert.success(dataset.ownerOnly
+        ? t('alert.changeToTrue')
+        : t('alert.changeToFalse'));
+    })
     .catch(noop);
 };
 </script>
@@ -128,6 +137,10 @@ const confirm = () => {
       "action": {
         "confirm": "Access own Entities"
       }
+    },
+    "alert": {
+      "changeToFalse": "App Users and Data Collectors will now have access to all Entities.",
+      "changeToTrue": "App Users and Data Collectors will now only have access to Entities they create."
     }
   }
 }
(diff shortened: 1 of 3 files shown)


src/components/dataset/owner-only.vue
```diff
@@ -0,0 +1,134 @@
...
+            <strong>{{ $t('ownerOnly') }}</strong>
+          </label>
+          <p id="dataset-owner-only-true-help" class="help-block">
+            {{ $t('radio.true') }}
+          </p>
+        </div>
+      </form>
+    </div>
+  </div>
+  <modal v-bind="confirmationModal" :hideable="!awaitingResponse" backdrop
```
matthew-white, line 42: We don't usually group the `<modal>` together with the HTML that opens the modal: we usually separate them out. However, I think it makes sense to group them together because it's all for the same feature. Before we started teleporting modals, I thought it was important to be careful about where in the component tree the modal was rendered. However, now the `.modal` element won't actually be rendered here and will instead be teleported to `#modals`. Given that, we have the freedom to render the `<modal>` _component_ here. I think doing so makes the overall logic/approach clearer. It also simplifies testing.

To compare to what we do for `approvalRequired`, there the `.panel` appears in `DatasetSettings`, while the modal is its own component (`DatasetPendingSubmissions`). That's more typical of what we've done in the past, but I think it has the downside of spreading the logic over multiple components. It also makes `DatasetSettings` more complicated. My vision for this new pattern is that a parent component like `DatasetSettings` can be very simple and can just stitch together multiple components like `DatasetOwnerOnly`.

src/components/dataset/owner-only.vue
```diff
@@ -0,0 +1,134 @@
...
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+<template>
+  <div class="panel panel-simple">
+    <div class="panel-heading">
+      <h1 class="panel-title">{{ $t('panel.title') }}</h1>
+    </div>
+    <div class="panel-body">
+      <form @change="confirmationModal.show()" @submit.prevent>
```
matthew-white, line 18: I could have put `@change` on each of the individual `<input>` elements, but I thought leveraging event bubbling might be nice.

src/components/form/web-forms-settings-confirmation.vue
```diff
@@ -53,7 +53,8 @@ except according to the terms contained in the LICENSE file.
         {{ $t('enketoConfirmation.description') }}
       </p>
       <div class="modal-actions">
-        <button class="btn btn-link" type="button" @click="$emit('hide')">
+        <button class="btn btn-link" type="button"
+          :aria-disabled="form.awaitingResponse" @click="$emit('hide')">
```
matthew-white, line 57: This change isn't really related to the rest of this PR. I just noticed while I was looking at prior art that this button didn't have an `aria-disabled` attribute.

test/components/dataset/owner-only.spec.js
matthew-white, line 1: Many tests like this have to use `load()` instead of `mount()` or `mockHttp()` because they have to get the route component (e.g., `DatasetSettings`) involved. However, because `DatasetOwnerOnly` is self-contained and includes the modal, the tests here can mount less of the app.

matthew-white (commented):
Notes for code review

matthew-white:
FYI @sadiqkhoja, I just pushed a new commit to show a toast message after a successful change to the flag. Related: https://github.com/getodk/central/issues/952#issuecomment-2956487372

PR #1293, opened 2025-06-09
Update UI for the filters

sadiqkhoja:
Closes getodk/central#1008

Updated Tests and manual verification.
Showed the UI changes to Nicole.

See in line comments where I have used some tricks to achieve desire result.

No

No


src/components/multiselect.vue
```diff
@@ -377,6 +378,12 @@ const toggleAfterEnter = ({ key }) => {
   if (key === 'Enter') $toggle.value.dropdown('toggle');
 };
 
+const apply = () => {
+  searchValue.value = '';
+  emitIfChanged();
+  $dropdown.value.removeClass('open');
```
sadiqkhoja, line 384: is there any other way to do this?

src/components/multiselect.vue
```diff
@@ -365,8 +367,7 @@ const $dropdown = computed(() => $(dropdown.value));
 onMounted(() => {
   $dropdown.value.on('shown.bs.dropdown', syncWithModelValue);
   $dropdown.value.on('hidden.bs.dropdown', () => {
```
matthew-white, line 369: Currently, it doesn't look like the `changes` set is cleared in this case. Before, it wasn't possible to hide the dropdown without emitting changes, so it was the sole responsibility of `emitIfChanged` to call `changes.clear()`. But now we need to clear `changes` in this case as well.

One question: should `changes.clear()` be called here or in the `syncWithModelValue()` function? Maybe it'd be better to call it here so that the values in `changes` aren't held onto unnecessarily.

src/components/multiselect.vue
```diff
@@ -365,8 +367,7 @@ const $dropdown = computed(() => $(dropdown.value));
...
-    searchValue.value = '';
-    emitIfChanged();
+    needsSync = true;
```
matthew-white, line 370:
```suggestion
    needsSync = changes.size !== 0;
    changes.clear();
```

src/components/multiselect.vue
matthew-white: I think we can call the [`.dropdown()`](https://getbootstrap.com/docs/3.4/javascript/#dropdowns-methods) method from the Bootstrap plugin:

```suggestion
  $dropdown.value.dropdown('toggle');
```

src/components/multiselect.vue
```diff
@@ -559,7 +579,8 @@ const emptyMessage = computed(() => (searchValue.value === ''
       // more selections. {all} has the text "All", and {none} has the text
       // "None". {all} and {none} will be translated separately based on what is
       // being selected.
-      "select": "Select {all} / {none}"
+      "select": "Select {all} / {none}",
```
matthew-white, line 582: This message can be removed now because it's not being shown anywhere.

matthew-white (commented):
Notes from interactive review

matthew-white (approved):

PR #1294, opened 2025-06-10
Handle null actee in upgrade.server action and add string

ktuite:
Frontend part of issue https://github.com/getodk/central/issues/971 and PR https://github.com/getodk/central-backend/pull/1511

There was an error in frontend while trying to display the `actee` for this event (which is null). The `upgrade` category gets treated as expecting a `form` actee (because most other upgrade events are about upgrading forms) and there's even a little comment about it.

```
// Presumably at some point, the actee of an upgrade audit might not be a form,
// at which point we will have to update this component (perhaps we would use
// the full action or a prefix instead of the category).
acteeSpeciesByCategory.upgrade = acteeSpeciesByCategory.form;
```

I didn't change this species handling because I wasn't sure how to untangle the `upgrade` category. I just added some code down below that switches to a different species if the actee is null.

Fixed:
Screenshot 2025-06-10 at 4 10 54 PM

### DIFF
--- a/test/components/audit/table.spec.js
+++ b/test/components/audit/table.spec.js
@@ -357,7 +357,7 @@ describe('AuditTable', () => {
     await testTarget(row, '');
   });
 
-  describe('system operation', () => {
+  describe('system logging without a target', () => {
     it('renders an analytics audit correctly', async () => {
       testData.extendedAudits.createPast(1, { action: 'analytics' });
       const row = mountComponent();


matthew-white (approved):
Thank you for looking into that! Sorry, I didn't realize that that would be an issue: I thought `AuditRow` already accounted for `null` actees. I made a slight change to the approach and pushed a commit along those lines. I didn't try it out locally, but I did add a test. Would you mind taking a look? If it looks good, I think we can go ahead and merge.

PR #1300, opened 2025-06-16
fix: adds response type when fetching edit instance's attachments

latin-panda:
Closes https://github.com/getodk/central-frontend/issues/1301

Test video: creates and edits submission of image question type 
 

https://github.com/user-attachments/assets/bd3c872c-72bb-4f5d-84e3-79570f50b6fc

Loads GeoJSON and select with images

Screenshot 2025-06-16 at 10 43 26 AM

### DIFF
--- a/src/components/web-form-renderer.vue
+++ b/src/components/web-form-renderer.vue
@@ -191,13 +191,7 @@ const fetchSubmissionXml = () => {
 
 const fetchSubmissionAttachments = () => {
   const requestUrl = apiPaths.submissionAttachments(form.projectId, form.xmlFormId, props.instanceId);
-  return submissionAttachments.request({
-    url: requestUrl,
-    alert: false,
-    responseType: 'blob', // Handle all file types for attachments.
-  })
-    .then(transformAttachmentResponse)
-    .catch(noop);
+  return submissionAttachments.request({ url: requestUrl, alert: false }).catch(noop);
 };
 
 const fetchSubmissionAttachment = (attachmentName) => {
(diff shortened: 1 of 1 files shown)


src/components/web-form-renderer.vue
```diff
@@ -193,7 +193,8 @@ const fetchSubmissionAttachments = () => {
   const requestUrl = apiPaths.submissionAttachments(form.projectId, form.xmlFormId, props.instanceId);
   return submissionAttachments.request({
     url: requestUrl,
-    alert: false
+    alert: false,
+    responseType: 'blob', // Handle all file types for attachments.
```
matthew-white, line 197: Do we need `responseType: 'blob'` here in `fetchSubmissionAttachments()`, or just in `fetchSubmissionAttachment()` below? The response here is an array, and I'm pretty sure it's just metadata. I don't think it contains any binary attachment content.

src/components/web-form-renderer.vue
```diff
@@ -193,7 +193,8 @@ const fetchSubmissionAttachments = () => {
...
   })
     .then(transformAttachmentResponse)
```
matthew-white, line 199: Maybe this line is part of what's confusing. I don't think it's actually an attachment response. The response isn't passed directly to Web Forms. I don't see anything using the return value from `fetchSubmissionAttachments()`, so I think this line could probably go away.

matthew-white (approved):

src/components/web-form-renderer.vue
matthew-white: It looks like it's used in the `editInstanceOptions` computed ref.

matthew-white: https://github.com/getodk/central-frontend/blob/6e6ab26ba34a96da9743949d9ec8a5249bda5005/src/components/web-form-renderer.vue#L213-L224

matthew-white: Central creates a submission attachment record for each expected attachment (for each attachment "slot"). That's true even if the attachment isn't uploaded or fails to upload. I guess here, the `editInstance` prop that's passed to Web Forms only includes submission attachments that have been successfully uploaded.

PR #1307, opened 2025-06-17
ci: build

alxndrsn:
Closes #

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,17 @@
+name: Build
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
+      with:
+        node-version: 22.12.0
+        cache: 'npm'
+    - run: npm run build


matthew-white (commented):
I like the approach here! It sounds good to me to add a separate Node script and to run a Bash command.

I think the main change I would make is to the file size metrics that the script is looking at. It looks like the script sums the file sizes, then checks that the sum is within expected bounds. However, it's normal for the sum to increase over time as we add more pages to the app. The main problem is if the part of the app that's loaded upfront (the initial bundle / the components that aren't loaded async) is too large. In other words, no single file should exceed some maximum file size. I don't feel like the sum matters as much.

We're interested in both .js and .css files. I think the initial bundle corresponds to one .js file and one .css file. I think we could ignore images if they're a problem for the script: I'm mainly just wanting to check that the initial bundle doesn't accumulate in size without us noticing.

matthew-white:
> Build timing is recorded by github by default. Keeping this build as a separate workflow should allow simple comparison over time

I think I'd also be interested in an automated check that the build doesn't take much longer than expected. How about we just add a timeout to the new GitHub action in this PR? Something that will alert us if the duration of `npm run build` or the duration of the action as a whole exceeds some maximum time.

matthew-white:
> How about we just add a timeout to the new GitHub action in this PR?

Oh wait, the action has one already. 😅 Sorry, I thought I checked that! Well, that's all I think we need. :+1:

alxndrsn:
> The main problem is if the part of the app that's loaded upfront (the initial bundle / the components that aren't loaded async) is too large. In other words, no single file should exceed some maximum file size. I don't feel like the sum matters as much.

I've updated the PR to remove bundle size checking - it would be great to achieve the initial step of actually checking the repo builds as part of CI.  Details/approaches to checking bundle size can then be discussed at greater length independently.

matthew-white (approved):

PR #1309, opened 2025-07-08
Fixes download filtered submissions: don't reset state on modal hide

sadiqkhoja:
Closes getodk/central#1168

This was kinda hard bug to fix. 

**What's happening:** When you hover on the hyperlink, you see the correct link but when you click on it, another function runs and hides the modal resulting in resetting of modal state/data, so when default event handler of browser is run it sees no odatafilter query parameter.

### DIFF
--- a/src/components/submission/list.vue
+++ b/src/components/submission/list.vue
@@ -78,7 +78,7 @@ except according to the terms contained in the LICENSE file.
     </div>
 
     <submission-download v-bind="downloadModal" :form-version="formVersion"
-      @hide="downloadModal.hide()"/>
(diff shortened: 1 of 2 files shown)


matthew-white:
When the `SubmissionDownload` modal is hidden, it doesn't reset the form fields in the modal other than the passphrase. But you're totally right that it ends up resetting the `odataFilter` prop that's passed to `SubmissionDownload`. In addition to the default event handler, that would be an issue for the "Try again" CTA in the toast. The CTA needs the `odataFilter` prop to continue to be set.

PR #1313, opened 2025-07-29
Fixes: pass query parameters as it is to the enketo iframe

sadiqkhoja:
Closes https://forum.getodk.org/t/when-prefilling-the-value-in-enketo-web-form-the-space-is-getting-replaced-with-a/55937 

Added tests and did manual verification

This solution passes provided query parameters as it is to the iframe, which will be no surprise for the user.

Previously I considered just replacing '+' with '%20' in the output of the `queryString()` but that didn't handle the case where user actually wants to pass `+` sign to the Enketo Form because `route.query` returns space for both space and `+`.

Should be tested with various special characters in prefill data to ensure no other encoding issues, compare the behavior of various special and multi-lingual characters with opening iframe in a new tab.

No

### DIFF
--- a/e2e-tests/tests/enketo.spec.js
+++ b/e2e-tests/tests/enketo.spec.js
@@ -250,4 +250,21 @@ test.describe('Enketo', () => {
 
     await expect(frame.getByRole('heading', { name: 'Successful' })).toBeVisible();
   });
+
+  test('default value is consistent between rendering enketo directly or via iframe', async ({ page }) => {
+    await login(page);
+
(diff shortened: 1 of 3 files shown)


lognaturel:
This looks good and safe to me. The logic to split up and restitch query params together makes sense to me. I think my review is sufficient unless you want to take a deeper look, @matthew-white 

Do you think we need QA? It feels isolated enough and easily verifiable enough to me that test + our sanity checks are sufficient.

matthew-white (approved):

PR #1316, opened 2025-08-13
Bulk delete Entities

sadiqkhoja:
TBA


src/components/entity/list.vue
```diff
@@ -458,6 +493,109 @@ export default {
...
       if (this.odataEntities.count < this.pageSizeOptions[0]) return;
       this.fetchChunk(false);
+    },
+    clearSelectedEntities() {
+      this.odataEntities.value.forEach(e => { e.__system.selected = false; });
+      this.selectedEntities.clear();
+    },
+    requestBulkDelete() {
+      const uuids = Array.from(this.selectedEntities);
+      // TODO: disable the whole table
```
sadiqkhoja, line 503: coming in the subsequent PR.

src/components/entity/list.vue
```diff
@@ -458,6 +498,105 @@ export default {
...
+      const onSuccess = () => {
+        if (this.deletedEntityCount.dataExists) this.deletedEntityCount.value += uuids.length;
+
+        uuids.forEach(uuid => this.odataEntities.removedEntities.add(uuid));
+
+        this.deletedEntities = [];
+        this.deletedEntities.push(...this.odataEntities.value.filter(e => uuids.includes(e.__id)));
+        this.odataEntities.value = this.odataEntities.value.filter(e => !uuids.includes(e.__id));
+        this.alert.success(this.$tcn('alert.bulkDelete', this.selectedEntities.size))
+          .cta(this.$t('undo'), () => this.requestBulkRestore(uuids));
```
matthew-white, line 528: I'm slightly concerned about the user doing something in the table (like changing a filter) right after the bulk delete, and only _then_ clicking Undo. I'm not as concerned about it as some of the other cases though, since the toast will automatically disappear along with the Undo button after a short amount of time. The only problem cases I can think of are:

1. Change a filter, then click Undo. The deleted entities will reappear even if the new filter doesn't apply to them.
2. View deleted entities, then click Undo. This one seems a little more likely (e.g., someone has second thoughts after viewing the trash). But again, it'd have to happen within 7 seconds.

Do you think we should make any changes in this area? Maybe this one is too much of an edge case?

src/components/entity/list.vue
```diff
@@ -458,6 +498,105 @@ export default {
...
+        }
+      });
+
+      const onSuccess = () => {
+        if (this.deletedEntityCount.dataExists) this.deletedEntityCount.value += uuids.length;
+
+        uuids.forEach(uuid => this.odataEntities.removedEntities.add(uuid));
+
+        this.deletedEntities = [];
+        this.deletedEntities.push(...this.odataEntities.value.filter(e => uuids.includes(e.__id)));
```
matthew-white, line 525: Instead of:

```js
this.deletedEntities = [];
this.deletedEntities.push(...anotherArray);
```

I think we can just do:

```js
this.deletedEntities = anotherArray;
```

src/components/entity/table.vue
```diff
@@ -87,6 +93,31 @@ const table = ref(null);
...
+  odataEntities.value.forEach(e => {
+    e.__system.selected = checked;
+  });
+  emit('selectionChanged', 'all', checked);
+};
+
+// When all rows are deleted, we want allSelected to be false
+watch(() => odataEntities.value?.length, (count) => {
+  if (count === 0) allSelected.value = false;
+});
```
matthew-white, line 119: How about adding a test of this case?

matthew-white (commented):
OK, I've reviewed everything now! It's looking great. The main thing I want to figure out before merging is what to do about concurrent actions like selecting rows during a slow background refresh. My feeling is that we could/should solve for those things in a follow-up PR. There's enough going on in this PR already. But I don't think we can close the issue and send it to QA until we solve those things, because the QA team will probably try concurrent actions.

src/components/entity/list.vue
```diff
@@ -60,6 +61,17 @@ except according to the terms contained in the LICENSE file.
...
       :awaiting-response="restoreModal.state && awaitingResponses.has(restoreModal.entity.__id)"
       @hide="restoreModal.hide()" @restore="requestRestore"/>
+    <action-bar :state="actionBarState"
+      :message="$tcn('actionBar.message', selectedEntities.size)"
+      @hide="clearSelectedEntities()">
+      <template #actions>
+        <button class="btn btn-primary" type="button" :aria-disabled="awaitingResponses.size > 0" @click="requestBulkDelete">
+            <span class="icon-trash"></span>
+            {{ $t('actionBar.cta') }}
+            <spinner :state="awaitingResponses.size > 0"/>
```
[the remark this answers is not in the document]
matthew-white: `awaitingResponses` is also used for `aria-disabled`. I think it's the same issue and that the button won't actually ever be disabled.

src/components/entity/list.vue
sadiqkhoja: it makes me nervous to assign one array to another because if I modify `anotherArray` then `this.deletedEntities` will be modified.

src/components/entity/list.vue
```diff
@@ -458,6 +498,105 @@ export default {
...
+        .then(onSuccess)
+        .catch((error) => {
+          const { cta } = this.alert.danger(requestAlertMessage(this.container.i18n, error));
+          cta(this.$t('action.tryAgain'), () => {
+            bulkDelete()
+              .then(() => {
+                this.alert.last.hide();
+                onSuccess();
+              })
+              .catch(noop);
```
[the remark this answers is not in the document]
sadiqkhoja: I don't think we need to do anything extra in case "try again" fails because alert with "try again" button is still on the screen and user can press it again.

src/components/entity/list.vue
```diff
@@ -458,6 +498,105 @@ export default {
...
+        this.odataEntities.value = this.odataEntities.value.filter(e => !uuids.includes(e.__id));
+        this.alert.success(this.$tcn('alert.bulkDelete', this.selectedEntities.size))
+          .cta(this.$t('undo'), () => this.requestBulkRestore(uuids));
+        this.selectedEntities.clear();
+      };
+
+      bulkDelete()
+        .then(onSuccess)
+        .catch((error) => {
+          const { cta } = this.alert.danger(requestAlertMessage(this.container.i18n, error));
```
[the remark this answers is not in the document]
sadiqkhoja: changed. But we still need to inject `container` to watch on openModal state

matthew-white (approved):
Looks great! There's a lot of functionality here, including row selection, bulk deletion, retry, and undo, yet the data flow and code is very clear. 💯

src/components/entity/feed-entry.vue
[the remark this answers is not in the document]
sadiqkhoja: https://github.com/getodk/central/issues/1358

PR #1319, opened 2025-08-19
Fix regex and log list of big files in check-bundle-size.js

sadiqkhoja:


bin/check-bundle-size.js
```diff
@@ -52,7 +53,7 @@ log('File sizes look OK.');
 
 function isTooBig({ path, size }) {
   // Special cases:
-  const simpleName = basename(path).replace(/-\w{8}\./, '.');
+  const simpleName = basename(path).replace(/-.{8}\./, '.');
```
sadiqkhoja, line 56: \w doesn't include `-` which can be part of the generated file. See `[check-bundle-size]  444452 dist/assets/index-DMq9Y-6m.js` in https://github.com/getodk/central-frontend/actions/runs/17078821495/job/48427136906

bin/check-bundle-size.js
```diff
@@ -41,6 +41,7 @@ const tooBigs = files.filter(f => f.tooBig);
 if (tooBigs.length) {
   log('!!!');
   log(`!!! ${tooBigs.length} file(s) are too big !!!`);
+  log(`!!! Files: ${tooBigs.map(f => f.path).join('; ')}`);
```
sadiqkhoja, line 44: Logs in Github doesn't show colors, so it is hard to see which files are too big

alxndrsn: Good catch; can you link to build logs where this occurs?

alxndrsn: If logging here, maybe:

```suggestion
  log(`!!! ${tooBigs.length} file(s) are too big:`);
  log('!!!');
  tooBigs.map(f => log(`!!! * f.path`));
  log('!!!');
```

?

Otherwise this could be fixed by renaming `redBold()` e.g. `flagViolation()` and e.g.

```js
if(enableColours) return [ RED, BOLD, text, RESET ].join('');
else return `${text}\t<-- TOO BIG`;
```

or something.

matthew-white (approved):
I'll let @alxndrsn take a look at the new PR #1328, but this one looks uncontroversial to me. :+1:

alxndrsn (approved):

PR #1321, opened 2025-08-21
Fix: %20 / space turns into `+`

sadiqkhoja:
Issue: After redirection whether from login page with `next` or from `/f/:enketoId` to canonical path, space character or `%20` is changed into `+` because internally $router.replace/push mimics the behavior of URLSearchParams.

### What has been done to verify that this works as intended?

Our only solution is to treat `+` in the value of default values coming from URL as space. `$router.replace` always converts space into `+` matching the behavior of URLSearchParams, replacing `$router.replace` with something written from scratch would be huge undertaking. 

Default/prefill values for Enketo should be tested thoroughly. 

Subtle change in behavior should be posted on Forum, previously when default value contained `+` then Enketo form showed the `+` sign but now it will show ` ` (space), to show `+` sign user would need to use `%2B`

### DIFF
--- a/e2e-tests/tests/enketo.spec.js
+++ b/e2e-tests/tests/enketo.spec.js
@@ -251,7 +251,7 @@ test.describe('Enketo', () => {
     await expect(frame.getByRole('heading', { name: 'Successful' })).toBeVisible();
   });
 
-  test('default value is consistent between rendering enketo directly or via iframe', async ({ page }) => {
+  test('default value is consistent between rendering enketo directly and via iframe', async ({ page }) => {
     await login(page);
 
     const queryWithSpaces = '?d[/data/first_name]=hello earth + hello mars %20 hello jupiter %2B hello saturn';
@@ -259,12 +259,13 @@ test.describe('Enketo', () => {
     await page.goto(`${appUrl}/enketo-passthrough/${publishedForm.enketoId}${queryWithSpaces}`);
     await expect(page.getByRole('heading', { name: publishedForm.name })).toBeVisible();
 
-    const defaultValueInEnketo = await page.getByLabel('First Name').inputValue();
+    await expect(page.getByLabel('First Name')).toHaveValue('hello earth + hello mars   hello jupiter + hello saturn');
 
     await page.goto(`${appUrl}/projects/${projectId}/forms/${publishedForm.xmlFormId}/submissions/new${queryWithSpaces}`);
     const iframe = await page.frameLocator('iframe');
     await expect(iframe.getByRole('heading', { name: publishedForm.name })).toBeVisible();
 
-    await expect(iframe.getByLabel('First Name')).toHaveValue(defaultValueInEnketo);
+    // except we transform + into space as well in iframe
+    await expect(iframe.getByLabel('First Name')).toHaveValue('hello earth   hello mars   hello jupiter + hello saturn');
   });
 });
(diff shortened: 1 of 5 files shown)


e2e-tests/tests/enketo.spec.js
```diff
@@ -251,20 +251,21 @@ test.describe('Enketo', () => {
...
+  test('default value is consistent between rendering enketo directly and via iframe', async ({ page }) => {
     await login(page);
 
     const queryWithSpaces = '?d[/data/first_name]=hello earth + hello mars %20 hello jupiter %2B hello saturn';
 
     await page.goto(`${appUrl}/enketo-passthrough/${publishedForm.enketoId}${queryWithSpaces}`);
     await expect(page.getByRole('heading', { name: publishedForm.name })).toBeVisible();
 
-    const defaultValueInEnketo = await page.getByLabel('First Name').inputValue();
+    await expect(page.getByLabel('First Name')).toHaveValue('hello earth + hello mars   hello jupiter + hello saturn');
```
matthew-white, line 262: (No change needed, just a note.) I was initially confused by the first `+` here. I thought, shouldn't the `+` in `queryWithSpaces` be converted to a space? But then I realized that this is going directly to /enketo-passthrough without involving Frontend. Enketo is unchanged and will still treat `+` as a plus, not a space. What's changed with this PR is that Frontend will never pass `+` to Enketo. Frontend will pass `%20` for a space or `%2B` for plus, but it won't pass `+`. A `+` passed to Frontend will end up getting passed to Enketo as `%20`. We can sort of see that happening in the assertion below.

test/components/enketo-iframe.spec.js
```diff
@@ -165,7 +174,26 @@ describe('EnketoIframe', () => {
...
+  it('should pass %20 to the iframe after login redirect', () => {
+    testData.extendedForms.createPast(1);
+    testData.extendedUsers.createPast(1, { email: 'test@email.com', role: 'none' });
+    return load('/login?next=%2Fprojects%2F1%2Fforms%2Ff%2Fsubmissions%2Fnew%3Fd%5Bfirstname%5D%3Djohn%2520doe')
+      .restoreSession(false)
+      .complete()
+      .request(submit)
+      .respondWithData(() => testData.sessions.createNew())
+      .respondWithData(() => testData.extendedUsers.first())
+      .respondWithData(() => ({})) //analytics
```
matthew-white, line 189: If `role: 'none'` is specified, a request for analytics shouldn't be sent. Maybe this is a different request? You can use `beforeEachResponse()` to log what's being sent.

matthew-white (approved):
Looks great! The tests help build confidence around the expected behavior.

matthew-white:
@sadiqkhoja, I'm trying to think through the easiest way to release this change. How about:

1. Rebase this PR onto the latest release tag (v2025.2.2). Right now, it's a few commits ahead.
2. I've created a branch named `v2025.2.x` that branches off the latest release tag. How about targeting that branch instead of `master` for this PR?
  - When we release, we can tag the tip of the `v2025.2.x` branch for the release. Then we can merge the branch into `master`.

As an alternative, we could also cherry-pick instead, but I kind of like it when the release tags end up getting on (merging into) `master`.

test/components/enketo-iframe.spec.js
sadiqkhoja: it was actually sending the analytics request because of `extendedUsers.first()` instead of `last()`

PR #1323, opened 2025-08-25
Show tag for repeat group on entity list properties page

ktuite:
Closes https://github.com/getodk/central/issues/1303

In backend, I added a property to the list of source forms that includes the name of the repeat group if the dataset is from a repeat group.

Leaving this as a draft because I still need to add tests. 

### DIFF
--- a/src/components/dataset/overview/connection-to-forms.vue
+++ b/src/components/dataset/overview/connection-to-forms.vue
@@ -25,7 +25,7 @@ except according to the terms contained in the LICENSE file.
                 :to="publishedFormPath(form.projectId, form.xmlFormId)"
                 v-tooltip.text/>
               <span v-if="form.repeatPath" class="repeat-tag"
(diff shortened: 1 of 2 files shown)


src/components/dataset/overview/connection-to-forms.vue
```diff
@@ -24,6 +24,10 @@ except according to the terms contained in the LICENSE file.
               <form-link :form="form"
                 :to="publishedFormPath(form.projectId, form.xmlFormId)"
                 v-tooltip.text/>
+              <span v-if="form.repeatPath" class="repeat-tag"
+                v-tooltip.aria-describedby="form.repeatPath">
```
ktuite, line 28: Can use `v-tooltip.no-aria` to explicitly not be read by screenreader (could be confusing to have path read out loud and field name)

Don't include beginning and trailing slashes in path

matthew-white (approved):

PR #1325, opened 2025-08-27
Fixes: ability to make multiple submission with public links

sadiqkhoja:
This PR addresses the issue of "not being able use last-saved instance with public links" mentioned in https://forum.getodk.org/t/public-access-links-with-last-saved-not-working-as-expected-in-enketo-in-central-v2025-2/56351/7

**Previous behaviour:**
Public link generated by Central had single in the URL, e.g. http://central-dev.localhost:8989/-/single/MbB0VvRZ7Xfl11D79MHPmDXU7AZkt0s?st=WQSP04Urm7OMPaHW2Tb55EVQYTPSkdEretsACKEdhmni5wgDbxBqhNQx4CVTewON . With that last-saved was not available and if user filled this Form then they saw thank-you page but if user removed single from the URL then they would see a success modal and they could continue to make multiple submissions.
I believe project managers used to share public links by removing single from the generated URL, when they wanted use last-saved and their users were able to make multiple submissions without refreshing the page.

**Current behaviour:**
Public link generated by Central doesn’t have single in the URL and when user fills the Form they always see thank-you page because the src of iframe always has single in it with the assumption that users of public link will always want to make one submission or they can refresh the page to make another submission if they want and if the link is not ‘Single Submission’. Additionally Enketo doesn't allow to access last-saved instance with `/single` prefix is given.

**Solution:**
Give users ability to add `/multiple` at the end of public link (before st query parameter), which will allow users to make multiple submissions and use last-saved instance

Added bunch of tests and manually verified that last-saved instance is working.

This is kinda temporary solution, we need to come up with better way of configuring end-of-form behavior in Central.

None

https://github.com/getodk/docs/issues/1985

### DIFF
--- a/src/components/enketo-iframe.vue
+++ b/src/components/enketo-iframe.vue
@@ -35,7 +35,6 @@ const props = defineProps({
     type: String,
     required: true
   },
-  endOfFormBehavior: String,
   instanceId: String
 });
 
@@ -78,7 +77,7 @@ const setEnketoSrc = () => {
     basePath = `/#${basePath}`;
   }
   let prefix = basePath;
-  const { return_url: _, returnUrl: __, ...query } = route.query;
+  const { return_url: _, returnUrl: __, single, ...query } = route.query;
 
   query.parentWindowOrigin = location.origin;
 
@@ -93,12 +92,15 @@ const setEnketoSrc = () => {
   if (props.actionType === 'offline') {
     return; // we don't render offline Enketo through central-frontend
   }
-  if ((props.actionType === 'public-link' && props.endOfFormBehavior !== 'multiple') || props.endOfFormBehavior === 'single') {
+  if (props.actionType === 'public-link' && single !== 'false') {
     prefix += '/single';
   } else if (props.actionType === 'preview') {
     prefix += `/${props.actionType}`;
+  // for actionType 'new', we add '/single' only if 'single' query parameter is true
+  } else if (props.actionType === 'new' && single === 'true') {
+    prefix += '/single';
   }
-  // for actionType 'new', we don't need to add anything to the prefix.
+
   // we no longer render Enketo for Edit Submission from central-frontend.
 
   if (props.enketoId === form.enketoOnceId) {
--- a/src/components/enketo-iframe.vue
+++ b/src/components/enketo-iframe.vue
@@ -70,6 +70,12 @@ const lastSubmitted = (enketoOnceId) => {
 
 const enketoSrc = ref();
 
+const single = computed(() => {
+  const { query } = route;
+  return (props.actionType === 'public-link' && query.single !== 'false') ||
+         (props.actionType === 'new' && query.single === 'true');
+});
+
 const setEnketoSrc = () => {
   let basePath = '/enketo-passthrough';
   // this is to avoid 404 warning
@@ -77,7 +83,7 @@ const setEnketoSrc = () => {
     basePath = `/#${basePath}`;
   }
   let prefix = basePath;
-  const { return_url: _, returnUrl: __, single, ...query } = route.query;
+  const { return_url: _, returnUrl: __, ...query } = route.query;
 
   query.parentWindowOrigin = location.origin;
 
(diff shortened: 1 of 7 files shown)


src/components/enketo-iframe.vue
```diff
@@ -131,7 +132,7 @@ const handleIframeMessage = (event) => {
     try { eventData = JSON.parse(event.data); } catch {}
 
     if (eventData?.enketoEvent === 'submissionsuccess') {
-      if (props.actionType === 'public-link' && redirectUrl.value) {
+      if ((props.actionType === 'public-link' || props.actionType === 'new') && redirectUrl.value) {
```
sadiqkhoja, line 135: this makes `return_url` / `returnUrl` query parameter functional for the logged in users (non public link)

sadiqkhoja: it is dependent on end of behavior (single/multiple), I can create a separate PR but that change should be released in the patch as well.

src/routes.js
```diff
@@ -716,24 +718,33 @@ const routes = [
...
     name: 'EnketoRedirector',
-    props: true,
+    props: (route) => {
+      const { suffix, ...params } = route.params;
+      return {
+        ...params,
+        actionType: suffix === 'preview' ? 'preview' : 'new',
+        endOfFormBehavior: suffix === 'new/single' ? 'single' : 'multiple'
+      };
+    },
```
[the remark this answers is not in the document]
sadiqkhoja: today `/f/:enketoId/new` goes to `/projects/:id/forms/:name/submissions/new`. 

With the latest commit, this change is not necessary. `single` query parameter will be passed by EnketoRedirector as it is.

src/components/form/submission.vue
```diff
@@ -78,6 +79,18 @@ const props = defineProps({
...
+  /**
+ * Only applicable for EnketoIFrame. Possible values are:
+ *
+ * `single`:   Adds the `single` prefix to the Enketo iframe `src`. After submission, the user is
+ *             redirected to a thank-you page or to the `return_url` query parameter. This is the
+ *             default behavior for public links.
+ *
+ * `multiple`: No prefix is added to the Enketo iframe `src`. After submission, the user is shown
+ *             a modal with the option to either close the modal and fill out another submission,
+ *             or close the window/tab. This is the default behavior for logged-in users.
```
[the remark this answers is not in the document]
sadiqkhoja: yes, previously for the logged-in users we used to send users to `/-/:enketoId` which allows multiple submissions.

src/components/enketo-iframe.vue
matthew-white: It seems reasonable to include it in the same PR given that it all has to do with behavior post-submission. :+1:

src/components/enketo-iframe.vue
```diff
@@ -133,7 +136,7 @@ const handleIframeMessage = (event) => {
...
```
matthew-white, line 139: What about a public link where ?single=false&return_url=something ? Which of the query parameters is relevant / takes priority? I.e., is the user able to submit a second submission, or are they redirected?

test/components/enketo-iframe.spec.js
```diff
@@ -49,6 +50,28 @@ describe('EnketoIframe', () => {
...
+
+    const iframe = app.find('iframe');
+    iframe.attributes('src').should.contain('/enketo-passthrough/single/');
+  });
+
+  it('renders iframe without /single prefix when single=false for public-link', async () => {
+    testData.extendedForms.createPast(1, { xmlFormId: 'a' });
+
+    const app = await load(`/f/${enketoId}?single=false&st=token`)
+      .respondWithProblem() // For session restore request - it's public link so respond with problem
```
matthew-white, line 67: You can also do `.restoreSession(false)`, it's fine the way it is though.

src/components/enketo-iframe.vue
matthew-white, line 139: How about adding a test that `return_url` is used in the /new case?

matthew-white (approved):

src/components/enketo-iframe.vue
sadiqkhoja: I think `single` should take priority, let me update the condition.

test/components/enketo-iframe.spec.js
```diff
@@ -109,6 +109,28 @@ describe('EnketoIframe', () => {
...
   });
 
+  it('redirects on submissionsuccess for new submission', async () => {
+    const fakeAssign = sinon.fake();
+    const wrapper = mountComponent({
+      props: { enketoId, actionType: 'new' },
+      container: {
+        router: mockRouter('/?return_url=http://example.com/projects/1'),
+        location: {
+          origin: window.location.origin,
```
matthew-white, line 119: No change needed, just FYI, I added a location mock such that `container.location.origin` automatically equals `window.location.origin` unless it's overridden, same for other properties. So you should only need to specify the property you need to mock (`assign`).

src/components/enketo-iframe.vue
matthew-white: That sounds good for `single` to take priority. :+1: So `return_url` will only be used in the single-submission case, when we are using a /-/single URL with Enketo — is that right?

matthew-white (approved):
Looks great! The test coverage is 💯

PR #1339, opened 2025-09-16
Show last login time on the users page

sadiqkhoja:
Part of getodk/central#1104

Just added a column following existing pattern.

No

We can update the screenshot, I don't know how often we do that. In any case I have added it in https://github.com/getodk/docs/issues/1983

### DIFF
--- a/src/components/user/list.vue
+++ b/src/components/user/list.vue
@@ -32,6 +32,7 @@ either is an Administrator or has no role. -->
           <th>{{ $t('header.displayName') }}</th>
           <th>{{ $t('header.email') }}</th>
           <th>{{ $t('header.sitewideRole') }}</th>
+          <th>{{ $t('header.lastActive') }}</th>
           <th class="actions">{{ $t('header.actions') }}</th>
         </tr>
       </thead>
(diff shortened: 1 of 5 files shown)


src/components/user/list.vue
```diff
@@ -174,7 +175,8 @@ export default {
       "Web Users have accounts on this website to oversee and administer the Projects on this server. Administrators can manage anything on the site. Users with no Sitewide Role can be given a Role on any Project, from that Project’s settings. Sitewide Administrators and some Project Roles can use a web browser to fill out Forms. To submit data through an application such as {collect}, create App Users for each Project."
     ],
     "header": {
-      "sitewideRole": "Sitewide Role"
+      "sitewideRole": "Sitewide Role",
+      "lastActive": "Last active"
```
matthew-white, line 179: How about adding a comment for translators describing what this column is? For some languages, I think translators will want to know what noun the adjective "active" refers to, since the noun and the adjective would need to match.

matthew-white, line 179:
```suggestion
      "lastActive": "Last Active"
```

I think we usually title-case column headers.

matthew-white (approved):

src/components/user/row.vue
matthew-white, line 1: I was just looking at the Figma file, and I noticed that the text "Invitation Pending" is kind of gray in Figma.

PR #1341, opened 2025-09-17
Track when requestData resource is set or patched

matthew-white:
This PR changes `requestData` resources so that they remember when they were last set or patched. The PR doesn't have any immediate user-facing effects, but it should facilitate the following issues:

- getodk/central#1300
  - As the user navigates between pages, we often preserve data. But that can lead to data becoming stale. It'd be nice to be able to clear data that has become stale. To do so, it'd be helpful to know when the data was last set.
- getodk/central#1106
  - This issue calls for displaying a timestamp for when data was last received.

New unit tests.

I'll add comments about specific decisions I made.

### DIFF
--- a/src/request-data/resource.js
+++ b/src/request-data/resource.js
@@ -45,6 +45,8 @@ class BaseResource {
   }
 
   get awaitingResponse() { return this[_store].awaitingResponse; }
+  get setAt() { return this[_store].setAt; }
+  get patchedAt() { return this[_store].patchedAt; }
 
   get dataExists() { return this[_store].data != null; }
   get initiallyLoading() { return this.awaitingResponse && !this.dataExists; }
@@ -62,14 +64,24 @@ const _container = Symbol('container');
 const _abortController = Symbol('abortController');
 class Resource extends BaseResource {
   constructor(container, name) {
-    const store = shallowReactive({ data: null, awaitingResponse: false });
+    const store = shallowReactive({
+      data: null,
+      awaitingResponse: false,
+      setAt: null,
+      patchedAt: null
+    });
     super(name, store);
     this[_container] = container;
     this[_abortController] = null;
   }
 
   get data() { return this[_store].data; }
-  set data(value) { this[_store].data = value; }
+
+  set data(value) {
+    this[_store].data = value;
+    this[_store].setAt = new Date();
+  }
+
   toRefs() { return { ...super.toRefs(), data: toRef(this[_store], 'data') }; }
 
   cancelRequest() { if (this.awaitingResponse) this[_abortController].abort(); }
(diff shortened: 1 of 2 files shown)


src/request-data/resource.js
```diff
@@ -62,14 +68,24 @@ const _container = Symbol('container');
...
     this[_container] = container;
     this[_abortController] = null;
   }
 
   get data() { return this[_store].data; }
-  set data(value) { this[_store].data = value; }
+
+  set data(value) {
+    this[_store].data = value;
+    this[_store].setAt = new Date();
```
matthew-white, line 86: `setAt` is only set: it's never cleared. I thought about setting it to `null` whenever `data` was set to `null`, but I ended up thinking that that wasn't necessary. I think it's simpler if `setAt` is always the timestamp of when `data` was last set, whether to `null` or anything else.

I took a similar strategy for `patchedAt`. I thought about setting `patchedAt` to `null` whenever `data` was set (either to `null` or to anything). But again, it ended up feeling unnecessary and just extra logic.

I'm happy to take another approach, this just seemed like the simplest approach.

src/request-data/resource.js
matthew-white, line 1: I always feel a little nervous about adding extra properties to a resource, since the resource properties like `awaitingResponse`, `data`, and now `setAt` and `patchedAt` are kind of mixed with the properties of `data` via proxy magic. If the `data` property itself had a property named `setAt` or `patchedAt`, things could get confusing. The behavior is well-defined (the proxy only falls back to `data` if the property doesn't exist on the resource itself), but it still seems potentially confusing. I feel OK about adding these two new properties, but this is something I'm thinking about. I thought about tucking these new properties away into a separate resource metadata object (something like `project.resource.setAt` or `project.response.setAt`), but I'm not quite there yet. I feel like that could be confusing in its own way.

test/request-data/resource.spec.js
```diff
@@ -22,7 +52,7 @@ describe('createResource()', () => {
       });
 
       it('passes the response and the resource to the callback', () => {
-        const projectData = testData.extendedProjects.createNew();
+        const projectData = testData.standardProjects.createNew();
```
matthew-white, line 55: This is not the most important change in the world, but the request in this test doesn't request extended metadata, so it feels better to return a response without metadata. I based at least one of the new tests on this one, which is why I was looking at it.

matthew-white (commented):
Notes for code review

sadiqkhoja (approved):

src/request-data/resource.js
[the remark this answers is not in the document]
matthew-white: I've created a PR to update `toRefs()` here: #1349.

PR #1347, opened 2025-09-18
Automatically reseed testData after testData.reset()

matthew-white:
This PR changes `testData.reset()` so that it automatically reseeds `testData` after resetting the data. Currently, resetting and reseeding are two separate operations. However, in practice, I don't think there's a reason to distinguish between the two.

Making this change allows us to remove a `beforeEach` hook in test/index.js. More importantly, it makes it easy for individual tests to reset and reseed `testData`. That need doesn't come up too often, but when it does, it's nice for there to be an easy way to accomplish it. This came up recently at https://github.com/getodk/central-frontend/pull/1344#discussion_r2356562529.

Tests continue to pass.

### DIFF
--- a/test/data/index.js
+++ b/test/data/index.js
@@ -45,7 +45,11 @@ const testData = Object.assign( // eslint-disable-line prefer-object-spread
   Users
 );
 
-testData.seed = seed;
-testData.reset = resetDataStores;
+seed();
+
+testData.reset = () => {
+  resetDataStores();
+  seed();
+};
 
 export default testData;
--- a/test/index.js
+++ b/test/index.js
@@ -28,8 +28,6 @@ window.expect = expect;
 // async components associated with the route.
 beforeAll(loadAsyncRouteComponents);
 
-beforeEach(testData.seed);
-
 enableAutoUnmount(afterEach);
 afterEach(() => {
   const app = document.querySelector('[data-v-app]');


test/data/index.js
```diff
@@ -45,7 +45,11 @@ const testData = Object.assign( // eslint-disable-line prefer-object-spread
...
 );
 
-testData.seed = seed;
-testData.reset = resetDataStores;
+seed();
+
+testData.reset = () => {
+  resetDataStores();
+  seed();
+};
```
matthew-white, line 53: If there's ever a need in the future to be able to reset without reseeding, we can add an optional parameter to `testData.reset()` along those lines. However, it's not something we've needed so far.

test/data/index.js
```diff
@@ -45,7 +45,11 @@ const testData = Object.assign( // eslint-disable-line prefer-object-spread
   Users
 );
 
-testData.seed = seed;
```
matthew-white, line 48: Given that `testData.reset()` now automatically reseeds, I don't think we need to expose `seed()` anymore. We could add it back in the future if we need to, but I think it makes things clearer to keep it encapsulated for now.

matthew-white (commented):
Notes for code review

sadiqkhoja (approved):

PR #1348, opened 2025-09-18
Disable table when bulk operations are in progress

sadiqkhoja:

### DIFF
--- a/src/components/disable-container.vue
+++ b/src/components/disable-container.vue
@@ -1,3 +1,15 @@
+<!--
+Copyright 2025 ODK Central Developers
+See the NOTICE file at the top-level directory of this distribution and at
+https://github.com/getodk/central-frontend/blob/master/NOTICE.
+
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+-->
+
 <template>
   <div class="disable-container">
     <div v-show="disabled" class="disable-container-overlay" tabindex="0" v-tooltip.sr-only></div>
--- a/test/components/disable-container.spec.js
+++ b/test/components/disable-container.spec.js
@@ -17,7 +17,6 @@ describe('DisableContainer', () => {
   describe('disabled prop', () => {
     it('shows overlay when disabled is true', () => {
       const component = mountComponent({ props: { disabled: true } });
-      component.find('.disable-container-overlay').should.exist;
       component.get('.disable-container-overlay').should.be.visible();
     });
 
@@ -61,10 +60,10 @@ describe('DisableContainer', () => {
       component.get('.disable-container-overlay').attributes('tabindex').should.equal('0');
     });
 
-    it('overlay has tooltip directive', () => {
-      const component = mountComponent({ props: { disabled: true } });
+    it('overlay has tooltip directive', async () => {
+      const component = mountComponent({ props: { disabled: true, disabledMessage: 'This section is disabled' } });
       const overlay = component.get('.disable-container-overlay');
-      overlay.element.should.exist;
+      await overlay.should.have.tooltip('This section is disabled');
     });
   });
 });
--- a/src/components/form-attachment/popups.vue
+++ b/src/components/form-attachment/popups.vue
@@ -158,6 +158,7 @@ export default {
 
 <style lang="scss">
 @import '../../assets/scss/variables';
+@import '../../assets/scss/mixins';
 
 $edge-offset: 25px;
 $popup-width: 300px;
@@ -249,14 +250,7 @@ $popup-width: 300px;
 }
 
 #form-attachment-popups-backdrop {
-  background-color: #000;
-  bottom: 0;
-  left: 0;
-  opacity: .8;
-  position: fixed;
-  right: 0;
-  top: 0;
-  z-index: $z-index-modal-backdrop;
+  @include overlay(fixed, 0.8);
 }
 
 @keyframes bob {


src/directives/disable-container.js
```diff
@@ -0,0 +1,82 @@
...
+  overlay.setAttribute('tabindex', '0'); // make it focusable with keyboard
+  vTooltip.mounted(overlay, {
+    modifiers: { 'aria-describedby': true },
+    value: 'Bulk operation in progress' // TODO: translation
+  });
+
+  return overlay;
+};
+
+const setInertOnChildren = (container, inert) => {
```
sadiqkhoja, line 28: It would be nice to avoid this and set insert on the container but if I do that then it would be impossible to make overlay focusable and show tooltip on it

src/directives/disable-container.js
```diff
@@ -0,0 +1,82 @@
...
+*/
+
+import vTooltip from './tooltip';
+
+
+const createOverlay = () => {
+  const overlay = document.createElement('div');
+  overlay.classList.add('disable-container-overlay');
+  overlay.setAttribute('tabindex', '0'); // make it focusable with keyboard
+  vTooltip.mounted(overlay, {
```
sadiqkhoja, line 20: is there a better way to attach tooltip here?

src/directives/disable-container.js
matthew-white, line 1: Do you think there's a way we could do something similar via a component or composable? I generally try to avoid directives where possible, since they're so different from typical patterns. I definitely see that there are cases where a directive just can't be avoided. For example, `v-tooltip` is used with a wide range of elements and components. I don't think it'd really work to render a component for each tooltip, and it wouldn't be ergonomic to call a composable for each tooltip. However, to me, `v-tooltip` is the exception to the rule of avoiding directives. Most patterns don't appear in as many places as tooltips, so they end up working OK as a component or composable.

What do you think about something like the following? Doing as much element insertion as possible via a component, but still calling `setInertOnChildren()` in a watcher as needed.

**src/components/disable-container.vue**

```vue
<template>
  <div ref="el" class="disable-container">
    <!-- .sr-only because I don't think aria-describedby is announced for divs.
    Though maybe it is if there's a tabindex 🤔 -->
    <div v-show="disabled" class="disable-container-overlay" tabindex="0" v-tooltip.sr-only></div>
    <span v-if="disabled && disabledMessage" class="sr-only">{{ disabledMessage }}</span>
    <div :inert="disabled"><slot></slot></div>
  </div>
</template>

<script setup>
import { useTemplateRef, watch } from 'vue';

const props = defineProps({
  disabled: Boolean,
  disabledMessage: String
});

const el = useTemplateRef('el');
// Maybe this isn't needed if we can wrap the slot in a <div :inert="disabled"> ?
watch(
  () => props.disabled,
  (value) => { setInertOnChildren(el.value, value); },
  { flush: 'post' }
);
</script>

<style lang="scss">
@import '../assets/scss/mixins';

.disable-container {
  position: relative;
}

.disable-container-overlay {
  @include overlay(absolute, 0.5);
}
</style>
```

**src/components/entity/list.vue**

```vue
<template>
  <disable-container :disabled="bulkOperationInProgress"
    :disabled-message="'Bulk operation in progress'">
    <entity-table/>
  </disable-container>
</template>
```

matthew-white:
By the way, `inert` is new to me, but it seems perfect for this task. 💯

src/components/disable-container.vue
matthew-white, line 1: Needs the file header with copyright information, etc.

src/assets/scss/_mixins.scss
```diff
@@ -128,3 +128,14 @@
 
   & > * { pointer-events: auto; }
 }
+
+@mixin overlay($position, $opacity) {
```
matthew-white, line 132: How about using this mixin in `FormAttachmentPopups` as well? Just to make it clear what the related usage sites are. Could be a follow-up PR as well.

src/components/disable-container.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <div class="disable-container">
+    <div v-show="disabled" class="disable-container-overlay" tabindex="0" v-tooltip.sr-only></div>
```
matthew-white, line 3: I forget why we have `tabindex` here. Why is it helpful to make the overlay focusable?

test/components/disable-container.spec.js
```diff
@@ -0,0 +1,70 @@
...
+describe('DisableContainer', () => {
+  it('renders slot content', () => {
+    const component = mountComponent();
+    component.get('button').text().should.equal('Test Content');
+  });
+
+  describe('disabled prop', () => {
+    it('shows overlay when disabled is true', () => {
+      const component = mountComponent({ props: { disabled: true } });
+      component.find('.disable-container-overlay').should.exist;
```
matthew-white, line 20:
```suggestion
      component.find('.disable-container-overlay').exists().should.be.true;
```

I think in Vue Test Utils, `.find()` will always return an object. So its return value will always exist / it will always not be `null`, which means the `.should.exist` assertion will definitely pass. I don't think the assertion delegates to the `exists()` method.

The `.get()` on the next line will automatically assert that the element exists, so the assertion on this line could probably also just be omitted if you prefer.

test/components/disable-container.spec.js
```diff
@@ -0,0 +1,70 @@
...
+    it('overlay has tabindex="0" for keyboard tabs', () => {
+      const component = mountComponent({ props: { disabled: true } });
+      component.get('.disable-container-overlay').attributes('tabindex').should.equal('0');
+    });
+
+    it('overlay has tooltip directive', () => {
+      const component = mountComponent({ props: { disabled: true } });
+      const overlay = component.get('.disable-container-overlay');
+      overlay.element.should.exist;
+    });
```
matthew-white, line 68:
```suggestion
    it('overlay has tooltip directive', async () => {
      const component = mountComponent({ props: { disabled: true, disabledMessage: 'This section is disabled' } });
      const overlay = component.get('.disable-container-overlay');
      await overlay.should.have.tooltip('This section is disabled');
    });
```

matthew-white (approved):

src/components/disable-container.vue
sadiqkhoja: without `tabindex` we can't focus it using only keyboard (pressing tab key to focus different elements on the page)

matthew-white: I'm more wondering, why is it important to be able to focus the overlay using the keyboard? The overlay doesn't have content, and it doesn't have an effect when clicked. Does allowing the user to focus the overlay help them better navigate the page?

By the way, I don't think this comment needs to block merge if it'd be helpful to merge the PR sooner. Even if we end up concluding that `tabindex="0"` isn't needed, we can always remove it in a follow-up PR.

sadiqkhoja: I find it helpful to know what overlay is all about when I use only keyboard. With mouse hover, one can quickly see the tooltip to know the reason but if someone is just using keyboard then without tabindex there is no way to see the tooltip.

PR #1349, opened 2025-09-19
Improve requestData toRefs()

matthew-white:
In #1341, I added two new properties to the reactive store underlying each resource object: `setAt` and `patchedAt`. The `toRefs()` method of the resource object is intended to return refs for reactive properties like these. Here are a couple of examples of how `toRefs()` is used:

https://github.com/getodk/central-frontend/blob/b24c3b7779789a80eb89f2e1e8b9642fad9b359f/src/components/project/archive.vue#L60

https://github.com/getodk/central-frontend/blob/b24c3b7779789a80eb89f2e1e8b9642fad9b359f/src/components/project/form-row.vue#L125

However, as mentioned at https://github.com/getodk/central-frontend/pull/1341#discussion_r2354231211, I decided not to add new refs to `toRefs()` for `setAt` and `patchedAt`. That's because I wanted to rework `toRefs()` in a couple of ways:

- Currently, `toRefs()` needs to be patched whenever a new reactive property is added to a resource object. I think it'd be better for `toRefs()` to be self-contained and not require maintenance/modification.
- Currently, whenever `toRefs()` is called, it generates all the refs that could possibly be accessed. However, components generally only need to grab a single ref off `toRefs()`. It seems wasteful to eagerly generate new refs for `setAt` and `patchedAt` every time `toRefs()` is called, knowing that those refs won't typically be used.

This PR solves for those concerns by changing `toRefs()` to return a proxy object. The proxy will generate refs on the fly rather than doing so eagerly.

The use of a proxy makes `toRefs()` a little more complicated. However, that logic is self-contained and won't need to be modified on a recurring basis. The proxy also eliminates the need for overriding `toRefs()` via inheritance, which was needed before and was a source of complexity. Overall, I think it's an improvement in clarity.

Existing tests continue to pass. I also added new unit tests of `toRefs()`, which I think help clarify exactly how the method behaves.

### DIFF
--- a/src/request-data/resource.js
+++ b/src/request-data/resource.js
@@ -9,7 +9,7 @@ https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
 including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 */
-import { computed, isRef, readonly, shallowReactive, toRef } from 'vue';
+import { computed, isRef, shallowReactive } from 'vue';
 
 import { isProblem, logAxiosError, requestAlertMessage, withHttpMethods } from '../util/request';
 import { noop } from '../util/util';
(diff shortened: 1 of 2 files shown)


matthew-white:
@sadiqkhoja, do you think you'd have time to take a look at this PR before regression testing? Not high-priority at all — other code reviews are more important. But if it fits before regression testing, it'd be nice to get it merged. Happy to discuss it interactively as well!

PR #1351, opened 2025-09-25
Render OpenLayers map

matthew-white:
This PR makes progress on submission mapping (getodk/central#1152). The biggest change is to the `GeojsonMap` component, which now renders an actual map using the OpenLayers library.

<details>
<summary>Here is a fuller list of what this PR does</summary>

- Render an OpenLayers map.
- Parse GeoJSON, adding the resulting geo features to the map.
- Initially fit the view so that the map shows all features.
- Show the number of features in view (e.g., "Showing 10 of 100").
- Update the map in place after a background refresh of the GeoJSON. The features on the map may change, but the center of the map and the zoom level will not.
- Hide the map when the GeoJSON is cleared.
  - This happens after filters are changed, but before the new response is received. During that time, the map is hidden, and a loading message is shown.
- Wait to show the map until conditions are met. The following two conditions are resolved async and can happen in either order:
  1. The GeoJSON response is received.
  2. All ancestor elements become visible. Much of `SubmissionList` is hidden until the `fields` response is received.
- Cluster features.
- Allow a feature or cluster to be selected.
- Style features based on whether they are selected or not.
- Show a popup after a feature is selected. The popup will send a request for OData.
- Deselect a feature
  1. After a click on an empty space on the map
  2. After the close button in the popup is clicked
</details>

---

This PR adapts the code in getodk/web-forms#491. It's been really helpful to be able to draw from that PR. In particular, the `GeojsonMap` component in this PR includes a lot of the code from the `useMapBlock()` composable in Web Forms. src/util/map-styles.js here closely matches map-styles.ts in Web Forms. There are some differences between the design/criteria for Central vs. Web Forms, some of which I mentioned in a [review](https://github.com/getodk/web-forms/pull/491#pullrequestreview-3270032257) I left on the Web Forms PR.

<details>
<summary>Here are some of the main differences between the two PRs</summary>

- Central fits the view to show all features. **UPDATE:** Actually, I think Web Forms does this now too.
- Central clusters features.
- Central shows the number of features in view.
- Central waits for ancestor elements to become visible before showing the map. As far as I know, that's not needed in Web Forms. Central also waits for the base layer to be loaded.
- Web Forms allows features to be saved, not just selected.
- Web Forms shows a button to center the map on the user's location. It shows an error message if the location is not available.
- Web Forms shows a button to make the map full-screen.
- Both Central and Web Forms load the OpenLayers dependency async. However, Web Forms seems to have more sophisticated logic around how that's done, surfacing the loading state. Central just uses `GeojsonMap` with `defineAsyncComponent()`, similar to other async components in Central. Nothing is shown until the async component is downloaded.
- Central and Web Forms both show popups, but they're different from one another and were implemented separately. For Central, the popup will be different from one page to another (i.e., for submissions vs. entities).
- I'm sure I'm missing other things that Web Forms does and other differences between the two PRs. I'll add things to this list as I think of them. Mostly I just wanted to give examples of how even though the two PRs are very similar in some ways, using OpenLayers similarly, they're also different in other ways.
</details>

---

This is an initial PR, and there's additional work that will need to be done. Most of that will probably come in follow-up PRs. I'm eager to get this PR merged sooner rather than later so that we can get it on staging and the QA team can begin testing it.

<details>
<summary>Here are some remaining to-dos</summary>

- Update the appearance of the count message.
- Add the buttons to the bottom of the popup.
- getodk/central#1391
- getodk/central#1392
- getodk/central#1383
- getodk/central#1384
- getodk/central#1389
</details>

---

I've been trying it out locally. I've tested points, lines, and shapes. I also faked the Backend response to return 100,000 random points to see what the performance would be like. I was happy to see that the map was still plenty responsive given that amount of data.

I've updated existing Frontend tests, and everything is passing. However, as of the time of writing, I haven't added new tests yet. I'm hoping to at least write tests of the popup as part of this PR. For tests of the `GeojsonMap` component, those are more likely to come in a follow-up PR: see getodk/central#1381.

Especially in the `GeojsonMap` component, I tried to add code comments explaining the approach. In a review I left on the Web Forms PR, I left comments about places where I ended up doing things differently than Web Forms: https://github.com/getodk/web-forms/pull/491#pullrequestreview-3270032257

Outside `GeojsonMap`, the main change is the popup. The popup largely follows existing patterns:

- `MapPopup` is a generic component that contains the code that will be shared between submissions and entities. It's pretty similar to `HoverCard`. I originally thought I might be able just to modify `HoverCard` and reuse that. But there ended up being enough differences that I decided to create a new component.
- The `SubmissionMapPopup` component is specific to submissions. It renders a `MapPopup` component. It's similar in some ways to `SubmissionBasicDetails`.

### DIFF
--- a/src/components/geojson-map.vue
+++ b/src/components/geojson-map.vue
@@ -185,8 +185,8 @@ const resize = () => {
 
 const fitView = (extent, options = undefined) => {
   mapInstance.getView().fit(extent, {
-    // Provide enough space for styled features.
-    padding: [40, 40, 40, 40],
+    // We need to provide enough space for styled features.
+    padding: [50, 50, 50, 50],
     // Avoid zooming in to an extreme degree.
     maxZoom: 16,
     ...options
@@ -303,7 +303,6 @@ const hide = () => {
 let selectedId;
 
 const hitDetectionOptions = {
-  hitTolerance: 6,
   layerFilter: (layer) => layer === featureLayer
 };
 
@@ -332,7 +331,7 @@ const selectCluster = (cluster) => {
   selectFeature(null);
 
   const features = cluster.get('features');
-  const duration = 600;
+  const duration = 1000;
   if (features.length < 1000) {
     const extent = createEmpty();
     for (const feature of features)
--- a/src/components/geojson-map.vue
+++ b/src/components/geojson-map.vue
@@ -291,6 +291,18 @@ const hide = () => {
 
 let selectedId;
 
+const hitDetectionOptions = {
+  hitTolerance: 5,
+  layerFilter: (layer) => layer === featureLayer
+};
+
+// Updates the cursor based on whether the user is moving over a feature.
+const moveOverFeature = (event) => {
+  if (!shown.value || event.dragging) return;
+  const hit = mapInstance.hasFeatureAtPixel(event.pixel, hitDetectionOptions);
+  mapContainer.value.style.cursor = hit ? 'pointer' : '';
+};
+
 // Selects an individual feature (not a cluster) or deselects the selected
 // feature (if feature is `null`).
 const selectFeature = (feature) => {
@@ -333,10 +345,7 @@ const selectCluster = (cluster) => {
 };
 
 const selectFeatureAtPixel = (pixel) => {
-  const hits = mapInstance.getFeaturesAtPixel(pixel, {
-    hitTolerance: 5,
-    layerFilter: (layer) => layer === featureLayer
-  });
+  const hits = mapInstance.getFeaturesAtPixel(pixel, hitDetectionOptions);
   if (hits.length === 0) {
     selectFeature(null);
   } else {
@@ -376,6 +385,7 @@ const olOn = (target, type, callback) => {
 };
 
 olOn(mapInstance, 'moveend', () => { if (shown.value) countFeaturesInView(); });
+olOn(mapInstance, 'pointermove', moveOverFeature);
 
 // OpenLayers has a `singleclick` event, but it lags the actual click by 250
 // milliseconds in order to exclude double-clicks. Here, we listen for `click`
@@ -463,10 +473,10 @@ onBeforeUnmount(() => {
     position: absolute;
     top: 5px;
     left: 63px;
-    pointer-events: none;
 
     color: #000;
     font-size: 12px;
+    user-select: none;
   }
 }
 </style>
--- a/src/components/geojson-map.vue
+++ b/src/components/geojson-map.vue
@@ -279,7 +279,7 @@ const show = async () => {
 
   shown.value = true;
   abortShow = noop;
-  countFeaturesInView();
+  inViewCount.value = featureCount.value;
   log('shown');
   emit('shown');
 };
@@ -426,9 +426,8 @@ watch(() => props.data, (newData, oldData) => {
       // set, even as the map remains transparent.
       show();
     else
-      // If the map isn't already shown, show() above will call
-      // countFeaturesInView(). But if the map is shown, we still need to call
-      // countFeaturesInView().
+      // If the map isn't already shown, then show() above will set
+      // inViewCount.value.
       countFeaturesInView();
   } else {
     hide();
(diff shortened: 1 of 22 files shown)


matthew-white (commented):
Round 2 of interactive review

src/components/geojson-map.vue
```diff
@@ -10,28 +10,484 @@ including this file, may be copied, modified, propagated, or distributed
...
+    ? { id, properties: feature.getProperties() }
+    : null);
+};
+
+const selectCluster = (cluster) => {
+  // Deselect any feature that's currently selected.
+  selectFeature(null);
+
+  const features = cluster.get('features');
+  const duration = 600;
```
matthew-white, line 335: It looks like this is 1000 in Web Forms, let's match it.

src/components/geojson-map.vue
```diff
@@ -10,28 +10,484 @@ including this file, may be copied, modified, propagated, or distributed
...
+
+
+
+////////////////////////////////////////////////////////////////////////////////
+// SELECT FEATURE
+
+let selectedId;
+
+const hitDetectionOptions = {
+  hitTolerance: 6,
```
matthew-white, line 306: This has been removed from Web Forms. Let's remove it here.

matthew-white (commented):
I just did another [review](https://github.com/getodk/web-forms/pull/491#pullrequestreview-3287613223) of the Web Forms PR. There are small changes that could be made to this PR to more closely match Web Forms.

test/components/submission/map-popup.spec.js
```diff
@@ -0,0 +1,148 @@
...
+      names: { first_name: 'Someone' },
+      p1: 'POINT (1 1)',
+      p2: 'POINT (2 2)'
+    });
+  });
+
+  it('does nothing if instanceId is not defined', () =>
+    mockHttp()
+      .mount(SubmissionMapPopup, mountOptions({
+        props: { instanceId: null }
```
matthew-white, line 52:
```suggestion
        props: { instanceId: null, fieldpath: null }
```

ktuite (approved):

matthew-white (commented):
Notes from interactive review

src/components/geojson-map.vue
```diff
@@ -10,28 +10,484 @@ including this file, may be copied, modified, propagated, or distributed
...
+const baseLayer = new TileLayer({ source: new OSM() });
+
+// We use WebGL when possible for performance reasons. If WebGL is not
+// available, we fall back to a basic 2D canvas. Our testing setup doesn't
+// support WebGL. Maybe some users don't as well.
+const supportsWebGL2 = document.createElement('canvas').getContext('webgl2') != null;
+if (!supportsWebGL2 && buildMode !== 'test')
+  // eslint-disable-next-line no-console
+  console.warn('WebGL2 not supported for map');
+const featureLayerClass = supportsWebGL2 ? WebGLVectorLayer : VectorLayer;
```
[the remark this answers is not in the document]
matthew-white: Done in e1138357c1928dc0090caa92f6ef2d6ede2c12f5.

src/components/geojson-map.vue
matthew-white: Done in c376a54fae2cc0fbce50b15009c65fe216e59a74

src/components/geojson-map.vue
```diff
@@ -10,28 +10,478 @@ including this file, may be copied, modified, propagated, or distributed
...
+  // OpenLayers style rules.
+  for (const feature of features) feature.set('id', feature.getId());
+
+  featureSource.addFeatures(features);
+  featureCount.value = features.length;
+  log('features added');
+};
+
+const removeFeatures = () => {
+  if (featureCount.value === 0) return;
```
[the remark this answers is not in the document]
matthew-white: I think I'm going to leave this check. It doesn't add much complexity. There are other functions in this file like this one that are called more often than they strictly need to be, and those other functions typically have an early return as well. Also, maybe it's nice not to call OpenLayers `clear()` if we don't need to.

PR #1353, opened 2025-09-26
Improve OdataLoadingMessage

matthew-white:
This PR makes a few changes to the `OdataLoadingMessage` component. Most of the changes are to facilitate mapping (getodk/central#1152).

- Replace the `odata` and `refreshing` props with a single `state` prop
  - The main impetus for this is that there will now be times when we show the loading message even after the response has been received. In `SubmissionMapView`, it takes a moment for the map to render after the response is received. Instead of leaving things completely blank during that time, I want to continue showing the loading message.
  - Even without that impetus, I think it's overall a simplifying change. It means that we don't need to pass a `refreshing` prop to `SubmissionMapView` or `SubmissionTableView`. There was also awkwardness in that `SubmissionMapView` was passing the `geojson` resource for the `odata` prop.
- Replace margin with padding
  - For mapping, I need to measure the `top` pixel of `SubmissionMapView` in order to correctly size the map. However, the top margin on `OdataLoadingMessage` was affecting where `SubmissionMapView` thought its `top` was.
- Make the `totalCount` and `top` props optional
  - `SubmissionMapView` doesn't have access to these values. Right now, it's passing `0` for both, but it'd be nice if it could just omit them.
- Other minor changes
  - One change that I'm happy about is that we can remove the `#odata-loading-spinner-container`. That's because the `Spinner` component now has an `inline` prop that we can use instead.

I've tried it out locally in the context of #1351. I've also updated tests.

### DIFF
--- a/src/components/odata-loading-message.vue
+++ b/src/components/odata-loading-message.vue
@@ -24,9 +24,6 @@ import Spinner from './spinner.vue';
 
 import { useI18nUtils } from '../util/i18n';
 
-const { t } = useI18n();
-const { tn } = useI18nUtils();
-
 defineOptions({
   name: 'OdataLoadingMessage'
 });
@@ -41,6 +38,9 @@ const props = defineProps({
   totalCount: Number,
   top: Number
 });
+
+const { t } = useI18n();
+const { tn } = useI18nUtils();
 
 const message = computed(() => {
   if (!props.state) return null;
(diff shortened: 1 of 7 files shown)


src/components/odata-loading-message.vue
```diff
@@ -23,87 +21,56 @@ import { computed } from 'vue';
...
+}
 
-  #odata-loading-message-text {
-    color: #555;
-    font-size: 12px;
-    padding-left: 24px;
-  }
+#odata-loading-message-text {
+  color: #555;
+  margin-left: 9px;
```
matthew-white, line 73:
```suggestion
  margin-left: 8px;
```

matthew-white (commented):
Notes from interactive review

PR #1355, opened 2025-10-01
Fixes : handle concurrent refresh and bulk delete gracefully

sadiqkhoja:
Closes getodk/central#1334

- Disable refresh button while bulk request is in progress
- Clear selection when refresh request is completed (was happening already because of watcher on odataEntities, added test)
- Abort refresh request on bulk request

Implemented all suggestions in original comment https://github.com/getodk/central-frontend/pull/1316/files#r2286984706

NA

NA 

### DIFF
--- a/src/components/entity/list.vue
+++ b/src/components/entity/list.vue
@@ -575,6 +575,7 @@ export default {
       this.allSelected = false;
     },
     cancelBackgroundRefresh() {
+      if (!this.refreshing) return;
       this.odataEntities.cancelRequest();
       this.deletedEntityCount.cancelRequest();
     },
--- a/src/components/entity/list.vue
+++ b/src/components/entity/list.vue
@@ -18,7 +18,7 @@ except according to the terms contained in the LICENSE file.
         :disabled="deleted" :disabled-message="deleted ? $t('filterDisabledMessage') : null"/>
       </form>
       <button id="entity-list-refresh-button" type="button"
-        class="btn btn-outlined" :aria-disabled="refreshing"
+        class="btn btn-outlined" :aria-disabled="refreshing || bulkOperationInProgress"
         @click="fetchChunk(false, true)">
         <span class="icon-refresh"></span>{{ $t('action.refresh') }}
         <spinner :state="refreshing"/>
@@ -341,6 +341,8 @@ export default {
 
       const $search = this.searchTerm ? this.searchTerm : undefined;
 
+      this.clearSelectedEntities();
+
       this.odataEntities.request({
         url: apiPaths.odataEntities(
           this.projectId,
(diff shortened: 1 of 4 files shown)


src/components/entity/list.vue
```diff
@@ -572,12 +574,17 @@ export default {
       this.odataEntities.value?.forEach(e => { e.__system.selected = false; });
       this.allSelected = false;
     },
+    cancelBackgroundRefresh() {
+      this.odataEntities.cancelRequest();
+      this.deletedEntityCount.cancelRequest();
```
matthew-white, line 579: (Just FYI, no change needed.) Initially I thought, maybe there's no need to cancel this request. It doesn't affect much. But then I remembered that the bulk operations mutate `deletedEntityCount.value`. There could be a race condition if this request weren't canceled. So I think the behavior is clearer this way. :+1:

src/components/entity/list.vue
```diff
@@ -341,6 +341,8 @@ export default {
 
       const $search = this.searchTerm ? this.searchTerm : undefined;
 
+      this.clearSelectedEntities();
```
matthew-white, line 344: In the PR description, you wrote:

> Clear selection when refresh request is completed (was happening already because of watcher on odataEntities, added test)

I see that this line is also clearing the selection right before the request is sent. If the watcher clears the selection after the response, am I thinking correctly that the selection is cleared both before and after? Is the selection cleared before the request just to make it clear as soon as possible that clicking the Refresh button clears the selection? Or to phrase the question differently, what are the benefits of clearing the selection both before and after vs. just after? I'm happy for us to keep this line, I just wanted to make sure I understood the reasoning.

matthew-white (approved):

src/components/entity/list.vue
```diff
@@ -572,12 +574,17 @@ export default {
       this.odataEntities.value?.forEach(e => { e.__system.selected = false; });
       this.allSelected = false;
     },
+    cancelBackgroundRefresh() {
+      this.odataEntities.cancelRequest();
+      this.deletedEntityCount.cancelRequest();
+    },
```
matthew-white, line 580:
```suggestion
    cancelBackgroundRefresh() {
      if (!this.refreshing) return;
      this.odataEntities.cancelRequest();
      this.deletedEntityCount.cancelRequest();
    },
```

What do you think about adding an early return to `cancelBackgroundRefresh()`? I know in most cases, such a check isn't necessary, but I think it makes it a little clearer to reason about the method and its possible side effects. I started thinking, even if a background refresh isn't in progress, I think it's possible for the initial request for `this.deletedEntityCount` to be in progress when `cancelBackgroundRefresh()` is called, in which case the request for `this.deletedEntityCount` will be canceled prematurely/unnecessarily. I know that case is pretty unlikely, but I think we can circumvent situations like that by checking `this.refreshing` upfront.

PR #1376, opened 2025-10-21
Procfile: run nginx via docker

alxndrsn:
Work done to get this project working on a new system.

This should help:

* standardise the nginx version running on different dev machines
* simplify dev environment setup
* standardise behaviour between central's deployed docker, and dev machines
* streamline testing frontend-related changes to nginx config

Tested locally - this is just for dev.

This subject was discussed previously at https://github.com/getodk/central-frontend/pull/1035.

The final suggestion there was to maintain the default behaviour of requiring devs to install and maintain their own `nginx` installations.

No change - this is dev-only.

No.  But dev documentation has been updated.


matthew-white:
I don't think my thoughts have changed that much about this since #1035. I'd personally prefer just to run Frontend locally without Docker, but I can see how it'd be helpful to support Docker as well. :+1: If it's feasible to support both setups, that sounds great to me.

One question I have is whether the Docker setup plays well with Vite HMR. If you change Frontend code while using Docker, does it immediately update in the browser? If so, making Docker the default sounds reasonable to me. I think HMR should be on by default, so if the Docker setup doesn't play well with HMR, I think Docker should get its own command (`npm run dev:docker`?). Note that right now, there's a difference between `npm run dev`, which uses HMR, and `npm run dev:build`, which doesn't use HMR and is more similar to the development setup pre-Vite.

> They're listed in order of descending desirability for me!

+1

Thanks for working on this, @alxndrsn!

alxndrsn:
I've had another go at this, and got close to zero changes with the nginx config, and supporting both with/without docker via:

```sh
npm run dev
npm run dev-no-docker
```

I haven't tested `npm run dev-no-docker`.  Hopefully the `-g "daemon off;"` switch works - when included in the config file it breaks nginx startup with:

```
nginx: [emerg] "daemon" directive is duplicate in /etc/nginx/nginx.conf:14
```

.gitignore
```diff
@@ -23,9 +23,6 @@ npm-debug.log*
 *.sln
 *.sw?
 
-/.nginx/*
-!/.nginx/.gitkeep
```
matthew-white: If the top-level .nginx/ will continue to be used for `npm run dev-no-docker`, won't we need to .gitignore the files within it?

matthew-white (commented):
> I've had another go at this, and got close to zero changes with the nginx config, and supporting both with/without docker

Excellent news, that sounds perfect to me.

I have a few small questions, but this approach is looking great to me.

package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "nf start vite,nginx",
+    "dev": "          nf start vite,nginx-docker",
+    "dev-no-docker": "nf start vite,nginx-host",
     "dev:build": "nf start build,nginx",
```
matthew-white:
```suggestion
    "dev-no-docker:build": "nf start build,nginx-host",
```

I want to look into getodk/central#1606 at some point, and this change will help in doing that. I'll probably be using `npm run dev-no-docker` at least for now.

We could also make this two commands (`dev:build` + `dev-no-docker:build`). I wouldn't be opposed to that, though neither will work until getodk/central#1606 is resolved. But the main thing is that I don't think it should reference `nginx` from the Procfile anymore if `nginx` has been replaced by `nginx-docker` + `nginx-host`.

alxndrsn: :+1: fixed

nginx-conf/.nginx/.gitignore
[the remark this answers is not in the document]
matthew-white: > The alternative is that `npm run dev` currently in `master` is already using `nginx-conf/.nginx/`.

It doesn't seem to be for me. I see it writing nginx.pid and nginx_access.log to the top-level .nginx/ directory.

<blockquote>

> Is there any way we could get the two to use the same directory?

I agree this would be preferable.  The following might work:

```diff
-nginx-host: nginx -g "daemon off;" -c "$PWD/nginx-conf/nginx.conf" -p "$PWD"
+nginx-host: nginx -g "daemon off;" -c "$PWD/nginx-conf/nginx.conf" -p "$PWD/nginx-conf"
```
</blockquote>

Mostly I think that would work. The only thing that I think wouldn't work are references in the config file to dist/, e.g., `root ./dist;`. Maybe that could become `root ../dist;`? Would that do alright with Docker?

> Maybe that folder should be called `nginx` instead of `nginx-conf`, and the temp dir should be `nginx/tmp` or something.

I like this idea if it ends up looking feasible, grouping all the nginx stuff in a single top-level directory.

matthew-white (approved):

PR #1411, opened 2025-11-07
i18n: detect unused keys

### LINKED ISSUE #829: Check for unused i18n messages
I'm noticing an unused i18n message in the `DatasetShow` component in Frontend. It'd be nice to have a way to search the codebase for unused i18n messages. All i18n messages are sent to Transifex for translation, so unused i18n messages lead to extra work for translators.

alxndrsn:
# Blocked by

* currently breaking `npm ci`
* https://github.com/getodk/central/issues/1578
* intlify/eslint-plugin-vue-i18n#643

---

Currently the ignore rules are _very_ general, and won't detect e.g. unused `back` keys mentioned in https://github.com/getodk/central/issues/1350.  This might be improved by something like https://github.com/intlify/eslint-plugin-vue-i18n/issues/668.

Closes getodk/central#829

It doesn't right now.

An alternative would be to implement something ourselves reusing the code in `bin/transifex`.

This PR also seems to introduce a bunch of other new vue eslint rules, which are then disabled in this PR.  That doesn't seem ideal; maybe they need to be evaluated and enabled separately.

Might reduce download sizes a tiny bit.  Should also prevent wasted time by translators.

No.


.eslintrc.js
```diff
@@ -15,6 +18,29 @@ module.exports = {
     defineModel: 'readonly'
   },
   rules: {
+    '@intlify/vue-i18n/no-unused-keys': [ 'error', {
+      src: 'src/',
+      callExpression: '^(\\$t|t|\\$tc|tc|\\$tcn|tcn|\\$tn|tn)$',
```
matthew-white: Once #1026 is complete, there won't be a `$tc()` or `tc()` anymore. At that time, we'll probably rename `$tcn()` to `$tn()`.

Currently, there is `$tcn()`, and there is `tn()`; there's no `$tn()` or `tcn()`. The asymmetry is because we've had `$tcn()` for a long time, while `tn()` was added around the time that we moved to Vue 3 and started using the Composition API. In the context of Vue I18n within the Composition API, `tc()` is deprecated, because `t()` now does what `$t()` and `$tc()` both used to do.

matthew-white: Re: being blocked by intlify/eslint-plugin-vue-i18n#643, I have two thoughts:

1. Maybe we just ignore `$tcn()` and `tn()` for now. They're not the most common way to access i18n messages.
2. What if in CI, we modified code files, e.g., replacing all strings containing `/\btn\(/` with `t(`? Would the Vue I18n check work then?

matthew-white (commented):
Just wanted to add some initial thoughts! As I wrote at https://github.com/getodk/central/issues/829#issuecomment-3504806196, I don't even know how we'd write our own check. The approach here seems promising to me. Even if it doesn't catch everything, it'd be much better than not checking at all.

PR #1417, opened 2025-11-18
Add pagination info in the query parameter

### LINKED ISSUE #1348: Add pagination to URL for submission/entity tables
Right now, when you filter the submission or entity table, the filter appears in the URL. That means that you can copy the URL and return to the table with the same filters applied. This issue is to add **pagination** to the URL:

- As you move from one page of the table to another, the URL should update (e.g., .../submissions?page=2, ?page=3, etc.).
- If you change the page size (e.g., from 250 to 500), the URL should update to reflect the change.

Adding pagination to the URL means that you'll be able to copy the URL to return to the same place in the table.

**Questions**

- What if the page in the URL no longer exists? E.g., some submissions have been deleted.
- What if the page size is invalid? E.g., someone copies the URL incorrectly and enters 50 instead of 500.

sadiqkhoja:
Closes getodk/central#1348

Manual verification + added few tests.

Please see inline comments.

No

No

### DIFF
--- a/apps/central/test/components/submission/list.spec.js
+++ b/apps/central/test/components/submission/list.spec.js
@@ -129,6 +129,25 @@ describe('SubmissionList', () => {
           });
         });
     });
+
+    it('makes the correct request after refresh button is pressed on second page', () => {
+      createSubmissions(251);
+      return loadSubmissionList()
+        .complete()
+        .request(component =>
+          component.find('button[aria-label="Next page"]').trigger('click'))
+        .respondWithData(() => testData.submissionOData(250, 250))
+        .complete()
+        .request(component =>
+          component.get('#refresh-button').trigger('click'))
+        .beforeEachResponse((_, { url }) => {
+          if (url.includes('.svc/Submissions')) {
+            const skip = relativeUrl(url).searchParams.get('$skip');
+            skip.should.be.equal('0');
+          }
+        })
+        .respondWithData(() => testData.submissionOData(250));
+    });
   });
 
   describe('refreshing keys', () => {
@@ -279,6 +298,26 @@ describe('SubmissionList', () => {
           $select(url).should.equal('__id,__system');
         })
         .respondWithData(testData.submissionOData));
+
+    // eslint-disable-next-line func-names
+    it('resets page if different field is selected on the second page', function() {
+      this.timeout(20 * 1000);
+      testData.extendedSubmissions.createPast(251);
+      return loadSubmissionList({ attachTo: document.body })
+        .complete()
+        .request(component =>
+          component.find('button[aria-label="Next page"]').trigger('click'))
+        .respondWithData(() => testData.submissionOData(250, 250))
+        .complete()
+        .request(changeMultiselect('#submission-field-dropdown', [19]))
+        .beforeEachResponse((_, { url }) => {
+          if (url.includes('.svc/Submissions')) {
+            const skip = relativeUrl(url).searchParams.get('$skip');
+            skip.should.be.equal('0');
+          }
+        })
+        .respondWithData(testData.submissionOData);
+    });
   });
 
   describe('deleted submissions', () => {
@@ -980,32 +1019,6 @@ describe('SubmissionList', () => {
         });
     });
 
-    it('adds page-size query parameter when page size is changed', () => {
-      createSubmissions(251);
-      return load('/projects/1/forms/f/submissions')
-        .complete()
-        .request(component => {
-          const sizeDropdown = component.find('.pagination select:has(option[value="500"])');
-          return sizeDropdown.setValue(500);
-        })
-        .respondWithData(() => testData.submissionOData(500))
-        .afterResponse(component => {
-          component.vm.$route.query['page-size'].should.equal('500');
-        });
-    });
-
-    it('adds page-number query parameter when next page is clicked', () => {
-      createSubmissions(251);
-      return load('/projects/1/forms/f/submissions')
-        .complete()
-        .request(component =>
-          component.find('button[aria-label="Next page"]').trigger('click'))
-        .respondWithData(() => testData.submissionOData(250, 250))
-        .afterResponse(component => {
-          component.vm.$route.query['page-number'].should.equal('2');
-        });
-    });
-
     it('displays the correct page when page-number is provided in URL', () => {
       createSubmissions(501);
       return load('/projects/1/forms/f/submissions?page-number=2', { root: false })
@@ -1033,16 +1046,9 @@ describe('SubmissionList', () => {
     it('selects last page when page-number is greater than last page in URL', () => {
       createSubmissions(501);
       return load('/projects/1/forms/f/submissions?page-number=999', { root: false })
+        .respondWithData(() => testData.submissionOData(250, 500))
         .afterResponse(component => {
           component.find('.pagination .form-group').text().should.be.eql('Row 501 of 501');
-        });
-    });
-
-    it('floors page-size to nearest valid value when invalid page-size is provided in URL', () => {
-      createSubmissions(251);
-      return load('/projects/1/forms/f/submissions?page-size=350', { root: false })
-        .afterResponse(component => {
-          component.find('.pagination select').element.value.should.be.eql('250');
         });
     });
 
--- a/apps/central/test/components/submission/list.spec.js
+++ b/apps/central/test/components/submission/list.spec.js
@@ -979,5 +979,97 @@ describe('SubmissionList', () => {
           ]);
         });
     });
+
+    it('adds page-size query parameter when page size is changed', () => {
+      createSubmissions(251);
+      return load('/projects/1/forms/f/submissions')
+        .complete()
+        .request(component => {
+          const sizeDropdown = component.find('.pagination select:has(option[value="500"])');
+          return sizeDropdown.setValue(500);
+        })
+        .respondWithData(() => testData.submissionOData(500))
+        .afterResponse(component => {
+          component.vm.$route.query['page-size'].should.equal('500');
+        });
+    });
+
+    it('adds page-number query parameter when next page is clicked', () => {
+      createSubmissions(251);
+      return load('/projects/1/forms/f/submissions')
+        .complete()
+        .request(component =>
+          component.find('button[aria-label="Next page"]').trigger('click'))
+        .respondWithData(() => testData.submissionOData(250, 250))
+        .afterResponse(component => {
+          component.vm.$route.query['page-number'].should.equal('2');
+        });
+    });
+
+    it('displays the correct page when page-number is provided in URL', () => {
+      createSubmissions(501);
+      return load('/projects/1/forms/f/submissions?page-number=2', { root: false })
+        .afterResponse(component => {
+          component.find('.pagination .form-group').text().should.be.equal('Rows 251–500 of 501');
+        });
+    });
+
+    it('displays correct number of rows when page-size is provided in URL', () => {
+      createSubmissions(600);
+      return load('/projects/1/forms/f/submissions?page-size=500', { root: false })
+        .afterResponse(component => {
+          component.find('.pagination select').element.value.should.be.eql('500');
+        });
+    });
+
+    it('selects first page when page-number is less than 1 in URL', () => {
+      createSubmissions(251);
+      return load('/projects/1/forms/f/submissions?page-number=0', { root: false })
+        .afterResponse(component => {
+          component.find('.pagination .form-group').text().should.be.eql('Rows 1–250 of 251');
+        });
+    });
+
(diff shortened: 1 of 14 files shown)


matthew-white:
A few quick thoughts from me!

1. Order of PRs

I think this PR will conflict with getodk/central#1434, just to the extent that I'll be adding an `EntityTableView` component for that. Should I go ahead and push a PR for that now? Or would it be easier to get this one merged first? I've been working on other aspects of entity mapping, so I'd be happy for this one to go first if it could get merged quickly.

2. Toggling to/from map view

If there are pagination-related query parameters, do those currently persist when toggling to map view? For example:

- Navigate to the table.
- Move to the second page of the table (`?page-number=2`).
- Move to map view.
    - I would expect the `page-number` query parameter to be removed at this point.
- Move back to table view.
    - I would expect to see the first page of the table.

3. Composable

How about we add a composable for the pagination-related query parameters? I'm thinking of something similar to `useDateRangeQueryRef()`. I could see such a composable being useful in the future beyond submissions and entities.

src/components/entity/list.vue
```diff
@@ -187,10 +187,32 @@ export default {
 
     const pageSizeOptions = [250, 500, 1000];
 
+    const pageNumber = useQueryRef({
+      fromQuery: (query) => {
+        const num = Number.parseInt(query['page-number'], 10);
+        if (!num || num < 0) return 0;
+        return num - 1;
+      },
+      toQuery: (value) => ({ 'page-number': value + 1 })
```
matthew-white: What do you think about removing the query parameter for the first page? Since the first page is the default — one less query parameter in the URL.

```suggestion
      toQuery: (value) => ({ 'page-number': value === 0 ? null : value + 1 })
```

src/components/entity/list.vue
```diff
@@ -43,7 +43,7 @@ except according to the terms contained in the LICENSE file.
         type="entity"
         :top="pagination.size"
         :filter="odataFilter != null || !!searchTerm"
-        :total-count="dataset.dataExists ? dataset.entities : 0"/>
+        :total-count="dataset.dataExists && !pagination.page ? dataset.entities : 0"/>
```
sadiqkhoja: I don't want to show "Loading first N entities...", with this change it will always display "Loading entiities..." when there is a pagination information in the query parameter on the first loading. Note that on change pages using pagination control doesn't show any loading message.

src/components/submission/table-view.vue
```diff
@@ -137,6 +161,11 @@ const fetchChunk = (clear, refresh = false) => {
   })
     .then(() => {
       pagination.count = odata.count;
+      const lastPage = Math.max(0, Math.ceil(pagination.count / pagination.size) - 1);
+      if (pagination.page > lastPage) {
+        pagination.page = lastPage;
+        fetchChunk(false);
```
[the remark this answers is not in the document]
sadiqkhoja: introduced `skipSettingSnapshotFilter` in fetchChunk

src/composables/pagination-query-ref.js
```diff
@@ -0,0 +1,41 @@
...
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+*/
+
+import useQueryRef from './query-ref';
+
+export default (pageSizeOptions = [250, 500, 1000]) => {
+  const pageNumber = useQueryRef({
+    fromQuery: (query) => {
+      const num = Number.parseInt(query['page-number'], 10);
```
sadiqkhoja: I using `page-number` for the readability sake, we can also go with just `page`.

src/composables/pagination-query-ref.js
```diff
@@ -0,0 +1,41 @@
...
+      if (!num || num < 0) return 0;
+      return num - 1;
```
sadiqkhoja: I guess page number starting from 1 is more acceptable than 0.

src/composables/pagination-query-ref.js
matthew-white: I definitely prefer `page` over `page-number`. It's much shorter, and I see it on many sites. `page` + `page-size` sounds like a nice combo to me. I don't have the strongest feelings about it though, we could run with `page-number` for now and see if anyone else on the team voices an opinion.

src/components/submission/table-view.vue
```diff
@@ -88,12 +90,10 @@ const odataSelect = computed(() => {
...
 // `clear` indicates whether this.odata should be cleared before sending the
-// request. `refresh` indicates whether the request is a background refresh
+// request. `refresh` indicates whether the request is a background refresh.
 // (whether the refresh button was pressed).
 const fetchChunk = (clear, refresh = false) => {
-  // Are we fetching the first chunk of submissions or the next chunk?
-  const first = clear || refresh;
-  if (first) {
+  if (refresh) {
     pagination.page = 0;
```
matthew-white: I'm trying to think through the ramifications of `pagination.page = 0;` for a background refresh. Wouldn't that cause the pagination control to immediately change to the first page? But the actual table wouldn't change until the request completes. If that takes a while, I think it'd be confusing for the pagination control to say that it's on the first page. Or even worse, if the request results in an error, the pagination control would incorrectly think that it's on the first page.

If it's a background refresh, I feel like `pagination.page` should be set to `0` only once the response is received.

It kind of feels like that would be changing the current behavior of background refresh though. Right now, does it refresh the current page rather than returning to the first page? If so, I feel like changing that behavior should maybe go in its own PR. Then this PR could just be focused on query parameters.

src/composables/pagination-query-ref.js
```diff
@@ -0,0 +1,41 @@
...
+Copyright 2025 ODK Central Developers
+See the NOTICE file at the top-level directory of this distribution and at
+https://github.com/getodk/central-frontend/blob/master/NOTICE.
+
+This file is part of ODK Central. It is subject to the license terms in
+the LICENSE file found in the top-level directory of this distribution and at
+https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
+including this file, may be copied, modified, propagated, or distributed
+except according to the terms contained in the LICENSE file.
+*/
```
matthew-white: We don't need these file headers for new files anymore. 🥳 So this can be removed.

matthew-white (commented):
It's looking good!

Probably the main question I still have is where/when `pagination.page` is set to 0.

In some cases, I made non-essential suggestions. I know we want to get this PR merged soon, so I'm happy with us skipping some of those suggestions or filing follow-up issues about them.

src/components/submission/table-view.vue
sadiqkhoja: It is existing behavior that on background refresh we navigate to the first page. Still I have moved the `pagination.page = 0;` to the `.then` block

src/components/submission/table-view.vue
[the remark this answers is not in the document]
sadiqkhoja: when I change the page number in the URL, I see whole page gets reloaded both in chrome and firefox, which sets the correct pagination options before request is made hence page itself shows the correct data. I hope I understood your comment correctly.

src/components/submission/table-view.vue
sadiqkhoja: I have tested on test.getodk.cloud and on pressing of refresh button it goes back to the first page.

What I have done now is that `$skip` would send 0 if `refresh` or `clear` is true otherwise use pagination options. And `pagination.page=0` happens only in the success response.

src/components/submission/table-view.vue
```diff
@@ -126,7 +131,10 @@ const fetchChunk = (clear, refresh = false) => {
     .catch(noop);
 };
 fetchChunk(true);
-watch([() => props.filter, () => props.deleted], () => { fetchChunk(true); });
+watch([() => props.filter, () => props.deleted], () => {
+  pagination.page = 0;
```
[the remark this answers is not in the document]
sadiqkhoja: Since I have move `pagination.page=0` in the `then` of the request, it will retain the existing behavior of changing columns.

test/util/router.js
matthew-white: I think these comments are out-of-date now:

- The mock router can now navigate.
- It's no longer read-only.

apps/central/test/util/router.js
```diff
@@ -75,7 +75,10 @@ export const mockRouter = (location = undefined) => (container) => {
       app.component('RouterLink', RouterLinkStub);
       app.component('RouterView', RouterViewStub);
     },
-    push: sinon.fake()
+    push: sinon.fake(),
+    replace: (loc) => {
+      currentRoute.value = router.resolve(loc);
+    },
```
matthew-white, line 85: I don't feel like it totally works to make this change on its own. For example, `$route` is currently set to a constant at install time under the assumption that it will never change. But that assumption won't be true anymore.

I'm thinking we could create a `Proxy` object that delegates to `currentRoute.value`. Something like:

```js
const routeProxy = new Proxy(currentRoute, {
  get: (route, prop) => route.value[prop]
});

app.config.globalProperties.$route = routeProxy;

app.provide(routeLocationKey, routeProxy);
```

test/util/router.js
```diff
@@ -75,7 +75,10 @@ export const mockRouter = (location = undefined) => (container) => {
       app.component('RouterLink', RouterLinkStub);
       app.component('RouterView', RouterViewStub);
     },
-    push: sinon.fake()
+    push: sinon.fake(),
```
matthew-white: What do you think about making `push()` the same as `replace()`? It feels a little unexpected that `replace()` will change the route, while `push()` won't. `push()` could remain a Sinon fake. So something like:

```js
push: sinon.fake(loc => {
  currentRoute.value = router.resolve(loc);
}),
```

I'm not sure how tests currently use `push()`, so maybe this would cause problems, but hopefully that'd work.

matthew-white (commented):
I haven't taken a look at all the changes yet, but I thought I'd leave some initial comments about `mockRouter()` and `replace()`. Related comment: https://github.com/getodk/central-frontend/pull/1417#discussion_r2936246221

apps/central/test/components/submission/list.spec.js
```diff
@@ -129,6 +129,25 @@ describe('SubmissionList', () => {
           });
         });
     });
+
+    it('makes the correct request after refresh button is pressed on second page', () => {
```
matthew-white, line 133: Could you add a similar test for `EntityList`? It looks to me like `SubmissionList` and `EntityList` work slightly differently in this area.

src/components/entity/table-view.vue
```diff
@@ -107,19 +112,21 @@ const fetchChunk = (clear, refresh = false) => {
...
   // This function is called for size change as well. So when the total number of entities are
   // less than the lowest size option, hence we don't need to make a request.
   if (odataEntities.count < pageSizeOptions[0]) return;
-  fetchChunk(false);
+  fetchChunk(false, pagination.page);
 };
 // For defineExpose()
 const exposedFetch = {
-  fetchData: (clear = true) => fetchChunk(clear, !clear),
+  fetchData: (clear = true) => fetchChunk(clear, clear ? 0 : pagination.page),
```
matthew-white:
```suggestion
  fetchData: (clear = true) => fetchChunk(clear),
```

I'm thinking that the second argument here should always be 0 (or should be omitted). At https://github.com/getodk/central-frontend/pull/1417#discussion_r3065291791, you wrote:

> I have tested on test.getodk.cloud and on pressing of refresh button it goes back to the first page.

But to me, that doesn't seem to be what this line is doing. If `clear` is `false` (if the refresh button has been pressed), then the user will remain on their current `pagination.page`, not go back to the first page.

src/components/submission/table-view.vue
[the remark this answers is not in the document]
matthew-white: Ah right, I forgot that's how it works. This might have been a concern under the old hash-based routing, but not anymore.

src/components/entity/list.vue
matthew-white: It'd be great if we could resolve this comment before merging. If this PR will make several i18n messages irrelevant, I think it'd be best to remove them. That could be done in a follow-up PR, but it'd be nice to track the need to do so in a new issue.

matthew-white:
I think this comment thread still needs to be resolved: https://github.com/getodk/central-frontend/pull/1417#discussion_r2566357871.

I glanced over all the discussion above, and with the exception of that thread and my review today, I didn't see any other open threads.

matthew-white (approved):
Looks great! 💯 I know we've been discussing this one for a while, but I think this PR has ended up in a really nice place. I think it says something that it now only modifies 8 files. I really like the unit tests of the `usePaginationQueryRef()` composable — they feel like they're at the right level.

There are still a few comments left to resolve. Nothing big though, so I'll go ahead and approve. @sadiqkhoja, feel free to merge once you've resolved those comments!

PR #1525, opened 2026-04-02
Update i18n messages related to dataset deletion

### LINKED ISSUE #1726: Revisit i18n of Delete Property modal
I think there are a few simple changes we could make to the Delete Property modal to improve its i18n:

**1. More translators comments**

Here's one example. In some languages, "Form(s)" will need to be inflected, because it's the object of "unlink". I think we should add a translator comment about `formsCount` that explicitly connects it to `unlinkForms`. The typical way to do that with component interpolation is using a `*.full` message that automatically adds translator comments.

```json
"unlinkForms": "unlink {forms} that set it",
"formsCount": "{count} Form | {count} Forms",
```

**2. Think about how to list actions**

```json
"intro": "Before you can delete the Property “{propertyName}”, you’ll need to {actions}.",
"clearEntities": "clear its value in {entities}",
"and": "and",
"unlinkForms": "unlink {forms} that set it",
```

Not all languages use a single word like "and" to connect verbs. For example, in Japanese, typically the verbs themselves are inflected when connected, which might not work with the fact that "and" is conditional here. I think it'd be better to list the actions as bullet points. So make the `intro` message very simple (something like "you'll need to do the following"), then list each action separately. In general, breaking a sentence into component parts like this often subtly introduces linguistic/grammatical assumptions.

CC @ktuite @sadiqkhoja

matthew-white:
Closes getodk/central#1726.

Updated tests. I haven't had the chance to try it out locally.

I'll add comments on individual lines to give more context about some of the decisions I made.

### DIFF
--- a/src/components/dataset/overview/delete-property-error.vue
+++ b/src/components/dataset/overview/delete-property-error.vue
@@ -15,22 +15,15 @@ except according to the terms contained in the LICENSE file.
     <template #title>{{ $t('title') }}</template>
     <template #body>
       <div v-if="state && errorObject" class="modal-introduction">
-        <i18n-t tag="p" keypath="intro">
-          <template #propertyName>{{ errorObject.details.propertyName }}</template>
-          <template #actions>
-            <i18n-t v-if="nonEmptyEntities" keypath="clearEntities">
-              <template #entities>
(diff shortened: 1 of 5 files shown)


sadiqkhoja (approved):

sadiqkhoja:
Tested locally:

image
image
image

PR #1526, opened 2026-04-03
Update text of SubmissionFeedEntry related to dataset deletion

matthew-white:
This PR is a follow-up to #1487. It makes some small changes to the text of `SubmissionFeedEntry`. I'll leave some line comments with additional details.

Updated tests.

### DIFF
--- a/src/components/submission/feed-entry.vue
+++ b/src/components/submission/feed-entry.vue
@@ -70,7 +70,7 @@ except according to the terms contained in the LICENSE file.
           </template>
           <template #dataset>
             <dataset-link v-if="entityDetails.entityAvailable" :project-id="projectId" :name="entityDetails.dataset"/>
-            <span v-else>{{ $t('title.entity.deletedDataset', { dataset: entityDetails.dataset }) }}</span>
+            <span v-else>{{ entityDetails.dataset }}</span>
           </template>
         </i18n-t>
       </template>
@@ -388,7 +388,7 @@ export default {
         "reprocess": "Previous Submission in offline update chain was received"
       }
     },
-    // Shown in the Submission feed when dataset or entity has been deleted.
+    // This text is shown next to an Entity when the Entity has been deleted.
     "deleted": "(deleted)"
   }
 }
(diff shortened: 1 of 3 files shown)


src/components/submission/feed-entry.vue
```diff
@@ -70,7 +70,7 @@ except according to the terms contained in the LICENSE file.
           </template>
           <template #dataset>
             <dataset-link v-if="entityDetails.entityAvailable" :project-id="projectId" :name="entityDetails.dataset"/>
-            <span v-else>{{ $t('title.entity.deletedDataset', { dataset: entityDetails.dataset }) }}</span>
```
matthew-white, line 73: When I search the codebase, I don't see a message named `deletedDataset`. When I log the title in the test, it doesn't look right:

> Updated Entity (deleted) xyz in title.entity.deletedDataset Entity List

This is for `entity.update.version`. When I look at `entity.create`, I see that it uses a simpler `<span v-else>`. I used the same pattern here.

src/components/submission/feed-entry.vue
```diff
@@ -388,7 +388,7 @@ export default {
         "reprocess": "Previous Submission in offline update chain was received"
       }
     },
-    // Shown in the Submission feed when dataset or entity has been deleted.
+    // This text is shown next to an Entity when the Entity has been deleted.
```
matthew-white, line 391: It makes sense to me that there are two different ways that an entity can end up being unavailable: if the entity is deleted vs. if the dataset is deleted. However, I don't think translators or users need to make that distinction. They can just understand that the entity is in the trash for some reason. Here, translators need to know what single noun the word "deleted" modifies. Hélène [wrote](https://getodk.slack.com/archives/D6BECFJCC/p1775197179434839?thread_ts=1774633067.096489&cid=D6BECFJCC) on Slack about this string in French:

> only works grouped this way if entity and entity list have the same gender.

It's shown next to the entity label for feed entries that are about entities (`entity.create`, `entity.update.version`), so I think we can say that it's specifically about entities.

sadiqkhoja (approved):

PR #1528, opened 2026-04-09
feature-flags: handle synthetic events cleanly

### LINKED ISSUE #1280: TypeError in useFeatureFlags()
Sometimes, but not always, I see an error like this when I open the browser console:

```
TypeError: undefined is not an object (evaluating 'n.key.toLowerCase')
```

The line that's flagged is:

https://github.com/getodk/central-frontend/blob/acd5cf674f3460335438f428ed280996676f4823/src/composables/feature-flags.js#L27

I'm not able to reproduce the error consistently. Looking at the code, I'm wondering whether it's possible for `event.key` to be `undefined` when `event` is a `keydown` or `keyup` event.

It'd be nice to resolve this error, but I don't think that it's user-facing. Even after seeing the error in the browser console, I'm able to press W+F to toggle Web Form buttons in Frontend.

alxndrsn:
Closes https://github.com/getodk/central/issues/1280

A new test, which fails before the code change.

The test is a bit different to others, but it has to avoid VueJS `test-utils`'s `component.trigger()`, as that function creates a KeyEvent, which seems to _always_ define `.key` as a `string`.

No effect expected.

No.

### DIFF
--- a/test/composables/feature-flags.spec.js
+++ b/test/composables/feature-flags.spec.js
@@ -35,7 +35,9 @@ describe('useFeatureFlags()', () => {
   });
 
   // Occasionally synthetic events can be fired without a .key prop, e.g.
-  // Chrome password manager filling form fields.
+  // Chrome password manager filling form fields.  Note that this *cannot*
+  // be recreated with `component.trigger()`, as that sends a `KeyEvent`,
+  // which _always_ has .key defined as a String.
   // See: https://github.com/getodk/central/issues/1280
   it('should not throw when a synthetic key event is fired', async () => {
     mountComponent();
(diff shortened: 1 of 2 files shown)


matthew-white (approved):
Thank you!

test/composables/feature-flags.spec.js
```diff
@@ -33,4 +33,13 @@ describe('useFeatureFlags()', () => {
 
     component.classes().should.be.empty;
   });
+
+  // Occasionally synthetic events can be fired without a .key prop, e.g.
+  // Chrome password manager filling form fields.
```
alxndrsn:
```suggestion
  // Occasionally synthetic events can be fired without a .key prop, e.g.
  // Chrome password manager filling form fields.  Note that this *cannot*
  // be recreated with `component.trigger()`, as that sends a `KeyEvent`,
  // which _always_ has .key defined as a String.
```

PR #1531, opened 2026-04-10
ci: ensure SVGs are minified

alxndrsn:
Ref https://github.com/getodk/central-frontend/pull/1529

This code is already in use in the website repo.

Duplicated code from website repo, so didn't consider alternatives much.

Should be no effect - just test code.  There's some chance that minifying an existing image might make it display differently.

No.


public/fonts/icomoon.svg
alxndrsn, line 1: What is this image even of?

karma.conf.js
```diff
@@ -71,7 +71,7 @@ module.exports = (config) => {
       '/v1/config/public/logo': '/base/src/assets/images/odk-logo.png',
       '/img/banner@1x.6c9e9f21.png': '/base/src/assets/images/whats-new/banner@1x.png', // Smaller resolution for circleCI test
       '/img/map-location.b523ce2d.svg': '/base/src/assets/images/geojson-map/map-location.svg',
-      '/img/fullscreen.37a932a6.svg': '/base/src/assets/images/geojson-map/fullscreen.svg'
+      '/img/fullscreen.334affb3.svg': '/base/src/assets/images/geojson-map/fullscreen.svg'
```
alxndrsn, line 74: @matthew-white any idea where these random strings come from?  I just copied this from the test failure, but would be good to explain here (in a comment) how to update them.

karma.conf.js
```diff
@@ -71,7 +71,7 @@ module.exports = (config) => {
       '/v1/config/public/logo': '/base/src/assets/images/odk-logo.png',
       '/img/banner@1x.6c9e9f21.png': '/base/src/assets/images/whats-new/banner@1x.png', // Smaller resolution for circleCI test
       '/img/map-location.b523ce2d.svg': '/base/src/assets/images/geojson-map/map-location.svg',
```
alxndrsn, line 73: `map-location.svg` has changed, so presumably this also needs to change.  Or it can be deleted?

alxndrsn: PR to remove at https://github.com/getodk/central-frontend/pull/1532

public/fonts/icomoon.svg
matthew-white: Our icons I think? 🤔 It's referenced in icomoon.css:

https://github.com/getodk/central-frontend/blob/993cf5feb9903c603ec8126cdd0abbb29c553bd2/src/assets/css/icomoon.css#L3-L6

src/assets/images/geojson-map/fullscreen.svg
matthew-white, line 1: We copied this .svg file from the Web Forms repo, so part of me is reluctant to change it. It's nice to be able to see that the two files are the same.

src/assets/images/geojson-map/map-location.svg
matthew-white, line 1: Similar comment for this one: we copied this .svg file from Web Forms. We did change its color, but that's it.

karma.conf.js
matthew-white: I'm not sure where the strings come from. I also just copy them from the test failure. 😅 To be honest, I don't love this image mapping setup we have for Karma. I'm hoping that this whole thing will be simpler with Vitest. I've put getodk/central#1267 in v2026.2, so hopefully we're able to make that switch soon.

matthew-white:
I'd say that I'm of two minds on this one. Conceptually, it makes sense to me to minimize SVG assets. On the other hand, we don't have that many SVG assets. When we have a big image, it tends to be PNG or JPEG. Two of the SVG files here were copied from Web Forms, and I'm a little hesitant to change them. The last SVG file was autogenerated by IcoMoon as part of a process that's already kind of involved (documented [here](https://github.com/getodk/central-frontend/blob/master/CONTRIBUTING.md#icons)).

How much are these files' sizes decreasing in practice? If it's appreciable, then it's probably worth doing. Maybe we could even share the reduced files back to Web Forms. But if the file sizes aren't decreasing by much, maybe it's not something we need to do quite yet.

alxndrsn:
> How much are these files' sizes decreasing in practice?

file | initial | minified
---|---|---
`fullscreen.svg` | `  165` | `  144`
`map-location.svg` | ` 1396` | ` 1137` 
`icomoon.svg` | `59857` | `56814`

> Two of the SVG files here were copied from Web Forms... The last SVG file was autogenerated by IcoMoon

All good arguments for retaining the existing versions.

An alternative would be to treat SVGs in git as source files, and minify them as part of the vite build.

matthew-white:
Now that I take a closer look, maybe Vite is already doing something fancy with some of the .svg files. When I run `npm run build`, the only .svg file I see is icomoon.svg. The other two seem to be inlined within the .js (specifically, geojson-map-*.js). Vite will definitely minify the JS. Hopefully it will also minify the SVG before inlining it in the JS. (I could check if that's of interest.)

In terms of why it's not inlining icomoon.svg, probably part of the reason is that it lives in public/, which I don't think the build transforms (it passes through files in public/ as static files). I'm not sure it even could be inlined given that its only reference is in CSS.

matthew-white:
> Hopefully it will also minify the SVG before inlining it in the JS. (I could check if that's of interest.)

I went ahead and checked, and it looks like the answer is "it does some minification, but not as much as it could." E.g., map-location.svg inlined is 1373 characters. That's less than the baseline of 1396 characters, and I can see that it's removing newline characters. But that's well afield of the 1137 achieved in this PR.

matthew-white:
I'm still feeling divided about this one.

- I like the idea to "treat SVGs in git as source files", but I'm not sure if that plays well with the magic that Vite is doing.
- We're definitely seeing some size reduction. Yet in absolute terms, it's not huge: at most 3 kB.
- Then again, why leave bytes on the table? We'll probably have more SVGs in the future.

Let me CC the Web Forms team for their opinion (@latin-panda @garethbowen). Do you all do anything special to minify SVG files? I'm wondering if the script in this PR could be helpful for Web Forms as well. Or maybe we just decide that SVG minification isn't needed for these repos quite yet.

latin-panda:
> Do you all do anything special to minify SVG files? I'm wondering if the script in this PR could be helpful for Web Forms as well. 

We haven’t prioritized SVG minification yet, and opting to keep the build process simple until it becomes necessary. Currently, we use [@mdi/js](https://www.npmjs.com/package/@mdi/js), which provides optimized icons (~150–350 bytes) and supports tree-shaking to ensure we only bundle what we use.

Our custom assets are also quite small. We have 5 SVGs from Figma (under 1.4kB) and the ODK logo (3kB). While we could look into optimizing the logo specifically, the rest of our assets are already at an acceptable size. That said, we're open to the script if we expect our custom SVG library to grow significantly in the future.

matthew-white:
I'm realizing that our monorepo plans (#1540) probably intersect with this PR. I think the vision is that check-images.sh will run for the entire repo, both Central Frontend and Web Forms. That means that it'd be good to get the Web Forms team's buy-in before making this change.

> we're open to the script if we expect our custom SVG library to grow significantly in the future.

At least on the Central side, I don't think we expect there to be many more .svg files or for there to be large .svg files. The .svg files for the map feel like a rare/special case.

For Central, the only .svg file that's sizable is icomoon.svg. I'm not totally sure what the conditions are that would cause icomoon.svg to load rather than icomoon.ttf or icomoon.woff. It could be that it's rarely loaded: ChatGPT tells me that SVG fonts aren't needed for modern browsers.

We're also hoping to change Central's icon set (getodk/central#931). At that point, I doubt we'll be using IcoMoon: icomoon.svg should go away. I like the sound of Web Forms' approach to icons: automatic tree-shaking sounds much nicer than IcoMoon.

> We haven’t prioritized SVG minification yet, and opting to keep the build process simple until it becomes necessary.

I think that's the direction I'm leaning as well. Maybe we keep the process simple for now and focus just on .png files. It's the .png files that are the largest images.

> We copied this .svg file from the Web Forms repo, so part of me is reluctant to change it. It's nice to be able to see that the two files are the same.

Just wanted to note that with the monorepo, this reasoning no longer applies. If we changed the file for Central, we would change the corresponding file for Web Forms as well (since check-images.sh would flag both), which means that the two files would continue to be the same.

Relatedly, at some point we should probably create a directory in the monorepo for shared assets like fullscreen.svg and the ODK logo.

---

As an alternative to the approach in this PR, I'm thinking we could reduce the file size limit of .svg files in check-bundle-size.js:

https://github.com/getodk/central-frontend/blob/01fe14fea5e558d2f7a052403b286784703ca927/bin/check-bundle-size.js#L73

I think we can make the default much smaller (5 kB?). Then create an exception specifically for icomoon.svg (or even just remove icomoon.svg if we're convinced it's unneeded). check-bundle-size.js can be our warning sign: if we ever start bundling large .svg files, that's probably a sign that we need to update our process along the lines of this PR.

@alxndrsn, what do you think about all this? Thank you for starting this conversation and for writing check-images.sh in the first place!

alxndrsn:
> I think we can make the default much smaller (5 kB?). Then create an exception specifically for icomoon.svg (or even just remove icomoon.svg if we're convinced it's unneeded). check-bundle-size.js can be our warning sign: if we ever start bundling large .svg files, that's probably a sign that we need to update our process along the lines of this PR.
> 
> @alxndrsn, what do you think about all this? Thank you for starting this conversation and for writing check-images.sh in the first place!

:+1: this sounds good to me

matthew-white:
I've filed getodk/central#1911 to keep track of that idea. I'm thinking it probably warrants its own PR, since it's a pretty different idea from the one here. @alxndrsn, what do you think about closing out this PR for now? We can always reopen later if we end up wanting to take the approach here.

PR #1532, opened 2026-04-10
karma: remove unnecessary image mapping

alxndrsn:
Ref https://github.com/getodk/central-frontend/pull/1531/changes#r3063568260

Tests still pass without it.

Less code, less to think about.

No effect - just test code.

No.


karma.conf.js
```diff
@@ -70,7 +70,6 @@ module.exports = (config) => {
       '/v1/config/public/hero-image': '/base/src/assets/images/whats-new/banner@1x.png',
       '/v1/config/public/logo': '/base/src/assets/images/odk-logo.png',
       '/img/banner@1x.6c9e9f21.png': '/base/src/assets/images/whats-new/banner@1x.png', // Smaller resolution for circleCI test
-      '/img/map-location.b523ce2d.svg': '/base/src/assets/images/geojson-map/map-location.svg',
```
matthew-white: Ah I bet Vite is inlining this SVG file, so it's no longer being requested and Karma no longer needs to serve it. I'm not sure though why map-location.svg would be inlined while fullscreen.svg isn't, given that fullscreen.svg is smaller. Maybe it's because of how map-location.svg is used (in src/util/map-styles.js). I'm also not sure why this line was added here in the first place, was there a point at which the file wasn't inlined? Not questions we need to answer, it's just interesting. In any case, things are definitely clearer with fewer images here, so LGTM. :+1:

apps/central/karma.conf.js
```diff
@@ -71,7 +71,6 @@ module.exports = (config) => {
       '/v1/config/public/hero-image': '/base/apps/central/src/assets/images/whats-new/banner@1x.png',
       '/v1/config/public/logo': '/base/apps/central/src/assets/images/odk-logo.png',
       '/img/banner@1x.6c9e9f21.png': '/base/apps/central/src/assets/images/whats-new/banner@1x.png', // Smaller resolution for circleCI test
-      '/img/map-location.b523ce2d.svg': '/base/apps/central/src/assets/images/geojson-map/map-location.svg',
```
matthew-white: Ah I bet Vite is inlining this SVG file, so it's no longer being requested and Karma no longer needs to serve it. I'm not sure though why map-location.svg would be inlined while fullscreen.svg isn't, given that fullscreen.svg is smaller. Maybe it's because of how map-location.svg is used (in src/util/map-styles.js). Not a question we need to answer, it's just interesting. In any case, things are definitely clearer with fewer images here, LGTM. :+1: Thank you, @alxndrsn!

matthew-white (approved):

matthew-white:
Unfortunately this PR probably will conflict with #1549. Might be worth waiting til #1549 is merged and then rebasing this one, just since #1549 is so large?

PR #1534, opened 2026-04-21
Update gitignore file to exclude context files for AI agents

sadiqkhoja:
This change allows individual developer to create their own context files for their preferred AI agent and do their own experimentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,9 @@ npm-debug.log*
 public/index.html
 test-results
 playwright-report
+
+# context files for AI agents
+/AGENTS.md
+/CLAUDE.md
+/.github/copilot-instructions.md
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -30,3 +30,9 @@ npm-debug.log*
 public/index.html
 test-results
 playwright-report
+
+# context files for AI agents
+AGENTS.md
+CLAUDE.md
+.github/copilot-instructions.md
```
matthew-white:
```suggestion
/.github/copilot-instructions.md
```

What do you think about prefixing with / so that it only matches the top-level directory? I think that makes sense for this line at least, maybe the others as well.

matthew-white (approved):

PR #1535, opened 2026-04-23
New Form page

### LINKED ISSUE #1827: Create new-form page with the content of the existing new form modal

sadiqkhoja:
Closes https://github.com/getodk/central/issues/1827

Tests are passing + manual verification.

- Just removed the modal stuff from the existing `new.vue` component and renamed it to `upload.vue`
- Created a new component called `new-page.vue`, current it just hosts `FormUpload`. Latter it will have links/img for the training/help content.
- On the edit draft page, form upload section is hide by default and it is shown on the press of 'upload new definition' button. Ultimately the upload section will be always visible in the following PR once we do the restyling of that page.

- Form upload flow (file selection, warnings, error handling) should work identically
- Navigation after successful form creation should still redirect to the draft page

Only if we have screenshots or videos there.

### DIFF
--- a/src/components/form/edit.vue
+++ b/src/components/form/edit.vue
@@ -122,7 +122,6 @@ provide('dragDisabled', dragDisabled);
 const dragHandler = ref(noop);
 provide('dragHandler', dragHandler);
 
-const uploadModal = modalData();
 const { router, alert } = inject('container');
 const { t } = useI18n();
 const afterUpload = () => {
@@ -130,7 +129,6 @@ const afterUpload = () => {
   draftAttachments.reset();
   formDraftDatasetDiff.reset();
 
-  uploadModal.hide();
   alert.success(t('alert.upload'));
 };
 
@@ -171,12 +169,6 @@ const afterAbandon = () => {
 <style lang="scss">
 #form-edit {
   padding-inline: 10px;
-
-  #form-upload {
-    p {
-      max-width: unset;
-    }
-  }
 }
 
 body:has(#form-edit) { background-color: #fff; }
--- a/src/components/form/edit.vue
+++ b/src/components/form/edit.vue
@@ -23,14 +23,13 @@ except according to the terms contained in the LICENSE file.
         </div>
       </div>
       <template v-if="formDraft.isDefined()">
-        <form-edit-def @upload="uploadModal.show()"/>
+        <form-edit-def @after-upload="afterUpload"/>
         <form-draft-testing/>
         <form-edit-draft-controls @publish="publishModal.show()"
           @abandon="abandonModal.show()"/>
       </template>
     </template>
 
-    <form-new v-bind="uploadModal" @hide="uploadModal.hide()" @success="afterUpload"/>
     <form-draft-publish v-if="formDraft.dataExists && formDraft.isDefined()"
       v-bind="publishModal" @hide="publishModal.hide()"
       @success="afterPublish"/>
@@ -51,7 +50,6 @@ import FormEditCreateDraft from './edit/create-draft.vue';
 import FormEditDef from './edit/def.vue';
 import FormEditDraftControls from './edit/draft-controls.vue';
 import FormEditPublishedVersion from './edit/published-version.vue';
-import FormNew from './new.vue';
 import Loading from '../loading.vue';
 
 import useRoutes from '../../composables/routes';
@@ -173,6 +171,12 @@ const afterAbandon = () => {
 <style lang="scss">
 #form-edit {
   padding-inline: 10px;
+
+  #form-upload {
+    p {
+      max-width: unset;
+    }
+  }
 }
 
 body:has(#form-edit) { background-color: #fff; }
--- a/src/components/form/edit/def.vue
+++ b/src/components/form/edit/def.vue
@@ -15,6 +15,12 @@ except according to the terms contained in the LICENSE file.
     <template #subtitle>{{ $t('subtitle') }}</template>
     <template v-if="changed" #tag>{{ $t('changed') }}</template>
     <template #body>
+      <div v-if="uploadSectionState">
+        <p class="form-edit-section-title">{{ $t('uploadSection.title') }}</p>
+        <div class="form-edit-section-body">
+          <form-upload @success="afterUpload" @cancel="setUploadSectionState(false)"/>
+        </div>
+      </div>
       <div id="form-edit-def-container">
         <div>
           <i18n-t keypath="versionName">
@@ -28,7 +34,7 @@ except according to the terms contained in the LICENSE file.
           <form-version-def-dropdown :version="formDraft" outlined
             @view-xml="viewXml.show()"/>
           <button id="form-edit-upload-button" type="button"
-            class="btn btn-primary" @click="$emit('upload')">
+            class="btn btn-primary" @click="setUploadSectionState(true)">
             <span class="icon-upload"></span>{{ $t('action.upload') }}
           </button>
         </div>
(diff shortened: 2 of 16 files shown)


src/components/form/new-page.vue
sadiqkhoja, line 1: I really wanted to name this file as `new.vue` but was having trouble with transifexKey :(

matthew-white: Let me know if you want to troubleshoot `transifexKey` together! FWIW recently I've been more likely to use `create` (a verb) rather than `new`.

src/components/form/edit/def.vue
```diff
@@ -15,6 +15,12 @@ except according to the terms contained in the LICENSE file.
     <template #subtitle>{{ $t('subtitle') }}</template>
     <template v-if="changed" #tag>{{ $t('changed') }}</template>
     <template #body>
+      <div v-if="uploadSectionState">
+        <p class="form-edit-section-title">{{ $t('uploadSection.title') }}</p>
+        <div class="form-edit-section-body">
```
matthew-white, line 20: I'm not sure how I feel about `form-edit-section-title` and `form-edit-section-body` being used here. I feel like it's better for them to stay local to the `FormEditSection` component.

- If this content is going to change soon anyway such that these classes will naturally go away, no need to worry about it.
- If the plan is for the classes to stay long-term: is the structure here intended to be nested? How about rendering a nested `FormEditSection` component instead?

src/components/form/edit.vue
```diff
@@ -173,6 +171,12 @@ const afterAbandon = () => {
 <style lang="scss">
 #form-edit {
   padding-inline: 10px;
+
+  #form-upload {
+    p {
+      max-width: unset;
+    }
+  }
```
matthew-white: How about moving this CSS to `FormEditDef`, since that's the component that renders `#form-upload`?

src/components/form/new-page.vue
```diff
@@ -0,0 +1,70 @@
...
+  await router.push(projectPath());
+};
+
+</script>
+
+<style lang="scss">
+#form-new {
+  p {
+    max-width: unset;
+  }
```
matthew-white: I remember seeing this in `FormEdit` as well. If `FormUpload` is always supposed to apply this style, how about moving the style there?

src/routes.js
matthew-white, line 1: Given that this file has changed, I'd expect to see changes to test/router.spec.js as well. It'd be good to assert that:

- Only users who can `form.create` can access the page.
- The title of the browser tab is as expected.

You can see what tests there are for another project page, like .../forms or .../settings, and add similar ones for .../new-form.

test/components/form/new.spec.js
```diff
@@ -33,115 +33,50 @@ const xlsForm = () => new File([''], 'my_form.xlsx', {
...
-        const text = modal.get('.modal-introduction p').text();
-        text.should.startWith('To create a Form,');
-      });
-
-      it('renders the paragraph about form attachments', () => {
-        const modal = mount(FormNew, mountOptions());
-        const text = modal.findAll('.modal-introduction p')[1].text();
-        text.should.include('Form Attachments');
-      });
-    });
```
matthew-white, line 73: I feel like these tests are still useful. This text is dynamic, so it's good to test it, I think.

test/components/form/new.spec.js
```diff
@@ -33,115 +33,50 @@ const xlsForm = () => new File([''], 'my_form.xlsx', {
...
-        const modal = mount(FormNew, mountOptions());
-        const text = modal.get('.modal-introduction p').text();
-        text.should.startWith('To update the Draft,');
-      });
-
-      it('does not render the paragraph about form attachments', () => {
-        const modal = mount(FormNew, mountOptions());
-        modal.findAll('.modal-introduction p').length.should.equal(1);
-      });
-    });
```
matthew-white, line 105: Similar comment as above about these tests, I feel like they're worth keeping.

src/components/form/new-page.vue
```diff
@@ -0,0 +1,70 @@
...
+  const message = t('alert.create', {
+    name: form.name != null ? form.name : form.xmlFormId
+  });
+  project.forms += 1;
+  await router.push(formPath(form.projectId, form.xmlFormId, 'draft'));
+  alert.success(message);
+};
+
+const afterCancel = async () => {
+  await router.push(projectPath());
```
matthew-white, line 45: Is there a test of this cancel behavior? Before, the modal would be hidden, but you'd stay on the page. Now though, you're redirected to a different page. I'm thinking that we should add a test of that new behavior even though the code here is pretty straightforward.

src/components/form/edit.vue
```diff
@@ -23,14 +23,13 @@ except according to the terms contained in the LICENSE file.
...
       <template v-if="formDraft.isDefined()">
-        <form-edit-def @upload="uploadModal.show()"/>
+        <form-edit-def @after-upload="afterUpload"/>
         <form-draft-testing/>
         <form-edit-draft-controls @publish="publishModal.show()"
           @abandon="abandonModal.show()"/>
       </template>
     </template>
 
-    <form-new v-bind="uploadModal" @hide="uploadModal.hide()" @success="afterUpload"/>
```
matthew-white, line 33: Is `uploadModal` not used for anything anymore? Could it be removed?

test/components/form/upload.spec.js
```diff
@@ -233,23 +168,23 @@ describe('FormNew', () => {
...
-      .mount(FormNew, mountOptions())
+      .mount(FormUpload, mountOptions())
       .testStandardButton({
-        button: '#form-new-upload-button',
+        button: '#upload-button',
         request: upload,
-        disabled: ['.modal-warnings .btn-primary', '.btn-link'],
-        modal: true
+        disabled: ['.form-upload-warnings .btn-primary', '.btn-link'],
+        component: true
```
matthew-white: I don't think `component: true` is an option for `testStandardButton()`, we can just remove `modal: true`.

test/components/form/upload.spec.js
```diff
@@ -666,29 +598,29 @@ describe('FormNew', () => {
...
         .request(upload)
         .respondWithProblem(xlsFormWarning)
         .complete()
         .testStandardButton({
-          button: '.modal-warnings .btn-primary',
-          disabled: ['#form-new-upload-button', '.btn-link'],
-          modal: true
+          button: '.form-upload-warnings .btn-primary',
+          disabled: ['#upload-button', '.btn-link'],
+          component: true
```
matthew-white: Similar comment as above, I think `component: true` can be removed here.

matthew-white (approved):
Looks great!

src/components/form/new-page.vue
matthew-white: OK, I think I understand the problem. `FormUpload` specifies `@transifexKey component.FormNew`, which means "move this entire object of messages" to the path `component.FormNew`. But the file here also defines messages, which by default would live at `component.FormNew` if this file were named `new.vue`. So there'd end up being two different objects being told to live at `component.FormNew`, which `@transifexKey` doesn't allow. `@transifexKey` will let you move/rekey an object of messages, but it won't merge it with another object at the same path.

In the end, I think I'm glad that this file gets a different name from `new.vue`. I feel like the diff and Git history of `new.vue` would have been confusing, since `new.vue` already existed. Right now, the diff is very nice and shows how `new.vue` has been moved to `upload.vue`. I don't think that would've happened if `new.vue` had been repurposed for the new form page. `upload.vue` would have shown up as an entirely new file.

src/components/form/edit/def.vue
sadiqkhoja: I added this block of code to keep the app functioning, this will go away.

test/components/form/upload.spec.js
sadiqkhoja: oh; find and replace all :(

PR #1536, opened 2026-04-25
router/canRoute(): handle missing validateData

### LINKED ISSUE #1850: [frontend] /account/ crashes
# Reproduce

1. log into https://test.getodk.cloud/
2. go to https://test.getodk.cloud/account/

# Observed

Page loads completely blank.

Checking the console logs shows:

```
TypeError: can't access property "every", e.meta.validateData is undefined
    Xu index-CRWN4AH-.js:formatted:2469
    Dv index-CRWN4AH-.js:formatted:20364
    wn index-CRWN4AH-.js:formatted:1213
    wn index-CRWN4AH-.js:formatted:1196
    Yn index-CRWN4AH-.js:formatted:1872
[index-CRWN4AH-.js:formatted:1781:65](https://test.getodk.cloud/assets/index-CRWN4AH-.js:formatted)
    bt index-CRWN4AH-.js:formatted:1781
    V index-CRWN4AH-.js:formatted:1639
Uncaught (in promise) TypeError: can't access property "every", e.meta.validateData is undefined
    Xu index-CRWN4AH-.js:formatted:2469
    Dv index-CRWN4AH-.js:formatted:20364
    wn index-CRWN4AH-.js:formatted:1213
    wn index-CRWN4AH-.js:formatted:1196
    Yn index-CRWN4AH-.js:formatted:1872
```

alxndrsn:
Closes getodk/central#1850

A new unit test.  Previously there were zero tests for `canRoute()`.

I'm not sure _why_ `validateData` is `undefined`.

No obvious risk of regression.

No.


matthew-white:
I have thoughts about a possible alternative approach to take, I'll post about that soon!

matthew-white:
> I have thoughts about a possible alternative approach to take, I'll post about that soon!

OK, not so soon after all, sorry for the delay!

I've pushed the alternative approach that I had in mind at #1703. Here's my thinking:

- The error message mentions `validateData`, yet it points to a more general problem: missing route meta fields.
- Route meta fields are defined in src/routes.js and are expected to exist on every route that the user is able to navigate to. They're only defined on bottom-level routes (routes that are not themselves a parent route). That's in part because it's never expected that users navigate directly to a parent route (e.g., /account instead of /account/edit). Honestly, I didn't even realize that was possible, so I'm glad you filed this issue.
- The approach here checks whether `to.meta.validateData` is falsy. However, I think the issue of missing route meta fields extends beyond the `validateData` mechanism. Many navigation guards in router.js make assumptions about the structure and default values of meta fields. That's why I think it's a good idea to solve this issue at a higher level — in a separate navigation guard that runs very early, rather than in `canRoute()`.
- routes.js defines default values for route meta fields, including `validateData`. So it's never expected that `to.meta.validateData` would be falsy. I think `canRoute()` can assume that `to.meta.validateData` is well-defined as long as we check the existence of route meta fields in general. Or another idea is that `canRoute()` could throw if it sees that `to.meta.validateData` is falsy. I'm less excited about the idea of `canRoute()` returning a normal value after encountering an invalid value of `to.meta.validateData`.

> A new unit test.  Previously there were zero tests for `canRoute()`.

Most router things are tested in test/router.spec.js, including `validateData`. I'd be open to unit tests as well, but I feel like there's good test coverage in test/router.spec.js.

alxndrsn:
@matthew-white thanks - great explanation.  Redirecting to `/` when an unknown route is accessed (https://github.com/getodk/central-frontend/pull/1703) looks like a much higher-level fix than this PR :+1:

PR #1540, opened 2026-05-03
chore: convert repo into monorepo

garethbowen:
NB: This is not ready to merge - there are other places where the directory structure is referenced (such as .md files) which I haven't yet changed, pending confirmation that this is the structure we want to go with.

The main change is moving app code into `/apps/central/`. This paves the way for the web-forms vue app to be added in `/apps/web-forms/`, and the rest of the web-forms packages to be moved to `/packages/*`.

An open question is how many of the other files in the root should also be moved to `/apps/central/`. Until web-forms is merged I won't be 100% confident that I've got the split right, but I don't have any concerns about moving files in a later PR if I didn't go far enough. In general if a file _could_ conceivably be shared between apps I've left it in root to make that more likely to happen.

The directory structure I'm proposing is...

```

apps/
-- central/
-- web-forms/ (the vue app)
packages/
-- common/
-- scenario/
-- web-forms/ (the vue component)
-- xforms-engine/
-- xpath/
package.json
...
```


apps/central/karma.conf.js
latin-panda, line 1: Would it make sense to move this into `apps/central`? The other packages (xforms-engine, xpath, etc.) are standardizing on the Vite/Vitest stack, so Central is the only place where Karma is still active. I'm thinking of keeping the root as clean as possible while Central eventually moves to Vite/Vitest.

package.json
```diff
@@ -6,21 +6,24 @@
...
-    "transifex:fix": "node bin/transifex/restructure.js > transifex/strings_en.json",
-    "transifex:lint": "bash -c 'diff transifex/strings_en.json <(node bin/transifex/restructure.js)'"
+    "transifex:fix": "node bin/transifex/restructure.js > apps/central/transifex/strings_en.json",
+    "transifex:lint": "bash -c 'diff apps/central/transifex/strings_en.json <(node bin/transifex/restructure.js)'"
   },
   "volta": {
     "node": "24.13.0"
   },
+  "workspaces": [
+    "apps/*"
```
latin-panda, line 22: Just a comment supporting the proposed structure:

I’m a huge fan of this `apps/` split. From a product management perspective, it decouples our two core personas (the Data Collector and the Form Builder). This gives us the flexibility to keep developing specialized roadmaps for each without one app’s constraints slowing down the other. 

Having a top-level `packages/` makes a lot of sense. It helps encourage code sharing while keeping the frontends independent and agile.

package.json
```diff
@@ -6,21 +6,24 @@
...
     "test:e2e": "e2e-tests/run-tests.sh",
-    "transifex:fix": "node bin/transifex/restructure.js > transifex/strings_en.json",
-    "transifex:lint": "bash -c 'diff transifex/strings_en.json <(node bin/transifex/restructure.js)'"
+    "transifex:fix": "node bin/transifex/restructure.js > apps/central/transifex/strings_en.json",
+    "transifex:lint": "bash -c 'diff apps/central/transifex/strings_en.json <(node bin/transifex/restructure.js)'"
   },
   "volta": {
     "node": "24.13.0"
   },
+  "workspaces": [
```
latin-panda, line 21: This is an excellent chance to clarify the distinction between the web-forms app and the web-forms vue-client. We often encounter situations where we want to differentiate these in our internal communication. Instead of this: 
```
apps/
  -- web-forms/ (the vue app)
packages/
  -- web-forms/ (the vue component)
  -- xforms-engine/
```

What about this?

```
apps/
  -- web-forms/ (the vue app)
packages/
  -- xforms-client/ <--- rename the web-forms vue-client
  -- xforms-engine/
```

latin-panda (changes requested):
I added a comment inline to open `threads` of conversations to avoid a large main timeline conversation.

package.json
garethbowen: I like that naming! I wonder if we should go one step further and rename the npm module too, ie: `@getodk/xforms-client`? I think it'd be confusing if the import and the folder were named differently...

Plus it'd be much better to rename before 1.0 than after!

apps/central/karma.conf.js
garethbowen: Good call. Done.

garethbowen:
@matthew-white @latin-panda I've updated all the paths I could find and made the change that Jenny suggested so I think this is ready for a full review. Thanks for volunteering!

index.html
matthew-white, line 1: Should this go in apps/central/? My assumption is that Central Frontend and the Web Forms Vue app will have two different index.html files (though maybe that's not true).

package.json
matthew-white, line 1: I'm not super knowledgable about how monorepos work, but won't there end up being multiple package.json files? Would it make sense to move the Central-specific part of this file (Central-specific dependencies, Central-specific commands like `transifex:fix`) into apps/central/?

apps/central/docs/CONTRIBUTING.md
matthew-white, line 1: I think I'd like to move this file into apps/central/. More than anything else, this file is an overview of the Central Frontend codebase — a guide for developers (internal or external) who want to contribute to Central Frontend.

I'm thinking that different subdirectories of the monorepo could each have its own CONTRIBUTING.md (and/or README.md). Maybe we also add an overarching CONTRIBUTING.md.

If we were to move this file to apps/central/, I think we would probably also move docs/img/icomoon*.png, which are referenced in CONTRIBUTING.md and are only relevant to Central.

.eslintrc.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
     curly: 'off',
     'implicit-arrow-linebreak': 'off',
     'import/first': 'off',
+    "import/no-extraneous-dependencies": ['error', {'devDependencies': ['**/test/**', 'vite.config.js']}],
```
matthew-white, line 25: Why is this change needed out of curiosity?

apps/central/bin/util/util.js
matthew-white, line 1: Does Web Forms have existing bin/ files? If so, is it desirable for Web Forms to share bin/util files with Central Frontend? I'm wondering: how about moving the .js files in bin/ to apps/central/bin/? I.e., how about each app get its own set of bin (scripting) files for now? We can always standardize later, but I feel like centralizing all bin files at the top level may require more immediate standardization.

It's definitely the case that some files in bin/ make sense to live at the repo-level, e.g., check-for-large-files.sh. But most of the files, especially all the .js files, are Central-specific. I think we could have both a top-level bin/ directory and app-specific bin/ directories.

In particular, the .js files in bin/ (including this util file) are just for the Vue I18n ↔ Transifex conversion, which isn't relevant to Web Forms.

apps/central/karma.conf.js
```diff
@@ -53,28 +53,29 @@ webpackConfigForKarma.module.rules.push({
 
 module.exports = (config) => {
   config.set({
+    basePath: '../../',
```
matthew-white, line 56: Why is this line needed out of curiosity? Because index.html lives in the top-level directory of the repository? Or because `npm run test` sets the working directory to the top-level directory?

I see a lot of `apps/central/` being added to this file, which made me wonder: if we didn't have `basePath: '../../'`, would we not need `apps/central/`?

package.json
latin-panda: Yes, I agree with the rename. It makes sense.

latin-panda (approved):
It looks good to me! Thanks!

index.html
matthew-white, line 1: Another question possibly related to the one above about index.html: should public/ also be moved into apps/central/? There are files in there that are only relevant to Central:

- blank.html
- fonts/icomoon.*

.eslintrc.js
matthew-white, line 1: How similar are the linting rules for Central Frontend vs. Web Forms? I'm thinking that many of these rules probably belong in apps/central/. How about moving this file to apps/central/, then adding a super generic .eslintrc.js to the top-level directory?

apps/central/src/assets/css/icomoon.css
matthew-white, line 1: (This comment isn't really relevant to this file, I just wanted to create a comment thread.)

How about moving icomoon.json into apps/central/? It's just a data file that we upload to the IcoMoon app when we use it. It doesn't drive behavior anywhere like in an IDE or CLI.

In general, my vision going forward is that the top-level directory of the monorepo should be as decluttered as possible, just to make it easier to understand the structure of the repo as a whole. How does that sound?

matthew-white (changes requested):
I'm leaving some comments with specific questions, but overall I'm really liking the new structure. The structure that's envisioned of apps/ and packages/ makes a lot of sense to me. :+1:

In terms of code review strategy, I support the idea of "this first PR doesn't have to get everything right; we can continue moving things later." That said, in some of my comments, there were places where it seemed reasonably clear to me that we would want to move the file into apps/central/.

index.html
garethbowen: That's true once the web-view app is created. Until then there will only be one index.html just like now.

I looked at making this change because you're right that it's cleaner, but doing that requires changing the central repo build-frontend script.

When we make a new web-forms app we'll also need to update the build-frontend script and it's not 100% clear to me what structure this will take until we do it, so rather than committing to a structure now I'd prefer to put this change off until we have a second app, and do it then. Does that work for you?

garethbowen: I agree ultimately that will be cleaner. As above, I'd prefer to put this off so it doesn't destabalise the central building process.

apps/central/docs/CONTRIBUTING.md
garethbowen: My plan was to review when web-forms lands, pull out the common pieces between the docs, and then link to the app specific contributing docs.

> a guide for developers (internal or external) who want to contribute to Central Frontend.

Developers who want to contribute to CF will need have the same dev environment set up for those who want to contribute to WF. If a CF contributor wants to, for example, test that editing forms works, this will require the WF app to be available either on a specific path or port. This differs from how it works right now with WF  installed from npm and built in to the CF app. Over time I anticipate there will be more cases where CF invokes the WF app, for example, there is a feature being designed for the draft form page to include a form preview using WF. For these sorts of uses cases WF and CF devs will have to be able to build both to have a working system, and therefore most of the contributing doc should be the same.

Is this more overlap than you were thinking?

.eslintrc.js
garethbowen: Because the base airbnb config only specifies test dirs at the root level: https://github.com/airbnb/javascript/blob/8ed19247bef145e3cc41b24a02dd1fe6d9b5e681/packages/eslint-config-airbnb-base/rules/imports.js#L71

apps/central/bin/util/util.js
garethbowen: > Does Web Forms have existing bin/ files?

Yes but it's minimal, and should be pushed down into a package.

> how about each app get its own set of bin (scripting) files for now? 

I think you're right. I think all the files except the transifex ones are useful. I'll move the transifex (and util) ones into /apps/central/bin

apps/central/karma.conf.js
garethbowen: I think once the public files are moved into apps/central that may be true. Right now karma can't find them unless basepath is set to root.

.eslintrc.js
garethbowen: That'll be necessary, for TS if nothing else. I was planning to leave this here until after the merge, and then it'll be obvious how many rules start failing. Then the specific rules can be pushed down, extended, or overwritten as appropriate.

I guess I'm defaulting to reuse and overwrite where necessary, rather than segregate and standardising some time later.

It'll become immediately clear when the repos are merged if I'm being overly optimistic!

apps/central/src/assets/css/icomoon.css
garethbowen: Where things only make sense in one package/app I agree (icomoon is a good example) - push them down and declutter the root. Where things are conceptually reusable I prefer to keep them in the root to encourage reuse, even if they're not be used in multiple packages yet.

I'll move the icomoon.json file down now.

package.json
garethbowen: The way I'd structure it once everything's said and done...

- `/package.json` has top level scripts and all dev dependencies. If you run, for example, `npm run build` it will build all packages and apps by invoking their individual `build` scripts in order. You might have a `transifex:lint` command at the top level which checks the translations in CF and WF in completely different ways.
- `/package-lock.json` - the one and only lock file in the repo
- `/[apps|packages]/*/package.json` contains the (non-dev) dependencies and scripts for this specific package, for example, `test` would invoke karma in cf and vitest in wf. Devs can run the scripts in a single workspace using the -w flag, eg: `npm run test -w=apps/central` from the root.

In terms of how we get there I was trying to disrupt cf development as little as possible. I could split the cf package.json into two as part of this PR, OR could wait until web-forms lands. Which do you prefer?

apps/central/docs/CONTRIBUTING.md
matthew-white: > Developers who want to contribute to CF will need have the same dev environment set up for those who want to contribute to WF.

That's true, but setup is covered by README.md, not CONTRIBUTING.md. I don't see much in CONTRIBUTING.md that's relevant to WF. Maybe the main thing would be the section about E2E Tests at the bottom.

Here's what I'd propose:

- Leave README.md where it is for now.
- Review README.md when web-forms lands.
- Move CONTRIBUTING.md to apps/central/docs/CONTRIBUTING.md right away.
  - I've been wanting to move CONTRIBUTING.md into a docs/ subdirectory anyway in order to declutter the top-level directory of CF.
- Once Web Forms has e2e tests, move the section about e2e tests in CONTRIBUTING.md to a shared .md file.

How does that sound?

This doesn't need to block the PR though. We could also move CONTRIBUTING.md once this PR is merged (and I'd be happy to do that).

.eslintrc.js
matthew-white: Couple of follow-up questions:

- Why does vite.config.js need to be in the list now? It doesn't seem to be in the airbnb list.
- Other files that are in the airbnb list aren't in the new list anymore, e.g., karma.conf.js and vue.config.js. I guess they don't need to be? I guess it's good for the list to be as minimal as possible.
- I don't really understand why this rule would be relevant to `/test/`, but not `/e2e-tests/`. It's not like `e2e-tests` was in the airbnb list before, so I'm not necessarily suggesting that we add it to the list now. But I guess I don't really understand how this rule is working. Happy to leave it a mystery, but if you know the answer, I'd be interested to learn more.

matthew-white: > I don't really understand why this rule would be relevant to `/test/`, but not `/e2e-tests/`.

Ah there's a separate e2e-tests/.eslintrc.js that specifies this rule there. I think that's why `e2e-tests` doesn't need to be in the new list.

index.html
matthew-white: I guess it's true currently that there's only one index.html, which renders both WF and the rest of CF. That sounds reasonable to move it once we have the Web Forms Vue app. :+1:

package.json
matthew-white: Thank you for the info, it's very helpful! That sounds like a good setup to me.

I am curious about how package.json will be split up exactly, but it doesn't need to happen today. :+1: Once web-forms lands, would you mind tagging me on the PR that ultimately splits up package.json?

apps/central/src/assets/css/icomoon.css
matthew-white: > Where things are conceptually reusable I prefer to keep them in the root to encourage reuse

That sounds great to me. :+1: I also like the idea of packages/common/ a lot.

.eslintrc.js
matthew-white: I do like the idea of aligning our lint rules (not all at once, but to the extent feasible). It feels like a relatively straightforward way to make the CF and WF codebases more alike.

package.json
garethbowen: > Once web-forms lands, would you mind tagging me on the PR that ultimately splits up package.json?

Absolutely. As soon as WF lands I'll get straight into a whole lot of merging and cleanup to minimise the time when it's messy. And I'd be happy to tag you on it!

apps/central/docs/CONTRIBUTING.md
garethbowen: Done!

apps/central/bin/util/util.js
matthew-white: Would you mind copying bin/.eslintrc.js to apps/central/bin/ as well?

matthew-white: Copying, not moving, since there will end up being .js files in both bin/ directories.

garethbowen: Nice catch! I also added `/apps/*/bin/` to the eslint command so those files are being linted again.

matthew-white (approved):
Looks good to me!

- I didn't review specific file moves. Instead, I checked out the branch and took a look at the resulting files. The directory structure looks good to me, as well as the list of files that are still in the top-level directory of the repository.
- I did review files whose content changed (many of which just involved adding `apps/central/`).
- I didn't search for broken references or more subtle ways that code might have made assumptions about the structure of the repo. In other words, my review was focused on things that changed, not things that we might have missed changing.
  - Yet on that point, I'm reassured that CI is passing. We have several verification steps in CI at this point, including linting code, checking Transifex files, running Vue component tests, e2e tests, etc.
  - I also think that if we did miss anything, we can always fix it later.

I still have a couple of open questions at https://github.com/getodk/central-frontend/pull/1540#discussion_r3197010485, but nothing terribly important / certainly nothing that needs to block merge. LGTM!

.eslintrc.js
garethbowen: vite.config.js was coming from here: https://github.com/vuejs/eslint-config-airbnb/blob/main/packages/eslint-config-airbnb/rules/imports.js#L48

It would be slightly cleaner to import from that package and then extend it with `**/test/**`. What do you think?

PR #1542, opened 2026-05-07
chore: merge web-forms into central-frontend

garethbowen:
This PR merges the all the web-forms code into central-frontend.
The code is completely ignored by existing scripts, linting, etc so it doesn't actually DO anything.
The next step will be having central-frontend actually use the inline code instead of the npm packages, and getting the web-forms CI up and running.

Run everything locally.

Discussed with the team at length....

Users shouldn't be impacted at all.

No.


bin/check-for-large-files.sh
```diff
@@ -2,7 +2,7 @@
 set -o pipefail
 shopt -s inherit_errexit || true
 
-git ls-files -z \
+git ls-files -z -- . ':!:packages/common/*' \
```
latin-panda: I only found these test files that break this check:
- whova_form.xml
- wards.xml

What do you think about moving them to a folder and skipping only that folder, instead of skipping the entire package?

packages/web-forms/bin/README.md
```diff
@@ -0,0 +1,36 @@
+# Scripts
```
latin-panda, line 1: So now, the "scripts" are in "bin". Can script/merge-translations.js also be moved to "bin" for the time being?

latin-panda (commented):
I didn't quite remember how many files WF has 😅 I've left a couple of comments below.

bin/check-for-large-files.sh
matthew-white: I agree that it'd be a good idea to make any exceptions here as narrow as possible. :+1: We've accidentally committed large files before, and they live in Git history forever. 🫠

How about specifying exceptions for those two specific files? Not even moving them into a folder, just skipping their current file paths specifically?

I'm also happy with @latin-panda's idea of moving them into a particular folder, I just wanted to throw out a second idea.

packages/common/src/constants/datetime.ts
matthew-white, line 1: I didn't realize that packages/common/ would already exist with this PR. I was thinking that it would be added in the future to hold files shared between Central and Web Forms. Right now, it seems more so a directory of files shared among Web Forms packages. Maybe in theory, we could start adding files shared between Central and Web Forms here, but I think we'd have a better chance of success by creating a fresh directory for that. Just the fact that all the files here are TypeScript is a pretty big shift for Central.

I'd propose that we rename this directory to `web-forms-common` or `xforms-common`. That feels like a more accurate description of what this directory is. It gives the Web Forms team complete control over the structure and setup of this directory, whereas I'd have opinions about a true common directory. In the future, we can add a separate `common` directory for shared files between Central and Web Forms. When that day comes, we can discuss things like directory structure, JS vs. TypeScript, etc.

packages/scenario/benchmarks/placeholder.bench.ts
matthew-white, line 1: What is the purpose of packages/scenario/? Is it specific to Web Forms? How about naming it `web-forms-scenario` or `xforms-scenario`?

matthew-white (changes requested):

garethbowen:
@matthew-white @latin-panda The recommended way to review this is just check the commits after the "Merge remote-tracking branch 'web-forms'" commit - everything else is literally identical as in web-forms. Sorry, that should have been part of the original instructions!

Other questions from Slack...

> How does the PR modify shared files?

The only shared files being modified are to get the build to pass and these changes are in the most recent commits.

> How does the PR modify the overall directory structure?

It adds the `/packages` directory which has all of web-forms code in it. It moves web-forms `/scripts` directory to `/packages/web-forms/bin/`. It adds a `/vendor-types` directory to the root to help build web-forms TS when the time comes.

packages/common/src/constants/datetime.ts
garethbowen: That makes sense to me. However, I'm trying to keep this PR very lean so reduce the chances of something going wrong. Renaming the common package would require changing a fair amount of source code which isn't currently CIed and can't be squashed. Would you comfortable with doing this move as part of the next PR which will get web-forms CI running and actually in use in CF?

packages/scenario/benchmarks/placeholder.bench.ts
garethbowen: Yes. As above, are you comfortable with me doing this post merge?

I'm also not convinced it needs to be a separate package. It's engine level integration tests so maybe it should just be in the `/packages/xforms-engine/test` dir. I'll follow up with the team to understand the logic about why it's separate, but potentially it was designed to be useful to test other engines (ie: collect)...

bin/check-for-large-files.sh
garethbowen: Done.

packages/web-forms/bin/README.md
garethbowen: Nice catch! Done.

packages/common/src/constants/datetime.ts
matthew-white: Sounds good, renaming it after this PR makes a lot of sense. :+1:

matthew-white: And no big rush in renaming it either, like if you want to get web-forms CI set up first, that sounds good too. It'd just be nice to rename it at some point soonish.

packages/scenario/benchmarks/placeholder.bench.ts
matthew-white: Doing it post-merge makes sense to me. :+1:

matthew-white (approved):
I checked the changes to check-for-large-files.sh and .image-hashes, which look good. The rest of the changes were to packages/ and vendor-types/, which don't currently affect Central.

packages/common/src/constants/datetime.ts
garethbowen: I'll make a note to do it in the same PR that gets CI running so I can verify the changes work and get it done ASAP. Thanks.

PR #1543, opened 2026-05-08
ci: consolidate GitHub Actions workflows

alxndrsn:
Move the existing GitHub Actions `build.yml` workflow into the `test.yml` workflow.

1. the current division seems artificial, and doesn't provide obvious benefit
2. "build" is being run as a means to test:
    1. large files in repo
    3. bundle size
    4. the build itself works
3. having these jobs within the same workflow will be helpful when building release artefacts as part of the CI process

### Old flow

<kbd>
Screenshot_2026-05-08_09-40-23

</kbd>

### New flow

<kbd>
Screenshot_2026-05-08_09-39-59

</kbd>

- [x] CI for this PR: https://github.com/getodk/central-frontend/actions/runs/25540863589?pr=1543

1. the current division seems artificial, and doesn't provide obvious benefit
2. "build" is being run as a means to test:
    1. large files in repo
    3. bundle size
    5. the build itself works
3. having these jobs within the same workflow will be helpful when building release artefacts as part of the CI process

No effect - just a build change.

No.


.github/workflows/tests.yml
```diff
@@ -31,8 +31,22 @@ jobs:
...
+  build:
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-node@v6
+      with:
+        node-version-file: package.json
+        cache: 'npm'
+    - run: ./bin/check-for-large-files.sh
```
garethbowen, line 43: This is in the wrong place (but it was before too).

My preferred approach is to have all forms of linting in one "job", ie: `./bin/check-for-large-files.sh`, `./bin/check-scripts.sh`, and `npm run lint`. I might even combine those into a single `npm` command so it's easy to check everything precommit too. Then I would have all jobs depend on the lint job so it fails fast and doesn't spawn long running jobs.

But, what you've done is a step in the right direction, so this isn't a blocker.

garethbowen (approved):
Definite improvement!

Optional suggestion inline...

.github/workflows/tests.yml
alxndrsn: Agreed.  One thing to note here is that unlike `check-for-large-files`, the later `check-bundle-size` has to be run post-build.

alxndrsn: Something like https://github.com/getodk/central-frontend/pull/1548?

matthew-white: > My preferred approach is to have all forms of linting in one "job", ie: `./bin/check-for-large-files.sh`, `./bin/check-scripts.sh`, and `npm run lint`. I might even combine those into a single `npm` command so it's easy to check everything precommit too.

Related: getodk/central#1561

garethbowen: Yes! Perfect.

PR #1544, opened 2026-05-08
ci: upload dist bundle to GitHub release

alxndrsn:

Tested on a forked repo:

- [x] normal push: https://github.com/alxndrsn/odk-central-frontend/actions/runs/25543174024
- [x] tagged push: https://github.com/alxndrsn/odk-central-frontend/actions/runs/25543176439

Pre-building the release distro allows skipping the `intermediate` build container in `central`'s `nginx` image build at https://github.com/getodk/central/blob/625eb988aecde0a2b30f69847a21aba3b86b7cb1/nginx.dockerfile#L1-L12

This can be replaced with e.g. `curl -L https://github.com/getodk/central-frontend/releases/download/v1.2.3/dist.tar.gz | tar -xzf -`.

On my machine this is ~32 seconds of the 34 second build process.

This also avoids pulling and executing significant amounts of NodeJS code & dependencies into prod machines.

It should not affect the final image size of central's `nginx` image.


.github/workflows/tests.yml
```diff
@@ -31,56 +35,31 @@ jobs:
...
-      if: failure()
-      uses: actions/upload-artifact@v4
-      with:
-        name: Playwright Artifacts
-        path: central/client/test-results
+    - run: ./bin/check-for-large-files.sh
+    - run: npm ci
+    - run: npm run build
+    - run: ./bin/check-bundle-size.js
+
```
garethbowen: Obviously will need to add back the e2e testing before the release. Were you thinking of just inlining the e2e testing lines into here so they run before publish all as the same job? But then how would the service fetch the frontend zip?

I think you probably need to have some path in the dockerfile which can take this zip as a environment variable so we can just pipe it in without publishing.

garethbowen (changes requested):
Love the automatic publishing!

Obviously we don't want to drop the e2e testing altogether - what were you thinking about doing with it?

alxndrsn:
> Obviously we don't want to drop the e2e testing altogether - what were you thinking about doing with it?

:facepalm: thanks for catching those.  Removed to save time while testing workflows!  Now re-added.

> Were you thinking of just inlining the e2e testing lines into here so they run before publish all as the same job?

The way the e2e tests are currently setup, they should probably _not_ be part of the main workflow - because the `odk-central-backend` version isn't pinned, there's no guarantee (or even expectation) that the e2e tests will pass for any given frontend commit.

.github/workflows/tests.yml
```diff
@@ -44,6 +48,15 @@ jobs:
...
     - run: npm run build
     - run: ./bin/check-bundle-size.js
+    - name: Upload Release Artefact
+      if: github.ref_type == 'tag'
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        tar -cvzf dist.tar.gz dist/
+        echo "Uploading dist bundle..."
+        gh release create ${{ github.ref_name }} dist.tar.gz --generate-notes
```
matthew-white: Usually right now, we create the GitHub release through the UI, which creates the Git tag at the same time. How will that change with this RP? Going forward, will we just create the Git tag, then this process will create the GitHub release automatically?

About `--generate-notes` specifically, I'm open to it, but I feel like we already do a good job of writing release notes summarizing user-facing changes. Usually we just link to the release in the main `central` repo. E.g., the latest Frontend release [v2026.1.2](https://github.com/getodk/central-frontend/releases/tag/v2026.1.2) just has these release notes linking back to the `central` repo:

> See release notes at https://github.com/getodk/central/releases/tag/v2026.1.2

matthew-white (commented):
The automatic publishing makes a lot of sense to me too!

matthew-white:
> because the `odk-central-backend` version isn't pinned, there's no guarantee (or even expectation) that the e2e tests will pass for any given frontend commit.

Yes, that's true unfortunately. Often the e2e tests would do fine with the latest stable version of Backend. It's not like we often change the Backend APIs that the Frontend e2e tests rely on. But sometimes the e2e tests do require API behavior that's been introduced in the current release under development.

Maybe we could introduce logic to the e2e tests along the lines of "check out the latest tagged version of Backend, unless a pinned commit has been specified that's later than the tagged version." Just an idea, and not really relevant to this PR, just wanted to throw it out there.

.github/workflows/tests.yml
alxndrsn: e2e tests were removed to speed up work on the PR - should have been re-added before asking for review

alxndrsn: Now reverted.

.github/workflows/tests.yml
alxndrsn: > Usually right now, we create the GitHub release through the UI, which creates the Git tag at the same time.

If that's a comfortable workflow, then this line can be changed to uploading the tar to the existing release :+1:

alxndrsn:
```suggestion
        gh release create ${{ github.ref_name }} dist.tar.gz
```

alxndrsn: > we already do a good job of writing release notes summarizing user-facing changes

:+1: `--generate-notes` removed

.github/workflows/tests.yml
```diff
@@ -44,6 +48,15 @@ jobs:
...
     - run: npm run build
     - run: ./bin/check-bundle-size.js
+    - name: Upload Release Artefact
+      if: github.ref_type == 'tag'
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        tar -cvzf dist.tar.gz dist/
+        echo "Uploading dist bundle..."
+        gh release create ${{ github.ref_name }} dist.tar.gz
```
alxndrsn: TODO:

1. test if this works when the release is already built;
    2. if so comment to reflect that;
    3. if not, fix

.github/workflows/tests.yml
alxndrsn: Changed from `gh release create $ref_name asset` to `gh release upload $ref_name asset`.

alxndrsn: Example release created via the GUI: https://github.com/alxndrsn/odk-central-frontend/releases/tag/v0.1.1-asdf123
Build: https://github.com/alxndrsn/odk-central-frontend/actions/runs/26097338123

.github/workflows/tests.yml
alxndrsn: Sorted.

.github/workflows/tests.yml
```diff
@@ -39,6 +43,15 @@ jobs:
...
     - run: npm run build
     - run: ./bin/check-bundle-size.js
+    - name: Upload Release Artefact
+      if: github.ref_type == 'tag'
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        tar -cvzf dist.tar.gz dist/
+        echo "Uploading dist bundle..."
+        gh release upload ${{ github.ref_name }} dist.tar.gz
```
alxndrsn: Include version in filename?

garethbowen: Nice to have if it's easy enough to get.

garethbowen (approved):

.github/workflows/tests.yml
alxndrsn: Added.

Test build: https://github.com/alxndrsn/odk-central-frontend/actions/runs/27558631502/job/81464274495
Test release: https://github.com/alxndrsn/odk-central-frontend/releases/tag/v0.0.0-oidjfroifgjroijgfoijdg

PR #1545, opened 2026-05-08
Feature: show publishing notes on form versions page

### LINKED ISSUE #1834: Show publish notes on form history page.

sadiqkhoja:
Closes https://github.com/getodk/central/issues/1834

### DIFF
--- a/apps/central/src/components/form-version/row.vue
+++ b/apps/central/src/components/form-version/row.vue
@@ -26,8 +26,7 @@ except according to the terms contained in the LICENSE file.
         @view-xml="$emit('view-xml')"/>
     </td>
   </tr>
-  <tr v-if="version.publishNotes && project.dataExists && project.permits('form.update')"
-    class="form-version-notes-row">
+  <tr v-if="version.publishNotes" class="form-version-notes-row">
     <td colspan="3">
       {{ $t('notes', { notes: version.publishNotes }) }}
     </td>
@@ -39,10 +38,6 @@ import EnketoPreview from '../enketo/preview.vue';
 import FormVersionDefDropdown from './def-dropdown.vue';
 import FormVersionString from './string.vue';
 import TimeAndUser from '../time-and-user.vue';
-
-import { useRequestData } from '../../request-data';
-
-const { project } = useRequestData();
 
 defineOptions({
   name: 'FormVersionRow'
--- a/apps/central/src/components/form-version/row.vue
+++ b/apps/central/src/components/form-version/row.vue
@@ -26,10 +26,10 @@ except according to the terms contained in the LICENSE file.
         @view-xml="$emit('view-xml')"/>
     </td>
   </tr>
-  <tr v-if="version.publishingNotes && project.permits('form.update')"
+  <tr v-if="version.publishNotes && project.dataExists && project.permits('form.update')"
     class="form-version-notes-row">
     <td colspan="3">
-      {{ $t('notes') }}: {{ version.publishingNotes }}
+      {{ $t('notes', { notes: version.publishNotes }) }}
     </td>
   </tr>
 </template>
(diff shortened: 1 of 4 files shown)


apps/central/src/components/form-version/row.vue
```diff
@@ -26,6 +26,12 @@ except according to the terms contained in the LICENSE file.
         @view-xml="$emit('view-xml')"/>
     </td>
   </tr>
+  <tr v-if="version.publishingNotes && project.permits('form.update')"
+    class="form-version-notes-row">
+    <td colspan="3">
+      {{ $t('notes') }}: {{ version.publishingNotes }}
```
sadiqkhoja: I guess colon doesn't need to be localized.

matthew-white: I think it'd be a little better to localize it. Japanese doesn't use the same colon character as English. I think you can incorporate the colon in the i18n message, something like:

```json
"notes": "Notes: {notes}"
```

apps/central/src/components/form-version/row.vue
```diff
@@ -26,6 +26,12 @@ except according to the terms contained in the LICENSE file.
         @view-xml="$emit('view-xml')"/>
     </td>
   </tr>
+  <tr v-if="version.publishingNotes && project.permits('form.update')"
```
matthew-white: Is it guaranteed here that `project.dataExists`? From a quick look, it looked to me like it wasn't. In that case, it'd be good to check for `project.dataExists` before `project.permits()`. Otherwise, there could be an error depending on the order of responses (if `formVersions` is received before `project`).

Another idea is to add `v-if="project.dataExists"` to an ancestor element (the `<tbody>`?).

apps/central/src/components/form-version/row.vue
```diff
@@ -68,13 +78,20 @@ defineEmits(['view-xml']);
 
   * + .form-version-def-dropdown { margin-left: 5px; }
 }
+
+.form-version-notes-row td {
```
matthew-white: I feel like we need some additional CSS to account for the fact that notes can contain multiple paragraphs. Something so that newlines don't become spaces.

apps/central/src/components/form-version/row.vue
```diff
@@ -26,6 +26,12 @@ except according to the terms contained in the LICENSE file.
         @view-xml="$emit('view-xml')"/>
     </td>
   </tr>
+  <tr v-if="version.publishNotes && project.dataExists && project.permits('form.update')"
```
matthew-white: Do we even need to check `project.permits('form.update')` in the frontend? I feel like Backend is taking care of that for us. As long as we don't render the row when `publishNotes` is falsey, I think that's the only thing we really need to check, not `project.permits()`.

apps/central/test/components/form-version/row.spec.js
```diff
@@ -95,4 +95,37 @@ describe('FormVersionRow', () => {
...
+          rows.length.should.equal(1);
+        }
+      });
+    });
+
+    it('does not show notes when user lacks form.update permission', async () => {
+      mockLogin.reset();
+      mockLogin({ role: 'none' });
+      testData.extendedProjects.createPast(1, { role: 'viewer' });
+      testData.extendedForms.createPast(1, { publishNotes: 'Some notes' });
```
matthew-white: I think of `testData` as mocking API responses, not what's in the Backend database. With that in mind, I feel like this is mocking an API response that shouldn't be possible. If the user is a `viewer`, they won't receive `publishNotes` from the backend. I feel like we can probably go ahead and omit this test. In the other comment above, I suggested that we remove the check of `project.permits()` entirely, which would also make this test not needed.

matthew-white (approved):

PR #1548, opened 2026-05-11
ci: run all static checks together

alxndrsn:
* group all quick bash test/check scripts together
* run all scripts in same CI job

CI: https://github.com/getodk/central-frontend/actions/runs/25657421372?pr=1548

Could call each cehck individually, but the approach in this PR makes it simpler to:

1. run all locally, and
2. ensure new checks are run in CI

Could run static checks as part of the unit test step, but a failure in `static-checks` is likely to indicate a misconfiguration rather than a break in the application logic.  So it seems logical to run them separately & in parallel.

No effect - just CI.

No.


matthew-white:
I tagged @garethbowen for review, since there was previous discussion at https://github.com/getodk/central-frontend/pull/1543#discussion_r3215620872. However, I'm happy to review as well. I took a brief look at the PR and nothing stood out to me as being a potential issue.

garethbowen (changes requested):
Definitely a step in the right direction.

There's a file conflict (as of now) which will need to be fixed, and the required jobs config somewhere in GH will need to be changed so it no longer expects the old jobs, and now expects "static-checks".

I would go one step further though, and have a single npm script which runs all of these things, and then the CI can just call that. I think this will resolve #1561 

It does mean you'll have to do an `npm ci` here which will slow it down a little, but then you can also do `npm run lint` here, so it will be a comprehensive lint step.

Finally I would make the `unit-tests` job dependant on `static-checks` so it can fail fast.

Things are about to get a little wild in CI with web-forms coming in. WF currently has 42 jobs designed to run massively parallel and how this will fit with CF remains to be seen...

alxndrsn:
> I would go one step further though, and have a single npm script which runs all of these things, and then the CI can just call that. I think this will resolve #1561

I agree, but I'd rather do that as an additional step as it will inevitably get slowed down by operating system differences.

alxndrsn:
> Finally I would make the `unit-tests` job dependant on `static-checks` so it can fail fast.

I deliberately didn't do this, as for me it's useful to know if code is working even if e.g. a file is too large, or a new image has been added.  However, the `static-checks` job is very fast, so I can see value in that, and perhaps even in having a dependency pipeline of:

1. `static-checks` (approx 10 seconds)
2. `build` (approx 40 seconds)
3. `unit-tests` (approx 3.5 minutes)

Maybe a separate discussion?

alxndrsn:
> There's a file conflict (as of now) which will need to be fixed

Conflicts have been resolved.

> and the required jobs config somewhere in GH will need to be changed so it no longer expects the old jobs, and now expects "static-checks".

:+1: I think this is best done once the PR is merged.

garethbowen:
>  ...perhaps even in having a dependency pipeline of...

That is my ideal but I also fully support incremental improvements in this direction so feel free to leave that for a future PR/discussion.

> I think this is best done once the PR is merged.

I may be wrong but I don't think it _can_ be done after merge. The point of "required" build steps is it'll block merging. So once you're ready to merge this you'll need to change what's required, then merge before anyone else needs to merge. Then everyone else will rebase and be good to go.

bin/run-static-checks.sh
```diff
@@ -0,0 +1,13 @@
...
+set -o pipefail
+shopt -s inherit_errexit || true
+
+log() { echo "[$(basename "$0")] $*"; }
+
+log "Running all static checks..."
+
+for check in ./bin/static-checks/*; do
+  "$check"
+done
```
garethbowen: I would rather the static checks were explicitly called rather than just running everything in the dir. My concern is that it would be quite easy for a malicious party to sneak a script in to a PR and it wouldn't be obvious that it was going to be executed. If the CI run were approved then it would executed in our GH which has had a few CVEs lately (cache poisoning for example).

As there are only 3 scripts and they're not likely to change often I think it's worth being explicit.

... and then if you agree to that, maybe we don't need the separate `/static-checks` dir which will clean up this pr a bit, but that's up to you.

garethbowen (changes requested):

bin/run-static-checks.sh
alxndrsn: Done.

garethbowen (approved):
Beautiful!

PR #1549, opened 2026-05-13
chore: web-forms packages build and test

garethbowen:
This PR...

- Gets web-forms build, lint, test, etc working in central frontend in development
- Adds WF CI steps so changes in WF are built, linted, unit and e2e tested as before the merge
- Rearranges dependencies so WF and CF both work regardless of their differing dependencies and versions
- Reworks the npm scripts to work on all packages, for example, `lint`, `build`, `test` at the root level now execute on all apps and packages in turn

I've run all commands locally and in CI.

Discussed at length

This shouldn't impact users at all. The main change is the web-forms code is now built from source (/packages/*), rather than downloaded from npm, which is done at build time.

No.


matthew-white:
@garethbowen what do you think about splitting this PR up into multiple smaller ones? I'm thinking we could have one that moves Central things around (e.g., moving the Central .eslintrc.js and adding a Central-specific package.json), then a follow-up PR that adds Web Forms CI. Personally, I find smaller PRs a bit easier to review.

apps/central/transifex/strings_en.json
matthew-white, line 1: Just randomly noticed that this file seems to have changed a lot when I wouldn't expect it to.

garethbowen: Good spotting!

I've fixed this with an update to the transifex utils script.

.github/workflows/tests.yml
```diff
@@ -28,9 +28,10 @@ jobs:
         node-version-file: package.json
         cache: 'npm'
     - run: npm ci
+    - run: npm run build
```
garethbowen, line 27: This is needed for the WF lint to pass.

.github/workflows/tests.yml
```diff
@@ -28,9 +28,10 @@ jobs:
...
     - run: npm run lint
     - uses: browser-actions/setup-chrome@v2
-    - run: npm run test
+    - run: npm run test -w=@getodk/central-frontend
```
garethbowen, line 30: The WF tests are run in a separate job.

apps/central/.eslintrc.js
```diff
@@ -94,6 +95,7 @@ module.exports = {
     'vue/no-template-target-blank': 'off',
     'vue/object-curly-newline': 'off',
     'vue/require-default-prop': 'off',
+    'vue/require-prop-types': 'off',
```
garethbowen, line 98: I'm not quite sure why these started firing - it's possible the package versions were slightly off. I felt turning them off was preferable to fixing the issues in the code for now.

apps/central/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "sass-loader": "^13.2.0",
+    "sinon": "~9",
+    "util": "^0.12.5",
+    "vite": "^7.3.1"
+  },
+  "engines": {
+    "node": "24",
+    "npm": "11"
+  }
+}
```
garethbowen, line 62: These dependencies are largely unchanged from before but they had to be pushed down because some of them conflict with WF, notably eslint. In the future we should look at getting on to the same version of these libraries, and ideally using the same linting rules, but this gets everything up and running for now.

packages/common/test-utils/xform-dsl/index.ts
```diff
@@ -231,7 +231,7 @@ export const proposed_rankDynamic = (ref: string, nodesetRef: string): XFormsEle
 	const rankAttributes = new Map<string, string>();
 	rankAttributes.set('ref', ref);
 
-	return new TagXFormsElement('odk:rank', rankAttributes, [itemset]) as XFormsElement;
+	return new TagXFormsElement('odk:rank', rankAttributes, [itemset]);
```
garethbowen, line 234: I'm also not sure why this started firing, but there are about a dozen examples like this further down in this PR. It's possible lint wasn't running on these files before, or that the versions have changed slightly so errors started firing.

packages/common/.prettierignore
```diff
@@ -0,0 +1,9 @@
+**/node_modules
+**/dist
+
+# Fixtures
+src/fixtures/*/**
+!src/fixtures/README.md
+!src/fixtures/xforms.ts
+src/fixtures/test-javarosa/**
+src/fixtures/test-scenario/**
```
garethbowen, line 9: These files used to be at the top level but we're not ready (or willing?) to apply prettier over the whole repo so I had to move them down into their respective repos.

packages/eslint.config.js
```diff
@@ -0,0 +1,396 @@
+// @ts-check
```
garethbowen, line 1: `/packages/` is a strange place for these files but it works to separate WF from CF code so I felt it was the best option.

Another option would be to have something like a top level `/config` dir that all the configs can go in, and then either refer to it in the CLI command, or have an inline config that extends the common one.

package.json
```diff
@@ -2,79 +2,64 @@
...
-    "build": "vite build",
-    "eslint": "eslint --max-warnings 0 --cache --ext .js,.vue apps/*/src/ apps/*/test/ e2e-tests/ bin/ apps/*/bin/ *.js",
-    "eslint:fix": "eslint --max-warnings 0 --fix --cache --ext .js,.vue apps/*/src/ apps/*/test/ bin/ apps/*/bin/ *.js",
-    "lint": "npm run eslint && npm run transifex:lint",
-    "lint:fix": "npm run eslint:fix && npm run transifex:fix",
-    "test": "./apps/central/test/run.sh",
-    "test:e2e": "e2e-tests/run-tests.sh",
-    "transifex:fix": "node apps/central/bin/transifex/restructure.js > apps/central/transifex/strings_en.json",
-    "transifex:lint": "bash -c 'diff apps/central/transifex/strings_en.json <(node apps/central/bin/transifex/restructure.js)'"
+    "build": "turbo run build",
```
garethbowen, line 11: I added turbo which we've used successfully in WF for monorepo building, and added CF to it. This is a change to CF development but I expect it'll make dev better because it caches the packages so you only have to rebuild if you've changed something.

package.json
```diff
@@ -2,79 +2,64 @@
...
+    "build:web": "vite build",
+    "lint": "npm run lint --workspaces",
+    "lint:fix": "npm run lint:fix --workspaces",
+    "test": "npm run test --workspaces",
```
garethbowen, line 16: These are convenience commands to run against all repos - the same commands exist in each package/app so it's easy to run a subset also.

package.json
```diff
@@ -2,79 +2,64 @@
...
+    "playwright": "^1.58.2",
+    "prettier": "^3.8.1",
+    "rimraf": "^6.1.3",
+    "turbo": "^2.8.16",
+    "typescript": "~5.9.3",
+    "typescript-eslint": "^8.57.0",
+    "vitest": "^4.0.18",
+    "vue": "^3.5.29",
+    "vue-eslint-parser": "^10.4.0",
+    "vue-tsc": "^3.1.5"
```
garethbowen, line 66: This is mostly WF dependencies. Because there are 5 packages of WF code and only 1 of CF code it made more sense to put the WF dependencies at root, and CF dependencies in its app.

Over time I believe this will improve as the two repos standardise on dependencies and versions.

garethbowen:
> ... what do you think about splitting this PR up into multiple smaller ones? ... Personally, I find smaller PRs a bit easier to review.

@matthew-white I completely agree - this PR is far from ideal. I spent some time thinking about how to split it up but almost every piece is co-dependant. For example, if I pulled out some of the Central changes I would end up having to modify them a second time in the subsequent PR to make the stack work again. I've managed to clean it up a little, and added some comments to the PR to hopefully help somewhat...

package.json
```diff
@@ -2,79 +2,64 @@
...
   },
+  "packageManager": "npm@11.6.1",
   "workspaces": [
-    "apps/*"
+    "apps/*",
+    "packages/*"
   ],
-  "dependencies": {
-    "@floating-ui/dom": "^1.1.0",
-    "@getodk/web-forms": "^0.24.0",
```
garethbowen, line 28: This npm dependency has been removed altogether.

garethbowen:
@matthew-white @latin-panda I believe this PR is ready for review. I've tagged the both of you because it touches both CF and WF code so would benefit from experience in both codebases. Apologies for the complexity of the PR and I promise never to do it again!

.github/workflows/wf-ci.yml
```diff
@@ -0,0 +1,320 @@
...
+            browser: chromium
+          - target: 'Web'
+            node-version: '24.13.0'
+            browser: firefox
+          - target: 'Web'
+            node-version: '24.13.0'
+            browser: webkit
+
+    steps:
+      - uses: 'actions/checkout@v5'
```
latin-panda: Can we use `actions/checkout@v6` to match Central's? Same with the other occurrences

.github/workflows/tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - run: sudo apt-get install shellcheck
     - run: ./bin/check-scripts.sh
   unit-tests:
-    timeout-minutes: 5
+    timeout-minutes: 20
```
latin-panda: Which test is taking long?

packages/web-forms/package.json
```diff
@@ -34,6 +34,12 @@
     "build:translations": "node bin/merge-translations.js",
     "dev": "vite",
     "dist-demo": "npm run build && vite serve dist-demo --port 5174",
+    "format:fix": "npm-run-all --print-name format:vue format:prettier",
+    "format:lint": "prettier -c \"**/*\" --ignore-unknown",
+    "format:prettier": "prettier -w \"**/*\" --ignore-unknown --cache",
+    "format:vue": "eslint ./**/*.vue  --report-unused-disable-directives --cache --fix",
+    "lint": "eslint . --report-unused-disable-directives --cache --max-warnings 0 && npm run format:lint",
+    "lint:fix": "eslint . --fix --report-unused-disable-directives --cache --max-warnings 0 && run format:fix",
```
latin-panda:
```suggestion
    "lint:fix": "eslint . --fix --report-unused-disable-directives --cache --max-warnings 0 && npm run format:fix",
```

package.json
```diff
@@ -2,79 +2,64 @@
...
-    "test": "./apps/central/test/run.sh",
-    "test:e2e": "e2e-tests/run-tests.sh",
-    "transifex:fix": "node apps/central/bin/transifex/restructure.js > apps/central/transifex/strings_en.json",
-    "transifex:lint": "bash -c 'diff apps/central/transifex/strings_en.json <(node apps/central/bin/transifex/restructure.js)'"
+    "build": "turbo run build",
+    "build:web": "vite build",
+    "lint": "npm run lint --workspaces",
+    "lint:fix": "npm run lint:fix --workspaces",
+    "test": "npm run test --workspaces",
+    "test:e2e": "e2e-tests/run-tests.sh"
```
latin-panda, line 17: I'm not familiar with this run-tests script. Are we adding WF's e2e to `run-tests.sh` as well, or keeping them separate with workspaces (like we do in L15)? It would be nice to be able to run them all or a subset.

package.json
```diff
@@ -2,79 +2,64 @@
...
+    "@vue/tsconfig": "^0.8.1",
+    "eslint": "^10.0.3",
+    "eslint-config-prettier": "^10.1.8",
+    "eslint-plugin-jsdoc": "^62.8.0",
+    "eslint-plugin-no-only-tests": "^3.3.0",
+    "eslint-plugin-prettier": "^5.5.5",
+    "eslint-plugin-vue": "^10.8.0",
+    "npm-run-all2": "^8.0.4",
+    "only-allow": "^1.2.1",
+    "organize-imports-cli": "^0.10.0",
```
latin-panda: I don't see this being called anywhere. I wonder if someone is using it manually. 
It seems like a VS Code-specific 😅 Probably my WebStorm is going to ignore it (?)

package.json
```diff
@@ -2,79 +2,64 @@
...
-    "sass-loader": "^13.2.0",
-    "sinon": "~9",
-    "util": "^0.12.5",
-    "vite": "^6.3.5"
+    "@changesets/changelog-github": "^0.5.1",
+    "@changesets/cli": "^2.29.7",
+    "@eslint/js": "^10.0.1",
+    "@playwright/test": "^1.58.2",
+    "@transifex/cli": "^7.1.5",
+    "@tsconfig/node20": "^20.1.8",
```
latin-panda, line 37: This project package is set to use node 24.13.0. Then do we update this to `@tsconfig/node24 `?

packages/eslint.config.js
latin-panda: I'd love to use TypeScript in `/app/webforms`, so the main `tsconfig` could go in `/config` 🤔

packages/common/test-utils/xform-dsl/index.ts
latin-panda: I’ve seen something like this before, when I modify old files and get them checked, only to have an error appear in a section I didn't even touch. 🤷🏽‍♀️

bin/check-bundle-size.js
latin-panda, line 1: Is this rename intentional? Is it because of the `type: module` in the package.json? 😲

.github/workflows/wf-ci.yml
```diff
@@ -0,0 +1,320 @@
...
+
+  odk-common:
+    name: '@getodk/common'
+    runs-on: 'ubuntu-latest'
+
+    strategy:
+      fail-fast: false
+      matrix:
+        target: ['Node']
+        node-version: ['20.19.3', '22.12.0', '24.13.0']
```
latin-panda, line 28: Central is set to v24. Do we still want to run all these versions? It'd be nice for anyone using the npm packages of WF, but on the other side, we could make the CI faster :) 
Same below

.github/workflows/wf-ci.yml
```diff
@@ -0,0 +1,320 @@
...
+        with:
+          node-version: '${{ matrix.node-version }}'
+
+      - uses: 'actions/cache@v5'
+        id: cache-install
+        with:
+          path: |
+            node_modules
+            **/node_modules
+          key: install-${{ matrix.node-version }}-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/ci.yml', 'examples/*/package-lock.json', 'packages/*/package.json', 'packages/*/package-lock.json') }}
```
latin-panda:
```suggestion
          key: install-${{ matrix.node-version }}-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/ci.yml', 'packages/*/package.json', 'packages/*/package-lock.json', 'apps/*/package.json', 'apps/*/package-lock.json') }}
```

Well, when we add `apps/web-app`, but we can remove `examples/*/....`
Same below

.github/workflows/wf-ci.yml
```diff
@@ -0,0 +1,320 @@
...
+          fail-on-cache-miss: true
+
+      - uses: 'actions/cache@v5'
+        id: cache-build
+        with:
+          path: |
+            examples/*/dist
```
latin-panda: I think this is an old directory that was removed a long time ago
Same below
```suggestion
```

.github/workflows/wf-ci-e2e.yml
```diff
@@ -0,0 +1,106 @@
...
+        with:
+          node-version: '${{ matrix.node-version }}'
+
+      - uses: 'actions/cache@v5'
+        id: cache-install
+        with:
+          path: |
+            node_modules
+            **/node_modules
+          key: install-${{ matrix.node-version }}-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/ci.yml', 'examples/*/package-lock.json', 'packages/*/package.json', 'packages/*/package-lock.json') }}
```
latin-panda:
```suggestion
          key: install-${{ matrix.node-version }}-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/ci.yml', 'packages/*/package.json', 'packages/*/package-lock.json') }}
```

.github/workflows/wf-ci-e2e.yml
```diff
@@ -0,0 +1,106 @@
...
+          fail-on-cache-miss: true
+
+      - uses: 'actions/cache@v5'
+        id: cache-build
+        with:
+          path: |
+            examples/*/dist
```
latin-panda:
```suggestion

```

.github/workflows/wf-ci-e2e.yml
```diff
@@ -0,0 +1,106 @@
...
+    runs-on: 'ubuntu-latest'
+
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: ['24.13.0']
+        browser: [chromium, firefox, webkit]
+
+    steps:
+      - uses: 'actions/checkout@v5'
```
latin-panda:
```suggestion
      - uses: 'actions/checkout@v6'
```
same with the ones below

.github/workflows/tests.yml
```diff
@@ -43,7 +44,37 @@ jobs:
     - run: ./bin/check-for-large-files.sh
     - run: npm ci
     - run: npm run build
-    - run: ./bin/check-bundle-size.js
+    - run: ./bin/check-bundle-size.cjs
+  build-wf:
```
latin-panda, line 43: Same feedback here below

latin-panda (changes requested):
Good job! I haven't finished reviewing, I'll do another pass tomorrow

.github/workflows/wf-ci.yml
garethbowen: Done.

.github/workflows/tests.yml
garethbowen: It's actually the `npm run build` which I had to add to get the TS linting to work. I set it to 20 to make sure that's what it was which was definitely too much, so I've reduced it to 10.

In a future PR I'd like to split linting from unit testing to make it more clear what this job is doing. This is being discussed in this PR: https://github.com/getodk/central-frontend/pull/1548

packages/web-forms/package.json
garethbowen: Great spotting! Fixed.

package.json
garethbowen: In this PR I opted to keep them separate for simplicity.

In the longer term I think we should aim to convert WF e2e scripts to run in the central-frontend context rather than the demo app because that's more representative of the real world. So in a future PR that would mean moving the `/packages/web-forms/e2e` files into `/e2e-tests` and modifying them to pass when run with this script.

It would be good to keep suites for different concepts so the existing WF e2e tests should probably start in their own suite run by their own CI job.

package.json
garethbowen: Looks like it was added in the [initial commit](https://github.com/getodk/web-forms/commit/5d4acb37f638d5ed303bce2e01dd0c7883813a7a#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) of web-forms 3 years ago.

`only-allow` is also not used any more. I'll get rid of them.

package.json
garethbowen: I think the tsconfig should be set to the minimum supported node version which for web-forms is 20. Is that your impression too?

I'm very open to bumping the minimum node to 22 at least.

bin/check-bundle-size.js
garethbowen: Yes exactly. I'm comfortable with it because it's just this one dev script. A better fix would be to convert it to module...

.github/workflows/wf-ci.yml
garethbowen: I've added this as a conversation in our WF meeting today. Let's at least drop v20, but TBH if nobody is using it today we could easily say 24+!

lognaturel: I'm onboard for switching to 24+ and updating the readme to say we currently only support the latest node version in active LTS, similar to Central (https://github.com/getodk/central#nodejs-version). It's aggressive for libraries, but like you say, we can expand when/if we learn others want to use the libraries in their systems.

.github/workflows/wf-ci.yml
garethbowen: Ok... I've left `/apps/web-forms/` out for now, but also removed examples. This is obviously quite fragile.

.github/workflows/wf-ci.yml
garethbowen: Done.

.github/workflows/wf-ci-e2e.yml
garethbowen: Done.

.github/workflows/wf-ci-e2e.yml
garethbowen: Done.

.github/workflows/wf-ci-e2e.yml
garethbowen: Done.

.github/workflows/tests.yml
garethbowen: Done.

package.json
matthew-white: For `lint` and `test`, those will probably just be used in CI, right? At least for the foreseeable future? Aren't we planning at some point soon to run individual workflows based on which apps/packages the PR touches? I feel like the approach here lumps workflows/steps together, whereas we generally want to keep them separate. I guess it's not much harm to add these, but I also don't see too much benefit. I'd just as well leave them out if that'd reduce the complexity of package.json.

apps/central/package.json
matthew-white: Would you mind adding more details about which Central dependencies changed and what their old/new versions are? Any major version changes?

Did only dev dependencies change, or prod as well? E.g., I see `vue` in this list.

I think this is an area where I'm feeling some of the most uncertainty. It feels hard to review exactly how Central's dependencies have changed.

package.json
matthew-white: We mostly use Vite in dev, and I don't think we've had any issues with caching. I'm not familiar with turbo. Could you say more about it? What implications will it have for Central? Does it need to be part of this PR?

packages/common/.prettierignore
matthew-white: We haven't discussed using prettier for Central Frontend, but I'd be open to it. Probably better to leave that for a follow-up PR though if possible.

apps/central/.eslintrc.js
matthew-white: Hmm I can't think of an example where we don't specify prop types in Central. Would you mind sharing a few example errors?

package.json
matthew-white: > This project package is set to use node 24.13.0.

Just FYI, we want to bump this soon to 24.14.1 to match Central prod: getodk/central#1806

.github/workflows/tests.yml
matthew-white: How about linting WF in a separate workflow? Like what if we kept this file as the Central-specific workflow, then did all WF things (linting, testing) in other workflows?

apps/central/package.json
matthew-white: > E.g., I see `vue` in this list.

Or rather, I see `vue` in both this file and the top-level package.json (where it's a dev dependency). Is Central on the same version of `vue` as before?

package.json
matthew-white: Is it possible for there to be a packages/package.json? Could the WF dependencies go there? It doesn't seem ideal that there would be so many root dependencies that Central doesn't need.

apps/central/.eslintrc.js
garethbowen: There are a handful of issues like this:

```
.../central-frontend/apps/central/src/components/geojson-map/dev-tools.vue
  61:17  warning  Prop "maxZoom" should define at least its type             vue/require-prop-types
  62:17  warning  Prop "cluster" should define at least its type             vue/require-prop-types.
```

They can be fixed by switching to use a different function, so instead of...

```js
const maxZoom = defineModel('maxZoom');
```

We would have

```js
const maxZoom = defineModel({ type: Number });
```

I can make this change if you prefer?

apps/central/package.json
garethbowen: Agreed - I wish there was a simple way to diff all the actual changes, but the package-lock is basically useless for this.

However, because I copied all the dependencies in the package.json down into this package npm _should_ respect the version ranges defined in this file. None of the ranges allow for a major version change any many are patches only (`~`) so if all of the dependencies have respected semver we should be ok.

I've checked this change against the original /package.json and the two changes I can see are...

- @vitejs/plugin-vue: ^5.2.3" -> ^6.0.6
-  vite: ^6.3.5 -> ^7.3.1

I think these are due to the building being done at the top level so they had to match. Arguably they could now be removed from the central package.json.

> Is Central on the same version of vue as before?

According to the package lock both before and after this change vue is resolved to 3.5.29, so yes!

package.json
matthew-white: > Over time I believe this will improve as the two repos standardise on dependencies and versions.

Would you mind saying a little more about this? What will improve exactly?

1. Are you thinking that CF will start using more of these root-level dependencies, replacing CF-only dependencies that are currently in apps/central/package.json? (E.g., Karma→Vitest)
2. Or are you thinking that WF will start using dependencies that are currently CF-only, at which time those dependencies will move from apps/central/package.json to the root package.json?

I think (1), but I'm not totally sure. But in either case, the root package.json won't ever shrink, right? The end result either way is that apps/central/package.json is expected to shrink over time?

An example or two might be helpful if any come to mind.

packages/common/.prettierignore
garethbowen: Yeah 100%. If you do do it it'll change many lines in many files so it's not something to include with any other changes!

package.json
matthew-white: This is kind of making me question the split between apps/ and packages/. What if we put all the Web Forms code in apps/web-forms/, including packages? Then there'd be a natural Web Forms root. We wouldn't need to use the repo root for shared configuration between the different Web Forms directories.

In addition to package.json, it seems like this is coming up for ESLint configuration as well. It sort of seems like there's a need for there to be a Web Forms root directory.

package.json
garethbowen: With the monorepo the top level build will build all packages not just central, which slows things down a bit. I think you'll be able to work in `/apps/central` most of the time and ignore wf altogether but if you want to do something like run e2e tests then you'll need to build wf or those tests will fail.

Turbo tries to solve this exact problem by caching the result of the build, so the first run through you build all packages, but from then on if you only modify files in /apps/central and then run `npm run build` it will only run the build on the central app.

I hadn't used it before joining ODK but it seems fairly reliable and reliably performant, and a relatively simple addition.

bin/check-bundle-size.js
matthew-white: If it's easy to convert it to module, that sounds good to me. We want to move all of Central Frontend to modules soon. It's only Karma that's holding us back from doing so. Related: getodk/central#1269

package.json
garethbowen: They're particularly useful if you're changing things in multiple packages (eg: /packages/web-forms and /apps/web-forms) but it's not too hard to type out the full command either.

I personally like having these here because it's a self documenting sort of thing - when I come to a new project I always check the package.json to see how to run basic stuff.

> Aren't we planning at some point soon to run individual workflows based on which apps/packages the PR touches?

Maybe, yes. I don't have any concrete plans on this. I have had experiences where it's difficult to get the partial build just right and it ends up not firing when you want it to. We _do_ have some immediate plans to make the build quicker and simple, in particular dropping node 20 and node 22 support which will reduce the build matrix a bit. If that still proves too slow, expensive, or complicated, then it's worth looking into how to do partial builds based on what's changed.

package.json
garethbowen: Hmmm yeah I see what you're saying. However once web-forms is the default then central will "need" these packages most of the time, for example the new draft experience has a web-forms preview pane which won't work without them. I envision WF and CF devs building and running both products most of the time because otherwise there will be parts of CF that just don't work.

I'm not sure that separating the package.json would solve your issue because npm hoists the dependencies to the top level so you still have to npm install everything from WF. The only difference would be that central wouldn't "see" the WF dependencies.

My hope is that the two codebases will standardise more over time so this will become less of an issue.

.github/workflows/wf-ci.yml
garethbowen: I've got a subsequent branch with this change ready to go as soon as this gets merged.

bin/check-bundle-size.js
garethbowen: Ok... turned out it was super easy to convert. Done.

apps/central/package.json
matthew-white: > if all of the dependencies have respected semver we should be ok.

Unfortunately, I haven't always found this to be the case. Even in the world of Vue dependencies, I've seen things break from one minor version to the next. Then again, if CI is passing, that may be all the reassurance we need.

Not necessarily advocating for this, but is there a process we could take that would minimize changes to Central dependencies? Something like:

1. Before adding any WF dependencies, move package.json and package-lock.json to apps/central/.
2. From apps/central/, uninstall only the packages that are destined for the root package.json.
  - The hope is that this will help us retain the existing package-lock.json rather than discarding it.

Another idea:

1. Change the semver ranges in apps/central/package.json to exactly match what's in package-lock.json.
2. Install.
3. Loosen the semver ranges back again.
  - This won't necessarily use the same subdependencies as before, but it will let us continue to use the same versions of the top-level dependencies as we do now.

<blockquote>

I've checked this change against the original /package.json and the two changes I can see are...

- @vitejs/plugin-vue: ^5.2.3" -> ^6.0.6
-  vite: ^6.3.5 -> ^7.3.1
</blockquote>

That sounds fine to change given that CI is passing. :+1:

> Arguably they could now be removed from the central package.json.

That sounds reasonable to me given that CF and WF both use Vite.

apps/central/.eslintrc.js
matthew-white: Ah I see, it's `defineModel()`. I think we do want to fix those at some point, but it doesn't have to happen as part of this PR. If it's easy for you to fix them now, that could work well. But equally good from my perspective to turn the rule off as you're doing, then file an issue for the Central team to fix things up.

package.json
matthew-white: Sounds very interesting, I could see how that could be useful. Is there any part of it that you need me to review? E.g., did you change the turbo configuration in order to incorporate Central? Or is more so that you're just copying an existing configuration from the `web-forms` repo?

.github/workflows/tests.yml
garethbowen: Yes I think so. Alex is working on a separate effort to get the linting in one place so expect another pass streamlining these jobs further.

apps/central/.eslintrc.js
garethbowen: I think I'll do the latter. It's not too hard to fix but you do need to know what types the models are going to hold which isn't always clear. I'll raise an issue now.

package.json
matthew-white: > I added turbo which we've used successfully in WF for monorepo building, and added CF to it. This is a change to CF development

OK right, you did say that you added CF to turbo. I can try to take a look at that part of it. If you have guidance about whether there's anything specific related to turbo that you want me to review, that could be helpful.

Just to clarify, you said "This is a change to CF development", but you also said, "you'll be able to work in `/apps/central` most of the time and ignore wf altogether". It sounds like turbo will mostly just affect CF development related to e2e tests — is that right?

We do run `npm run build` in Docker as part of the ops in the `central` repo, but I don't imagine that turbo would disrupt that.

package.json
matthew-white: > once web-forms is the default then central will "need" these packages most of the time, for example the new draft experience has a web-forms preview pane which won't work without them.

I see what you're saying, but the Central Vue app won't need any of these dependencies. The Central Vue app and the WF Vue app will be separate. To me, the fact that Central will render WF in an iframe isn't a reason for the Central Vue app to include dependencies that it doesn't need.

> I envision WF and CF devs building and running both products most of the time

Yes, I think that's true. :+1:

> I'm not sure that separating the package.json would solve your issue because npm hoists the dependencies to the top level so you still have to npm install everything from WF. The only difference would be that central wouldn't "see" the WF dependencies.

If separating package.json based on app/package isn't important, why not install all dev dependencies at the top level? Why not have everything see all dependencies?

To me, it's nice to make it very clear which apps/packages rely on which dependencies, including which packages Central relies on vs. which are WF-only.

> My hope is that the two codebases will standardise more over time so this will become less of an issue.

I also expect some standardization over time. :+1: However, it's not clear to me that we ever intend the codebases to fully align on things like TypeScript. Many of the dependencies in this list are related to TypeScript.

I'm open to being convinced that this isn't a problem in practice, but it just seems weird that Web Forms has a bunch of shared configuration/dependencies, yet no way to share it without also affecting Central. I feel like it hints at a larger issue with the structure of the repo.

apps/central/package.json
garethbowen: >  Something like: ... Another idea:

I think in both of these cases the change would undo itself the next time anyone touched the package-lock (eg: added, removed or updated a dependency). npm seems to be very aggressive about checking for new versions of packages when anything has changed.

> That sounds reasonable to me given that CF and WF both use Vite.

I've cleaned up the package.json files where we have consistent dependencies. `npm install` didn't change the package-lock.json at all so should be safe!

package.json
garethbowen: It's all configured in the top level turbo.json. It's a bit esoteric to read through but basically you define which tasks depend on which. It's mostly copied over from web-forms with an entry added which invokes `npm run build:web` which depends on the other packages being built first. It's hard to review but feel free to clone my repo and try it out for yourself!

apps/central/package.json
garethbowen: When I removed vue and vue-router from apps/central eslint started failing. I think there's an old bug that impacts the eslint version that central is using but is fixed in the version web-forms is using. I've put vue and vue-router back for now, and we should look at getting on to the same version of eslint at some point (but that has its own issues...)

package.json
garethbowen: > Just to clarify, you said "This is a change to CF development", but you also said, "you'll be able to work in /apps/central most of the time and ignore wf altogether". It sounds like turbo will mostly just affect CF development related to e2e tests — is that right?

I think you'll have the option to work either way. If you're doing something in a common package _and_ the central app then building at the top level would be the way to go. But if you're just living in central land then it'd be quicker to `cd apps/central` and just do npm commands in there. In my experience developers all do things their own way so I probably haven't tested each unique combination of commands, but I'm confident we can get there!

> We do run npm run build in Docker as part of the ops in the central repo, but I don't imagine that turbo would disrupt that.

Turbo will still run in that situation but nothing will be cached because it's a new build. Once prod installs docker images then it'll impact prod even less.

package.json
garethbowen: > If separating package.json based on app/package isn't important, why not install all dev dependencies at the top level? Why not have everything see all dependencies?

This doesn't work if central and web-forms need different versions. One example is eslint where WF uses 10 and CF uses 8. I tried to upgrade CF but unfortunately the airbnb eslint config hasn't been updated in years and is not compatible. For situations like that we need to keep them separate, but where possible I agree the dependencies should be at the top level. One of my recent commits does a better job at cleaning that up.

> I'm open to being convinced that this isn't a problem in practice, but it just seems weird that Web Forms has a bunch of shared configuration/dependencies, yet no way to share it without also affecting Central. I feel like it hints at a larger issue with the structure of the repo.

Let's discuss this in the call in 12 hours!

apps/central/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "raw-loader": "^4.0.2",
+    "sass": "~1",
+    "sass-loader": "^13.2.0",
+    "sinon": "~9",
+    "util": "^0.12.5"
+  },
+  "engines": {
+    "node": "24",
+    "npm": "11"
+  }
```
matthew-white: Is it possible/desirable to specify `node` and `npm` in one place, in the root package.json?

.gitignore
```diff
@@ -11,8 +11,11 @@
 .DS_Store
 node_modules
 npm-debug.log*
-/dist
+**/dist
```
matthew-white, line 14: Why was this change needed out of curiosity?

- Because the Central `dist` directory is now in `apps/central`?
- Or because there will be multiple `dist` directories?

.gitignore
```diff
@@ -11,8 +11,11 @@
...
+packages/web-forms/dist-demo
```
matthew-white, line 15: This will probably go away at some point soonish, right?

package.json
```diff
@@ -2,79 +2,66 @@
...
-    "sass": "~1",
-    "sass-loader": "^13.2.0",
-    "sinon": "~9",
-    "util": "^0.12.5",
-    "vite": "^6.3.5"
+    "@changesets/changelog-github": "^0.5.1",
+    "@changesets/cli": "^2.29.7",
+    "@eslint/js": "^10.0.1",
+    "@playwright/test": "^1.58.2",
+    "@transifex/cli": "^7.1.5",
```
matthew-white, line 36: Ooh I didn't realize there was a node package for the Transifex CLI. I'm pretty sure I installed the CLI some other way. I'd be interested in making use of this for Central. I may file an issue or ask more about that at some point.

.gitignore
garethbowen: Currently each package has its own dist including apps/central, so yes to both. There's a top level dist which has everything rolled up.

.gitignore
garethbowen: Yeah very soon. I'm working on separating out the demo app into its own repo then this will go.

package.json
```diff
@@ -2,79 +2,66 @@
...
+    "@vue/eslint-config-typescript": "^14.7.0",
+    "@vue/tsconfig": "^0.8.1",
+    "eslint": "^10.0.3",
+    "eslint-config-prettier": "^10.1.8",
+    "eslint-plugin-jsdoc": "^62.8.0",
+    "eslint-plugin-no-only-tests": "^3.3.0",
+    "eslint-plugin-prettier": "^5.5.5",
+    "eslint-plugin-vue": "^10.8.0",
+    "npm-run-all2": "^8.0.4",
+    "playwright": "^1.58.2",
```
matthew-white: Question from call: what is the difference between `playwright` here and `@playwright/test` in the Central package.json?

matthew-white: In this PR, `@playwright/test` has been removed from the Central package.json. So Central will be using `playwright`. It'd be interesting to learn more about the difference between the two.

package.json
```diff
@@ -2,79 +2,66 @@
...
+    "prettier": "^3.8.1",
+    "rimraf": "^6.1.3",
+    "turbo": "^2.8.16",
+    "typescript": "~5.9.3",
+    "typescript-eslint": "^8.57.0",
+    "vite": "^7.3.1",
+    "vitest": "^4.0.18",
+    "vue-eslint-parser": "^10.4.0",
```
matthew-white: Question from call today: what does this package do in addition to `eslint-plugin-vue`? Should we be using this in Central?

turbo.json
matthew-white, line 1: Question from call today: Is the `dist` directory and its contents that's generated by this PR the same as the `dist` directory that's generated by the `master` branch? The files in the `dist` directory will be served by nginx, so we only want it to contain files that we want to serve and that are needed for Central.

matthew-white: It's possible that the `dist` directory is relative to the package. That may be what's happening (and what's necessary) even for how Web Forms works today.

matthew-white (commented):
Notes from call today between @garethbowen and me

turbo.json
garethbowen: I've checked against `master` and the dist directories are identical, except for: (a) the file sizes are slightly different which is probably to be expected, (b) the web-forms translation files exist in this pr but don't in master. I checked and the translation values only exist in the translation files (ie: aren't also inlined into the js) so I think this is ok, and in fact, desirable because we only want the user to download languages when they need them. I've tested and translation still works in the app. So while I can't tell what changed to cause this difference in the dist dir but it seems like it's fine or even possibly better than before.

package.json
garethbowen: It turns out `vue-eslint-parser` is a dependency of `eslint-plugin-vue` ([source](https://github.com/vuejs/eslint-plugin-vue/blob/91a136cfdd743b1e77f63f824d1fdfe9c39f261e/package.json#L128)) so you don't need to list it explicitly unless you need a different version for some reason. I've removed it and everything still works.

package.json
garethbowen: Similarly `@playwright/test` depends on `playwright` so no harm in removing it.

latin-panda (approved):
I've reviewed it again, and everything looks great! If you have more changes you'd like to make after yesterday's call, please don't hesitate to ask for another review :)

apps/central/package.json
matthew-white: @garethbowen and I discussed this topic on a call earlier this week.

- Central Frontend's package-lock.json will be blown away by this change, but as @garethbowen has said, packages will still respect the semver ranges specified in package.json.
  - We wouldn't usually blow away package-lock.json, but this seems like a one-off/special occasion.
  - It's probably nice to be on more recent versions of dependencies anyway.
  - The fact that Central Frontend's CI is passing is a strong vote of confidence that things continue to basically work after this change. We'll also have QA do a round of regression testing shortly before release.

> I think in both of these cases the change would undo itself the next time anyone touched the package-lock (eg: added, removed or updated a dependency). npm seems to be very aggressive about checking for new versions of packages when anything has changed.

I mentioned to @garethbowen that I haven't observed this behavior from npm. When I `npm install` a new package or `npm uninstall` an existing one, I don't see unrelated packages change in package-lock.json. I often review the diff for package-lock.json when it changes, and I haven't observed that behavior. npm probably would do that if you ran `npm update` without any arguments, but we don't generally do that in Central Frontend.

> I've put vue and vue-router back for now, and we should look at getting on to the same version of eslint at some point

I agree. :+1: I've filed getodk/central#1906 about this.

matthew-white:
Just wanted to make a note about things that I'm not planning to review. I'm happy to let @latin-panda's review stand for these files:

- Changes to Web Forms (the packages/ directory)
- Web Forms workflows/CI
- I'm only skimming turbo.json and am not trying to fully understand it at this point

package.json
matthew-white: I've filed an issue about this at getodk/central#1907.

package.json
matthew-white: > I'm open to being convinced that this isn't a problem in practice, but it just seems weird that Web Forms has a bunch of shared configuration/dependencies, yet no way to share it without also affecting Central. I feel like it hints at a larger issue with the structure of the repo.

@garethbowen has convinced me that it's best to keep the current structure/setup. :+1: In principle, things are flexible enough that we're able to structure the monorepo directories however we want. We have the option to create a clean split between Web Forms and Central Frontend if we wanted. That would make it easy to do things strictly within Web Forms (like sharing dependencies) without touching Central. However, it would arguably be too easy: a clean split like that would disincentivize standardization, as WF and CF would be able to walk down separate paths without checking in. The current structure may lead to friction, but I think a healthy kind of friction that prompts us to start conversations and seek standardization.

At the moment, the only point of friction seems to be package.json, so practically speaking, I don't think it's a big deal.

An alternative approach we could take without changing the directory structure would be for each Web Forms package to repeat the same set of dev dependencies. That sounds like needless effort though.

> it's not clear to me that we ever intend the codebases to fully align on things like TypeScript. Many of the dependencies in this list are related to TypeScript.

This still seems true to me. Hopefully, it's also easy enough to ignore the TypeScript-related dependencies while skimming the list in package.json. In the end, the list of dependencies in the root package.json isn't that long, and in most cases, it's pretty clear why the dependency is on the list.

apps/central/.eslintrc.js
matthew-white: Just to link things: getodk/central#1898

apps/central/package.json
matthew-white: @garethbowen, just wanted to make sure you saw this question.

apps/central/package.json
```diff
@@ -0,0 +1,68 @@
...
+  },
+  "devDependencies": {
+    "@faker-js/faker": "^9.7.0",
+    "@intlify/unplugin-vue-i18n": "^6.0.8",
+    "@vitejs/plugin-vue": "^6.0.6",
+    "@vue/cli-service": "^5.0.8",
+    "@vue/compiler-sfc": "~3",
+    "@vue/eslint-config-airbnb": "^7.0.0",
+    "@vue/test-utils": "^2.4.6",
+    "chai": "^5.1.1",
```
matthew-white, line 41: For some reason, I thought Web Forms also used chai. What does Web Forms use for assertions?

package.json
```diff
@@ -2,79 +2,64 @@
...
+    "eslint-plugin-no-only-tests": "^3.3.0",
+    "eslint-plugin-prettier": "^5.5.5",
+    "eslint-plugin-vue": "^10.8.0",
+    "npm-run-all2": "^8.0.4",
+    "prettier": "^3.8.1",
+    "rimraf": "^6.1.3",
+    "turbo": "^2.8.16",
+    "typescript": "~5.9.3",
+    "typescript-eslint": "^8.57.0",
+    "vite": "^7.3.1",
```
matthew-white, line 64: @garethbowen you wrote at https://github.com/getodk/central-frontend/pull/1549#discussion_r3263468621 that this will be a major version change from the perspective of Central Frontend, as Central was previously on v6 of `vite` (also v5 of `@vitejs/plugin-vue`, not v6). Do you feel like we should review Vite release notes or otherwise look into breaking changes stemming from this major version change? Or is it sufficient that CI passes? My instinct is that it wouldn't be a bad idea to check Vite docs / release notes for breaking changes.

apps/central/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "vue": "^3.5.29",
+    "vue-flatpickr-component": "~9",
+    "vue-i18n": "^10.0.7",
+    "vue-router": "^4.6.3",
+    "xml-formatter": "~2"
+  },
+  "devDependencies": {
+    "@faker-js/faker": "^9.7.0",
+    "@intlify/unplugin-vue-i18n": "^6.0.8",
+    "@vitejs/plugin-vue": "^6.0.6",
```
matthew-white: I thought the plan was to move this dependency to the root package.json. Isn't the dependency shared with Web Forms?

apps/central/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "@vue/cli-service": "^5.0.8",
+    "@vue/compiler-sfc": "~3",
+    "@vue/eslint-config-airbnb": "^7.0.0",
+    "@vue/test-utils": "^2.4.6",
```
matthew-white, line 40: Does Web Forms not use Vue Test Utils? What does it use instead?

apps/central/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "chai": "^5.1.1",
+    "chai-as-promised": "^8.0.0",
+    "comment-json": "~4",
+    "eslint": "^8.28.0",
+    "eslint-plugin-import": "~2",
+    "eslint-plugin-vue": "~9",
+    "foreman": "^3.0.1",
```
matthew-white: Since the Procfile is in the root of the repository, I feel like this dependency belongs in the root package.json. Its main purpose is to run Vite and nginx at the same time / in the same terminal. That seems relevant to Web Forms going forward, since Web Forms development locally and especially the future Web Forms Vue app will require nginx to be running.

apps/central/package.json
matthew-white: @garethbowen, just wanted to make sure you saw this question.

package.json
```diff
@@ -2,79 +2,64 @@
...
   "version": "0.1.0",
   "private": true,
+  "type": "module",
   "scripts": {
     "dev": "          nf start vite,nginx-docker",
     "dev-no-docker": "nf start vite,nginx-host",
     "dev:build": "nf start build,nginx-docker",
     "dev-no-docker:build": "nf start build,nginx-host",
-    "build": "vite build",
-    "eslint": "eslint --max-warnings 0 --cache --ext .js,.vue apps/*/src/ apps/*/test/ e2e-tests/ bin/ apps/*/bin/ *.js",
```
matthew-white, line 11: This used to lint bin/ and e2e-tests/, but now I don't think anything is doing that, since those directories are outside a workspace. Could you please double-check that we're linting files in those directories?

We should also lint the .js files in the top-level directory of the repository (*.js from this line).

package.json
```diff
@@ -2,79 +2,64 @@
...
+    "build:web": "vite build",
+    "lint": "npm run lint --workspaces",
+    "lint:fix": "npm run lint:fix --workspaces",
+    "test": "npm run test --workspaces",
+    "test:e2e": "e2e-tests/run-tests.sh"
   },
   "volta": {
     "node": "24.13.0"
   },
+  "packageManager": "npm@11.6.1",
```
matthew-white, line 22: Could you say more about the addition of this line? Is it pinning the version of npm? If so, why is that necessary to do now? Historically, we haven't done that for Central Frontend.

Does this have implications for the `central` ops repo? E.g., will we need to ensure that this particular version of npm is installed? Will we need to update this line whenever we update Node?

Why specify the version of npm here rather than inside the `volta` object?

matthew-white (approved):
LGTM 🚀

Thank you, @garethbowen!

I left a comment above about what I reviewed and what I didn't review. I also wanted to note that my review was focused on what changed. I didn't put much thought about what *didn't* change, i.e., what else (if anything) might need to change. If issues come up along those lines, I'm sure we can address.

I still have some open questions, which you can see above. However, I don't feel like any of those needs to block merge. I'd like to continue those conversations and fully resolve those questions, but I'm happy for this PR to be merged at any time. One option for the questions above is to file one or more follow-up issues and continue discussing there.

I'm excited that Web Forms can now be built and tested in the new world of the monorepo!

apps/central/package.json
garethbowen: Sorry I forgot to reply.

For the published packages (/packages/*) it's useful to have for people installing as a dependency. For the apps, I think you're right. Removed.

apps/central/package.json
garethbowen: WF uses vitest assertions: https://vitest.dev/api/expect.html

package.json
garethbowen: Vite changes are mostly node and browser support changes which isn't within our support requirements: https://vite.dev/blog/announcing-vite7

Similarly for the plugin - node versions and a configuration change that doesn't impact central: https://github.com/vitejs/vite-plugin-vue/blob/main/packages/plugin-vue/CHANGELOG.md

package.json
garethbowen: Nice spotting!

This is problematic because of the discrepancy between eslint versions so we don't really have a config at the top level. If I add one then /apps/central uses that instead of its own.

In the meantime I've created a basic config in /bin and added a command that uses that config for `bin/*`, `e2e-tests/*`, and `*.js`.

I'll update the eslint issue for standardisation so these dirs use the common config once we have one.

package.json
garethbowen: packagemanager is used by corepack for compatibility with package managers other than npm. It was useful when web-forms used yarn but I don't think it's needed now that we've standardised on npm because it's the default package manager. I'll remove it and confirm.

apps/central/package.json
garethbowen: Done.

apps/central/package.json
garethbowen: This is the one that fails lint, so I have to keep it here in the meantime.

apps/central/package.json
garethbowen: You're right! Consolidated at the root.

package.json
garethbowen: It turns out turbo requires it to build! I've added it back again.

PR #1550, opened 2026-05-13
Project Custom Properties tab and ability to create properties

### LINKED ISSUE #1870: Project-level "Custom Properties" tab with user properties management screen
Add a project-level screen where project managers can create and list user properties. This is the primary place to manage user properties; inline creation from the App User / PAL create modals is a separate follow-up.

- New screen under project settings (exact location per Figma)
- Create a new user property (name only — values are free text and set per actor elsewhere)
- List all user properties defined on the project
- Validation per backend rules (same as dataset property names, cannot be `displayName`)

## Out of scope
- Deleting user properties (future)
- Inline creation from App User / PAL create modals (tracked as a follow-up)
- Assigning values to users (handled in other screens)

Depends on:
- https://github.com/getodk/central/issues/1868

ktuite:
Closes https://github.com/getodk/central/issues/1870

Screenshot 2026-05-13 at 4 02 28 PM

I've used it along with the backend in https://github.com/getodk/central-backend/pull/1824 

We've discussed a lot of ways to show user properties, and this was the one we decided to start with!

This is part of a larger effort to give more flexibility and enable entity segmentation.

We will want to add user documentation about what these properties are and how to use them. 

### DIFF
--- a/apps/central/src/components/project/custom-properties/list.vue
+++ b/apps/central/src/components/project/custom-properties/list.vue
@@ -6,18 +6,17 @@
         <button v-if="project.dataExists && project.permits('project.update')"
           id="custom-properties-list-new-button" type="button" class="btn btn-primary"
           @click="createModal.show()">
-          <span class="icon-plus-circle"></span>{{ $t('new') }}
+          <span class="icon-plus-circle"></span>{{ $t('action.create') }}&hellip;
         </button>
       </template>
     </page-section>
     <div class="page-body-heading">
-      <p>{{ $t('heading[0]') }}</p>
       <p>
-        <span>{{ $t('heading[1]') }}</span>
+        <span>{{ $t('heading[0]') }}</span>
         <sentence-separator/>
         <i18n-t keypath="moreInfo.clickHere.full">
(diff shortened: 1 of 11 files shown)


apps/central/src/components/project/custom-properties/list.vue
```diff
@@ -0,0 +1,105 @@
...
+      // A brief introduction to Custom Properties shown for the current project
+      "Some helper intro text",
+      "Custom Properties are custom fields that can be attached to project app users and public links to store additional metadata."
+    ],
+    "emptyTable": "No custom properties have been defined for this project.",
+    // Shown after a custom property is successfully created.
+    "created": "Custom property created successfully.",
+    // This is the text of a button that is used to create a new Custom Property for users.
+    // It is shown next to a heading whose text is "Custom Properties".
+    "new": "New"
```
ktuite: should fix text before committing, at least remove placeholder text that shouldn't be translated

sadiqkhoja (approved):
rebased, updated copy, and fixed the tests.

PR #1551, opened 2026-05-14
Allow user to upload entity CSV with missing properties

### LINKED ISSUE #1787: Allow user to import Entity CSV with missing properties in header
If a user tries to import an Entity CSV whose header does not include all properties, they see an error like:

Image

Sometimes it's completely legitimate to omit a column in an import. For example, in the case above I had a `consent` property that only ever gets set by forms.

It's hard to bulk backfill properties so we definitely do want to warn. Maybe something like "The following 3 properties are missing: foo, bar, baz. If you want to give them values later, you will need to do so manually for every Entity. Are you sure you want to continue?"

[Figma file](https://www.figma.com/design/ULpCuTKkIys6Y99YGyMcrF/Central?node-id=4711-3615&t=PMDCYeAiF6LkLuOa-1)

matthew-white:
Closes getodk/central#1787.

New tests. I also tried it out locally.

The existing `EntityUpload` modal is pretty complicated, so I tried not to change too much of the existing code. Instead, I added a new ref to store the warning about missing properties (`headerWarnings`) and a new component to show the warning (`EntityUploadHeaderReview`). As I continue working on getodk/central#1784, some of this approach may change.


apps/central/src/components/entity/upload.vue
```diff
@@ -58,10 +58,12 @@ except according to the terms contained in the LICENSE file.
           <entity-upload-data-error v-if="dataError != null"
             :message="dataError"/>
         </entity-upload-file-select>
+        <entity-upload-header-review v-if="headerWarnings != null && headerErrors == null"
```
matthew-white:
```suggestion
        <entity-upload-header-review v-if="headerWarnings != null"
```

apps/central/src/components/entity/upload.vue
```diff
@@ -245,16 +261,17 @@ const rowToEntity = (values, columns) => {
   return obj;
 };
 const { i18n: globalI18n, redAlert } = inject('container');
-const parseEntities = async (file, headerResults, signal) => {
```
matthew-white, line 248: Instead of passing header `warnings` here as part of `headerInfo`, let's move all the setting that this function does into `selectFile()` below. This function could then just return the results of parsing the data. We wouldn't need to distinguish between `headerResults` and `headerInfo`.

matthew-white (commented):

ktuite (approved):

apps/central/src/components/entity/upload.vue
matthew-white: Done in b3a712722aeb20e4d82889bfcdb296dec9ca8a46

apps/central/src/components/entity/upload.vue
matthew-white: Done in 2398bbb5441d50285e7563b5312a4222121edc02

PR #1552, opened 2026-05-20
feat: adds support to annotation, drawing, and signature appearances in Web Forms

### LINKED ISSUE #15: Annotation question type
## Description

Implement the [**Annotation** question type](https://docs.getodk.org/form-question-types/#image-widget-with-annotation) with the following requirements:

### Core Features
* **Full-screen mode:** Support for immersive annotation.
* **Color picker:** Allow users to select stroke colors.
* **Undo:** Unlimited undo actions within the current drawing session.
* **Stroke:** Unlimited strokes per session. Only one fixed stroke thickness supported.
* **Zoom:** Zoom in and out functionality for fine-detail work.
* **Info button:** Explains icon buttons.
* **Image Flattening:** Merge strokes into the base image on edit to create a new JPG (add debounce X ms and monitor performance).
  * **Edit Submission Case:** Existing strokes cannot be modified; only new strokes can be added or undone within the current session.

#### Mobile
* Selecting or taking a picture immediately opens the full-screen canvas.
* Drawing is only possible when full-screen mode is activated.

#### Desktop
* Does not default to full-screen; displays an editable preview first.
* Users can draw in both the preview and full-screen modes.
* Does not include a "Mark up image" button (drawing is enabled by default).

---

### Future Scope (needs user feedback)
- [ ] Clear all strokes button
- [ ] Shapes: Lines, arrows, rectangles, text, symbols, emojis
- [ ] Eraser tool
- [ ] Variable stroke thickness
- [ ] Compatibility with multi-page forms (hopefully not a relevant concern in this world but making explicit due to experience with Enketo)
- [ ] Grid layout support

---

## Design
[Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=4817-1587&t=7MeM1X4lkiQ6a9Wr-4)

## User Stories

* **As a form user**, I want to be able to draw freely on an image to annotate it.
* **As a form user**, I want to be able to see the details of a large image to annotate them even if my device is small.
* **As a form creator**, I want to be able to specify a default image (or an image from a preceding question) to annotate so that users don't have to select the same image multiple times.
* **As a data analyst**, I want annotated images to have the same resolution as the background image so that I don't lose any detail.

## Related

* https://github.com/getodk/web-forms/issues/699

### LINKED ISSUE #698: Draw question type
### Description

Implement the `draw` question type ([specs](https://docs.getodk.org/form-question-types/#draw-widget)) to allow users to capture freehand sketches and annotations. This feature should provide a canvas where users can draw with a stylus or finger and save the result as an image file. Key functionalities include:

* **Blank Canvas:** Implementation of the standard `draw` widget for creating new sketches from scratch.
* **Canvas Clearing:** Ability for users to reset the canvas if they need to restart their drawing.
* **Media Integration:** Ensuring the resulting sketch is saved and associated with the form submission as a standard image attachment.

---

### Checklist
  - [x] Yes. Describe the UI/UX requirements or attach a Figma link in the "Design" section.
  - [ ] No
  - [ ] No
  - [ ] No
  - [ ] No
---

### User Stories

* As a data collector, I want to draw a freehand sketch of a site map so that I can provide a visual representation of the area I am surveying.
* As a researcher, I want to collect hand-drawn diagrams from participants to better understand their spatial perceptions or ideas.
* As a form designer, I want to include a drawing prompt so that users can sign their names or provide non-textual responses.
* As a user, I want to be able to clear my drawing and start over if I make a mistake, ensuring the submitted sketch is accurate.

---

### Design

- [**Figma design**](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=4895-6052&t=RMqVWFAykeUdIuCf-4)

---

### Dependencies

-

---

### Additional Notes

-

---

### Related Issues

-

### LINKED ISSUE #699: Signature widget
### Description

Implement the `signature` widget ([spec](https://docs.getodk.org/form-question-types/#signature-widget)), which is a specialized version of the `draw` question type designed specifically for capturing handwritten signatures. This feature should present a canvas with a distinct horizontal line for signing and include a "Gather Signature" button to initiate the capture. Key functionalities include:

* **Specialized Canvas:** A signature-focused drawing area with a baseline.
* **Image Output:** Saving the captured signature as a standard image attachment (typically `.png` or `.jpg`).
* **Signature Appearance:** Proper handling of the `signature` appearance attribute to distinguish it from a general drawing widget.

---

### Checklist
  - [ ] No
  - [ ] No
  - [ ] No
  - [ ] No
---

### User Stories

* As a data collector, I want to capture a formal signature from a respondent so that I can verify consent or completion of an agreement.
* As a form designer, I want to use the `signature` appearance so that the user is presented with a clear signing area rather than a general-purpose sketchpad.
* As a researcher, I want to ensure that signatures are stored as image files linked to the submission for audit and verification purposes.
* As a user, I want to sign my name using my finger or a stylus and be able to redo the signature if it is not legible.

---

### Design

- [**Figma Design**](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=4905-4046&t=RMqVWFAykeUdIuCf-4)

---

### Dependencies

-

---

### Additional Notes

-

---

### Related Issues
- https://github.com/getodk/web-forms/issues/15
- https://github.com/getodk/web-forms/issues/698

latin-panda:
Closes https://github.com/getodk/web-forms/issues/15
Closes https://github.com/getodk/web-forms/issues/698
Closes https://github.com/getodk/web-forms/issues/699

- Manual testing of signature, draw, and annotate upload appearances using the new `image-drawing.xml` test form, covering: empty start, full-screen drawing flow, undo, clear, color picker, pan/zoom, default `jr://` background image, edit-submission...
- Verified that annotate continues to resize on save when `orx:max-pixels` is set, and that the resized version is what gets submitted while the canvas keeps drawing against the original-resolution image.

- Chosen Konva + vue-konva over a raw `<canvas>` approach or `Fabric.js` because it provides a layered scene graph (perfect for managing the background image + stroke layer separately) and native Vue bindings. It also benefits from a large community and active maintenance.

- We are lazy-loading Konva because this question type is rarely used, meaning only a subset of forms require it. By utilizing AsyncCanvas + loadCanvas.ts to dynamically import `vue-konva` and `CanvasBlock.vue`, the canvas code lands in its own chunk and avoids bloating the initial bundle.
  - Note: A promise cache was implemented to work around a Safari bug where concurrent dynamic imports can incorrectly resolve `.default` to undefined.

- Implemented `fetchFormAttachment` on `PrimaryInstance` so default images referenced by `jr://images/...` on an annotate upload can be loaded as the canvas background. Routing this through `InstanceAttachmentsState` that centralizes the logic.

- Upload controls with signature, draw, or annotate appearances are now rendered as interactive canvases.

[PR DOCS](https://github.com/getodk/docs/pull/2065)


packages/web-forms/package.json
latin-panda, line 1: ToDo: Update root's package-lock.json with new deps for package/web-forms when [this other PR is merged](https://github.com/getodk/central-frontend/pull/1549).

packages/common/src/fixtures/upload/image-drawing.xml
latin-panda, line 1: This, along with other test forms, will be moved later, but I'm adding it here for now so everything can be moved together.

packages/web-forms/src/assets/styles/panel-controls.scss
latin-panda, line 1: Map and Canvas components share some code; this was renamed to reuse in both places.

packages/web-forms/src/components/common/canvas/loadCanvas.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * IMPORTANT: The dynamic imports here keep vue-konva and CanvasBlock in a separate lazy chunk.
+ * Do not replace them with static imports.
+ *
+ * The module-level cache serialises concurrent imports. Safari returns undefined for ".default"
```
latin-panda, line 5: The Safari bug is real, and we hit it during testing with iPhone devices. Keep the module-level `cache.promise` even if it looks redundant.
(It was not an issue in Firefox and Chrome when using desktop and Android )

packages/web-forms/src/components/common/canvas/AsyncCanvas.vue
```diff
@@ -0,0 +1,45 @@
...
+import { getCurrentInstance, inject, shallowRef } from 'vue';
+
+const props = defineProps<{
+	mode: CanvasMode;
+	baseImageSrc: `blob:${string}` | null;
+	isDisabled?: boolean
+}>();
+
+const emit = defineEmits<(e: 'saveImage', file: File | null) => void>();
+const isFullScreen = defineModel<boolean>('isFullScreen', { default: false });
```
latin-panda, line 21: When choosing or dropping an image, it will load full-screen on mobile devices so the user can start drawing. This is not the behavior for default values or values from "edit submission".

packages/web-forms/src/components/common/canvas/CanvasBlock.vue
```diff
@@ -0,0 +1,429 @@
...
+onUnmounted(() => {
+	document.removeEventListener('keydown', onKeyDown);
+	resizeObserver.current?.disconnect();
+	cancelFlatten();
+	abortPendingLoad();
+});
+</script>
+
+<template>
+	<!-- eslint-disable vue/no-undef-components -- vue-konva registers v-stage and other components on lazy loading -->
```
latin-panda, line 242: I couldn't find a way to fix this ESLint error. Because the component is lazy-loaded, it is not being picked up by the linter.

packages/web-forms/src/components/common/canvas/CanvasComponent.i18n.json
```diff
@@ -0,0 +1,54 @@
+{
```
latin-panda, line 1: Translations into Spanish, French, and Indonesian will be completed after the PR is merged.

packages/web-forms/src/components/common/canvas/CanvasControls.vue
```diff
@@ -0,0 +1,241 @@
...
+				:disabled="disableUndo"
+				@click="emit('undo')"
+			>
+				<IconSVG :name="CONTROL_ICONS.undo" />
+			</button>
+		</div>
+	</div>
+
+	<div v-if="isColorPickerVisible" ref="colorPickerPanel" class="color-picker-panel">
+		<ColorPicker :model-value="selectedColor" inline @change="onColorChange" />
```
latin-panda, line 185: The color picker was surprisingly challenging at first. I tried to use the native HTML element, but it behaves differently in each browser, offers little customization, and has many bugs in Safari. So at the end I decided to go with PrimeVue.

packages/web-forms/src/components/common/canvas/useCanvasDrawing.ts
```diff
@@ -0,0 +1,359 @@
...
+const OUTPUT_FILE_NAME = 'canvas.jpg';
+const MIN_ZOOM = 0.25;
+const MAX_ZOOM = 4;
+const ZOOM_STEP = 0.25;
+
+/**
+ * Scale image to fit container (if portrait then fit height, if landscape then fit width). Cap at 1 to avoid
+ * upscaling small images, since users can zoom in manually.
+ */
+const calculateBaseScale = (container: Size, image: Size): number => {
```
latin-panda, line 69: When saving, the original image is used, not a scaled version.

packages/web-forms/src/components/common/canvas/useCanvasDrawing.ts
```diff
@@ -0,0 +1,359 @@
...
+	crop?: BoundingBox
+): Promise<File | null> => {
+	if (baseScale <= 0) {
+		return null;
+	}
+
+	const blob = (await stage.toBlob({
+		pixelRatio: 1 / baseScale,
+		mimeType: OUTPUT_MIME_TYPE,
+		quality: 0.9, // Visible quality to 100% but lowers file size significantly.
```
latin-panda, line 94: This was a huge improvement in the final filesize! Image quality is still the same.

packages/web-forms/src/components/common/ActionsInfoDialog.vue
```diff
@@ -1,40 +1,40 @@
+<script lang="ts">
+export interface ActionInfo {
```
latin-panda, line 2: This might be a Vue thing, but this export can't be in the setup script below. I considered creating a "types" file, but it seemed unnecessary just for one interface.

latin-panda:
@garethbowen This is still a draft because I need to rebase it once [your PR ](https://github.com/getodk/central-frontend/pull/1549)is merged to update the package-lock with the new deps and run the CI for Web Forms. But the code is ready to review

packages/web-forms/src/assets/styles/panel-controls.scss
```diff
@@ -1,38 +1,45 @@
...
+}
+
+@mixin panel-control-button {
 	display: flex;
 	align-items: center;
 	justify-content: center;
-	height: 42px;
-	width: 42px;
+	height: 40px;
+	width: 40px;
```
garethbowen, line 42: Is it worth adding a variable for `odk-icon-l`?

packages/web-forms/src/components/common/canvas/CanvasBlock.vue
```diff
@@ -0,0 +1,429 @@
...
+	draw.resetZoom();
+	cancelFlatten();
+	void emitSaveImage();
+};
+
+const onKeyDown = (event: KeyboardEvent) => {
+	if (event.key === 'Escape') {
+		closeFullScreen();
+	}
+};
```
garethbowen: I thought full screen only worked on mobile? If so it's unlikely people will hit the "escape" key?

packages/web-forms/src/components/common/canvas/CanvasComponent.i18n.json
```diff
@@ -0,0 +1,54 @@
...
+  "canvas_block.load_image.error": {
+    "string": "Failed to load image.",
+    "developer_comment": "Error message shown when the background image fails to load in the canvas."
+  },
+  "canvas_block.draw_instruction.message": {
+    "string": "Start drawing here",
+    "developer_comment": "Instruction message shown in draw mode before the user has made any strokes."
+  },
+  "canvas_info_dialog.header.title": {
+    "string": "How to use this tool?",
```
garethbowen: I don't think this should have a question mark. If it were "how do I use this tool?" then it would, but the way it's phrased is a statement not a question.

packages/web-forms/src/components/common/canvas/CanvasComponent.i18n.json
```diff
@@ -0,0 +1,54 @@
...
+    "developer_comment": "Explains the canvas control button that activates or deactivates pan mode. When active, dragging moves the canvas instead of drawing."
+  },
+  "canvas_controls.zoom_in.description": {
+    "string": "Zoom in",
+    "developer_comment": "Explains the canvas control button that zooms in."
+  },
+  "canvas_controls.zoom_out.description": {
+    "string": "Zoom out",
+    "developer_comment": "Explains the canvas control button that zooms out."
+  },
```
garethbowen: There are a few translations which mean literally the same thing as the ones on maps. Is it worth putting them in the main translation json for reuse? It'd make things easier for the translators.

"how to use this tool" and "undo" both jump to mind as well.

packages/web-forms/src/components/common/canvas/CanvasControls.vue
```diff
@@ -0,0 +1,241 @@
...
+import { DEFAULT_STROKE_COLOR } from '@/components/common/canvas/useCanvasDrawing.ts';
+import type { CanvasModeConfig } from '@/components/common/canvas/getModeConfig.ts';
+
+const CONTROL_ICONS = {
+	clearAll: 'mdiTrashCanOutline',
+	closeFullScreen: 'mdiArrowCollapse',
+	colorPicker: 'mdiPalette',
+	openInfo: 'mdiInformationSlabCircleOutline',
+	pan: 'mdiHandBackRightOutline',
+	undo: 'mdiArrowULeftTop',
```
garethbowen, line 18: Similar to the translation question - do you think these should be shared somewhere? In this case I'm thinking about if we decide to change the undo icon to something else it you'd want to make sure you changed it everywhere.

garethbowen (commented):
Ran out of time to finish this morning but will come back this afternoon and finish!

packages/xforms-engine/src/instance/PrimaryInstance.ts
```diff
@@ -168,7 +170,15 @@ export class PrimaryInstance<
...
-		const { mode, initialState, scope, model, secondaryInstances, config } = options;
+		const {
+      mode,
+      initialState,
+      scope,
+      model,
+      secondaryInstances,
+      fetchFormAttachment,
+      config
+    } = options;
```
garethbowen: Spaces not tabs.

garethbowen (changes requested):
Code looks great. I haven't tested this on my machine but have had a look on your demo app. I'll have a closer look once it's running inside central when the other PR is merged.

packages/web-forms/src/assets/styles/panel-controls.scss
latin-panda: I would use `odk-icon-?` for the icon inside. I wouldn't tie it to the button element (I see it different from an "icon" element).
I do see a couple of other buttons with fixed sizes. We could introduce variables for buttons (`odk-btn-?`). Would it be okay to do that in a following PR? I want to see if that makes sense or if we just make it more general (`odk-element-?`). Thinking out loud.

packages/web-forms/src/components/common/canvas/CanvasBlock.vue
latin-panda: Oh yes, old code. Removed

packages/web-forms/src/components/common/canvas/CanvasComponent.i18n.json
latin-panda: Thanks! Fixed

packages/xforms-engine/src/instance/PrimaryInstance.ts
latin-panda: Ah it's my IDE, I'll check if we have lint for that

packages/web-forms/src/components/common/canvas/CanvasControls.vue
latin-panda: I was considering this when I moved the CSS to a shared file.
Because the overlap is small and can be easily identified and replaced by IDE find & replace, I decided to leave it as is. 
Alternatively, we could have a `common/panelControlIcons.ts` that exports these constants.

I'm leaning toward leaving it as is, but what do you think about the alternative?

packages/web-forms/src/components/common/canvas/CanvasComponent.i18n.json
latin-panda: Yes, we could have a `common-actions.i18n.json`.

Actually, do you think these labels should be more descriptive?

Change "Zoom in" -> "Zoom in to draw details easily"

Change "How to use this tool" -> "Guide to using the canvas"

packages/web-forms/src/assets/styles/panel-controls.scss
garethbowen: Fine to leave it for a subsequent PR.

packages/web-forms/src/components/common/canvas/CanvasComponent.i18n.json
garethbowen: No I don't think we need to get too descriptive. I hope people know why they want to zoom in, for example.

I think "canvas" may be a technical term and regular users wouldn't necessarily think of this as a canvas. Most labels for the "how to use this tool" guide in other apps are very simplistic, like "help". So we could just go super simple?

packages/web-forms/src/components/common/canvas/CanvasControls.vue
garethbowen: Yeah I think leave it as is, at least for now. Thanks!

packages/web-forms/src/components/common/canvas/CanvasComponent.i18n.json
latin-panda: fixed

latin-panda:
@garethbowen This is ready for another look. I'm going to add the changeset once the other PR that adds it is merged.

garethbowen:
Don't forget to update the feature matrix json!

Some things I noticed in testing this locally...

### rounded corners

On desktop, the canvas has square corners but the panel has rounded corners which overlaps slightly: 

image

This can be fixed by either rounding the corners of the canvas or adding overflow:hidden somewhere in the dom.

### double top border

On panels without a header (eg: draw) there is 1px border on the panel, and 1px top border on the content, so the border is 2px thick but should be 1.

image

Fix this by removing the border on the content when there is no header content. I think this might only apply when the panel is in a group, because the group has `p-panel-toggleable` which adds the border style. The "all question types" form can reproduce it.

### drawing outside of canvas

I don't know if this is easy or even possible to fix, but I frequently draw outside of the box just by accident. Particularly with the signature. When that happens the drawing doesn't resume when you enter the canvas again - you have to release the mouse button and go again. This only happens on desktop - on mobile it seems to keep working as I'd expect.

### disabled styling

Disabled draw and signature widgets look the same as enabled ones except nothing happens when you mousedown.

image

Consider greying out the canvas or something to give some idea that it's not broken.

### editing a form to remove the picture

I don't know if this is by design but if you submit a picture and then edit the form, you can't delete the original image, you can only clear any additional annotations you've made. This is inconsistent with other fields including the image upload field which lets you delete the previously submitted image. My instinct was the trash can icon should delete the original image and not just my changes. It might make sense for the first click to remove changes you've made, and then the second click to remove the underlying image.

packages/xforms-engine/src/parse/body/appearance/uploadAppearanceParser.ts
```diff
@@ -0,0 +1,5 @@
+import { TokenListParser, type ParsedTokenList } from '../../../lib/TokenListParser.ts';
+
+export const uploadAppearanceParser = new TokenListParser(['annotate', 'draw', 'signature']);
```
garethbowen, line 3: There's overlap with the inputAppearanceParser which also defines these tokens. I think they should be removed from that file now that we have them here?

garethbowen (changes requested):
One minor code comment. A few UX suggestions added in the comment on the PR.

packages/xforms-engine/src/parse/body/appearance/uploadAppearanceParser.ts
latin-panda: Good catch! Done

latin-panda:
> overlap with the inputAppearanceParser 

Fixed!

> rounded corners

Thanks, it's fixed, the attached video shows test in Chrome, Firefox and Safari

> double top border

After fixing the "rounded corners," I didn't notice this issue. Could you please try again? If it occurs again, let me know the steps, environment, or any relevant data that might help. 

> drawing outside of canvas

It's tricky! I haven't found a fix yet. We can keep an eye on user feedback once we publish in the forum to see if more people run into this, and create an issue.

> disabled styling

I've grayed out the canvas for all for consistency: annotate, draw, signature.

Test video:

https://github.com/user-attachments/assets/b6577fb7-72bd-4404-855c-a5d488a360cc

lognaturel:
> lets you delete the previously submitted image

We’re not very consistent about this across question types. For example, you also can’t clear a select one and that has already come up on the forum. I think we should leave that aside for now and consider having a consistent way to clear across question types at some point. It doesn’t feel urgent.

Related Figma threads: [decision to not have X](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn?node-id=4948-14006#1755066519), [FAQ](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn?node-id=4817-1981#1770830040)

latin-panda:
> **editing a form to remove the picture**
I don't know if this is by design but if you submit a picture and then edit the form, you can't delete the original image, you can only clear any additional annotations you've made. This is inconsistent with other fields including the image upload field which lets you delete the previously submitted image. My instinct was the trash can icon should delete the original image and not just my changes. It might make sense for the first click to remove changes you've made, and then the second click to remove the underlying image.

You're right that it behaves differently from the standard image upload, but this was a design choice. Here is a bit of context on why it works this way:

- We originally had an "X" button to remove the image (see screenshot below), but we realized the `annotate` is functionally closer to `draw` and `signature`. So, we updated `annotate` to match, because `draw` and `signature` need a "Clear All" function to remove active strokes rather than deleting the base image (edit submission's answer), aligning with Collect's "reset" option.


- Having an "X" (delete image), a trash can (clear annotations), and an undo button all on the same screen creates too much confusion.

- In Collect, you can long-press on any question type to remove the answer, which is how you remove the selected image. However, this feature is hard to discover. Much like your double-tap idea for deleting the base image, it relies on a hidden way to "delete image". Ultimately, we decided to avoid hidden interactions that are difficult for users to find, opting instead to keep all actions visible.

- We assumed people would be more likely to replace an image, and rarely want to delete (based on Collect, which doesn't have a visible delete image feature) 

However, if we determine that this is needed, we can design a kebab menu (like the one in Repeat Group) with a dropdown and use labels instead of icons ("Clear all", "delete image", etc.)

garethbowen:
Happy to leave it for a later implementation, but I definitely think we'll need some way around it, for example if I accidentally submit a signature that I then want to remove there's no way to do that (other than scribble all over it?).

garethbowen:
> After fixing the "rounded corners," I didn't notice this issue. Could you please try again?

Fixed on my machine too. Freebie!

lognaturel:
> if I accidentally submit a signature that I then want to remove there's no way to do that (other than scribble all over it?)

Thanks for describing what you have in mind, that's super helpful! Currently there's no way for someone to edit their own submission after sending it. Same in Enketo and Collect: the submitter can't generally make edits after finalization (queuing for send while offline). In Collect, it's possible for a project manager to opt in to allowing edits after finalization/submission. That allows users to edit submissions they still have on their device only.

I do think that part of improving support for self-report cases will likely involve some kind of access to edit one's own submissions. But I don't think it will be a priority -- it can make things really complex to manage when you're not sure when a submission is closed for editing.

I don't think the issue you described is present for signature, though, because an editor can use the trashcan icon to clear the canvas. I think it's unique to annotate where it's only possible to replace the annotated image, not clear it altogether.

packages/web-forms/src/assets/styles/panel-controls.scss
```diff
@@ -1,38 +1,45 @@
...
 
-@mixin map-control-button {
+@mixin panel-control-bar-horizontal {
+	@include panel-control-bar;
+	flex-direction: row;
+	left: var(--odk-spacing-m);
+	bottom: var(--odk-spacing-m);
+	background: var(--odk-base-background-color);
+	border: 1px solid var(--odk-border-color);
+	border-radius: var(--odk-spacing-m);
```
garethbowen: Nit: This looks funny to me because it's right beside the panel radius which is 6px so they don't line up. I propose we use --odk-radius for all rounded corners.

packages/web-forms/src/components/common/canvas/AsyncCanvas.vue
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.canvas-disabled-overlay {
+  position: absolute;
+  inset: 0;
+  background-color: rgba(from var(--odk-muted-background-color) r g b / 0.7);
+  z-index: var(--odk-z-index-top-overlay);
+  pointer-events: none;
+}
+</style>
```
garethbowen, line 66: Spaces not tabs...

We're going to have to figure out linting for this sooner rather than later.

garethbowen (approved):
A couple of minor issues inline. Approved to unblock.

PR #1554, opened 2026-05-22
fix: Scroll to top when a new instance is requested in Web Forms

### LINKED ISSUE #410: Scroll to top when a new instance is requested
Currently if a new instance is requested after submission, that has no effect on the UI. The user should be scrolled to the top.

latin-panda:
Closes https://github.com/getodk/web-forms/issues/410

##### Before fix
https://github.com/user-attachments/assets/93cd5e61-7ba1-43f5-843a-4abdc50c4223

##### After fix
https://github.com/user-attachments/assets/4bf2db6e-93f9-43e1-ba88-a1bf6ad5b263


latin-panda:
@garethbowen This is still a draft because I need to rebase it once https://github.com/getodk/central-frontend/pull/1549 is merged. Please feel free to review now or later, once the branch is rebased.

lognaturel:
It's probably in practice OK to do this just through DOM/UI manipulation in this context because we should be pretty confident that after submission no control is active, no validation errors are shown, user intent is always to go to the top. But I think we should carefully consider whether we want to set the precedent of the UI layer determining form position.

I guess this comes down to how closely related this is to other concepts that involve form position:
* directly jumping to the first question with an error
* navigating between questions in `pages` mode
* direct linking to a question in the URL
* jumping to a question from the table of contents
* resuming a draft at the last question that had focus

I feel strongly that all of those should involve the engine. That is, the user expresses intent in the UI (e.g. clicks on a question link), the engine gets that request, the engine updates the active question, the UI reacts to that change and scrolls as needed. So "jump to question/form location X" is a navigation state transition that the UI layer responds to by scrolling/page flipping.

Up to you and @garethbowen to consider whether you feel this approach is ok for now given that longer-term goal.

garethbowen:
I'm in two minds here...

The case for putting it in the web-forms package is that the visual positioning is not the concern of the engine. Some of those examples could be opinionated and different clients may want different behaviour in those situations. Furthermore the engine doesn't (currently) have all the information it needs to make the right call in all situations, for example it doesn't know which question currently has focus.

The case for putting it in the engine is that every single use case in your comment has the same outcome, which is to navigate so that a particular field is in view. One exception is the scroll to top (this exact issue) but arguably scrolling to the first element is better because if the user is filling in multiple forms they don't need to re-read the preamble and if the focus ends up on the first element they can just start typing. Given this, a single client side callback for scroll element into view would make it very easy to solve all of these cases in a consistent way. If the client does want to have custom control this can probably be handled in the callback.

Given we only have one client right now I'm leaning towards putting it in the engine which is more work up front but will make the other issues trivial to implement.

lognaturel:
> scrolling to the first element is better because if the user is filling in multiple forms they don't need to re-read the preamble

I think we probably want an engine level "start of form" concept (if we don't already) that the client can then interpret as it wants. "Fill another" is most likely to be useful in an enumerator-mediated context and then the preamble might be something they need to read out loud. I think the end result of this PR is what we want (scroll to the very top).

> putting it in the engine which is more work up front but will make the other issues trivial to implement.

Exactly. My concern is that if we start with the UI being able to affect the form position directly and then we also give the engine some of that responsibility we could end up in states where the UI and engine believe they're in different places and that could be hard to troubleshoot.

latin-panda:
@lognaturel  @garethbowen Thanks for the review and for sharing your thoughts! 

I agree the engine should own logical navigation state: which question is active, where the validation errors are, what the suggested next step is after submission. That's clearly engine territory.

But visual presentation, including scroll, is a client concern. The engine can publish state and intents (e.g. "submission complete, next = new instance", "active question changed to X"), and the client decides what to do with that: whether to scroll, how (smooth vs. instant), focus, timing, whether the current UX even calls for scrolling at all.

 It's two clean responsibilities, and since they don't overlap, there's no path for the two layers to disagree on logical position (that would only happen if both tried to mutate it). 

If different clients legitimately want different behavior, the decision belongs in the client, not in a callback the engine invokes. The engine just needs to make its state observable enough that the client can react.

For this PR specifically, `submissionResult.next === POST_SUBMIT__NEW_INSTANCE` is already the kind of intent signal that lets the client do its job. What's here is the client responding to a published state transition from the engine. 

One more thing on shape: as more of these cases come up, I think the right place for this on the client is a composable (useFormNavigation()) that subscribes to engine state/intents and centralizes the logic (scroll-to-top on new instance, scroll-to-first-error on validation, scroll-to-node on TOC jump, etc.). So it's one interpreter reacting to engine state, testable in isolation, and swappable if other clients want different behavior. I wouldn't build it for this PR (4 lines doesn't justify the abstraction yet), but I'd extract it the moment a second case lands.

pd. I'll push the changeset after the other PR adding it is merged :)

garethbowen:
@latin-panda If I understand your comment correctly you're agreeing (more or less) with the eventual state that @lognaturel proposed, but suggesting we go with the client only implementation for now, then when we develop the next navigation event we'll refactor this one to comply?

latin-panda:
Mostly, yes! Just one clarification on the "eventual state":
  
I'm on board with the engine handling logical navigation state. However, I'd keep scroll/visual behavior on the client side in both phases, not just for now. Here's the plan I'm thinking:

- In this PR: all client-side, reacting to submissionResult.next.
- Once a second case lands: still client-side, but organized into a useFormNavigation() composable that listens to engine state and intents and centralizes scroll/focus management. At that point, we'll have enough cases to design and test the abstraction properly.

 Either way, the engine's interface stays simple: it just needs to expose its state in a way the client can observe and react to.

garethbowen:
> still client-side, but organized into a useFormNavigation() composable that listens to engine state and intents and centralizes scroll/focus management.

This is exactly what I was thinking too.

> In this PR: all client-side, reacting to submissionResult.next.

Works for me.

packages/web-forms/src/components/OdkWebForm.vue
```diff
@@ -83,6 +84,9 @@ const hostSubmissionResultCallbackFactory = (
 			trackDevice: props.trackDevice,
 		};
 		state.value = updateSubmittedFormState(submissionResult, currentState, options);
+		if (submissionResult?.next === POST_SUBMIT__NEW_INSTANCE) {
+			document.scrollingElement?.scrollTo(0, 0);
```
garethbowen: Have you thought about using animation?

```suggestion
			document.scrollingElement?.scrollTo({ top: 0, left: 0, behavior: 'smooth' });
```

I prefer this because it gives a clue as to where the page jumped to.

garethbowen (commented):
One suggestion inline, and it looks like the e2e tests are failing?

latin-panda:
@garethbowen I've applied the feedback, it's ready for another review. 

I'll rebase the PR once the e2e tests are fixed.

garethbowen (approved):
Very nice.

FYI lint is now failing on whitespace issues.

PR #1555, opened 2026-05-22
Form Draft Redesign Part 1: Remove icon, warnings and dots from FormEditSection component

sadiqkhoja:
This just removes the icon, warnings and dots from FormEditSection component and adds rounded border around it.

image

In the next PRs:
-  Make border conditional
-  Move FormEditAttachments and FormEditEntities from FormEditDef to FormEdit
-  Move version number from body to section heading
-  Hide actions bar when there is no file in FormUpload component
-  Set fullWidth to false for draft page
-  Show sections on FormEdit vertically when there is no active draft
-  Do not use FormEditSection in FormEditDraftControl and align center


apps/central/src/components/form/edit/def.vue
```diff
@@ -39,9 +39,6 @@ except according to the terms contained in the LICENSE file.
           </button>
         </div>
       </div>
-      <div id="form-edit-def-within">
-        <span class="icon-file-code-o"></span>{{ $t('withinDef') }}
```
matthew-white, line 43: Is it possible to remove this i18n message `withinDef`?

apps/central/src/components/form/edit/def.vue
```diff
@@ -39,9 +39,6 @@ except according to the terms contained in the LICENSE file.
           </button>
         </div>
       </div>
-      <div id="form-edit-def-within">
```
matthew-white, line 42: There are styles related to `#form-edit-def-within` below. Can those be removed now?

apps/central/src/components/form/edit/section.vue
```diff
@@ -11,106 +11,39 @@ except according to the terms contained in the LICENSE file.
...
-  // 44px = 10px of padding on the left and right + 24px for the icon. The width
-  // of the icon varies significantly, but the height of the icon is a
-  // consistent 24px. If the icon is wider than 24px, that will just eat into
-  // the padding.
-  width: 44px;
-  height: 44px;
-
-  // Use flexbox to center the icon.
+  border-radius: 12px;
+  border: 1px solid var(--Central-Grey-2, #E3E4E4);
```
matthew-white: Do we have a CSS property `--Central-Grey-2`? How about using the Sass variable with the same name?

apps/central/src/components/form/edit/section.vue
```diff
@@ -11,106 +11,39 @@ except according to the terms contained in the LICENSE file.
...
+  padding: 20px;
+  margin-bottom: 10px;
+  column-gap: 15px;
```
matthew-white: If the `flex-direction` is `column` (as below), I think `column-gap` will have no effect. Is that right? Are there still multiple columns?

apps/central/src/components/form/edit/section.vue
matthew-white: Now that the `.icon-circle` has been removed, I think this `column-gap` can go away. `.form-edit-section-tag` only has one child now.

matthew-white (approved):

apps/central/src/components/form/edit/section.vue
sadiqkhoja: yeah `column-gap` has no effect

apps/central/src/components/form/edit/def.vue
matthew-white: Ah it looks like it's removed in the follow-up PR #1556, that sounds fine to me. :+1:

PR #1556, opened 2026-05-22
Form Draft Redesign Part 2: Restructure FormEditDef component

sadiqkhoja:
Dependent on https://github.com/getodk/central-frontend/pull/1555

- Moved FormEditAttachments and FormEditEntities from FormEditDef to FormEdit
- Show FileUpload always
- Moved preview and definition button to the section header area
- Remove version block, will add it to the title in the next PR

image


apps/central/src/components/form/edit/def.vue
```diff
@@ -14,48 +14,28 @@ except according to the terms contained in the LICENSE file.
...
-        </div>
-        <div>
-          <enketo-preview :form-version="formDraft" outlined/>
-          <form-version-def-dropdown :version="formDraft" outlined
-            @view-xml="viewXml.show()"/>
-          <button id="form-edit-upload-button" type="button"
-            class="btn btn-primary" @click="setUploadSectionState(true)">
-            <span class="icon-upload"></span>{{ $t('action.upload') }}
-          </button>
+          <form-upload @success="afterUpload"/>
```
matthew-white, line 24: It looks like there's no `@cancel` anymore? Should we hide the Cancel button in `FormUpload`? (That could be done in a follow-up PR, just wondering.)

apps/central/transifex/strings_en.json
```diff
@@ -3466,14 +3466,7 @@
...
       "versionName": {
-        "string": "Version name: {name}",
-        "developer_comment": "This is shown for a Form version."
-      },
-      "action": {
-        "upload": {
-          "string": "Upload new Form Definition",
-          "developer_comment": "This is the text for an action, for example, the text of a button."
-        }
+        "string": "Version name: {name}"
```
matthew-white: Wasn't this message removed in the component? I wonder why it's still showing up in this file. ~If there's a problem with our Transifex scripts/workflow, would you mind filing an issue about that?~

matthew-white (approved):

apps/central/transifex/strings_en.json
matthew-white: Ah I think the message is being reused/copied to `FormEditPublishedVersion` via `@transifexKey`. That explains why it's not being removed here. Given that, could you please move the comment that the message used to have and add it to `FormEditPublishedVersion`? ("This is shown for a Form version.") The end result is that you should see that nothing about the string has changed in transifex/strings_en.json, including the `developer_comment`.

apps/central/src/components/form/edit/def.vue
```diff
@@ -76,65 +56,25 @@ const changed = computed(() =>
...
-  }
-
-  > :nth-child(2) {
-    flex-shrink: 0;
-    margin-left: auto;
-  }
-
-  .form-version-def-dropdown, #form-edit-upload-button { margin-left: 5px; }
-}
+.form-edit-section-actions .form-version-def-dropdown { margin-left: 5px; }
```
matthew-white:
```suggestion
#form-edit-def .form-edit-section-actions .form-version-def-dropdown { margin-left: 5px; }
```

I think we should add an id selector here. Otherwise this style could affect other components as well. Maybe not a problem in practice in this case, I just think it's good practice to reference the component in the selector somehow.

matthew-white (approved):
One more comment, it's still looking great.

apps/central/src/components/form/edit/def.vue
sadiqkhoja: In the upcoming PR, I will hide submit and cancel button by default and they will be visible once there is a file to be uploaded.

PR #1558, opened 2026-05-24
chore: update supported node version to 24.14.1

### LINKED ISSUE #809: Remove support for Node 20 and Node 22
### Description

We support and test against Node 20 (EOL) and 22 (old). We should drop these and only support Node 24 from now on. We will start testing against Node 26 soon so keep the CI matrix so it's easy to add that later. This should be done before 1.0.0 so library users have a clear indication of what Node version to use.

### LINKED ISSUE #1806: Update Node version in dev to match prod
Right now, the version of Node in prod is 24.14.1. However, in dev, it's 24.13.0. This issue is to update the version in dev to match prod.

I've made a note in #1805 about updating Node in dev so that we update prod and dev at the same time for v2026.2. CC @sadiqkhoja, who will likely be the one to work on #1805.

garethbowen:
Closes getodk/central#1806
Closes getodk/web-forms#809


README.md
```diff
@@ -31,7 +31,7 @@ We need your help to make ODK Central Frontend as useful as possible! Please see
 
 Follow these steps to set up your development environment:
 
-- Install Node.js 20.
+- Install Node.js 24+
```
alxndrsn: Should this say


```suggestion
- Install [Volta](https://volta.sh/)
```

alxndrsn (approved):
Looks great.  Some ideas for the future:

* remove matrix builds - I don't think there's any benefit testing on multiple node versions, as they're only used in dev
* some kind of check to ensure that the node version is consistent across all different sources

garethbowen:
> remove matrix builds

Well theoretically web-forms packages have been designed to be used as npm libraries so a wider support matrix makes sense, so I've left the matrix there so it's easy to add the next current version of Node.

> some kind of check to ensure that the node version is consistent across all different sources

There is a way to have a variable which would make this easier in future which would be my preference. I had a brief look into it before backing away again.

PR #1559, opened 2026-05-24
chore: update README with feature matrix

garethbowen:
This puts the feature matrix markdown in the /packages/web-forms/README.md which is probably the right place.


latin-panda (approved):
Looks good!

PR #1561, opened 2026-05-24
chore: merge scenario package into xforms-engine package

garethbowen:

Mostly CI.

This was done to remove a package that only existed for integration testing xforms-engine. Moving this into an integration testing folder in xforms-engine reduces complexity and overhead of managing an additional package which provide no value.

None.

No.


latin-panda (approved):
Everything looks good to me

PR #1562, opened 2026-05-25
feat(#448): support encrypting submission payload

### LINKED ISSUE #448: Add submission data encryption and display encryption status
### Description
Web Forms currently does not encrypt submission data, though the [base64EncryptedKey is already extracted from the XML](https://github.com/getodk/web-forms/blob/8dc572af77d00ee71d351d6c4d980245181bf0fe/packages/xforms-engine/src/parse/model/SubmissionDefinition.ts#L37) and passed to the host app in the payload. 

Consider implementing the encryption of submission data in Web Forms and adding a UI indicator to inform users if their submissions are encrypted. 

Encryption could be implemented in Web Forms or in the host app (e.g. Central Frontend). However, implementing it in Web Forms would benefit other clients by eliminating the need for them to develop their own encryption solutions.

---

### Checklist
  - [x] Yes. Describe the UI/UX requirements or attach a Figma link in the "Design" section.
  - [ ] No
  - [ ] No
  - [x] No
---

### User Stories

- **As a** program manager
  **I want** to know if the form's data is being encrypted 
  **so that** I can add the encryption key in the form 

---

### Design

- **Figma**: [Insert URL]

---

### Dependencies

-

---

### Additional Notes
- [Enketo has an encryptor](https://github.com/enketo/enketo/blob/ba2bec44acf561f278fd022aa361447674c1d9a8/packages/enketo-express/public/js/src/module/encryptor.js#L99) that we can use as reference. 
- Specs: https://getodk.github.io/xforms-spec/#submission-attributes
- Some notes at https://github.com/getodk/web-forms/issues/318

---

### Related Issues

- Central: [Cleartext submissions can be accepted to an encrypted form](https://github.com/getodk/central/issues/782#top)

garethbowen:
Closes getodk/web-forms#448

Various testing, manual and automatic

Encryption which was previously ignored silently is now applied when filling out forms using web-forms

### DIFF
--- a/e2e-tests/global.setup.js
+++ b/e2e-tests/global.setup.js
@@ -1,4 +1,5 @@
 import { test as setup, expect } from '@playwright/test';
+import { ENCRYPTION_SECRET } from './util';
 
 const appUrl = process.env.ODK_URL;
 const user = process.env.ODK_USER;
@@ -40,7 +41,7 @@ const createProject = async(request, encrypted) => {
           Authorization: `Basic ${credentials}`
         },
         data: {
-          passphrase: "encryptionsecret",
+          passphrase: ENCRYPTION_SECRET,
           hint: "it was a secret"
         }
       });
--- a/e2e-tests/global.setup.js
+++ b/e2e-tests/global.setup.js
@@ -5,7 +5,7 @@ const user = process.env.ODK_USER;
 const password = process.env.ODK_PASSWORD;
 const credentials = Buffer.from(`${user}:${password}`, 'utf-8').toString('base64');
 
-setup('create new project', async ({ request }) => {
+const createProject = async(request, encrypted) => {
   try {
     const createProjectResponse = await request.post(`${appUrl}/v1/projects`, {
       headers: {
@@ -32,7 +32,30 @@ setup('create new project', async ({ request }) => {
     const project = await createProjectResponse.json();
 
     expect(project.id).not.toBeFalsy();
-    process.env.PROJECT_ID = project.id;
+
+    if (encrypted) {
+      const encryptionResponse = await request.post(`${appUrl}/v1/projects/${project.id}/key`, {
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Basic ${credentials}`
+        },
+        data: {
+          passphrase: "encryptionsecret",
+          hint: "it was a secret"
+        }
+      });
+      expect(encryptionResponse).toBeOK();
+      const getProjectResponse = await request.get(`${appUrl}/v1/projects/${project.id}`, {
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Basic ${credentials}`
+        }
+      });
+      expect(getProjectResponse).toBeOK();
+      const getProjectBody = await getProjectResponse.json();
+      process.env.ENCRYPTION_KEY_ID = getProjectBody.keyId;
+    }
+    return project.id;
   } catch (err) {
     if (err.message.includes('ECONNREFUSED')) {
       // eslint-disable-next-line preserve-caught-error
@@ -48,4 +71,9 @@ setup('create new project', async ({ request }) => {
       throw err;
     }
   }
+};
+
+setup('create new project', async ({ request }) => {
+  process.env.PROJECT_ID = await createProject(request, false);
+  process.env.PROJECT_ID_ENCRYPTED = await createProject(request, true);
 });
--- a/e2e-tests/global.teardown.js
+++ b/e2e-tests/global.teardown.js
@@ -19,6 +19,6 @@ teardown('delete project', async () => {
       .then(res => res.json())
       .then(res => res.success);
   });
-  const results = await Promise.all(promises);
+  const results = await Promise.allSettled(promises);
   results.forEach(result => expect(result).toEqual(true));
 });
--- a/e2e-tests/global.teardown.js
+++ b/e2e-tests/global.teardown.js
@@ -8,10 +8,8 @@ const projectId = process.env.PROJECT_ID;
 const projectIdEncrypted = process.env.PROJECT_ID_ENCRYPTED;
 
 teardown('delete project', async () => {
-  for (const project of [ projectId, projectIdEncrypted ]) {
-    if (!project) return;
-
-    const result = await fetch(`${appUrl}/v1/projects/${project}`, {
+  const promises = [ projectId, projectIdEncrypted ].map((project) => {
+    return fetch(`${appUrl}/v1/projects/${project}`, {
       method: 'DELETE',
       headers: {
         'Content-Type': 'application/json',
@@ -19,8 +17,8 @@ teardown('delete project', async () => {
       }
     })
       .then(res => res.json())
-      .then(r => r.success);
-
-    expect(result).toEqual(true);
-  }
+      .then(res => res.success);
+  });
+  const results = await Promise.all(promises);
+  results.forEach(result => expect(result).toEqual(true));
 });
--- a/e2e-tests/global.teardown.js
+++ b/e2e-tests/global.teardown.js
@@ -20,5 +20,5 @@ teardown('delete project', async () => {
       .then(res => res.success);
   });
   const results = await Promise.allSettled(promises);
-  results.forEach(result => expect(result).toEqual(true));
+  results.forEach(result => expect(result.value).toEqual(true));
 });
--- a/e2e-tests/global.teardown.js
+++ b/e2e-tests/global.teardown.js
@@ -5,19 +5,22 @@ const user = process.env.ODK_USER;
 const password = process.env.ODK_PASSWORD;
 const credentials = Buffer.from(`${user}:${password}`, 'utf-8').toString('base64');
 const projectId = process.env.PROJECT_ID;
+const projectIdEncrypted = process.env.PROJECT_ID_ENCRYPTED;
 
 teardown('delete project', async () => {
-  if (!projectId) return;
+  for (const project of [ projectId, projectIdEncrypted ]) {
+    if (!project) return;
 
-  const result = await fetch(`${appUrl}/v1/projects/${projectId}`, {
-    method: 'DELETE',
-    headers: {
-      'Content-Type': 'application/json',
-      Authorization: `Basic ${credentials}`
-    }
-  })
-    .then(res => res.json())
-    .then(r => r.success);
+    const result = await fetch(`${appUrl}/v1/projects/${project}`, {
+      method: 'DELETE',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Basic ${credentials}`
+      }
+    })
+      .then(res => res.json())
+      .then(r => r.success);
 
-  expect(result).toEqual(true);
+    expect(result).toEqual(true);
+  }
 });
--- a/e2e-tests/tests/encrypted-submission.spec.js
+++ b/e2e-tests/tests/encrypted-submission.spec.js
@@ -1,9 +1,8 @@
 import { test, expect } from '@playwright/test';
 import BackendClient from '../backend-client';
+import { login, ENCRYPTION_SECRET } from '../util';
 
 const appUrl = process.env.ODK_URL;
-const user = process.env.ODK_USER;
-const password = process.env.ODK_PASSWORD;
 const projectId = process.env.PROJECT_ID_ENCRYPTED;
 
 const FIRSTNAME_COLUMN_INDEX = 2;
@@ -16,33 +15,20 @@ test.beforeAll(async ({ playwright }, testInfo) => {
   backendClient = new BackendClient(playwright, `${testInfo.project.name}_wf`, projectId);
   await backendClient.alwaysHideModal();
   publishedForm = await backendClient.createForm();
-  await backendClient.setWebForms(publishedForm.xmlFormId, true); // this shouldn't be necessary - remove before pushing
+  await backendClient.setWebForms(publishedForm.xmlFormId, true);
 });
 
 test.afterAll(async () => {
   await backendClient.dispose();
 })
 
-const login = async (page) => {
-  await page.goto(appUrl);
-  await expect(page.getByRole('heading', { name: 'Welcome to ODK Central' })).toBeVisible();
-
-  await page.getByPlaceholder('email address').fill(user);
-  await page.getByPlaceholder('password').fill(password);
-
-  await page.getByRole('button', { name: 'Log in' }).click();
-
-  await page.waitForURL(appUrl);
-};
-
-const checkResponse = async (response, column, values) => {
+const assertCsvResponse = async (response, column, expected) => {
   expect(response).toBeOK();
   const text = await response.text();
   const rows = text.split('\n');
-  for (let i = 0; i < values.length; i++) {
-    const actual = rows[i + 1].split(',')[column];
-    expect(actual).toBe(values[i]);
-  }
+  rows.shift(); // remove the header row
+  const actual = rows.map(r => r.split(',')[column]);
+  expect(actual).to.deep.equal(expected);
 };
 
 test.describe('ODK Web Forms Encrypted', () => {
@@ -63,15 +49,15 @@ test.describe('ODK Web Forms Encrypted', () => {
 
     // attempt to download the file with no key
     const unencryptedResponse = await backendClient.downloadCsv(publishedForm.xmlFormId);
-    await checkResponse(unencryptedResponse, STATUS_COLUMN_INDEX, ['not decrypted', 'not decrypted']);
+    await assertCsvResponse(unencryptedResponse, STATUS_COLUMN_INDEX, ['not decrypted', 'not decrypted']);
 
     // attempt to download the csv with wrong passphrase
     const wrongPassphraseResponse = await backendClient.downloadCsv(publishedForm.xmlFormId, 'hax');
     expect(wrongPassphraseResponse).not.toBeOK();
 
     // attempt to download the csv with correct passphrase
-    const rightPassphraseResponse = await backendClient.downloadCsv(publishedForm.xmlFormId, 'encryptionsecret');
-    await checkResponse(rightPassphraseResponse, FIRSTNAME_COLUMN_INDEX, ['Jane Doe', 'John Doe']); // csv is reverse date order
+    const rightPassphraseResponse = await backendClient.downloadCsv(publishedForm.xmlFormId, ENCRYPTION_SECRET);
+    await assertCsvResponse(rightPassphraseResponse, FIRSTNAME_COLUMN_INDEX, ['Jane Doe', 'John Doe']); // csv is reverse date order
   });
 
 });
--- a/e2e-tests/tests/encrypted-submission.spec.js
+++ b/e2e-tests/tests/encrypted-submission.spec.js
@@ -46,7 +46,7 @@ const checkResponse = async (response, column, values) => {
 };
 
 test.describe('ODK Web Forms Encrypted', () => {
-  test('allows multiple submission', async ({ page }) => {
+  test('requires passphrase to download content', async ({ page }) => {
     await login(page);
 
     await page.goto(`${appUrl}/projects/${projectId}/forms/${publishedForm.xmlFormId}/submissions/new`);
--- a/e2e-tests/tests/encrypted-submission.spec.js
+++ b/e2e-tests/tests/encrypted-submission.spec.js
@@ -28,7 +28,7 @@ const assertCsvResponse = async (response, column, expected) => {
   const rows = text.split('\n');
   rows.shift(); // remove the header row
   const actual = rows.map(r => r.split(',')[column]);
-  expect(actual).to.deep.equal(expected);
+  expect(actual).toEqual(expected);
 };
 
 test.describe('ODK Web Forms Encrypted', () => {
--- a/e2e-tests/tests/encrypted-submission.spec.js
+++ b/e2e-tests/tests/encrypted-submission.spec.js
@@ -1,42 +1,72 @@
 import { test, expect } from '@playwright/test';
 import BackendClient from '../backend-client';
 import { login, ENCRYPTION_SECRET } from '../util';
+import Zip from 'adm-zip';
 
 const appUrl = process.env.ODK_URL;
 const projectId = process.env.PROJECT_ID_ENCRYPTED;
 
-const FIRSTNAME_COLUMN_INDEX = 2;
-const STATUS_COLUMN_INDEX = 8;
-
-let publishedForm;
+let simpleForm;
+let formWithAttachment;
 let backendClient;
 
 test.beforeAll(async ({ playwright }, testInfo) => {
   backendClient = new BackendClient(playwright, `${testInfo.project.name}_wf`, projectId);
   await backendClient.alwaysHideModal();
-  publishedForm = await backendClient.createForm();
-  await backendClient.setWebForms(publishedForm.xmlFormId, true);
+  simpleForm = await backendClient.createForm();
+  await backendClient.setWebForms(simpleForm.xmlFormId, true);
+  formWithAttachment = await backendClient.createAttachmentForm();
+  await backendClient.setWebForms(formWithAttachment.xmlFormId, true);
 });
 
 test.afterAll(async () => {
   await backendClient.dispose();
-})
+});
 
-const assertCsvResponse = async (response, column, expected) => {
-  expect(response).toBeOK();
-  const text = await response.text();
+const assertCsv = (text, column, expected) => {
   const rows = text.split('\n');
   rows.shift(); // remove the header row
   const actual = rows.map(r => r.split(',')[column]);
   expect(actual).toEqual(expected);
 };
 
+const assertCsvResponse = async (response, column, expected) => {
+  expect(response).toBeOK();
+  const text = await response.text();
+  assertCsv(text, column, expected);
+};
+
+const assertZipResponse = async (response, csvColumn, csvExpected, attachmentExpected) => {
+  expect(response).toBeOK();
+
+  const body = await response.body();
+  const zip = new Zip(body);
+  const entries = zip.getEntries();
+  let count = 0;
+  for (const entry of entries) {
+    const name = entry.entryName;
+    const text = zip.readAsText(entry);
+    if (name.endsWith('.csv')) {
+      assertCsv(text, csvColumn, csvExpected);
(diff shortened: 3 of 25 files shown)


garethbowen:
@alxndrsn I think this is ready for review. It's a direct port of the PR you reviewed in web-forms with the addition of an e2e test to actually show it works.

e2e-tests/global.setup.js
```diff
@@ -32,7 +32,30 @@ setup('create new project', async ({ request }) => {
...
+      expect(encryptionResponse).toBeOK();
+      const getProjectResponse = await request.get(`${appUrl}/v1/projects/${project.id}`, {
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Basic ${credentials}`
+        }
+      });
+      expect(getProjectResponse).toBeOK();
+      const getProjectBody = await getProjectResponse.json();
+      process.env.ENCRYPTION_KEY_ID = getProjectBody.keyId;
```
alxndrsn, line 57: Storing runtime globals as environment variables seems odd.

e2e-tests/global.setup.js
```diff
@@ -48,4 +71,9 @@ setup('create new project', async ({ request }) => {
       throw err;
     }
   }
+};
+
+setup('create new project', async ({ request }) => {
+  process.env.PROJECT_ID = await createProject(request, false);
+  process.env.PROJECT_ID_ENCRYPTED = await createProject(request, true);
```
alxndrsn, line 79: ...but perhaps with precedent?

e2e-tests/global.teardown.js
```diff
@@ -5,19 +5,22 @@ const user = process.env.ODK_USER;
...
+        'Content-Type': 'application/json',
+        Authorization: `Basic ${credentials}`
+      }
+    })
+      .then(res => res.json())
+      .then(r => r.success);
 
-  expect(result).toEqual(true);
+    expect(result).toEqual(true);
+  }
```
alxndrsn: Probably not significant, but this teardown looks like it'll throw or fail when e.g. the first project fails to delete.

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/symmetric.test.ts
```diff
@@ -0,0 +1,96 @@
...
+				const wordArray = CryptoJS.lib.WordArray.create(words, 6);
+
+				const result = wordArrayToArrayBuffer(wordArray);
+
+				expect(result.length).toBe(6);
+				expect(Array.from(result)).toEqual([0x11, 0x22, 0x33, 0x44, 0x55, 0x66]);
+			});
+
+			it('should handle high-bit (negative) signed integers correctly', () => {
+				const words = [0xaabbccdd >>> 0];
```
alxndrsn: What's the purpose of the 0-bit shift here?  I can't work out what it achieves:

```
> 0xaabbccdd >>> 0
2864434397
> 0xaabbccdd
2864434397
```

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/symmetric.test.ts
```diff
@@ -0,0 +1,96 @@
...
+				expect(wordArray.sigBytes).toBe(8);
+			});
+
+			it('should handle data that is not a multiple of 4 bytes (padding)', () => {
+				const data = new Uint8Array([0xaa, 0xbb, 0xcc]);
+
+				const wordArray = arrayBufferToWordArray(data);
+
+				expect(wordArray.words.length).toBe(1);
+				const unsignedWord = wordArray.words[0]! >>> 0;
```
alxndrsn: As below - unclear what the shift is achieving.

e2e-tests/global.setup.js
garethbowen: Agreed. It's the convention these tests are using. I'm hoping to add many more e2e tests now that wf can be developed alongside cf and have a few ideas for making the suite easier to work with.

e2e-tests/global.teardown.js
garethbowen: Nice catch. I've converted it to run all the requests in parallel, then wait, then expect all results to be true.

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/symmetric.test.ts
garethbowen: You're absolutely right, with this number the `>>> 0` does nothing. Removed.

packages/xforms-engine/test/lib/client-reactivity/instance-state/quarantine/symmetric.test.ts
garethbowen: Here it actually does change the number from a negative to its positive counterpart, in this case `-1430533120` => `2864434176` . It's effectively underflowing the negative number into the unsigned 32 bit integer.

I've added a comment to explain what it's actually doing.

alxndrsn: Feels ignorant to have asked - thanks for clarifying.  An alternative to a comment could be a function, e.g.:

```js
function asUnsignedInt(n) {
  return n >>> 0;
}
```

e2e-tests/global.teardown.js
```diff
@@ -5,19 +5,20 @@ const user = process.env.ODK_USER;
...
+      method: 'DELETE',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Basic ${credentials}`
+      }
+    })
+      .then(res => res.json())
+      .then(res => res.success);
+  });
+  const results = await Promise.all(promises);
```
alxndrsn: For even more defence, `Promise.allSettled()`

e2e-tests/tests/encrypted-submission.spec.js
```diff
@@ -0,0 +1,77 @@
...
+const STATUS_COLUMN_INDEX = 8;
+
+let publishedForm;
+let backendClient;
+
+test.beforeAll(async ({ playwright }, testInfo) => {
+  backendClient = new BackendClient(playwright, `${testInfo.project.name}_wf`, projectId);
+  await backendClient.alwaysHideModal();
+  publishedForm = await backendClient.createForm();
+  await backendClient.setWebForms(publishedForm.xmlFormId, true); // this shouldn't be necessary - remove before pushing
```
alxndrsn: Did this turn out to be necessary in the end?

e2e-tests/tests/encrypted-submission.spec.js
```diff
@@ -0,0 +1,77 @@
...
+});
+
+test.afterAll(async () => {
+  await backendClient.dispose();
+})
+
+const login = async (page) => {
```
alxndrsn: This function seems to be duplicated in a few places:

* `e2e-tests/tests/encrypted-submission.spec.js`
* `e2e-tests/tests/enketo.spec.js`
* `e2e-tests/tests/web-forms.spec.js`
* `e2e-tests/util.js`

e2e-tests/tests/encrypted-submission.spec.js
```diff
@@ -0,0 +1,77 @@
...
+
+  await page.getByPlaceholder('email address').fill(user);
+  await page.getByPlaceholder('password').fill(password);
+
+  await page.getByRole('button', { name: 'Log in' }).click();
+
+  await page.waitForURL(appUrl);
+};
+
+const checkResponse = async (response, column, values) => {
```
alxndrsn:
```suggestion
const assertCsvResponse = async (response, columnIdx, expectedValues) => {
```

Maybe more idiomatic?

e2e-tests/tests/encrypted-submission.spec.js
```diff
@@ -0,0 +1,77 @@
...
+  expect(response).toBeOK();
+  const text = await response.text();
+  const rows = text.split('\n');
+  for (let i = 0; i < values.length; i++) {
+    const actual = rows[i + 1].split(',')[column];
+    expect(actual).toBe(values[i]);
+  }
```
alxndrsn: I'm unclear how this is likely to fail, but I wonder if it could be more revealing to compare the full values simultaneously, something like:

```js
expect(rows.map(r => r.split(',')[column])).to.deep.equal(expectedValues);
```

e2e-tests/tests/encrypted-submission.spec.js
```diff
@@ -0,0 +1,77 @@
...
+    // attempt to download the file with no key
+    const unencryptedResponse = await backendClient.downloadCsv(publishedForm.xmlFormId);
+    await checkResponse(unencryptedResponse, STATUS_COLUMN_INDEX, ['not decrypted', 'not decrypted']);
+
+    // attempt to download the csv with wrong passphrase
+    const wrongPassphraseResponse = await backendClient.downloadCsv(publishedForm.xmlFormId, 'hax');
+    expect(wrongPassphraseResponse).not.toBeOK();
+
+    // attempt to download the csv with correct passphrase
+    const rightPassphraseResponse = await backendClient.downloadCsv(publishedForm.xmlFormId, 'encryptionsecret');
```
alxndrsn: Should `encryptionsecret` be another env var? 🙃

e2e-tests/tests/encrypted-submission.spec.js
```diff
@@ -0,0 +1,77 @@
...
+    await checkResponse(rightPassphraseResponse, FIRSTNAME_COLUMN_INDEX, ['Jane Doe', 'John Doe']); // csv is reverse date order
```
alxndrsn: Test cases look good :+1:

Would it be weird to have done the setup in `beforeAll()` (or equivalent), and then write each test case as a separate `test(...)` function?

packages/xforms-engine/test/integration/submission-encryption.test.ts
```diff
@@ -0,0 +1,81 @@
...
+		const scenario = await buildSubmissionPayloadScenario({
+			submissionElements: [t(`submission base64RsaPublicKey="${base64RsaPublicKey}"`)],
+		});
+		const submissionResult = await scenario.prepareWebFormsInstancePayload();
+
+		expect(submissionResult.submissionMeta).toMatchObject({
+			encryptionKey: base64RsaPublicKey,
+		});
+
+		expect(submissionResult.data.length).to.equal(1);
```
alxndrsn: From here on looks like additional assertions which are not currently present in this test :+1:

packages/xforms-engine/test/integration/submission-encryption.test.ts
```diff
@@ -0,0 +1,81 @@
...
+	): Promise<Scenario> => {
+		const scenario = await Scenario.init(
+			'Encrypted',
+			html(
+				head(
+					title('Encrypted'),
+					model(
+						mainInstance(t('data id="encrypted"', t('inp', 'test'), t('meta', t('instanceID')))),
+						...(options?.submissionElements ?? []),
+						bind('/data/inp').required(),
```
alxndrsn, line 45: This looks like it's changed subtly from the previous implementation in `packages/xforms-engine/test/integration/submission.test.ts`; is there anything worth noting about the difference?

packages/xforms-engine/test/integration/submission-encryption.test.ts
```diff
@@ -0,0 +1,81 @@
...
+					)
+				),
+				body(input('/data/rep/inp', label('inp')))
+			)
+		);
+
+		return scenario;
+	};
+
+	it('includes a form-specified `base64RsaPublicKey` as encryptionKey', async () => {
```
alxndrsn: With the additional assertions, does the test title also need an update?

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/wordArrayUtils.ts
```diff
@@ -0,0 +1,24 @@
+import * as CryptoJS from 'crypto-js';
+
+export const arrayBufferToWordArray = (buffer: Uint8Array<ArrayBuffer>) => {
+	const words = [];
+	for (let i = 0; i < buffer.length; i += 4) {
+		words.push(
+			((buffer[i] ?? 0) << 24) |
+				((buffer[i + 1] ?? 0) << 16) |
+				((buffer[i + 2] ?? 0) << 8) |
+				(buffer[i + 3] ?? 0)
```
alxndrsn, line 10: I feel like prettier is making this harder to read than necessary.

packages/xforms-engine/src/lib/client-reactivity/instance-state/prepareInstancePayload.ts
```diff
@@ -37,15 +52,47 @@ class InstanceFile extends File implements ClientInstanceFile {
...
 		});
 	}
 }
 
+export interface Submission {
+	readonly instanceXML: string;
+	readonly attachments: readonly File[];
+}
+
+const collectInstanceFiles = async (
```
alxndrsn: I'm struggling with this function, especially:

* how it relates to `collectInstanceAttachmentFiles()`
* what it's return value represents

Would the following be more accurate?

```suggestion
const buildInstance = async (
```

Should the function be returning the `InstanceFile` instance instead of the `instanceXML`?

e2e-tests/global.teardown.js
garethbowen: Sold!

e2e-tests/tests/encrypted-submission.spec.js
garethbowen: Actually this one is! There was another one I removed. Yes this one is necessary because otherwise it tries to load in enketo which is the default (for another month at least).

Comment removed.

e2e-tests/tests/encrypted-submission.spec.js
garethbowen: Lol. Fixed.

e2e-tests/tests/encrypted-submission.spec.js
garethbowen: Done.

e2e-tests/tests/encrypted-submission.spec.js
garethbowen: Fair enough. I've gone with this with a couple of changes (eg: the first row is the column names so I throw that away).

e2e-tests/tests/encrypted-submission.spec.js
garethbowen: I've exported a constant from util instead.

e2e-tests/tests/encrypted-submission.spec.js
garethbowen: I had a look at that. The first problem is that you don't have access to the `page` in the beforeAll (by design). I could create a new page but that's got a smell about it. And I think the smell is because the thing I'm actually testing is the web-forms submission - the `checkResponse` is just the assertions that the submission worked as expected. So it feels weird to move the thing I'm testing into the beforeAll.

I think for now I prefer to leave it as is, if that works for you?

packages/xforms-engine/test/integration/submission-encryption.test.ts
garethbowen: I don't think so, though it is complicated. I think the main change was because the comment that "Note: `t()` fails here..." was no longer correct so I could use a cleaner alternative.

packages/xforms-engine/test/integration/submission-encryption.test.ts
garethbowen: Done.

e2e-tests/tests/encrypted-submission.spec.js
alxndrsn: > The first problem is that you don't have access to the `page` in the beforeAll

I guess the answer would be to extract a function, so you get:

```js
  test('values are obscured in CSV when no key provided', async ({ page }) => {
    await submitFormWith(page, { 'First Name':'John Does' });
    await submitFormWith(page, { 'First Name':'Jane Doesn\'t' });

    const res = await backendClient.downloadCsv(publishedForm.xmlFormId);
    await assertCsvResponse(res, STATUS_COLUMN_INDEX, ['not decrypted', 'not decrypted']);
  }));
  
  test('download fails if incorrect key provided', async ({ page }) => {
    await submitFormWith(page, { 'First Name':'John Does' });
    await submitFormWith(page, { 'First Name':'Jane Doesn\'t' });

    const res = await backendClient.downloadCsv(publishedForm.xmlFormId);
    expect(res).not.toBeOK();
  }));
  
  test('values are unencrypted when correct key supplied', async ({ page }) => {
    await submitFormWith(page, { 'First Name':'John Does' });
    await submitFormWith(page, { 'First Name':'Jane Doesn\'t' });

    const res = await backendClient.downloadCsv(publishedForm.xmlFormId);
    await assertCsvResponse(res, FIRSTNAME_COLUMN_INDEX, ['Jane Doesn\'t', 'John Does']); // csv is reverse date order
  }));
```

> I think for now I prefer to leave it as is, if that works for you?

:+1: no big deal

packages/xforms-engine/src/lib/client-reactivity/instance-state/prepareInstancePayload.ts
garethbowen: I've changed the function to return the instanceFile and renamed it to `createSubmission` which is now more accurate IMO.

e2e-tests/util.js
```diff
@@ -16,4 +17,7 @@ const login = async (page) => {
   await page.waitForURL(appUrl);
 };
 
-export { login };
+export {
+  login,
+  ENCRYPTION_SECRET
+};
```
alxndrsn, line 23: For my education: is there any reason to export this mutable(?) object vs exporting the `const`s directly?

garethbowen: Not as far as I know.

garethbowen: One thing is in a big file it can make the exported functions easier to spot.

alxndrsn: Aha!  I think I was confused with:

```js
export default {
  login,
  ENCRYPTION_SECRET
};
```

packages/xforms-engine/test/integration/submission-encryption.test.ts
```diff
@@ -0,0 +1,81 @@
...
+		expect(submission).to.contain(
+			'<data xmlns="http://www.opendatakit.org/xforms/encrypted" encrypted="yes" id="encrypted">'
+		);
+		expect(submission).to.contain('<encryptedXmlFile>submission.xml.enc</encryptedXmlFile>');
+		expect(submission).to.contain(
+			'<meta xmlns="http://openrosa.org/xforms"><instanceID>uuid:TODO-mock-xpath-functions</instanceID></meta>'
+		);
+
+		const [encodedFilename] = entries.next().value!;
+		expect(encodedFilename).to.equal('submission.xml.enc');
```
alxndrsn: Would this be the right place to check for other **required** fields from https://getodk.github.io/xforms-spec/encryption#submission-manifest?

I think the only check missing may be for `base64EncryptedKey`.

packages/xforms-engine/test/integration/submission-encryption.test.ts
```diff
@@ -0,0 +1,81 @@
...
+	const buildSubmissionPayloadScenario = async (
+		options?: BuildSubmissionPayloadScenario
+	): Promise<Scenario> => {
+		const scenario = await Scenario.init(
+			'Encrypted',
+			html(
+				head(
+					title('Encrypted'),
+					model(
+						mainInstance(t('data id="encrypted"', t('inp', 'test'), t('meta', t('instanceID')))),
```
alxndrsn:
```suggestion
						mainInstance(t('data id="encrypted-form-1"', t('inp', 'test'), t('meta', t('instanceID')))),
```

`encrypted` is overloaded (also e.g. a required attribute of the `<data>`).  It might be less ambiguous to give the form itself a different name.

packages/xforms-engine/test/integration/submission-encryption.test.ts
```diff
@@ -0,0 +1,81 @@
...
+		});
+
+		expect(submissionResult.data.length).to.equal(1);
+		const entries = submissionResult.data[0].entries();
+
+		const [submissionFilename, file] = entries.next().value!;
+		expect(submissionFilename).to.equal('xml_submission_file');
+		const submission = await getBlobText(file);
+		expect(submission).to.contain(
+			'<data xmlns="http://www.opendatakit.org/xforms/encrypted" encrypted="yes" id="encrypted">'
```
alxndrsn: Maybe an additional test for a versioned form?

packages/xforms-engine/test/integration/submission-encryption.test.ts
```diff
@@ -0,0 +1,81 @@
...
+		);
+		expect(submission).to.contain('<encryptedXmlFile>submission.xml.enc</encryptedXmlFile>');
+		expect(submission).to.contain(
+			'<meta xmlns="http://openrosa.org/xforms"><instanceID>uuid:TODO-mock-xpath-functions</instanceID></meta>'
+		);
+
+		const [encodedFilename] = entries.next().value!;
```
alxndrsn: Would it be helpful to get the `file` here, and check it doesn't contain a known plaintext value that was contained in the form submission?

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
```diff
@@ -0,0 +1,80 @@
...
+		readonly symmetricKey: CryptoJS.lib.WordArray
+	) {
+		const md = CryptoJS.algo.MD5.create();
+		md.update(instanceId);
+		md.update(symmetricKey);
+		this.ivSeedArray = wordArrayToArrayBuffer(md.finalize());
+	}
+
+	next(): string {
+		++this.ivSeedArray[this.counter % this.ivSeedArray.length]!;
```
alxndrsn: From the spec:

> calculate index as remainder of the counter modulo 16
> increment byte in seed array at index with 1

is `this.ivSeedArray.length` guaranteed to be 16?

e2e-tests/tests/encrypted-submission.spec.js
```diff
@@ -0,0 +1,63 @@
...
+    await expect(page.getByRole('heading', { name: publishedForm.name })).toBeVisible();
+
+    await page.getByLabel('First Name').fill('John Doe');
+    await page.getByRole('button', { name: 'send' }).click();
+    await expect(page.getByRole('heading', { name: 'Successful' })).toBeVisible();
+
+    await page.getByRole('button', { name: 'fill out again' }).click();
+    await page.getByLabel('First Name').fill('Jane Doe');
+    await page.getByRole('button', { name: 'send' }).click();
+    await expect(page.getByRole('heading', { name: 'Successful' })).toBeVisible();
```
alxndrsn, line 120: would it be possible to also include an example including one or a few attachments?

alxndrsn (changes requested):

packages/xforms-engine/test/integration/submission-encryption.test.ts
garethbowen: Added.

packages/xforms-engine/test/integration/submission-encryption.test.ts
garethbowen: Changed.

packages/xforms-engine/test/integration/submission-encryption.test.ts
garethbowen: Surely that's unrelated to encryption?

packages/xforms-engine/test/integration/submission-encryption.test.ts
garethbowen: Done.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
garethbowen: The ivSeedArray is created from the md5 hash of the instanceId and symmetricKey (line 22).

md5 is guaranteed to be 128 bit therefore 16 bytes.

e2e-tests/tests/encrypted-submission.spec.js
garethbowen: Yes, but as soon as I do that it'll download as a zip instead of a csv. We don't currently have any package which can unzip. Do you think it's worth adding a dependency for this test?

packages/xforms-engine/test/integration/submission-encryption.test.ts
alxndrsn: I think it changes the behaviour of the submission manifest:

 https://getodk.github.io/xforms-spec/encryption:

> `<element>`/attribute | description
> ---|---
> `version` | Required on the root element with same value as the `version` attribute in the [Primary Instance](https://getodk.github.io/xforms-spec/#primary-instance) of the XForm but only if it exist there.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
alxndrsn: :+1:

I think it would be clearer to hard-code `16` per the spec, rather than deriving the constant from a property of MD5.

e2e-tests/tests/encrypted-submission.spec.js
alxndrsn: > Do you think it's worth adding a dependency for this test?

I do, as it can have significant impact on the encryption algorithm.  And probably worth testing both with & without attachments.

packages/xforms-engine/test/integration/submission-encryption.test.ts
garethbowen: Ahh right. Making sure it's copied over. Will do.

garethbowen: Done.

packages/xforms-engine/src/lib/client-reactivity/instance-state/quarantine/symmetric.ts
garethbowen: I've hardcoded the mod operand to 16.

e2e-tests/tests/encrypted-submission.spec.js
garethbowen: Done.

garethbowen:
@alxndrsn Tests added. Green build.

alxndrsn (approved):
Looking good.

PR #1563, opened 2026-05-25
check-for-large-files: add lower, more-selective limits

### LINKED ISSUE #1911: Check for large .svg files
Filing an issue based on https://github.com/getodk/central-frontend/pull/1531#issuecomment-4488549141

We expect most .svg files to be quite small. If they weren't small, we'd want to think about minimizing them along the lines of getodk/central-frontend#1531. But given that expectation, we should lower the size ceiling in check-bundle-size.js. Right now it allows .svg files to be up to 60 kB.

We do have one large .svg file in that range, icomoon.svg, but we can make an exception for that file. It's not clear that users ever actually even use that file. After #931, hopefully we'll be able to do away with that file completely.

This issue is to:

- Lower the size ceiling for .svg files (to 5 kB?)
- While making an exception for icomoon.svg

alxndrsn:
Closes getodk/central#1911?

Run the script with some different values in it and seen it fail as expected.

No effect - just a CI/test script.

No.


matthew-white:
What do you think about making this change to check-bundle-size.js instead? check-bundle-size.js already has logic around limits being different for different file types and there being exceptions for specific files. I was thinking we could add icomoon.svg there under "Special cases":

https://github.com/getodk/central-frontend/blob/2c087b6acd4ffbc83fc282e6ba4fe5ee71622c3a/bin/check-bundle-size.js#L55-L63

Then reduce the limit for .svg files here:

https://github.com/getodk/central-frontend/blob/2c087b6acd4ffbc83fc282e6ba4fe5ee71622c3a/bin/check-bundle-size.js#L72

I feel like the main purpose of check-for-large-files.sh is just to make sure that big files aren't committed to the repo, causing them to live on in Git history forever. In that scenario, the file type doesn't matter so much — we just don't want any super big files committed. Whereas check-bundle-size.js is more about checking the size of files that users will download, where expectations of file size do differ by file type.

alxndrsn:
> I feel like the main purpose of check-for-large-files.sh is just to make sure that big files aren't committed to the repo, causing them to live on in Git history forever. In that scenario, the file type doesn't matter so much — we just don't want any super big files committed. Whereas check-bundle-size.js is more about checking the size of files that users will download, where expectations of file size do differ by file type.

I agree that bundle size & repo size are different things, but I think that reducing the repo size limits is still a positive thing, even if it means introducing more specific expectations.

Maybe this PR doesn't close https://github.com/getodk/central/issues/1911, but still has value?

matthew-white:
I could see how there could be value in reducing repo size limits. E.g., not all .png files in the repo are included in the bundle that check-bundle-size.js checks, so this change would enforce a reduced size limit for those .png files.

On the other hand, as we make progress on things like #1544, the repo size will begin to matter less, since end users will no longer download the repo. In that case, it might be enough just to have a single, simple size limit. I'm slightly concerned about an increase in maintenance/chores. E.g., if there needs to be an exception for a particular file, that exception may need to be added in two places: check-bundle-size.js and now check-for-large-files.sh.

We also have check-for-large-files.sh in `central-backend` and `central`, and maybe there's a way in which it's nice for them to be as similar as possible.

I don't have strong feelings here, so I'm happy to run with it if you think it'd be an improvement. We should probably double-check with the Web Forms team though, since the check could also affect them.

alxndrsn:
> E.g., if there needs to be an exception for a particular file, that exception may need to be added in two places: check-bundle-size.js and now check-for-large-files.sh.

Hopefully these exceptions are infrequent.  If the burden gets too big, we could remove one or both checks, or move to Git LFS?

PR #1564, opened 2026-05-25
fix(web-forms#805): allow labels with computed media urls

### LINKED ISSUE #805: Outputs in text range media chunks do not work
### Expected Behaviour

It should be possible to set the value of a media chunk based on a calculation. For example in the xlsx setting the `image` column to `${myimg}`.

### Observed Behaviour

Instead the form fails to load with an error, similar to "URL constructor:   is not a valid URL."

### Reproduction Steps

Form (XML or XLSX file):

```xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms"
  xmlns:h="http://www.w3.org/1999/xhtml"
  xmlns:ev="http://www.w3.org/2001/xml-events"
  xmlns:xsd="http://www.w3.org/2001/XMLSchema"
  xmlns:jr="http://openrosa.org/javarosa"
  xmlns:orx="http://openrosa.org/xforms"
  xmlns:odk="http://www.opendatakit.org/xforms">
  <h:head>
    <h:title>images-choice</h:title>
      <itext>
        <translation lang="default" default="true()">
          <text id="/data/note:label">
            <value>Reference image</value>
            <value form="image"> jr://images/<output value=" /data/animal "/> </value>
          </text>
        </translation>
      </itext>
      <instance>
        <data id="images-choice" version="20240611120218">
          <animal/>
          <note/>
          <meta>
            <instanceID/>
          </meta>
        </data>
      </instance>
      <instance id="animals">
        <root>
          <item>
            <itextId>animals-0</itextId>
            <name>tiger</name>
            <img>tiger.jpg</img>
          </item>
          <item>
            <itextId>animals-1</itextId>
            <name>camel</name>
            <img>camel.jpg</img>
          </item>
        </root>
      </instance>
      <bind nodeset="/data/animal" type="string"/>
    </model>
  </h:head>
  <h:body>
    <select1 ref="/data/animal">
      <label>Animal</label>
      <itemset nodeset="instance('animals')/root/item">
        <value ref="img"/>
        <label ref="img"/>
      </itemset>
    </select1>
    <input ref="/data/note">
      <label ref="jr:itext('/data/note:label')"/>
    </input>
  </h:body>
</h:html>
```

Steps:
1. Load the form.
2. See error

NB: This is the XML produced from xls2xform: `<value form="image"> <output value=" /data/animal "/> </value>`. If you manually remove the whitespace around the output then you don't get the error but the image doesn't load either.

### Environment

- Browser version:
- OS:
- Web Forms version:

### Additional Context

Reported on the forum: https://forum.getodk.org/t/limiting-entity-access-for-app-users-and-data-collectors/54386/23

garethbowen:
Closes getodk/web-forms#805

Integration testing.

This bug probably isn't a blocker but I think the code ends up cleaner so it's a win-win.

A feature that works in Collect is now supported in Web Forms

No.

### DIFF
--- a/.changeset/rich-teams-push.md
+++ b/.changeset/rich-teams-push.md
@@ -1,0 +1,6 @@
+---
+'@getodk/xforms-engine': patch
+'@getodk/web-forms': patch
+---
+
+Fixed bug where computed image urls blocked form loading
--- a/packages/common/src/fixtures/select/7-images-choice-computed.xml
+++ b/packages/common/src/fixtures/select/7-images-choice-computed.xml
@@ -1,0 +1,59 @@
+<?xml version="1.0"?>
+<h:html xmlns="http://www.w3.org/2002/xforms"
+  xmlns:h="http://www.w3.org/1999/xhtml"
+  xmlns:ev="http://www.w3.org/2001/xml-events"
+  xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+  xmlns:jr="http://openrosa.org/javarosa"
+  xmlns:orx="http://openrosa.org/xforms"
+  xmlns:odk="http://www.opendatakit.org/xforms">
+  <h:head>
+    <h:title>images-choice</h:title>
+    <model odk:xforms-version="1.0.0">
+      <itext>
+        <translation lang="default" default="true()">
+          <text id="/data/note:label">
+            <value>Reference image</value>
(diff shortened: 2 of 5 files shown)


garethbowen:
@latin-panda This is ready for review. It's 99% the same code as you've already reviewed on the web-forms side.

latin-panda (approved):
Ready to ship 🚀

PR #1566, opened 2026-05-27
chore: enforce spaces over tabs in Web Forms packages

latin-panda:
- Switches indentation from tabs to spaces across web forms packages' files 
- Adds an .editorconfig to enforce spaces going forward.

CI passes

Standard setup for code styles

No

No


.editorconfig
latin-panda, line 1: New, for IDE

latin-panda:
@garethbowen the relevant changes are in these files:
prettier.config.js
.editorconfig
README.md

After updating the Prettier config, the linter flagged all files using tabs, and I ran `format:fix` to correct them. All the other files you see in this diff were converted from tabs to spaces. Matching app/central that uses 2-space indent.

garethbowen (changes requested):
You must be as crazy as I am! :D 

I think this is the right approach. We'll need to standardise eventually and there's no gentle way to do it. It's crazy to me that even the web-forms files weren't standard, so this is a win.

I think it would be worth checking with at least some of the team. Some people really dislike wholesale changes in the git history. Personally I think it's much more important for "head" to be consistent than for the history to be "clean" so it's ok with me.

Would you mind splitting the README changes out so this change is just about one thing?

latin-panda:
@garethbowen Ready!
- The update to the README file was extracted in this PR: https://github.com/getodk/central-frontend/pull/1569 

- The team has been notified [on Slack](https://getodk.slack.com/archives/C01TKJ7TG7J/p1779952605550799)

garethbowen (approved):
Thanks!

Obviously I haven't reviewed all files, but what I've seen looks good.

latin-panda:
I did a final pass, and looks good

PR #1567, opened 2026-05-27
fix(#811): align geopoint altitude/accuracy formatting with ODK Collect

### LINKED ISSUE #811: Discrepancy in format of user selected geolocation data between Collect and Web Forms
When a user selects a location manually Collect sets the accuracy and altitude to 0 ([source](https://github.com/getodk/collect/blob/master/location/src/main/java/org/odk/collect/location/Location.kt)) whereas Web Forms leaves them blank. We should decide which is correct and be consistent.

The [spec](https://getodk.github.io/xforms-spec/#data-type:geopoint) implies that all four values are required.

Reported by user via DM. The use case is they want to have a constraint to require user selection and disallow the GPS recorded by the device. Their forms currently check if the accuracy value is set to 0. The workaround is to check for 0 or `undefined`.

The `placement-maps` appearance works as well and is a better solution because it updates the UI to not have the automatic collection, but there's no equivalent for lines or polygons so it's not a complete solution. We could consider also providing an equivalent for those input types as well, but we should still produce output consistent with Collect regardless.

latin-panda:
Closes https://github.com/getodk/web-forms/issues/811

Updated unit tests and scenario tests. Manually tested the following cases:

- Question types with default values where altitude and longitude are missing.
- Tested geolocation when granted and when denied.
- Tested manually entered points.
- Compared Collect data vs Web Forms in central by downloading submissions CSV.

- Saved data will always include the 4 parts (lat, lon, alt, and acc)
- Altitude zero is displayed as a valid measurement (sea level)
- Accuracy zero is hidden because zero accuracy is not a real measurement, and to keep the ui simple, since space is limited.

No


latin-panda:
@garethbowen Could you please review when you have a chance? :)

.changeset/itchy-shrimps-call.md
```diff
@@ -0,0 +1,6 @@
+---
+"@getodk/xforms-engine": patch
+"@getodk/web-forms": patch
```
garethbowen, line 4: In general, do we want to also include changes to web-forms in the central-frontend release notes? Previously I think we had decided to do that for changes in web-forms dependencies so that would be consistent.

packages/web-forms/src/components/common/map/map-helpers.ts
```diff
@@ -13,25 +13,23 @@ import type {
...
+		return [
+			latitude,
+			longitude,
+			Number.isFinite(altitude) ? altitude! : 0,
+			Number.isFinite(accuracy) ? accuracy! : 0,
+		];
 	}
 
-	return coords;
+	return [];
```
garethbowen: I think it's easier to read if you return  the null/error case first, eg:

```js
if (!Number.isFinite(latitude) || !Number.isFinite(longitude)) {
  return [];
}
...
```

garethbowen (changes requested):
Minor change requested inline.

The form that was failing was never shared with me but in a private DM the issue was...

> I use this constraint to prevent automatic geopoints (have a different but similar one for line/shape) and enforce manual placement. Works fine in Collect & Enketo but manually placing a point by clicking on the map in the browser still triggers the constraint in Webforms.
> `selected-at(${geopoint}, 3) = 0`

Can you manually test to ensure that the constraint above now passes with this change?

.changeset/itchy-shrimps-call.md
latin-panda: Do you mean adding `@getodk/central-frontend: patch`. That's going to be hard to remember 😅

If the library is an explicit dependency in `apps/central`'s package.json, it will automatically add the reference, for example, like this: 
Screenshot 2026-05-29 at 11 05 35 PM

We can open a ticket to improve these notes, like adding links to commits and other packages' release notes.

packages/web-forms/src/components/common/map/map-helpers.ts
latin-panda: Done! Thanks!

latin-panda:
> The form that was failing was never shared with me but in a private DM the issue was...
>
>> I use this constraint to prevent automatic geopoints (have a different but similar one for line/shape) and enforce manual placement. Works fine in Collect & Enketo but manually placing a point by clicking on the map in the browser still triggers the constraint in Webforms.
selected-at(${geopoint}, 3) = 0
>
> Can you manually test to ensure that the constraint above now passes with this change?

Yes, I used the form below. Manual point's altitude and accuracy are zero (no constraints are shown). 





<details><summary>Form: prevents automatic points</summary>

```xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms"
        xmlns:h="http://www.w3.org/1999/xhtml"
        xmlns:jr="http://openrosa.org/javarosa"
        xmlns:odk="http://www.opendatakit.org/xforms">
  <h:head>
    <h:title>Geopoint Constraint Test</h:title>
      <instance>
        <data id="geopoint-constraint-test">
          <geopoint/>
          <meta>
            <instanceID/>
          </meta>
        </data>
      </instance>
      <bind nodeset="/data/geopoint" type="geopoint"
            constraint="selected-at(., 3) = 0"
            jr:constraintMsg="Constraint failed: must place manually (accuracy must be 0)"/>
    </model>
  </h:head>
  <h:body>
    <input ref="/data/geopoint" appearance="placement-map">
      <label>Place a point manually on the map</label>
      <hint>Constraint: selected-at(geopoint, 3) = 0 — accuracy must be 0; manual placement stores accuracy=0 and passes</hint>
    </input>
  </h:body>
</h:html>
```

 </details> 

I also tested the same thing with `maps` appearance and got the constraints with automatic points:

latin-panda:
@garethbowen this is ready for another round of review :)

.changeset/itchy-shrimps-call.md
garethbowen: We'll figure this out as we go, I'm just wondering if it's easier to build the release notes if one changelog (central-frontend) has everything. It's probably not going to work, particularly once we have a second app, so fine for now. Maybe we can script a release note generator for central which combines the release notes of frontend, backend, and web-forms...

garethbowen (approved):
Nailing it!

.changeset/itchy-shrimps-call.md
latin-panda: That sounds nice, and shouldn't be hard. I'll play with the options next week

PR #1568, opened 2026-05-27
Entity properties can now be deleted

lognaturel:
I noticed some text that is no longer true!

Visual code inspection and updating tests only.

It's a removal so I don't think there are any alternatives here.

Biggest risk is probably if I misunderstood the templating and removed more than I should have.

No.

### DIFF
--- a/apps/central/src/components/form/edit/entities.vue
+++ b/apps/central/src/components/form/edit/entities.vue
@@ -20,10 +20,6 @@ except according to the terms contained in the LICENSE file.
           <p v-if="datasetDiff.counts.updatedDatasets !== 0"
             id="form-edit-entities-diff-counts">
             <span>{{ diffCounts }}</span>
-            <template v-if="datasetDiff.counts.newProperties !== 0">
-              <sentence-separator/>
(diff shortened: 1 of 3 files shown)


lognaturel:
@sadiqkhoja tagging you for review since you're in this part of the code, feel free to close if you're already taking care of it elsewhere!

lognaturel:
No changeset needed, this is a trivial change with no real user-facing consequences.

apps/central/src/components/form/edit/entities.vue
```diff
@@ -112,7 +106,6 @@ const diffCounts = computed(() => {
         "properties": "{count} Eigenschaft | {count} Eigenschaften"
       }
     },
-    "cannotDeleteProperties": "Eigenschaften können nicht gelöscht werden.",
```
sadiqkhoja, line 115: I usually don't delete/modify any line below `<!-- Autogenerated by destructure.js -->`, any deleted key automatically gets deleted after we pull translations from Transifex and then `destructure` the result.

sadiqkhoja (approved):

PR #1571, opened 2026-05-28
npm run dev:build: fix vue pages

alxndrsn:
Fixes error when running `npm run dev:build`:

	rewrite or internal redirection cycle while internally redirecting to "/index.html", client: 127.0.0.1, server: localhost, request: "GET / HTTP/1.1", host: "127.0.0.1:8989"

Ref https://github.com/getodk/central/issues/1919

- [x] tested `npm run dev:build` locally
- [ ] got another dev (preferably on MacOS) to test `npm run dev:build` with & without this change

* simple
* fixes a problem

No effect - just in dev.

No.


nginx-conf/nginx.conf
```diff
@@ -153,7 +153,7 @@ http {
 
     # Only relevant for `npm run dev:build`, not `npm run dev`.
     location / {
-      root ./dist;
+      root /dist;
```
matthew-white, line 148: Naively, I would think that `./dist` is a relative path, which is what we want, as opposed to `/dist`, an absolute path. Is that not how `root` works?

I'm pretty sure this worked at some point not too long ago. Has the behavior of `root` changed recently?

matthew-white: Now I'm wondering whether there could be a difference in this area between `npm run dev:build` and `npm run dev-no-docker:build`. Like maybe it needs to be `/dist` for `dev:build` and `./dist` for `dev-no-docker:build`.

matthew-white: I like the `dev-no-docker` stuff, but I'm also open to doing away with it. Wicher liked to avoid Docker where possible, but it could be that the current team is fine with Docker-only. Maybe worth polling Slack? Maybe the solution here is to go Docker-only, then run with `/dist` instead of `./dist`. Not sure about that, maybe there's an alternative fix that accommodates both `dev:build` and `dev-no-docker:build`, I just wanted to throw out these thoughts.

alxndrsn: https://github.com/getodk/central/issues/1935

alxndrsn: Confirm if this change is necessary, as others report that docker+nginx setup is already working on `master`.

matthew-white: Confirmation sounds good, but my guess is that it is necessary. People have probably been trying `npm run dev`, not `npm run dev:build`.

matthew-white (approved):

nginx-conf/nginx.conf
alxndrsn: > People have probably been trying `npm run dev`, not `npm run dev:build`.

Of course :facepalm:

PR #1572, opened 2026-05-28
dev/nginx: remove copyright notice from config

alxndrsn:
No functional change, but makes it easier to compare with production config at https://github.com/getodk/central/blob/next/files/nginx/odk.conf.template

Ref https://github.com/getodk/central/issues/1601

Run `npm run dev` and `npm run dev:build` without the copyright notice.

No functional change, but makes it easier to compare with production config at https://github.com/getodk/central/blob/next/files/nginx/odk.conf.template

No change - this config file is only used in dev.

No.


nginx-conf/nginx.conf
```diff
@@ -1,16 +1,3 @@
...
-# https://github.com/getodk/central-frontend/blob/master/NOTICE.
-#
-# This file is part of ODK Central. It is subject to the license terms in
-# the LICENSE file found in the top-level directory of this distribution and at
-# https://www.apache.org/licenses/LICENSE-2.0. No part of ODK Central,
-# including this file, may be copied, modified, propagated, or distributed
-# except according to the terms contained in the LICENSE file.
-
-# This configuration file is for development only. For production, see
-# https://github.com/getodk/central.
```
matthew-white, line 12: Could we keep these comments? It might make the diff with the production config a little noisier, but I think it's an important message to convey. We also mention in the readme that this nginx config is only for development.

matthew-white (approved):
Based on the comments quoted in getodk/central#1906, it should be OK to remove this copyright notice. I agree that it'd make it easier to compare with the production config. I feel like it reduces noise in this file.

nginx-conf/nginx.conf
alxndrsn: :+1: good call

PR #1575, opened 2026-05-28
Disable CSP for e2e tests

sadiqkhoja:
Disable CSP for e2e tests; it is a temporary workaround for https://github.com/getodk/central/issues/1920 so that we can unblock PR merges.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -100,6 +100,7 @@ jobs:
         sed -i 's|\${BASE_URL}|http://${DOMAIN}|g' files/enketo/config.json.template
         sed -i 's|\${BASE_URL}|http://${DOMAIN}|g' files/service/config.json.template
         sed -i 's/\$scheme/https/g' files/nginx/odk.conf.template
+        sed -i 's/\Content-Security-Policy/Content-Security-Report-Only/g' files/nginx/odk.conf.template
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -100,6 +100,7 @@ jobs:
         sed -i 's|\${BASE_URL}|http://${DOMAIN}|g' files/enketo/config.json.template
         sed -i 's|\${BASE_URL}|http://${DOMAIN}|g' files/service/config.json.template
         sed -i 's/\$scheme/https/g' files/nginx/odk.conf.template
+        sed -i 's/\Content-Security-Policy/Content-Security-Report-Only/g' files/nginx/odk.conf.template
```
lognaturel: The backslash isn't needed before the C but I don't think it causes any problems either. It is necessary to make sure a literal `$` is used in the other expressions.

lognaturel (approved):

.github/workflows/tests.yml
sadiqkhoja: oh right

sadiqkhoja:
```suggestion
        sed -i 's/Content-Security-Policy/Content-Security-Report-Only/g' files/nginx/odk.conf.template
```

PR #1576, opened 2026-05-29
fix(#808): XPath and XForms-engine memory leaks in repeat

### LINKED ISSUE #808: Investigate likely memory leak in form with hundreds of repeat instances
[update_multi_select.xlsx](https://github.com/user-attachments/files/27787447/update_multi_select.xlsx) with [transactions.csv](https://github.com/user-attachments/files/27787466/transactions.csv)

This form has 400 repeat instances. I clicked around in it a bunch and then left it open while my computer slept. When I came back to my computer, my memory was full and I was using a lot of swap. Once I closed the tab, things got better but memory was not fully released, I had to restart the browser.

Firefox 149.0.2 (aarch64)
macOS 15.7.4

latin-panda:
Closes https://github.com/getodk/web-forms/issues/808

### What changed?

 Fixed memory leaks in the XPath namespace-resolver cache and the `xforms-revalidate` listener map, and improved `position(node)` performance on nodes with many same-name siblings.

### Performance Comparison

**Note:** The final master heap snapshot (the full-cycle capture) could not be completed because the browser crashed before it finished, and is the master third data point used throughout this report. 

Master leaks memory on every toggle; the branch does not.

Master accumulates ~**4,150 `UpsertableMap` objects per toggle** (each retaining a backing array, a closure, and a V8 Context), growing the heap without bound. On the branch the same structure stays flat at **3 instances** for the entire run, and total heap **decreases** across the cycle as expected. The branch is the fix.

**Heap trajectory**

| Stage | Branch nodes | Branch heap | Master nodes | Master heap |
|---|---|---|---|---|
| Loaded | 3,317,689 | **217.4 MB** | 3,471,396 | **226.9 MB** |
| After 80 toggles | 3,265,170 | **213.3 MB** | 4,791,399 | **277.5 MB** |
| Further into cycle | 3,098,358 | **205.8 MB** | 5,806,101 | **318.2 MB** |

- **Branch:** **decrease** (217 → 213 → 206 MB), node count falling. Healthy steady state.
- **Master:** **increase** (227 → 278 → 318 MB), +2.3M nodes. Note the master third snapshot is only the *half-way* point of the cycle and is already **+112 MB** above the branch's *full*-cycle heap.

Detached DOM nodes stayed at noise level in both (6–12), so this is a retained-JS-object leak, not a detached-DOM leak.

**Root cause: leaked `UpsertableMap`**

On master, four items grow. Indicates a leaked object dragging its dependencies:

| Constructor | Δ over 80 toggles | Δ size | Per toggle |
|---|---|---|---|
| `object UpsertableMap$1` | +331,616 | +5.2 MB | ~4,145 |
| `array` (its backing store) | +331,645 | +30.9 MB | ~4,146 |
| `closure` (anonymous) | +332,162 | +9.1 MB | ~4,152 |
| `system / Context` | +332,435 | +9.1 MB | ~4,155 |

**~0.66 MB leaked per toggle**, never reclaimed.

`UpsertableMap` instance count by stage:

| | Loaded | After 80 toggles | Further |
|---|---|---|---|
| **Master** | 37,890 | 369,506 | 657,355 |
| **Branch** | 3 | 3 | 3 |

Each toggle on master constructs a fresh `UpsertableMap` (plus the closure + Context that captures it) and retains the previous ones — consistent with a per-toggle subscription/watcher/cache entry that is never disposed. The branch eliminates the per-toggle allocation entirely.

**Ruling out false positives**

The only constructors that *grew* on the branch were DevTools/profiler artifacts, present in **both** runs and unrelated to app logic:

- `PerformanceScriptTiming`, `PerformanceEventTiming`, `PerformanceLongAnimationFrameTiming`, `TaskAttributionTiming` — recorded because the Performance panel was active during capture.
- `CSSTransition` / `KeyframeEffect` — animations in flight at snapshot time.
- `WeakArrayList`, `FunctionTemplateInfo` — V8 internals.

These appear identically in master and do not accumulate with toggle count, so they don't affect the comparison.

On the branch, the real app structures all **shrank** loaded→cycle: `InternalNode` −61,893, `Context` −16,791, `Object` −14,680, and PrimeVue style objects (`CSSStyleRule`/`CSSRuleList`/`StylePropertyMap`) roughly **halved** — i.e. the app cleans up after itself.


packages/xpath/test/xforms/position.test.ts
```diff
@@ -81,5 +81,19 @@ describe('#position()', () => {
       contextNode,
     });
   });
+
+  it('position(node) returns correct position among many same-name siblings', () => {
```
latin-panda, line 85: More like a sanity test

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -112,10 +112,15 @@ const postloadValue = (
...
-  const ref = context.contextReference();
-  context.definition.model.registerXformsRevalidateListener(ref, () => {
+  const { model } = context.definition;
+
+  model.registerXformsRevalidateListener(context, () => {
     setValueIfPreloadDefined(context, setValue, preload);
   });
+
+  onCleanup(() => {
+    model.unregisterXformsRevalidateListener(context);
```
latin-panda, line 122: Without this, removed repeat instances leak listener entries.

packages/xpath/src/functions/xforms/node-set.ts
```diff
@@ -345,23 +345,30 @@ export const position = new NumberFunction(
...
 
-    let currentNode = value;
-    let result = 0;
-
-    do {
-      result += 1;
-
-      const previousNode = domProvider.getPreviousSiblingElement(currentNode);
+    const siblings = domProvider.getChildNodes(parent);
+    let blockPosition = 0;
```
latin-panda, line 354: Single O(n) forward scan over parent's children, counting position within the target's contiguous block of same-name siblings.

latin-panda:
@garethbowen this fixes the memory leak which is the main concern from the issue. The UI is slow when having hundreds of repeats and select options. I have a couple of ideas for how to address it, but they're bigger changes (pulldata, etc) I'll open a separate issue tomorrow with the details.

packages/xpath/src/functions/xforms/node-set.ts
```diff
@@ -345,23 +345,30 @@ export const position = new NumberFunction(
...
-      currentNode = previousNode;
-    } while (domProvider.getQualifiedName(currentNode) === nodeName);
+      blockPosition += 1;
+      if (sibling === value) {
+        return blockPosition;
+      }
+    }
 
-    return result;
+    return blockPosition;
```
garethbowen, line 371: Is this faster simply because you're getting the siblings once outside the loop, rather than every iteration?

packages/xpath/test/xforms/position.test.ts
```diff
@@ -81,5 +81,19 @@ describe('#position()', () => {
...
+    const items = Array.from({ length: repeatCount }, () => `<item/>`).join('\n');
+    testContext = createXFormsTestContext(`<root>${items}</root>`);
+
+    const items$ = testContext.document.querySelectorAll('item');
+
+    // First, middle, and last item positions
+    testContext.assertNumberValue('position(.)', 1, { contextNode: items$[0] });
+    testContext.assertNumberValue('position(.)', 200, { contextNode: items$[199] });
+    testContext.assertNumberValue('position(.)', 400, { contextNode: items$[399] });
+  });
```
garethbowen, line 117: There's a few more things to this function that aren't currently covered by unit tests, in particular...

- position of the root node
- when not isQualifiedNamedNode
- when previous siblings have different qualified names

These are quirks that I didn't realise the position function had and would be worth testing for.

garethbowen (changes requested):
Nice result!

packages/xpath/src/functions/xforms/node-set.ts
latin-panda: In the worst case, the old loop called `getPreviousSiblingElement`, which went through the parent's child list every time. With `n` children that's n × n work total. The new version goes through the child list once and counts the same-name block up to our node (O(n)).

packages/xpath/test/xforms/position.test.ts
latin-panda: I've added test cases for those scenarios

latin-panda:
@garethbowen this is ready for another look :)

garethbowen (approved):
Amazing! Thanks.

PR #1579, opened 2026-05-29
Form Draft Redesign Part 5: Hide upload and cancel button by default

sadiqkhoja:

### DIFF
--- a/apps/central/src/components/form/upload.vue
+++ b/apps/central/src/components/form/upload.vue
@@ -103,9 +103,9 @@ definition for an existing form -->
         {{ file != null ? file.name : '' }}
       </div>
     </file-drop-zone>
-    <div class="actions">
+    <div v-if="file != null" class="actions">
       <button type="button" class="btn btn-link" :aria-disabled="awaitingResponse"
-        @click="$emit('cancel')">
+        @click="clear()">
         {{ $t('action.cancel') }}
       </button>
       <button id="upload-button" type="button"
--- a/apps/central/src/components/form/upload.vue
+++ b/apps/central/src/components/form/upload.vue
@@ -130,7 +130,7 @@ export default {
   name: 'FormUpload',
   components: { DocLink, FileDropZone, SentenceSeparator, Spinner },
   inject: ['redAlert'],
-  emits: ['cancel', 'success'],
+  emits: ['success'],
   setup() {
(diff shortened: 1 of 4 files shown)


apps/central/src/components/form/upload.vue
matthew-white: Is it impossible now for `this.file == null`? If so, how about:

- Removing this `if` block
- Removing the i18n message below (`alert.fileRequired`) so that it's not sent to translators

I saw that you removed the test about this alert. If the alert is impossible now, it'd be nice to simplify the code a little.

matthew-white: Does anything emit `cancel` anymore? If not, how about removing `cancel` here?

matthew-white: Also, how about removing `@cancel` in `FormNewPage`? If that case is now impossible, we can remove `afterCancel` there. I see you removed the related test, which makes me think it's impossible now.

apps/central/test/components/form/upload.spec.js
```diff
@@ -198,8 +185,11 @@ describe('FormUpload', () => {
   it('implements some standard button things for the upload button', () => {
     mockLogin();
     testData.extendedProjects.createPast(1);
+    // await setFiles(component.get('input'), [xlsForm()]);
```
matthew-white: How about removing this line instead of commenting it out?

apps/central/src/components/form/upload.vue
```diff
@@ -103,9 +103,9 @@ definition for an existing form -->
         {{ file != null ? file.name : '' }}
       </div>
     </file-drop-zone>
-    <div class="actions">
+    <div v-if="file != null" class="actions">
```
matthew-white, line 106: How about adding a test that verifies this `v-if`?

- Initially, `.actions` is not rendered
- After a file is selected, it is rendered

apps/central/src/components/form/upload.vue
```diff
@@ -103,9 +103,9 @@ definition for an existing form -->
...
       <button type="button" class="btn btn-link" :aria-disabled="awaitingResponse"
-        @click="$emit('cancel')">
+        @click="clear()">
```
matthew-white, line 108: How about adding a test of this `@click`?

matthew-white (approved):

PR #1580, opened 2026-05-29
Form Draft Redesign Part 6: Miscellaneous

sadiqkhoja:
- Brought back CSS to hide tag element if no slot is provided
- Removed fullWidth from the draft page
- Aligned sections on draft page vertically when there is no active draft
- Restyle FormEditControl

image

### DIFF
--- a/apps/central/src/components/form/edit.vue
+++ b/apps/central/src/components/form/edit.vue
@@ -14,14 +14,9 @@ except according to the terms contained in the LICENSE file.
     @dragenter="dragHandler" @dragleave="dragHandler" @drop="dragHandler">
     <loading :state="formDraft.initiallyLoading"/>
     <template v-if="formDraft.dataExists">
-      <div class="row">
-        <div v-if="formDraft.isEmpty()" class="col-xs-6">
-          <form-edit-create-draft @success="fetchDraft(true)"/>
-        </div>
-        <div v-if="form.dataExists && form.publishedAt != null" class="col-xs-6">
-          <form-edit-published-version/>
-        </div>
-      </div>
+      <form-edit-published-version v-if="form.dataExists && form.publishedAt != null"/>
+      <form-edit-create-draft v-if="formDraft.isEmpty()" @success="fetchDraft(true)"/>
+
       <template v-if="formDraft.isDefined()">
         <form-edit-def @after-upload="afterUpload"/>
         <form-edit-attachments/>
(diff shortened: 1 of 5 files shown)


apps/central/src/components/form/edit/section.vue
```diff
@@ -77,8 +77,13 @@ $heading-margin-bottom: 10px;
   border-radius: 6px;
   color: $color-accent-primary;
   padding: 5px 9px;
+
+  // Hide the entire element if no tag slot is provided. We don't want the icon
+  // to be shown in that case.
```
matthew-white: > We don't want the icon to be shown in that case.

We no longer show an icon here, right? How about removing or updating this comment?

apps/central/src/components/form/edit/section.vue
```diff
@@ -77,8 +77,13 @@ $heading-margin-bottom: 10px;
...
+  &:has(> :nth-child(2):empty) { display: none; }
```
matthew-white: I don't see that `.form-edit-section-tag` has a second child, so I don't see how this selector would ever apply.

- **Idea 1: Select on the `<span>` being empty.** Something like `&:has(> span:empty) { display: none; }` or `&:has(> :first-child:empty) { display: none; }`
- **Idea 2: Remove the `<span>` altogether.** The `<span>` was helpful when we rendered an icon, but now I wonder whether it's doing anything. Can it just be removed? Then the selector here could be even simpler: `&:empty { display: none; }`

matthew-white (approved):

apps/central/src/components/form/edit.vue
```diff
@@ -14,14 +14,9 @@ except according to the terms contained in the LICENSE file.
...
-          <form-edit-create-draft @success="fetchDraft(true)"/>
-        </div>
-        <div v-if="form.dataExists && form.publishedAt != null" class="col-xs-6">
-          <form-edit-published-version/>
-        </div>
-      </div>
+      <form-edit-published-version v-if="form.dataExists && form.publishedAt != null"/>
+      <form-edit-create-draft v-if="formDraft.isEmpty()" @success="fetchDraft(true)"/>
+
       <template v-if="formDraft.isDefined()">
```
matthew-white: How about grouping these two closer together and using `v-if` + `v-else`? Right now, they use two separate `v-if`s, but their conditions (`formDraft.isEmpty()` for one, `formDraft.isDefined()` for the other) are mutually exclusive.

PR #1582, opened 2026-06-02
transifex:destructure: improve plural error message

alxndrsn:
Previous:

	Error: Expected the plural categories [one, few, many, other], but found [few, many, one, other]. Did you download the translations "to translate"?

This branch:

	Error: .appUser in locale "cs" expected the plural categories [one, few, many, other], but found [few, many, one, other]. Did you download the translations "to translate"?

Tested locally - `master` currently fails to destructure.

This should make it much clearer where the problematic JSON is - currently you have to divine it from a combination of log messages and grepping for the bad translation string.  Depending on how unique the translated text is, this might be more or less complicated.

Ideally the full key path would be supplied, but that would require rewriting the parsing/visiting code in `destructure()` as `JSON.parse()` doesn't provide full key context to the `reviver` function.

No effect for users.

No.

### DIFF
--- a/apps/central/bin/util/transifex.js
+++ b/apps/central/bin/util/transifex.js
@@ -100,7 +100,7 @@ class PluralForms {
   }
 
   // Transifex uses ICU plurals.
-  static fromTransifex(string, locale) {
+  static fromTransifex(key, string, locale) {
     const forms = [];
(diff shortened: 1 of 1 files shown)


matthew-white (approved):

matthew-white:
Definitely seems like an improvement, thank you @alxndrsn!

PR #1569, opened 2026-05-28
chore: updates the README with troubleshooting notes

latin-panda:
- Adds setup guidance to the README: Notes that `npm run dev` requires Docker host networking to be enabled (or to use `npm run dev-no-docker` instead), and adds a troubleshooting note for the "error code 500" when it's missing.

#### Before submitting this PR, please make sure you have:

- [ ] run `npm run test` and `npm run lint` and confirmed all checks still pass OR confirm CircleCI build passes
- [ ] verified that any code or assets from external sources are properly credited in comments or that everything is internally sourced
- [ ] run `npm run changeset` to generate a changeset file for changes that should be included in the release notes

PR #1587, opened 2026-06-02
dev/docker: tighten docker network binding

alxndrsn:

Tested locally.

Generally need to expose nginx beyond localhost.  If testing e.g. mobile devices on the local network, an option can be added.

No effect - dev only.

No.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
 vite: vite dev
 build: vite build --mode development --watch
 nginx-host: nginx -g "daemon off;" -c "$PWD/nginx-conf/nginx.conf" -p "$PWD"
-nginx-docker: docker run --rm --network=host -v "$PWD/nginx-conf":/etc/nginx/ -v "$PWD/.nginx":/etc/nginx/.nginx -v "$PWD/dist":/dist nginx:1.27.4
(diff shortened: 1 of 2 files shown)


matthew-white (approved):

Procfile
```diff
@@ -1,4 +1,4 @@
 vite: vite dev
 build: vite build --mode development --watch
 nginx-host: nginx -g "daemon off;" -c "$PWD/nginx-conf/nginx.conf" -p "$PWD"
-nginx-docker: docker run --rm --network=host -v "$PWD/nginx-conf":/etc/nginx/ -v "$PWD/.nginx":/etc/nginx/.nginx -v "$PWD/dist":/dist nginx:1.27.4
+nginx-docker: docker run --rm --publish 127.0.0.1:8686:8686 -v "$PWD/nginx-conf":/etc/nginx/ -v "$PWD/.nginx":/etc/nginx/.nginx -v "$PWD/dist":/dist nginx:1.27.4
```
matthew-white, line 4: #1569 updated the readme to mention `--network=host`. I'm thinking the readme will probably need some tweaks as part of or after this PR?

matthew-white:
> If testing e.g. mobile devices on the local network, an option can be added.

It's been a while, but in the past, I've used `ngrok` to test Collect / mobile devices. That approach is written up in the readme. I think that approach will continue to work after this change. It makes sense to me that if we ever wanted to test mobile devices in some other way, we could add an option for that.

Procfile
alxndrsn: Thanks, removed references to host networking.

PR #1589, opened 2026-06-03
dev: drop no-docker support

### LINKED ISSUE #1935: [frontend] drop support for non-dockerised nginx
Frontend dev relies on nginx to simulate production routing between API, frontend, OIDC etc.

Currently both system nginx and docker-based nginx options are supported:

* https://github.com/getodk/central-frontend/blob/a7f1bfc72a03888b74263a1def9bda47120cd11a/package.json#L7-L10
* https://github.com/getodk/central-frontend/blob/a7f1bfc72a03888b74263a1def9bda47120cd11a/Procfile#L3-L4

However, it's difficult to ensure both work.

Similar to other dependencies, pinning the nginx version gives:

* simpler dev environment setup
* consistent experience for different developers
* more similar dev & prod environments

Drop support for non-docker nginx for frontend dev.

alxndrsn:
Closes getodk/central#1935

* tested locally

* reduces maintenance burden
* increases consistency between dev environments
* increases similarity between dev & prod environments

No effect - dev only.

No.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Follow these steps to set up your development environment:
 
 - Install [Volta](https://volta.sh/)
 - Install dependencies by running `npm install`.
-- Install docker.
+- Install docker (for `nginx`).
 - Set up [ODK Central Backend](https://github.com/getodk/central-backend).
   - You will need to create a user using an ODK Central Backend command line script.
   - You will probably also want to promote that user to a sitewide administrator.
--- a/README.md
+++ b/README.md
@@ -33,8 +33,7 @@ Follow these steps to set up your development environment:
 
 - Install [Volta](https://volta.sh/)
 - Install dependencies by running `npm install`.
-- Install docker or NGINX.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -33,8 +33,7 @@ Follow these steps to set up your development environment:
 
 - Install [Volta](https://volta.sh/)
 - Install dependencies by running `npm install`.
-- Install docker or NGINX.
-  - The `npm run dev` runs NGINX in a Docker. If you'd rather skip that, install NGINX directly and use `npm run dev-no-docker` instead.
+- Install docker.
```
matthew-white:
```suggestion
- Install docker in order to run NGINX.
```

Just thinking that it could be nice to explain why Docker is needed. We talk more about NGINX below, so maybe it's nice to preview here that NGINX is part of the local setup.

matthew-white (approved):

README.md
alxndrsn: Added.

PR #1593, opened 2026-06-03
Web Forms setting on draft Form page

### LINKED ISSUE #1693: Show Web Forms opt-in on new form draft
As a project manager, I want to opt into Web Forms while creating an initial form draft so I don't have to remember to go to settings to opt in later and so that I can test with Web Forms while iterating on my draft.

[Figma](https://www.figma.com/design/ULpCuTKkIys6Y99YGyMcrF/Central?node-id=4110-4935&t=bmuJSpVh24PH66iX-0)

These might be prerequisites to making WF easier to opt into:
* https://github.com/getodk/central/issues/1101
* https://github.com/getodk/central/issues/1004

sadiqkhoja:
Closes https://github.com/getodk/central/issues/1693

image

Follows existing patterns

No

No

### DIFF
--- a/.changeset/quiet-bottles-like.md
+++ b/.changeset/quiet-bottles-like.md
@@ -1,0 +1,5 @@
+---
+"@getodk/central-frontend": patch
+---
+
+Web Forms setting (enketo/ODK-Web-Forms) can be set for the draft Forms.
--- a/.changeset/quiet-bottles-like.md
+++ b/.changeset/quiet-bottles-like.md
@@ -2,4 +2,4 @@
 "@getodk/central-frontend": patch
 ---
 
-Web Forms setting (enketo/ODK-Web-Forms) can be set for the draft Forms.
+Web Forms setting (enketo/ODK-Web-Forms) can be set for the draft Forms. (getodk/central#1693)
--- a/apps/central/src/components/form/edit.vue
+++ b/apps/central/src/components/form/edit.vue
@@ -13,6 +13,7 @@ except according to the terms contained in the LICENSE file.
   <div id="form-edit">
     <loading :state="formDraft.initiallyLoading"/>
     <template v-if="formDraft.dataExists">
+      <form-edit-web-form v-if="formDraft.isDefined()"/>
       <form-edit-published-version v-if="form.dataExists && form.publishedAt != null"/>
 
       <form-edit-create-draft v-if="formDraft.isEmpty()" @success="fetchDraft(true)"/>
@@ -47,6 +48,7 @@ import FormEditCreateDraft from './edit/create-draft.vue';
 import FormEditDef from './edit/def.vue';
 import FormEditDraftControls from './edit/draft-controls.vue';
 import FormEditPublishedVersion from './edit/published-version.vue';
+import FormEditWebForm from './edit/web-form.vue';
 import Loading from '../loading.vue';
 
 import useRoutes from '../../composables/routes';
--- a/apps/central/src/components/form/edit/web-form.vue
+++ b/apps/central/src/components/form/edit/web-form.vue
@@ -9,7 +9,7 @@
       </i18n-t>
       <radio-field v-model="webformsEnabled" :options="options"
         class="form-edit-web-form-options" :disabled="awaitingResponse"
-        :disabled-message="$t('updateInProgress')"
+        :disabled-message="$t('common.updateInProgress')"
         @update:model-value="update"/>
     </template>
   </form-edit-section>
@@ -43,10 +43,12 @@ const { request, awaitingResponse } = useRequest();
 const webformsEnabled = ref(formDraft.webformsEnabled);
 
 const { t } = useI18n();
+const { alert } = inject('container');
+
 
 const options = computed(() => [
   { value: true, text: t('option.webForms') },
-  { value: false, text: t('option.enketo') }
+  { value: false, text: t('common.useEnketo') }
 ]);
 
 const update = (newVal) => {
@@ -56,6 +58,7 @@ const update = (newVal) => {
   )
     .then(() => {
       formDraft.webformsEnabled = newVal;
+      alert.success(t('alert.settingSaved'));
     })
     .catch(() => {
       webformsEnabled.value = formDraft.webformsEnabled;
@@ -81,7 +84,6 @@ const update = (newVal) => {
     display: inline;
 
     .radio {
-      cursor: pointer;
       margin-bottom: 6px;
       margin-top: 6px;
       padding-left: 20px;
@@ -96,14 +98,15 @@ const update = (newVal) => {
   "en": {
     "title": "New default web form experience",
     "description": {
-      "full": "Newly created forms now use the new ODK web forms experience by default for previews, web submissions and submission editing. Some advanced form features may not yet be fully supported, {learnMore}.",
+      "full": "Newly created Forms now use the new ODK Web Forms experience by default for previews, web submissions and submission editing. Some advanced form features may not yet be fully supported, {learnMore}.",
       "learnMore": "learn more"
     },
     "option": {
-      "webForms": "New ODK Web Forms (recommended)",
-      "enketo": "Enketo compatibility mode"
+      "webForms": "New ODK Web Forms (recommended)"
     },
-    "updateInProgress": "Update request in progress"
+    "alert": {
+      "settingSaved": "Draft Form setting saved."
+    }
   }
 }
 </i18n>
--- a/apps/central/src/components/form/edit/web-form.vue
+++ b/apps/central/src/components/form/edit/web-form.vue
@@ -1,0 +1,109 @@
+<template>
+  <form-edit-section id="form-edit-web-form" no-border>
+    <template #title>{{ $t('title') }}</template>
+    <template #body>
+      <i18n-t tag="p" keypath="description.full" class="form-edit-web-form-description">
+        <template #learnMore>
+          <doc-link to="/web-forms-intro#features-not-yet-supported">{{ $t('description.learnMore') }}</doc-link>
+        </template>
+      </i18n-t>
+      <radio-field v-model="webformsEnabled" :options="options"
+        class="form-edit-web-form-options" :disabled="awaitingResponse"
+        :disabled-message="$t('updateInProgress')"
+        @update:model-value="update"/>
+    </template>
+  </form-edit-section>
+</template>
+
+<script setup>
+import { computed, inject, ref } from 'vue';
+import { useI18n } from 'vue-i18n';
+
+import DocLink from '../../doc-link.vue';
+import FormEditSection from './section.vue';
+import RadioField from '../../radio-field.vue';
+
+import useRequest from '../../../composables/request';
+import { apiPaths } from '../../../util/request';
+import { useRequestData } from '../../../request-data';
+
+defineOptions({
+  name: 'FormEditWebForm'
+});
+
+const projectId = inject('projectId');
+const xmlFormId = inject('xmlFormId');
+
+const { resourceView } = useRequestData();
+const formDraft = resourceView('formDraft', (data) => data.get());
+
+const { request, awaitingResponse } = useRequest();
+
+// formDraft exists and is defined.
+const webformsEnabled = ref(formDraft.webformsEnabled);
+
+const { t } = useI18n();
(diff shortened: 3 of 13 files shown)


.changeset/quiet-bottles-like.md
```diff
@@ -0,0 +1,5 @@
+---
+"@getodk/central-frontend": patch
+---
+
+Web Forms setting (enketo/ODK-Web-Forms) can be set for the draft Forms.
```
matthew-white: Would you mind referencing the issue here as well? In GitHub release notes, I generally try to reference the issue, since that's usually where design and release criteria are discussed.

apps/central/src/locales/en.json5
```diff
@@ -136,6 +136,7 @@
         "update": "@:audit.action.default.update",
         "update_draft_set": "Create or Update Draft",
         "update_draft_replace": "Replace Draft",
+        "update_draft_update": "Draft Settings Changed",
```
matthew-white: How about adding this new action as an option in `AuditFiltersAction`? No test needed, just adding it to the `<select>` options.

matthew-white: I think we should test that `AuditRow` is working correctly for this new audit log action. How about adding a test along those lines to test/components/audit/table.spec.js?

apps/central/src/components/form/edit/web-form.vue
```diff
@@ -0,0 +1,109 @@
...
+{
+  "en": {
+    "title": "New default web form experience",
+    "description": {
+      "full": "Newly created forms now use the new ODK web forms experience by default for previews, web submissions and submission editing. Some advanced form features may not yet be fully supported, {learnMore}.",
+      "learnMore": "learn more"
+    },
+    "option": {
+      "webForms": "New ODK Web Forms (recommended)",
+      "enketo": "Enketo compatibility mode"
```
matthew-white: This string also appears in #1598. It'd be good to put it in en.json5 or to use `@transifexKey` so that it's not translated twice.

apps/central/src/components/form/edit/web-form.vue
```diff
@@ -0,0 +1,109 @@
...
+      width: max-content;
+    }
+  }
+}
+</style>
+
+<i18n lang="json5">
+{
+  "en": {
+    "title": "New default web form experience",
```
matthew-white, line 99: I'm thinking we should capitalize "web form" or at least "form".

apps/central/src/components/form/edit/web-form.vue
```diff
@@ -0,0 +1,109 @@
...
+    "description": {
+      "full": "Newly created forms now use the new ODK web forms experience by default for previews, web submissions and submission editing. Some advanced form features may not yet be fully supported, {learnMore}.",
```
matthew-white: I feel like some words or phrases here need to be capitalized, things like "Web Forms", "Form features", "Submission editing".

apps/central/src/components/form/edit/web-form.vue
```diff
@@ -0,0 +1,109 @@
...
+    margin-bottom: 18px;
+    max-width: unset;
+  }
+
+  .form-edit-web-form-options {
+    margin-top: 12px;
+    display: inline;
+
+    .radio {
+      cursor: pointer;
```
matthew-white: Why is `cursor: pointer` needed here? It's a little surprising to me that other places wouldn't need that style as well. My instinct is that a style like that would be best placed in the `RadioField` component or in app.scss.

apps/central/src/components/form/edit/web-form.vue
```diff
@@ -0,0 +1,109 @@
...
+    "title": "New default web form experience",
+    "description": {
+      "full": "Newly created forms now use the new ODK web forms experience by default for previews, web submissions and submission editing. Some advanced form features may not yet be fully supported, {learnMore}.",
+      "learnMore": "learn more"
+    },
+    "option": {
+      "webForms": "New ODK Web Forms (recommended)",
+      "enketo": "Enketo compatibility mode"
+    },
+    "updateInProgress": "Update request in progress"
```
matthew-white: What do you think about moving this to en.json5? I feel like we're likely to use it elsewhere in the future.

apps/central/src/components/form/edit/web-form.vue
```diff
@@ -0,0 +1,109 @@
...
+]);
+
+const update = (newVal) => {
+  request.patch(
+    apiPaths.formDraft(projectId, xmlFormId),
+    { webformsEnabled: newVal }
+  )
+    .then(() => {
+      formDraft.webformsEnabled = newVal;
+    })
```
matthew-white, line 62: Should there be a success message/toast?

apps/central/src/components/form/edit/web-form.vue
```diff
@@ -0,0 +1,109 @@
...
+@import '../../../assets/scss/variables';
+
+#form-edit-web-form {
+  background-color: $color-success-light;
+  border-radius: 12px;
+  padding: 20px;
+
+  .form-edit-web-form-description {
+    margin-bottom: 18px;
+    max-width: unset;
```
matthew-white, line 79: I feel like we're beginning to `unset` this in a bunch of places. I don't have a good sense of when that's needed vs. not. A thought for the future: should we remove `max-width` from the `text-block` Sass mixin if we no longer prefer its early wrapping?

apps/central/test/components/form/edit/web-form.spec.js
```diff
@@ -0,0 +1,112 @@
...
+      testData.extendedForms.createPast(1, { xmlFormId: 'a', draft: true, webformsEnabled: true });
+      return load('/projects/1/forms/a/draft')
+        .complete()
+        .request(app => {
+          const enketoRadio = app.findAll('#form-edit-web-form input[type="radio"]')[1];
+          return enketoRadio.setValue(true);
+        })
+        .respondWithData(() => {
+          testData.extendedFormDrafts.update(-1, { webformsEnabled: false });
+          return { success: true };
```
matthew-white:
```suggestion
          return testData.standardFormDrafts.last();
```

I think it's a good idea to try to match the API response, even though right now it's not actually used by the component.

This comment also applies to other lines in this file.

apps/central/test/components/form/edit/web-form.spec.js
```diff
@@ -0,0 +1,112 @@
...
+        })
+        .afterResponse(app => {
+          const radios = app.findAll('#form-edit-web-form input[type="radio"]');
+          radios[0].element.disabled.should.be.false;
+          radios[1].element.disabled.should.be.false;
+        });
```
matthew-white, line 92: Is there anything else in Frontend that will use `formDraft.webformsEnabled`? It'd be cool if we could see the change to `webformsEnabled` reflected in another component as well.

Or will nothing else in Frontend use `webformsEnabled`? Maybe the main consumer of `formDraft.webformsEnabled` will be the Web Forms web app. So that means that the Web Forms web app will need to be careful about the difference between `form.webformsEnabled` and `formDraft.webformsEnabled`, which may differ from one another.

apps/central/src/locales/en.json5
matthew-white:
```suggestion
        "update_draft_update": "Update Draft Settings",
```

1. Almost all of these messages use the format/order "[verb] [noun]", so I think we should try to match that.
2. Generally in other messages here, the word "update" is used, not "change(d)".

matthew-white (approved):
Looks good! I'm leaving several comments, but mostly just small things. @sadiqkhoja, feel free to make those changes, then merge. But feel free to tag me if there's anything you'd like me to take another look at.

apps/central/src/components/form/edit.vue
```diff
@@ -14,6 +14,7 @@ except according to the terms contained in the LICENSE file.
     @dragenter="dragHandler" @dragleave="dragHandler" @drop="dragHandler">
     <loading :state="formDraft.initiallyLoading"/>
     <template v-if="formDraft.dataExists">
+      <form-edit-web-form v-if="formDraft.isDefined()"/>
```
matthew-white, line 16: What's the status of Web Forms and encryption? In the issue, there's discussion about what to render in the case that the form is encrypted. In the Backend PR, I see a reference to `keyId`. Should we check `formDraft.get().keyId` here? Or will Web Forms have implemented encryption by the time of release? (But if that's the case, why does Backend need to reference `keyId`?)

If you do end up wanting to add logic and/or tests related to `keyId`, it'd be great if that could go in a separate commit or even its own follow-up PR. That's logic that I'd probably want to take a look at (even if it's just a quick look).

apps/central/src/components/form/edit/web-form.vue
lognaturel: This is something @nicoleorlowski and I have been discussing. I think in this context we mean "web form functionality" generically so lowercase feels appropriate to me. We're not talking about ODK Web Forms here and we're not talking about the Form abstraction in Central either, we're talking about the colloquial "web form" concept.

apps/central/src/components/form/edit/web-form.vue
lognaturel: ODK Web Forms definitely, in that context we're talking about the technology/product.

I don't know how recently we've talked about this but I do want to state that at some point I'd love to back out of capitalizing system nouns. It doesn't localize well at all and it adds a lot of cognitive overhead to all ODK-related writing.

For now I propose
> "full": "Newly created Forms now use the new ODK Web Forms experience by default for previews, web submissions and submission editing. Some advanced form features may not yet be fully supported, {learnMore}."

I'd tend to leave "form features" because there we're talking about XLSForm features really, not features of the Central `Form` concept.

I agree with @matthew-white that "web submissions and Submission editing" would probably be the most "correct" but it's just hard to read.

lognaturel: I think "submission editing" is also generic capability, really.

apps/central/src/components/form/edit.vue
lognaturel: To crosslink things, encryption is merged and will be included: https://github.com/getodk/central-backend/pull/1840#discussion_r3376298202

apps/central/src/components/form/edit/web-form.vue
matthew-white: > I don't know how recently we've talked about this but I do want to state that at some point I'd love to back out of capitalizing system nouns.

I'd also love to stop capitalizing. Maybe I'll stop flagging it in code review. Even once a PR is merged, we can always tweak text/capitalization that looks wrong on staging.

lognaturel: Yes, that seems aligned with our new approach to text in general!

apps/central/src/components/form/edit/web-form.vue
sadiqkhoja: added to RadioField. It was not needed before because we use RadioField with button appearance and they have `.btn` class, which already has `cursor: pointer`

apps/central/src/components/form/edit/web-form.vue
sadiqkhoja: created https://github.com/getodk/central/issues/2008

apps/central/test/components/form/edit/web-form.spec.js
sadiqkhoja: Yes, `formDraft.webformsEnabled` is consumed by Web Forms web app, so I'm thinking to add an e2e test for this. Logged an issue to track this https://github.com/getodk/central/issues/2009

e2e-tests/tests/web-forms.spec.js
```diff
@@ -29,9 +28,6 @@ test.describe('ODK Web Forms', () => {
       {
         description: 'New Submission',
         url: ({ enketoId }) => `/-/${enketoId}`, requireLogin: true
-      }, {
-        description: 'Edit Submission',
-        url: ({ enketoId, instanceId }) => `/-/edit/${enketoId}?instance_id=${instanceId}`, requireLogin: true
```
sadiqkhoja, line 34: in case of `webformsEnabled`, backend doesn't pass the submission data to Enketo hence this endpoint won't work even we call EditSubmission backend point. 

PS: Previously this URL was showing form with Enketo.

PR #1598, opened 2026-06-04
Fixes: Text and style of web forms section on form settings page updated

sadiqkhoja:
Related https://github.com/getodk/central/issues/1681

image

### DIFF
--- a/apps/central/src/components/form/settings.vue
+++ b/apps/central/src/components/form/settings.vue
@@ -56,9 +56,9 @@ except according to the terms contained in the LICENSE file.
                 <label>
                   <input v-model="webformsEnabled" name="webformsEnabled" type="radio" :value="false"
                     @change="confirmationModal.show({ webformsEnabled: false })">
-                    <span class="label-title">
-                      {{ $t('webFormsSetting.enketoDefault') }}
-                    </span>
+                  <span class="label-title">
+                    {{ $t('common.useEnketo') }}
+                  </span>
                   <br>
                   {{ $t('webFormsSetting.enketoDescription') }}
                 </label>
@@ -209,8 +209,7 @@ const hideAndReset = () => {
       "webForms": "Web Forms",
       // Description of a section. {formName} is replaced with the name of the Form
       "description": "Fill out, preview and edit your “{formName}” Form using",
-      // The word "Enketo" should not be translated
-      "enketoDefault": "Enketo compatibility mode",
+      // This text is shown under the option of Enketo for Form settings.
       "enketoDescription": "No longer actively developed. Use for specialized features not yet supported by the new system.",
       "odkWebFormsDescription": "A faster, updated experience that will replace the current version over time.",
       // Success message when Enketo is selected as web form technology. {formName} is replaced with the name of the Form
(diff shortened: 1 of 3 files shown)


sadiqkhoja:
I don't think it is necessary to create changelog for this.

apps/central/src/components/form/settings.vue
```diff
@@ -190,7 +210,9 @@ const hideAndReset = () => {
       // Description of a section. {formName} is replaced with the name of the Form
       "description": "Fill out, preview and edit your “{formName}” Form using",
       // The word "Enketo" should not be translated
-      "enketoDefault": "Enketo (default)",
+      "enketoDefault": "Enketo compatibility mode",
+      "enketoDescription": "No longer actively developed. Use for specialized features not yet supported by the new system.",
```
matthew-white, line 213: How about adding a comment for translators that this refers to Enketo? In some languages, words like "developed" will need to match "Enketo".

apps/central/src/locales/en.json5
```diff
@@ -475,7 +475,7 @@
     "versionShort": "v{version}",
     // This text is shown on its own before details about a warning.
     "warning": "Warning",
-    "beta": "Beta",
+    "recommended": "recommended",
```
matthew-white, line 479: It might be possible to keep this in `common`, but I feel like translators of certain languages will wonder what this refers to. Since "recommended" is an adjective/participle, some languages will inflect it to match the corresponding noun. So translators may wonder what the noun is.

- Option 1: Don't keep this message in `common`. Move it to the component and add a comment there saying what it refers to specifically (what the noun is).
- Option 2: Keep the message in `common`, but add a translator comment saying that this is a generic message that may be shown next to any recommendation. It's standalone text that isn't part of a longer phrase.

Not sure if one option or the other is better, I bet either would work well. But I think some translator comment will be needed either way.

apps/central/src/components/form/settings.vue
```diff
@@ -190,7 +210,9 @@ const hideAndReset = () => {
       // Description of a section. {formName} is replaced with the name of the Form
       "description": "Fill out, preview and edit your “{formName}” Form using",
       // The word "Enketo" should not be translated
-      "enketoDefault": "Enketo (default)",
+      "enketoDefault": "Enketo compatibility mode",
```
matthew-white: How about changing the key? I feel like the `Default` part doesn't fit anymore. Something like `enketoTitle` or `enketoMode` or `enketoCompatibility`.

apps/central/src/components/form/settings.vue
```diff
@@ -43,16 +40,27 @@ except according to the terms contained in the LICENSE file.
...
                 </label>
               </div>
               <div class="radio">
                 <label>
-                  <input v-model="webformsEnabled" name="webformsEnabled" type="radio" :value="true"
-                    @change="confirmationModal.show({ webformsEnabled: true })">
-                  ODK Web Forms
+                  <input v-model="webformsEnabled" name="webformsEnabled" type="radio" :value="false"
+                    @change="confirmationModal.show({ webformsEnabled: false })">
+                    <span class="label-title">
```
matthew-white: This `<span>` looks overly indented to me, it's not a child of `<input>`.

matthew-white (approved):

PR #1599, opened 2026-06-04
fix(#738): revokes media's object URL on Web Forms unmount

### LINKED ISSUE #738: Memory leak caused by object url never being revoked
### Description

In [MediaBlockBase](https://github.com/getodk/web-forms/blob/f6372a02c6146a8baf3f63c3c95dcaec655fad46/packages/web-forms/src/components/common/media/MediaBlockBase.vue#L55) we create an object url for a blob which is never [revoked](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static). This creates a potentially large blob which is referenced by the page and not GCed until leaving the WF SPA.

Currently you can't really navigate to another page so it's not causing any noticeable issues.

At the same time ensure the `mediaCache` is being GCed to ensure that the URLs are free to be GCed.

latin-panda:
Closes https://github.com/getodk/web-forms/issues/738

- Manually verified using a temporary button to simulate unmount: the hook ran and the cache was cleared.

 - The cache is intentionally not cleared on form submission because the same form definition (and its media attachments) is reused when the user fills the form again (avoiding unnecessary re-fetching of the same blobs). 
 - The cache stores form definition media, no user uploads. 
- Revoking on unmount covers all cases where the component is destroyed (e.g. navigating away, or a redirect), without interfering with the cache during a session.

No

No

### DIFF
--- a/.changeset/eager-teeth-mate.md
+++ b/.changeset/eager-teeth-mate.md
@@ -1,0 +1,6 @@
+---
+"@getodk/central-frontend": patch
+"@getodk/web-forms": patch
+---
+
+Revoke media's object URLs on Web Forms unmount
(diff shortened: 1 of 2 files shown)


latin-panda:
@garethbowen What do you think about this solution?

garethbowen (approved):
Looks good.

It's hard to track down every usage and ensure they're cleaned up but I think this solves this particular issue.

PR #1603, opened 2026-06-05
ci/e2e-tests: do not retry failed tests

alxndrsn:
Retries in CI has been set to `2` since these tests were introduced in #1200.

Tests are not currently flaky, so retrying failed tests just wastes time and hides new issues.

Ref https://github.com/getodk/central/issues/1923

CI.

Can always reintroduce retries if this causes build instability, or fix the underlying issues.

No effect - just tests.

No.

### DIFF
--- a/e2e-tests/playwright.config.js
+++ b/e2e-tests/playwright.config.js
@@ -12,7 +12,7 @@ export default defineConfig({
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
-  retries: process.env.CI ? 2 : 0,
+  retries: 0,
   /* Opt out of parallel tests on CI. */
   workers: process.env.CI ? 1 : undefined,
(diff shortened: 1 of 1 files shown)


e2e-tests/playwright.config.js
```diff
@@ -12,7 +12,7 @@ export default defineConfig({
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
```
matthew-white: How about removing this comment as well?

matthew-white (approved):

e2e-tests/playwright.config.js
alxndrsn: :+1:

alxndrsn:
```suggestion
  /* Do not retry failed tests. */
```

alxndrsn: Comment updated.

PR #1605, opened 2026-06-05
ci/e2e-tests: bail after first failure

alxndrsn:
This should lead to quicker pipeline failure, and so faster feedback for developers.

Ref https://github.com/getodk/central/issues/1923

CI.

Flaky tests should be fixed, not glossed over.  Also currently there's no flake, so this should help prevent regressions.

No effect - just the test suite.  But it should lead to more solid tests/product.

No.

### DIFF
--- a/e2e-tests/playwright.config.js
+++ b/e2e-tests/playwright.config.js
@@ -11,6 +11,8 @@ export default defineConfig({
   fullyParallel: false,
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
+  /* Bail after first test failure in CI */
+  maxFailures: process.env.CI ? 1 : undefined,
   /* Retry on CI only */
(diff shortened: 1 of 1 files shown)


matthew-white:
On rare occasion, a test only fails in CI, not locally. In such a case, it can be helpful to see the full list of failing tests, not just the first one. But I guess it should always be possible to temporarily loosen this configuration while debugging such a case.

matthew-white:
I wonder if we should make a similar change for `unit-tests` as well (bailing after the first failure) 🤔

matthew-white (approved):

alxndrsn:
> On rare occasion, a test only fails in CI, not locally. In such a case, it can be helpful to see the full list of failing tests, not just the first one. But I guess it should always be possible to temporarily loosen this configuration while debugging such a case.

:+1:

We could also set `maxFailures` a little higher in future.

PR #1606, opened 2026-06-06
dev/nginx: fix central-backend hostname

alxndrsn:
Due to a50d496ee24f02cee090de2d3abb2d969246370e, central-backend is no longer available on `localhost`, so it must be accessed via the explicit `host.docker.internal` hostname.

Note:

* this approach may fail if `backend` is also run via Docker.  If that's a use-case we need to support then we may need to revert to --network=host until a more universal approach is implemented
* a similar fix may be required for `enketo`

- [x] tested locally on linux
- [x] tested locally on macos

See above.

No effect for end users.

No.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
 vite: vite dev
 build: vite build --mode development --watch
-nginx: docker run --rm --publish 127.0.0.1:8686:8686 -v "$PWD/nginx-conf":/etc/nginx/ -v "$PWD/.nginx":/etc/nginx/.nginx -v "$PWD/dist":/dist nginx:1.27.4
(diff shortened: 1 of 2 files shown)


sadiqkhoja (approved):
thank you

Procfile
```diff
@@ -1,3 +1,3 @@
 vite: vite dev
 build: vite build --mode development --watch
-nginx: docker run --rm --publish 127.0.0.1:8686:8686 -v "$PWD/nginx-conf":/etc/nginx/ -v "$PWD/.nginx":/etc/nginx/.nginx -v "$PWD/dist":/dist nginx:1.27.4
+nginx: docker run --rm --publish 127.0.0.1:8686:8686 --add-host=host.docker.internal:host-gateway -v "$PWD/nginx-conf":/etc/nginx/ -v "$PWD/.nginx":/etc/nginx/.nginx -v "$PWD/dist":/dist nginx:1.27.4
```
sadiqkhoja, line 3: this is not strictly required for MacOS (running OrbStack). `host.docker.internal` is always added by default

alxndrsn: Required for Linux, it seems.  Glad it doesn't break macOS :+1:

ktuite (approved):
works for me, too

PR #1607, opened 2026-06-07
ci/e2e-tests: show logs at end of run

alxndrsn:
Eases debugging issues with docker containers.

CI run at https://github.com/getodk/central-frontend/actions/runs/27203096300/job/80312833078?pr=1607:

<kbd>
Screenshot_2026-06-20_10-24-29

</kbd>

It's simple, and consistent with other projects, e.g. https://github.com/getodk/central/blob/6b1ac3b8a8f4b21c74eda22d4773a332a4c724e3/.github/workflows/test.yml#L53-L61

No effect on users.

No.


.github/workflows/tests.yml
```diff
@@ -5,74 +5,7 @@ on:
   pull_request:
 
 jobs:
-  static-checks:
```
alxndrsn, line 12: revert

e2e-tests/tests/encrypted-submission.spec.js
alxndrsn, line 1: revert

.github/workflows/tests.yml
```diff
@@ -136,6 +136,10 @@ jobs:
       with:
         name: Playwright Artifacts
         path: central/client/test-results
+    - if: always()
+      name: Docker Container Logs
+      working-directory: central
+      run: docker compose logs || true
```
latin-panda, line 142: Consider displaying about 100 lines to avoid overwhelming when debugging 
```
docker compose logs --tail=100
```

latin-panda (approved):
LGTM! Just adding one suggestion below.

.github/workflows/tests.yml
alxndrsn: I'm not sure when this would be helpful, but it risks cutting off relevant logs.

PR #1608, opened 2026-06-08
fix(web-forms#535): separate enketo and web-forms into their own vue app

garethbowen:
Things left to do...

- [x] The asset bundles are still combined so users of either app end up downloading all the code for both. I think I have a way forward but this doesn't impact the majority of this review so I thought it was worth a first look
- [x] The unit tests are being difficult, because they rely on a lot of the infrastructure of central to run. I got some running on vitest but the solution might be to go back to karma and do the bare minimum.
- [x] Depends on associated central patch: https://github.com/getodk/central/pull/1960

https://github.com/getodk/web-forms/issues/535


apps/forms/src/components/submission.vue
```diff
@@ -0,0 +1,243 @@
...
+    }
+    webFormsEnabled.value = false;
+  }
+  return formConfig;
+
+};
+
+const permits = (project: Project, verbs: string[]) => {
+  return !verbs.every(verb => project.verbs.includes(verb));
+}
```
latin-panda, line 149: This negation got me thinking for a bit. After checking `master`, it seems we can remove the negation 
```suggestion
 const permits = (project: Project, verbs: string[]) => verbs.every(verb => project.verbs.includes(verb));
```
Then update L140, to include `!` before the `permits(...)` 
```js
if ((route.name === 'SubmissionNew' || route.name === 'DraftSubmissionNew') && !permits(project, ['submission.create'])) {
    return false;
  }
```

nginx-conf/nginx.conf
```diff
@@ -143,6 +143,18 @@ http {
       return 404;
     }
 
+    # Only relevant for `npm run dev:build`, not `npm run dev`.
+    location ~ ^/f/.+|/projects/\d+/forms/[^/]+(?:/draft)?(?:/preview|/submissions/new(?:/offline)?/?|/submissions/[^/]+/edit)/?$ {
```
latin-panda, line 156: Same feedback [from here](https://github.com/getodk/central/pull/1960#pullrequestreview-4450229635)

vite.config.js
```diff
@@ -47,11 +47,29 @@ const devServer = {
   cors: false
 };
 
+// used to route requests to the right app in development
+const devAppRouter = () => ({
+  name: 'dev-app-router',
+  configureServer(server) {
+    server.middlewares.use((req, res, next) => {
+      // must match the regex paths defined in Nginx
+      const newSubmissionUrlRegex = /^\/projects\/\d+\/forms\/[^/]+(?:\/draft)?(?:\/preview|\/submissions\/new(?:\/offline)?\/?|\/submissions\/[^/]+\/edit)\/?/;
```
latin-panda: Same feedback [from here](https://github.com/getodk/central/pull/1960#pullrequestreview-4450229635)

Is there a way to define this in one place and share it in the different configs?

apps/forms/src/components/enketo-iframe.vue
```diff
@@ -14,35 +14,35 @@ except according to the terms contained in the LICENSE file.
...
-    required: true
-  },
-  actionType: {
-    type: String,
-    required: true
-  },
-  instanceId: String
-});
-
 const emit = defineEmits(['loaded']);
```
latin-panda: Who calls this emit? Is it dead code?

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -416,42 +237,156 @@ const handleSubmit = async (payload, callback) => {
...
-    if (styleTag.textContent.includes('form-initialization-status') ||
-        // WF's reset.css
-        styleTag.textContent.replace(/\s+/g, '').includes('body{all:revert;')) {
-      // eslint-disable-next-line no-param-reassign
-      styleTag.disabled = disabled;
+const transformAttachmentResponse = async (response: Response) => {
+  const { status, statusText, headers } = response;
+
+  const fetchHeaders = new Headers();
+  for (const [key, value] of Object.entries(headers)) {
```
latin-panda: The response comes from the native `fetch`, so the headers are a [Headers type](https://developer.mozilla.org/en-US/docs/Web/API/Headers/entries#examples) that should be iterated using its entries function.

```suggestion
  for (const [key, value] of headers.entries()) {
```

apps/central/src/router.js
```diff
@@ -85,7 +85,13 @@ router.afterEach(unlessFailure(to => {
   // gives percentage encoded value - central#919.
   // Note: to.path is '/#/some/path' when the route has query parameter after '#'
   // e.g. /#/some/path?next=/ - central#939
-  router.beforeEach(to => (to.fullPath.startsWith('/#/') ? to.fullPath.substring(2) : true));
+  router.beforeEach(to => {
+    if (to.fullPath.startsWith('/#/')) {
+      // This must do a full page refresh because the page may end up in a different vue app
+      window.location.href = to.fullPath.substring(2);
```
latin-panda: The user's back button takes them to the hash URL, which redirects again. The `replace` swaps the entry.
Also, it can return false here to cancel any forward loading in central (if any callback is reacting to navigation events)
```suggestion
     window.location.replace(to.fullPath.slice(2));
     return false; 
```

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -416,42 +237,156 @@ const handleSubmit = async (payload, callback) => {
...
+      resolveAttachment: fetchSubmissionAttachment
+    };
+  }
+  return null;
+}); 
+
 
-#web-form-renderer-submission-modal pre {
-  white-space: pre-wrap;
+if (isEdit.value) {
```
latin-panda: Preferably wrap in an init function or onMount hook

apps/forms/index.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<!--
+Copyright 2017 ODK Central Developers
```
latin-panda: There have been conversations in Slack about this (maybe in the team channel?). I think we resolved to drop this for new development, since this is a new app. It could be dropped.

latin-panda (changes requested):
It's great to see this coming along! I did an initial review, and tomorrow I'll take another deep look.

apps/forms/src/components/submission.vue
garethbowen: Oops! I think I was testing to ensure it worked but removed the wrong `!` when I was done. Fixed.

nginx-conf/nginx.conf
garethbowen: Done.

vite.config.js
garethbowen: I don't know if that'll be possible between nginx and vue, but I'm sure we can at least get the nginx using the same config. There's an ongoing convo about this because the mismatch between dev and prod is risky. I think the docker work will help us move in that direction.

I've added comments and renamed the regexes to be more clear.

apps/forms/src/components/enketo-iframe.vue
garethbowen: Yes! Thank you.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Done.

apps/central/src/router.js
garethbowen: Nice. Thanks.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Done.

apps/central/src/router.js
garethbowen: It turns out returning `false` cancels the request so which caused playwright to bail. Because this isn't a security issue I've left it continuing on which might be a little bit of a waste but means the request completes cleanly.

apps/forms/index.html
garethbowen: Done.

garethbowen:
@latin-panda I've resolved all your comments. I've also...
- fixed the bundling so the central app no longer downloads web-forms
- added error handling so if your session times out you get redirected to the login page

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -416,42 +225,160 @@ const handleSubmit = async (payload, callback) => {
...
+      fetchHeaders.append(key, value.replace('; charset', ';charset'));
+    } else {
+      fetchHeaders.append(key, value);
     }
+  }
+
+  let body;
+  const contentType = headers.get('content-type');
+  if (contentType && (contentType.includes('application/json') || contentType.includes('application/geo+json'))) {
+    body = await response.json();
```
latin-panda: In L254, this body is passed to the [Response constructor](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response#body), but it doesn't accept the data type that `.json()` produces and ends up converting it to a string (`"[object Object]"`). That's probably why [it initially used JSON.stringify(body)](https://github.com/getodk/central-frontend/blob/984593142dd2b8b3a31ab792e8f68d6e03bcb053/apps/central/src/components/web-form-renderer.vue#L159).

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -1,368 +1,132 @@
...
+const postPrimaryInstance = async (file:File) => {
+  const draftPath = props.form.draft ? '/draft' : '';
+  let url;
+  let method;
   if (isEdit.value) {
-    url = apiPaths.submission(form.projectId, form.xmlFormId, props.instanceId);
+    url = `/v1/projects/${props.form.projectId}/forms/${props.form.xmlFormId}/submissions/${props.instanceId}`;
+    method = 'PUT';
+  } else {
+    url = `/v1/projects/${props.form.projectId}/forms/${props.form.xmlFormId}${draftPath}/submissions`;
```
latin-panda, line 57: When does it append a `st` query parameter for public-link?

latin-panda: Same question for L59, L151, L262, 268, 279. 

What do you think about having a URL builder and unit testing it? Looking to centralize it

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -1,368 +1,132 @@
...
-    submissionResult.attachmentResult.set(attachment.name, {
-      success: false
-    });
-  });
-
-  clearForm = () => {
-    clearFormCallback({ next: POST_SUBMIT__NEW_INSTANCE });
-  };
+const submissionPath = () => {
+  return `/v1/projects/${props.form.projectId}/forms/${props.form.xmlFormId}/submissions/${props.instanceId}`;
```
latin-panda: The [suffix is used only for api](https://github.com/getodk/central-frontend/blob/984593142dd2b8b3a31ab792e8f68d6e03bcb053/apps/central/src/composables/routes.js#L115) calls, for L116 we need it without `/v1/`

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -373,25 +137,53 @@ const handleResult = () => {
...
+  try {
+    const headers = {
+      'Content-Type': attachment.type,
+      'X-Requested-With': 'XMLHttpRequest'
+    };
+    const response = await fetch(url, { body: attachment, headers, method: 'POST' });
+    if (response.ok) {
+      const data = await response.json();
+      result = { success: true, data };
+    }
```
latin-panda: It should assign `result` when `response.ok` is false to ensure consistent handling in subsequent flows after this code.

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -1,368 +1,132 @@
...
+const visibleModal = ref();
 
-/**
- * Hide the modal
- */
-const hideModal = () => {
-  submissionModal.hide();
+const getAttachment = (requestUrl: URL) => {
+  const encodedName = encodeURIComponent(requestUrl.pathname.split('/').pop()!);
+  const draftPath = '';
```
latin-panda: In L64, there's a check for `props.form.draft` that we could use here as well.

Or does it even need `draftPath`? What happens if there are attachments for form drafts?

apps/forms/src/components/submission.vue
```diff
@@ -0,0 +1,237 @@
...
+*/
+const route = useRoute();
+const router = useRouter();
+
+const projectId: number | null = route.params.projectId ? Number.parseInt(route.params.projectId as string) : null;
+const formId: string | null = route.params.xmlFormId ? encodeURIComponent(route.params.xmlFormId as string) : null;
+const instanceId: string | null = route.params.instanceId ? encodeURIComponent(route.params.instanceId as string) : null;
+const enketoId: string | null = route.params.enketoId ? encodeURIComponent(route.params.enketoId as string) : null;
+const useWebForms: boolean = route.query.webforms === 'true';
+const offline: boolean = route.params.offline === 'offline';
```
latin-panda: All of these are only set to a new value on page reload (on mount), instead of being reactive to the app route (`computed(() => route.params...`) Is that okay? It might surprise others with little context.

apps/forms/src/components/submission.vue
```diff
@@ -0,0 +1,237 @@
...
+    if (!access) {
+      window.location.replace('/');
+    } else {
+      form.value = formConfig;
+      loadingState.value = false;
+    }
+  } catch (e) {
+    if (e instanceof RequestError && e.statusCode >= 401 && e.statusCode < 404) {
+      // not logged in
+      window.location.href = '/login?next=/wf' + window.location.pathname;
```
latin-panda: I think it needs to preserve `window.location.search` for [the navigation after login](https://github.com/getodk/central-frontend/blob/984593142dd2b8b3a31ab792e8f68d6e03bcb053/apps/central/src/components/account/login.vue#L192)

apps/forms/src/components/submission.vue
```diff
@@ -0,0 +1,237 @@
...
+    throw new RequestError('Form not found', 404);
+  }
+
+  redirectEnketoUrls(formConfig);
+
+  if (formConfig.webformsEnabled || useWebForms) {
+    xform.value = await getFormXml(formConfig.projectId, formConfig.xmlFormId, formConfig.draft, st)
+    webFormsEnabled.value = true;
+  } else {
+    if (offline) {
```
latin-panda: Little comment to not lose track. 
```suggestion
    if (offline) {
    // TODO: Update once Web Forms has support for offline 
```

apps/forms/src/components/submission.vue
```diff
@@ -0,0 +1,237 @@
...
+  </div>
+  <div v-else-if="errorState" class="form-load-error">
+    {{ $t('formNotFound') }}
+  </div>
+  <template v-else-if="webFormsEnabled">
+    <WebFormRenderer :form="form!" :xform="xform!" :instance-id="instanceId" :action-type="props.actionType ?? 'new'"/>
+  </template>
+  <template v-else>
+    <EnketoIframe :form="form!" :enketo-id="enketoId" :action-type="props.actionType ?? 'new'"/>
+  </template>
```
latin-panda, line 225: It doesn't seem right to default to new here. Instead of silently prompting the user to create a new submission, it would be better to display an error message and an option to close the page. For example, if editing isn't supported by Enketo, it could show: `Editing is only supported for Web Forms. Please contact the administrator.`

apps/forms/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "lint": "eslint . --report-unused-disable-directives --cache --max-warnings 0",
+    "lint:fix": "eslint . --fix --report-unused-disable-directives --cache --max-warnings 0",
+    "test": "BROWSER_NAME=chromium vitest run"
+  },
+  "dependencies": {
+    "vue": "^3.5.29",
+    "vue-i18n": "^10.0.7",
+    "vue-router": "^4.6.3"
+  },
+  "devDependencies": {
```
latin-panda, line 17: Is there a way we can declare primevue and the local packages (web-forms, xforms-engine) as direct dependencies here? It helps a lot with readability so we know exactly what the app consumes.

apps/forms/src/utils/api.ts
```diff
@@ -0,0 +1,97 @@
...
+  return qs !== '' ? `?${qs}` : qs;
+};
+
+export const getFormXml = async (projectId: number, formId: string, draft: boolean, st?: string | null) => {
+  const draftPath = draft ? '/draft' : '';
+  const qs = queryString({ st });
+  const url = `/v1/projects/${projectId}/forms/${formId}${draftPath}.xml${qs}`;
+  const response = await fetch(url);
+  if (!response.ok) {
+    const result = await response.json();
```
latin-panda: If `response.json()` breaks (response.ok is falsy, maybe corrupted), is there any other catch to handle it nicely?

latin-panda (changes requested):
I haven't finished yet, but I'm submitting the first part of today's review

apps/forms/src/components/enketo-iframe.vue
latin-panda:
```suggestion
 if (returnUrlPascalCase && typeof returnUrlPascalCase === 'string') {
   return returnUrlPascalCase;
 }
 if (returnUrl && typeof returnUrl === 'string') {
   return returnUrl;
 }
 return null;
```

apps/forms/src/components/submission.vue
```diff
@@ -0,0 +1,237 @@
+<script setup lang="ts">
+import { ref, defineAsyncComponent } from 'vue';
+import { useRoute, useRouter } from 'vue-router';
+import { type Form, getFormByEnketoId, getFormByFormId, getFormXml, getProject, type Project, queryString, RequestError } from '../utils/api.ts';
```
latin-panda: Is lint or prettify running? This is very long. It should have been caught and fixed automatically

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -1,368 +1,132 @@
...
-const { i18n } = inject('container');
-
-const { initiallyLoading, dataExists } = props.actionType === 'edit' ? resourceStates([formVersionXml, submissionAttachments]) : resourceStates([formVersionXml]);
-
-watch(() => initiallyLoading.value, (value) => {
-  if (!value) emit('loaded');
-});
+if (inst) {
+  inst.appContext.app.use(PrimeVue, { theme: { preset: odkThemePreset, options: { darkModeSelector: false } } });
+  // TODO this should work instead of the above, but haven't got it working yet...
```
latin-panda: Is this planned for this round or for a follow-up PR?

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -400,14 +192,31 @@ const submitData = async () => {
...
+
+  clearForm = () => {
+    clearFormCallback({ next: POST_SUBMIT__NEW_INSTANCE });
+  };
+};
+
+const handleSubmit = async (
+  payload: MonolithicInstancePayload,
+	clearFormCallback: Function
+) => {
```
latin-panda, line 227: The spacing looks off to me 
```suggestion
const handleSubmit = async (
  payload: MonolithicInstancePayload,
  clearFormCallback: Function
) => {
```

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -416,42 +225,160 @@ const handleSubmit = async (payload, callback) => {
...
+
+const closeWindow = () => {
+  window.close();
+};
+
+onMounted(() => {
+  if (isEdit.value) {
+    fetchSubmissionAttachments().then(() => {
+      loading.value = false;
+    })
```
latin-panda: I think lint catches these too
```suggestion
    });
```

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -737,6 +687,10 @@ onUnmounted(() => {
     }
   },
   "zh-Hant": {
+    "action": {
+      "close": "關閉",
+      "tryAgain": "再試一次",
+    },
```
latin-panda, line 759: This file is so big 😅, we can split it later

apps/forms/src/utils/api.ts
```diff
@@ -0,0 +1,97 @@
...
+    super(message);
+    this.statusCode = statusCode;
+    Object.setPrototypeOf(this, new.target.prototype);
+  }
+}
+
+export const queryString = (query:any) => {
+  if (query == null) return '';
+  const entries = Object.entries(query);
+  if (entries.length === 0) return '';
```
latin-panda:
```suggestion
  if (query == null) {
    return '';
  }

  const entries = Object.entries(query);
  if (entries.length === 0) {
    return '';
   }
   
```

apps/forms/src/utils/api.ts
```diff
@@ -0,0 +1,97 @@
...
+    draft: !result.publishedAt,
+    enketoOnceId: result.enketoOnceId,
+    webformsEnabled: !!result.webformsEnabled
+  };
+};
+
+export const getFormByEnketoId = async (enketoId: string, st?: string | null): Promise<Form> => {
+  const qs = queryString({ st });
+  const url = `/v1/form-links/${enketoId}/form${qs}`;
+   return getForm(url);
```
latin-panda: Spacing looks off here

apps/forms/src/i18n.js
```diff
@@ -0,0 +1,201 @@
...
+    } catch (e) {}
+  }
+};
+
+// Returns the user's preferred locale based on their previous selection and
+// their browser settings. Returns `null` if there is no locale that matches
+// their preferences.
+export const userLocale = () => {
+  const storageLocale = localStore.getItem('locale');
+  if (storageLocale != null && locales.has(storageLocale)) return storageLocale;
```
latin-panda:
```suggestion
  if (storageLocale != null && locales.has(storageLocale)) {
    return storageLocale;
  }
```

latin-panda: Same as all the ones below, not too many, improves readability

apps/forms/src/i18n.js
```diff
@@ -0,0 +1,201 @@
...
+// NUMBER FORMATS
+
+const numberFormats = {
+  default: {},
+  noGrouping: { useGrouping: false },
+  percent: { style: 'percent' }
+};
+
+for (let i = 1; i < 15; i += 1)
+  numberFormats[`maximumFractionDigits${i}`] = { maximumFractionDigits: i };
```
latin-panda:
```suggestion
for (let i = 1; i < 15; i += 1) {
  numberFormats[`maximumFractionDigits${i}`] = { maximumFractionDigits: i };
}
```

apps/forms/src/i18n.js
```diff
@@ -0,0 +1,201 @@
...
+
+
+
+////////////////////////////////////////////////////////////////////////////////
+// EXPORT
+
+export const i18n = createI18n({
+  locale: fallbackLocale,
+  fallbackLocale,
+  messages: { },
```
latin-panda, line 212: Is it correct that `messages` is empty?

apps/forms/src/main.ts
```diff
@@ -0,0 +1,14 @@
...
+import { createApp } from 'vue';
+
+import { webFormsPlugin } from '@getodk/web-forms';
+import Forms from './Forms.vue';
+
+import router from './router';
+import { i18n } from './i18n'
+
+const app = createApp(Forms as Component);
+app.use(webFormsPlugin);
```
latin-panda, line 14: Why defining Primevue in `web-form-renderer.vue L35` instead of here `main.ts`? 🤔

apps/forms/index.html
latin-panda: Did you decide to leave it here and remove it from the other files? Or perhaps it has not been pushed yet? 😅

apps/forms/package.json
latin-panda: Maybe not... 🤔 
Leaving the comment open just in case there's an idea

bin/check-bundle-size.js
```diff
@@ -54,10 +54,13 @@ log('File sizes look OK.');
...
   // Special cases:
   const simpleName = basename(path).replace(/(-[\w-]{8})+\./, '.');
+  // TODO double check these - some have been removed or replaced
   switch (simpleName) {
     case 'icomoon.svg':    return size >    60_000;
     case 'index.js':       return size > 2_300_000;
-    case 'web-form.js':    return size > 2_000_000;
+    case 'main.js':        return size >   400_000;
+    case 'forms.js':       return size > 2_000_000;
+    case 'web-form.js':    return size > 2_000_000; // Matches web-form-renderer.js
```
latin-panda: Is this L63 the web forms plugin or the web-forms renderer of `app/forms`? If the latter, it can be lowered. If the former, we need to update the comment.

When running the script:
//:build:web: dist/assets/web-form-renderer-CycNBHBr.js             985.19 kB │ gzip: 291.12 kB
//:build:web: dist/forms/forms-fTznoukg.js                        1,848.76 kB │ gzip: 779.64 kB

latin-panda (changes requested):
This is the second part of today's review.
Mainly small safe feedback that improves readability

apps/forms/index.html
garethbowen: Ok! Removed that one too.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Oh wow... looking at this now I don't think we need this function at all. It was used to transform an axios response into a native response, but because it's all native that's not necessary.

Removed!

apps/forms/src/components/web-form-renderer.vue
garethbowen: As far as I can tell it's only used for fetching the formXml which is handled in the api. After that the st isn't needed.

Regarding centralising it, I think that's a good idea. At the moment these are only used once each so I just inlined them.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Good catch! Amazing that it worked at all... removed now.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Yes! Done.

apps/forms/src/components/web-form-renderer.vue
garethbowen: You're right! It is expected by the API. Fixed.

apps/forms/src/components/submission.vue
garethbowen: You're right. I've switched them all to computed.

apps/forms/src/components/submission.vue
garethbowen: Nice. Done.

apps/forms/src/components/submission.vue
garethbowen: Done.

apps/forms/src/components/submission.vue
garethbowen: I think defaulting to 'new' is correct in the case when no action type is provided.

However you're right that we should catch the case where they're editing in Enketo. I've caught that out and given a non-specific error above (404). A specific error is new functionality - I'm happy to raise an issue if you think it's worth that.

apps/forms/package.json
garethbowen: I've added primevue because that one we should definitely have.

I'm unsure whether we should include monorepo packages in the dependencies...

apps/forms/src/utils/api.ts
garethbowen: I don't think so. We could catch it here and wrap it up with a 500 error or something? But I think letting it go and be handled by the caller is ok too.

apps/forms/src/components/enketo-iframe.vue
garethbowen: Done.

apps/forms/src/components/submission.vue
garethbowen: At the moment it's using central's config for eslint, and no prettify at all. As most of the code came from central I took the path of least resistance. I've manually broken the line here.

I'm looking forward to getting web-forms, forms, and central on the same setting.

apps/forms/src/components/web-form-renderer.vue
garethbowen: At the moment it's working so I'm ok leaving it in. If you have any ideas please let me know!

apps/forms/src/components/web-form-renderer.vue
garethbowen: Done.

apps/forms/src/utils/api.ts
garethbowen: Fixed. Many of these are just copied right out of central...

apps/forms/src/utils/api.ts
garethbowen: Done.

apps/forms/src/i18n.js
garethbowen: Fixed.

apps/forms/src/i18n.js
garethbowen: All the messages required are inline in the vue files. This will change if/when we modernise the i18n.

apps/forms/src/main.ts
garethbowen: Because primevue isn't needed for enketo rendering so I wanted to leave it out if possible.

bin/check-bundle-size.js
garethbowen: Good catch. I reduced the size to 1MB.

garethbowen: Funny side note: The reason it matches `web-form-renderer.js` is because the word "renderer" is 8 letters long which exactly matches the regex which is designed to strip off the hash suffix.

apps/forms/src/i18n.js
garethbowen: Done.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Linting is a mess... I'm using the old central eslint config but now because I'm using typescript it's not happy. I'm going to look at how hard it would be to conform with web-forms style...

garethbowen: Ok I've switched to using the same config as web-forms packages...

garethbowen:
@latin-panda Thank you for your continued reviews! I've responded to your comments, and this is good to go again.

apps/central/src/components/account/login.vue
```diff
@@ -186,9 +186,10 @@ export default {
       }
       if (url.origin !== window.location.origin || url.pathname === '/login')
         return internal('/');
-
       if (url.pathname.startsWith(`${enketoBasePath}/`))
         return external(url.href);
+      if (url.pathname.startsWith(`${webFormsBasePath}/`))
+        return external(url.pathname.slice(webFormsBasePath.length + 1));
```
latin-panda: The `url.pathname` doesn't include the `url.search` and `url.hash` pieces. 
Does it make sense like this?

```suggestion
         const path = url.pathname.slice(webFormsBasePath.length + 1);
        return external(path + url.search + url.hash);
```

L190 passes everything, including search and hash

apps/forms/src/components/web-form-renderer.vue
latin-panda: I think it's fine like this. I see it as lower priority since we will remove Enketo in the near future. Actually, can you remove the comment at L30 since it's not lazy-loaded to prevent confusion?

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -416,42 +221,126 @@ const handleSubmit = async (payload, callback) => {
...
+
+const closeWindow = () => {
+  window.close();
+};
+
+onMounted(() => {
+  if (isEdit.value) {
+    fetchSubmissionAttachments().then(() => {
+      loading.value = false;
+    });
```
latin-panda: In case of an error, it doesn't stay loading state indefinitely
```suggestion
    fetchSubmissionAttachments().finally(() => loading.value = false )
```

.github/workflows/tests.yml
```diff
@@ -87,8 +89,9 @@ jobs:
         path: client
         fetch-depth: 0
     - name: Clone getodk/central repo
+    # TODO revert the URL below once nginx changes are pushed to central/next
       run: |
-        git clone -b next https://github.com/getodk/central.git
+        git clone -b separate-forms-app https://github.com/garethbowen/central.git
```
latin-panda: Remember to revert this once the other PR is merged

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -1,368 +1,131 @@
...
 
   // Error handler - Primary Instance
   if (!submissionResult.primaryInstanceResult.success) {
     const error = submissionResult.primaryInstanceResult.data;
     if (error.response && isProblem(error.response.data) && error.response.data.code === 401.2) {
-      showModal({ type: 'sessionTimeoutModal' });
+      visibleModal.value = { type: 'sessionTimeoutModal', hideable: true };
     } else {
-      showModal({ type: 'errorModal', errorMessage: requestAlertMessage(i18n, error) });
+      visibleModal.value = { type: 'errorModal', errorMessage: 'errormsg', hideable: true }; // TODO requestAlertMessage(i18n, error) });
```
latin-panda: Is this TODO about translating 'errormsg'?

apps/forms/index.html
```diff
@@ -0,0 +1,49 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>ODK Web Forms</title>
```
latin-panda, line 6: Central dynamically was setting the titles when opening the form

Image

In this branch

Image

latin-panda (changes requested):
We're close to wrapping up! I did another deep dive and left some feedback below.

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -416,42 +225,160 @@ const handleSubmit = async (payload, callback) => {
...
+  <template v-else="!loading">
+    <OdkWebForm
+      :form-xml="props.xform"
+      :edit-instance="editInstanceOptions"
+      :fetch-form-attachment="getAttachment"
+      :track-device="true"
+      @submit="handleSubmit"/>
+  </template>
+
+  <Dialog :visible="visibleModal" :draggable="false" :closable="visibleModal?.hideable" @update:visible="visibleModal = null">
```
sadiqkhoja: This needs to have an overlay to block the interaction with the Form

apps/central/src/locales/en.json5
```diff
@@ -218,7 +218,6 @@
     "saveSettings": "Save settings",
     // This text is shown next to a dropdown menu or other way to change the sorting of Projects or Forms
     "sort": "Sort",
-    // @transifexKey component.WebFormRenderer.retryModal.action.tryAgain
```
sadiqkhoja, line 221: This line needs to stay here so that `tryAgain`'s key remain same and translators don't need to translate it again.

@matthew-white am I understanding it correctly?

apps/central/src/router.js
```diff
@@ -85,7 +85,13 @@ router.afterEach(unlessFailure(to => {
...
   // Note: to.path is '/#/some/path' when the route has query parameter after '#'
   // e.g. /#/some/path?next=/ - central#939
-  router.beforeEach(to => (to.fullPath.startsWith('/#/') ? to.fullPath.substring(2) : true));
+  router.beforeEach(to => {
+    if (to.fullPath.startsWith('/#/')) {
+      // This must do a full page refresh because the page may end up in a different vue app
+      window.location.replace(to.fullPath.slice(2));
+    }
+    return true;
+  });
```
sadiqkhoja, line 94: I don't think anyone would have bookmarked `/#/f/....` or `/#/project/:id/forms/:formId/submissions/new` (+similar) because those routes were introduced after the web-history based navigation change. Therefore this change is not necessary but not really harmful.

Seeking @matthew-white 's opinion here as well.

.changeset/lemon-deserts-report.md
```diff
@@ -0,0 +1,6 @@
+---
+"@getodk/central-frontend": minor
+"@getodk/forms": minor
+---
+
+Separated web-forms and enketo into a new app
```
sadiqkhoja, line 6: We probably don't need to include this release notes.

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -1,368 +1,131 @@
...
+    
     if (isPublicLink.value) {
-      showModal({ type: 'thankYouModal', hideable: false });
+      visibleModal.value = { type: 'thankYouModal', hideable: false };
     } else if (isEdit.value) {
-      showModal({ type: 'editSubmissionModal', hideable: false });
+      visibleModal.value = { type: 'editSubmissionModal', hideable: false };
       setTimeout(() => {
-        router.push(submissionPath(form.projectId, form.xmlFormId, props.instanceId));
+        router.push(submissionPath());
```
sadiqkhoja: this might not work because `submissionPath` points to central-frontend app

apps/forms/src/components/submission.vue
```diff
@@ -0,0 +1,266 @@
...
+      window.location.replace('/');
+    } else {
+      form.value = formConfig;
+      loadingState.value = false;
+    }
+  } catch (e) {
+    if (e instanceof RequestError && e.statusCode >= 401 && e.statusCode < 404) {
+      // not logged in
+      const relativeUrl = window.location.href.substring(window.location.origin.length);
+      window.location.href = '/login?next=/wf' + relativeUrl;
```
sadiqkhoja, line 192: Need to test (manually or otherwise) that query paramter passed to enketo are lost when user is redirected to the login page.

apps/forms/src/utils/api.ts
```diff
@@ -0,0 +1,118 @@
+export interface Form {
+  xmlFormId: string;
+  projectId: number;
+  enketoId: string;
+  state: string;
+  enketoOnceId?: string;
+  draft: boolean;
+  webformsEnabled: boolean;
+};
```
sadiqkhoja: nit: 
```suggestion
}
```

apps/forms/index.html
```diff
@@ -0,0 +1,49 @@
...
+      .spinner {
+        width: 60px;
+        height: 60px;
+        border: 8px solid #f1f5f9;
+        border-bottom-color: #3e9fcc;
+        border-radius: 50%;
+        display: inline-block;
+        box-sizing: border-box;
+        animation: rotation 1s linear infinite;
+      }
```
sadiqkhoja, line 26: I guess the duplicate css styles like `.loading-container` and `.spinner` will be removed or a spinner component will be created in the followup PR

apps/forms/src/components/preview.vue
```diff
@@ -0,0 +1,127 @@
...
+  draft: Boolean,
+});
+
+const WebFormRenderer = defineAsyncComponent(() => import('./web-form-renderer.vue'));
+const EnketoIframe = defineAsyncComponent(() => import('./enketo-iframe.vue'));
+
+const route = useRoute();
+
+const projectId = computed(() => Number.parseInt(encodeURIComponent(route.params.projectId as string)));
+const formId = computed(() => encodeURIComponent(route.params.xmlFormId as string));
```
sadiqkhoja, line 17: I think parameters coming from `route` are already encoded, need to double check.

apps/forms/src/components/enketo-iframe.vue
```diff
@@ -147,10 +136,10 @@ const handleIframeMessage = (event) => {
         try {
           const normalizedUrl = new URL(redirectUrl.value);
           if (['http:', 'https:'].includes(normalizedUrl.protocol)) {
-            if (normalizedUrl.origin === location.origin) {
+            if (normalizedUrl.origin === window.location.origin) {
               router.push(normalizedUrl.pathname);
```
sadiqkhoja: For editing a submission in Enketo we redirect user back to the submission detail page, which is in central-frontend so this will not work.

apps/forms/src/router.ts
```diff
@@ -0,0 +1,95 @@
...
+    },
+  },
+  {
+    path: '/projects/:projectId([1-9]\\d*)/forms/:xmlFormId/draft/submissions/new/:offline(offline)?',
+    component: Submission as Component,
+    name: 'DraftSubmissionNew',
+    props: (route) => {
+      const { offline, ...params } = route.params;
+      return {
+        ...params,
```
sadiqkhoja: `Submission` component reads projectId and xmlFormId from the route directly so we can remove it from here.

Same for preview route.

vite.config.js
```diff
@@ -66,7 +89,32 @@ export default defineConfig(({ mode }) => ({
...
+          if (moduleId && moduleId.includes('/apps/forms/')) {
+            return 'forms/[name]-[hash].js';
+          }
+          return 'central/[name]-[hash].js';
+        }
+      },
+    },
+  },
+  test: {
+    plugins: [vue(), devAppRouter()],
```
sadiqkhoja: wouldn't plugins defined at the root be used by the test?

sadiqkhoja (commented):
Overall it's looking good, here are few comments:

I am planning to deploy this branch to dev environment to do some testing, the routing part can be tricky to judge just be looking at the code, there are so many execution branches. I'll be following last section of [this](https://docs.google.com/document/d/19DYSIYXUY8ouizn1MAKz0j_M_17IpVFQXXwIwHzBco4/edit?tab=t.0) to do verification.

apps/central/src/components/account/login.vue
garethbowen: Yes true! Thanks. Fixed.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Done.

apps/forms/src/components/web-form-renderer.vue
garethbowen: I had a look at your suggestion, but felt it was important to report the error. But then trying to show the error meant adding translations and error handling etc etc.

In the end I moved the `fetchSubmissionAttachments` into submission.vue and load it as part of the loading of the form (if editing). This means it's handled by the error handling the same way if the form isn't found. It also means we can request them in parallel which speeds the page load up a little. And finally it means we don't need the `loading` state in web-form-renderer at all.

apps/forms/src/components/web-form-renderer.vue
garethbowen: It used to call a utility function in central which was based on an axios response which no longer worked. I initially left it because we have a separate issue for improving error messages, but after you called it out I changed my mind and decided this PR should at least not make things worse. I've ported the util function and the translations from central and now the functionality should be identical.

apps/forms/index.html
garethbowen: Snazzy! Done.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Done.

apps/central/src/locales/en.json5
garethbowen: Because the "WebFormRenderer" component doesn't exist any more this causes an error, so I removed it.

In a later commit I'll have to resolve the translations because right now they're not updated from transifex at all - I've just copied the current values. We haven't decided yet if we want to follow central's translation setup or the one used by web-forms packages, so I'm deferring this to a later date.

apps/central/src/router.js
garethbowen: I was tempted to remove it as well, but @lognaturel pushed back a little. It's currently tested in e2e tests.

.changeset/lemon-deserts-report.md
garethbowen: I'm going to do some performance testing once it's on a prod like machine. I _think_ page load speed be noticeably faster, and if so, would be good to include in the release notes (IMO).

apps/forms/src/components/web-form-renderer.vue
garethbowen: Good spotting! Updated.

apps/forms/src/components/submission.vue
garethbowen: I've just tested it and it works for me.

apps/forms/src/utils/api.ts
garethbowen: Fixed.

apps/forms/index.html
garethbowen: The spinner is a special case because in the index.html we want it to be styled before the css assets have been downloaded. However because I've inlined it into the head of index.html it's not needed in the vue files as well.

So I've removed the duplicated css and verified it still works, and then also pulled out a Spinner component with an HTML comment documenting this approach.

apps/forms/src/components/preview.vue
garethbowen: Googled and tested - they're _decoded_ automatically, so `a%20b` -> `a b`. But because we use them in outgoing requests I encode them here to be safe.

Arguably it might make more sense to have the api handle all routes and all encoding rather than doing it up front...

apps/forms/src/components/enketo-iframe.vue
garethbowen: True! Thanks. It actually simplifies the logic because now it does the `assign` either way.

apps/forms/src/router.ts
garethbowen: Thanks.

vite.config.js
garethbowen: Right! I think this was committed while trying to get tests to work but it made no difference. Removed.

apps/central/src/router.js
lognaturel: I might have gotten confused! They’re covered by new tests or previously-existing tests?

Am I understanding correctly that # was dropped before /-/ became /f/ and the RESTful routes were supported? That sounds right to me. Are these combinations newly-supported or continuing what was supported before?

garethbowen: These are previously supported URLs. Here is the e2e test in `master` which ensures they work: https://github.com/getodk/central-frontend/blob/a16775dd5494d3f6d9548344194f24c867e1c755/e2e-tests/tests/enketo.spec.js#L51-L53

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -416,42 +230,111 @@ const handleSubmit = async (payload, callback) => {
...
+			<span role="heading">{{ $t(visibleModal.type + '.title') }}</span>
+		</template>
+
+    <template #default>
+      <Translation v-if="visibleModal.type === 'errorModal'" tag="p" keypath="errorModal.body">
+        <template #errorMessage>
+          <pre>{{ $t(visibleModal.errorMessage) }}</pre>
+        </template>
+        <template #supportEmail>
+          <a href="emailto:support@getodk.org">support@getodk.org</a>
```
latin-panda:
```suggestion
          <a href="mailto:support@getodk.org">support@getodk.org</a>
```

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -416,42 +230,111 @@ const handleSubmit = async (payload, callback) => {
...
+        </template>
+      </Translation>
+      <Translation v-else-if="visibleModal.type === 'retryModal'" tag="p" keypath="retryModal.body">
+        <template #supportEmail>
+          <a href="emailto:support@getodk.org">support@getodk.org</a>
```
latin-panda: Same:
```suggestion
          <a href="mailto:support@getodk.org">support@getodk.org</a>
```

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -1,413 +1,227 @@
...
-    submissionResult.attachmentResult.set(attachment.name, {
-      success: false
-    });
-  });
+const isSessionTimeout = (error) => error.response && isProblem(error.response.data) && error.response.data.code === 401.2;
 
-  clearForm = () => {
-    clearFormCallback({ next: POST_SUBMIT__NEW_INSTANCE });
-  };
+const getErrorMessage = (data) => {
```
latin-panda, line 99: Could it be undefined or null?
```suggestion
const getErrorMessage = (data) => {
  if (!data) {
    return;
  }
```

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -1,413 +1,227 @@
...
+  return `/projects/${props.form.projectId}/forms/${props.form.xmlFormId}/submissions/${props.instanceId}`;
+};
 
-  submissionResult.attachmentResult = new Map();
-  data.attachments.forEach(attachment => {
-    submissionResult.attachmentResult.set(attachment.name, {
-      success: false
-    });
-  });
+const isSessionTimeout = (error) => error.response && isProblem(error.response.data) && error.response.data.code === 401.2;
```
latin-panda: Could anything here be undefined or null?

apps/central/src/router.js
lognaturel: The RESTful routes could very well go because they're not likely good candidates for bookmarking -- folks with auth would navigate to them. But the ones with enketoFormIDs are much more likely to be bookmarked since they can be public access links. If there are existing tests, probably best to keep.

latin-panda (changes requested):
Just small feedback below. I didn't spot anything weird while testing, I'll have another test round tomorrow with fresh eyes

apps/forms/src/components/web-form-renderer.vue
garethbowen: Oh wow! Eagle eyes.

Hilariously this is an existing bug: https://github.com/getodk/central-frontend/blob/73042d7a1638e608013d8b565a8a2d8e69ae7c9b/apps/central/src/components/web-form-renderer.vue#L33

I wonder if we've been missing out on support emails all this time...

Fixed now.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Done.

apps/forms/src/components/web-form-renderer.vue
garethbowen: I don't think so, but I've added this block to be defensive.

apps/forms/src/components/web-form-renderer.vue
garethbowen: I've added a defensive check which I think covers everything.

sadiqkhoja:
Found following issue on dev.getodk.cloud while testing this PR with corresponding `central` PR

-  Can't submit the Form using public link:
  Public links use `st` query parameter as their auth but on the submit button I don't see token being passed to the backend.
  Happens only when ODK Web Forms is select, works correctly for Enketo.

Tested other edge cases, that all worked 🎉

- Offline submissions in Enketo
- Once-only Submissions in Enketo
- Submission editing
- Redirection to a third party thank-you page
- Form preview for draft and published Form
- Prefilled values in Enketo

apps/central/src/locales/en.json5
sadiqkhoja: I don't think it is necessary for the "WebFormRenderer" component to exist for the `transifexKey`. I added this line back locally and ran `npm run transifex:fix` and this the git diff:

```diff
diff --git a/apps/central/src/locales/en.json5 b/apps/central/src/locales/en.json5
index 3dcff9c59..48523858e 100644
--- a/apps/central/src/locales/en.json5
+++ b/apps/central/src/locales/en.json5
@@ -218,6 +218,7 @@
     "saveSettings": "Save settings",
     // This text is shown next to a dropdown menu or other way to change the sorting of Projects or Forms
     "sort": "Sort",
+    // @transifexKey component.WebFormRenderer.retryModal.action.tryAgain
     "tryAgain": "Try again",
     "undo": "Undo",
     "update": "Update",
diff --git a/apps/central/transifex/strings_en.json b/apps/central/transifex/strings_en.json
index 20e428398..6b72f8b14 100644
--- a/apps/central/transifex/strings_en.json
+++ b/apps/central/transifex/strings_en.json
@@ -451,10 +451,6 @@
       "string": "Sort",
       "developer_comment": "This text is shown next to a dropdown menu or other way to change the sorting of Projects or Forms"
     },
-    "tryAgain": {
-      "string": "Try again",
-      "developer_comment": "This is the text for an action, for example, the text of a button."
-    },
     "undo": {
       "string": "Undo",
       "developer_comment": "This is the text for an action, for example, the text of a button."
@@ -5539,6 +5535,16 @@
         "string": "Invitation Pending"
       }
     },
+    "WebFormRenderer": {
+      "retryModal": {
+        "action": {
+          "tryAgain": {
+            "string": "Try again",
+            "developer_comment": "This is the text for an action, for example, the text of a button."
+          }
+        }
+      }
+    },
     "WhatsNew": {
       "title": {
         "string": "🎨 Branded login pages, Entity List cleanups, and file uploads in Web Forms"
```

garethbowen: That's not desirable because the `action.tryAgain` key is used in other places in central like `entity/list.vue` so I wanted to leave the key unchanged.

garethbowen:
@sadiqkhoja 

> Can't submit the Form using public link:

Ouch... yes. I think I must've had a session cookie when testing this locally so it gave a false positive.

I've fixed this now and added a unit test to ensure the st is being passed to the server.

latin-panda (approved):
I did another pass through the code and tested the app with fresh eyes, and it looks good to me! 
Remember to revert this [URL](https://github.com/getodk/central-frontend/pull/1608/changes#r3389962509) before merging :)

PR #1609, opened 2026-06-08
Feature getodk/central#1790: Show custom properties on app-users table

sadiqkhoja:
Towards getodk/central#1790

Added tests and manual verification.

Follows the existing pattern.

None.

Will be added along with other changes in https://github.com/getodk/central/issues/1866


sadiqkhoja:
I should have created two separate PRs; first for just adding custom props and second for using TableFreeze component to facilitate the code review :(

PS: I'll submit separate PR for the changes in public-link list.

apps/central/src/components/table/freeze.vue
```diff
@@ -172,7 +172,7 @@ defineExpose({ getRowPair });
   .btn-group {
     // Setting the background color in case an action is transparent.
     background-color: $color-page-background;
-    left: -1000px;
+    left: -2000px;
```
sadiqkhoja, line 175: 1000px was not enough to hide the button.

apps/central/src/components/table/freeze.vue
```diff
@@ -10,7 +10,7 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <div class="table-freeze">
+  <div class="table-freeze" :class="{ 'frozen-only': frozenOnly }">
```
sadiqkhoja, line 13: added this class so that parent could set the width of the frozen table to 100% when there is no custom properties.

apps/central/src/components/field-key/list.vue
```diff
@@ -132,6 +147,7 @@ export default {
   },
   methods: {
     fetchData(resend) {
+      this.$emit('fetch-actor-properties', false);
```
sadiqkhoja: move this to the `created` method

apps/central/src/components/field-key/row.vue
sadiqkhoja, line 1: maybe rename to metadata-row

apps/central/src/components/custom-props-data-row.vue
sadiqkhoja, line 1: Move this component out of `field-key` directory so that it can be reused by public-link-list component as well.

apps/central/src/components/field-key/row.vue
```diff
@@ -14,10 +14,6 @@ except according to the terms contained in the LICENSE file.
     <td class="display-name">
       <span v-tooltip.text>{{ fieldKey.displayName }}</span>
     </td>
-    <td>
-      <time-and-user :iso="fieldKey.createdAt" :user="fieldKey.createdBy"/>
```
sadiqkhoja, line 18: check if this is now a dead component.

ktuite (approved):

apps/central/src/components/field-key/row.vue
sadiqkhoja: nope

apps/central/src/components/field-key/row.vue
sadiqkhoja: decided not to rename this now, will create an issue to be consider later.

PR #1610, opened 2026-06-09
feat(#803): add Sentry error reporting and performance monitoring

### LINKED ISSUE #803: Implement error reporting
### Description

Some points to consider...

latin-panda:

I've tested in Sentry:

<details> <summary> Automatically tracking exceptions in browser's console log </summary>

</details>

<details> <summary> Manually tracking error using `captureException` from '@sentry/vue'</summary>

```
import { captureException } from '@sentry/vue'; // anywhere in the app
captureException(new Error('sentry test: Manually tracking an error that is not in the console log'));
```

</details>

<details> <summary> Tracking app/forms </summary>

First error is with web-forms tag
Second is central

</details>

<details> <summary> Performance tracking </summary>

</details>

<details> <summary> I've activated the debug mode to see what it would print </summary>

</details>

Initializes Sentry in `apps/central` via a new `util/sentry.js` utility. Captures Vue component errors (`attachErrorHandler`) and page load timing (`browserTracingIntegration`). Silent no-op when `ODK_CENTRAL_FRONTEND_SENTRY_DSN` is absent, so self-hosters are unaffected.

The DSN is a build-time env var (`ODK_CENTRAL_FRONTEND_SENTRY_DSN`), exposed via a custom `envPrefix` added to `vite.config.js`. Set it in `.env.local` for local testing (see README). I don't see that we are doing things like this, but worth considering for simplicity. 

I've set a low `tracesSampleRate` to prevent any performance or network issues when users are using the app. See code comment for more details. 

No necessarily, it has a note in the README file already. 


apps/central/src/main.js
```diff
@@ -19,12 +19,12 @@ import App from './components/app.vue';
...
 // because the plugins require jQuery.
 import './jquery';
 import './bootstrap';
 
-createApp(App)
-  .use(createContainer())
-  .directive('tooltip', vTooltip)
-  .mount('#app');
+const app = createApp(App);
+initSentry(app, 'central-frontend');
```
latin-panda, line 29: 1. Once the `app/forms` from the "app split work" is merged, we can just do the same setup but with the web forms tag: `initSentry(app, 'web-forms');`
2. Ideally `Sentry.init()` would run before `createApp()` for earliest possible error capture. However, `attachErrorHandler` from `apps/central/src/util/sentry.js` requires the app instance, so both calls must come after `createApp()`. In practice, the gap is negligible; no user code runs between the two lines.

.env.example
```diff
@@ -0,0 +1,2 @@
+# Sentry DSN for error reporting. Leave empty to disable monitoring.
+ODK_CENTRAL_FRONTEND_SENTRY_DSN=
```
latin-panda: This is new in the repo. A simple alternative to `/client-config.json`, definitely open to change. 

After this merge, the `central` repo requires a small follow-up PR to pass `ODK_CENTRAL_FRONTEND_SENTRY_DSN` as a Docker build argument in both `nginx.dockerfile` and `docker-compose.yml`. This ensures Vite embeds the DSN into the production bundle. Without it, Sentry remains disabled in production regardless of the environment settings.

latin-panda:
@matthew-white, this is my first attempt and still a draft :) What do you think?

How do we normally test these things? Do we have a test project in our Sentry server for experiments?

apps/central/src/util/sentry.js
```diff
@@ -0,0 +1,22 @@
...
+const initSentry = (app, source) => {
+  const dsn = import.meta.env.ODK_CENTRAL_FRONTEND_SENTRY_DSN;
+  if (!dsn) {
+    return;
+  }
+
+  init({
+    app,
+    dsn,
+    integrations: [browserTracingIntegration()],
```
latin-panda, line 12: This line automatically tracks uncaught exceptions.

For errors that we want to report and that are handled in the code. We can manually track them by using:

```ts
import { captureException } from '@sentry/vue'; 
....
captureException(new Error('sentry test: Manually tracking an error that is not in the console log'));
```

apps/central/.eslintrc.js
```diff
@@ -21,7 +21,10 @@ module.exports = {
     curly: 'off',
     'implicit-arrow-linebreak': 'off',
     'import/first': 'off',
-    "import/no-extraneous-dependencies": ['error', {'devDependencies': ['**/test/**', 'vite.config.js']}],
+    "import/no-extraneous-dependencies": ['error', {
+      'devDependencies': ['**/test/**', 'vite.config.js'],
+      'packageDir': ['.', '../..'],
```
latin-panda, line 26: Since @sentry/vue was installed in the root package.json, eslint didn't find it. This makes it consider the root package.json too

apps/forms/src/components/submission.vue
```diff
@@ -191,7 +192,7 @@ const load = async () => {
       const relativeUrl = window.location.href.substring(window.location.origin.length);
       window.location.href = '/login?next=/wf' + relativeUrl;
     } else {
-      // unknown error
+      captureException(e);
```
latin-panda, line 195: Manual tracking of exceptions to Sentry since these don't make it to the browser console.

apps/forms/src/utils/sentry.ts
```diff
@@ -0,0 +1,21 @@
+import type { App } from 'vue';
+import { browserTracingIntegration, init, setTag } from '@sentry/vue';
+
+const initSentry = (app: App, source: string) => {
+  const dsn = import.meta.env.ODK_CENTRAL_FRONTEND_SENTRY_DSN as string | undefined;
```
latin-panda, line 5: Almost the same as app/central, except for the typing required by the type check.

This small script could be extracted to a common place, like /packages/central-common or similar. At the moment, it is so small that it is left in each app/utils for now.

latin-panda:
@garethbowen When you have time, could you please review?

garethbowen (approved):
LGTM!

Consider including instructions for deploying to production in the central repo somewhere, and tag Yaw to roll this out for cloud.

latin-panda:
@garethbowen To update the central Docker files to include the new variable and docs, do you know whether I should branch from `next` or `master`? I think `next`, right?

lognaturel:
Amazing to see this merged @latin-panda 🎉 🎉! Yes, `next` is the way to go.

PR #1614, opened 2026-06-11
chore(web-forms): enable Transifex push/pull for translations

latin-panda:

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -8,3 +8,10 @@ source_lang  = en
 type         = STRUCTURED_JSON
 minimum_perc = 0
 
+[o:getodk:p:web_forms:r:strings]
+file_filter  = packages/web-forms/locales/strings_<lang>.json
+source_file  = packages/web-forms/locales/strings_en.json
+source_lang  = en
+type         = STRUCTURED_JSON
+minimum_perc = 0
+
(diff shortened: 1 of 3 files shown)


.tx/config
```diff
@@ -8,3 +8,10 @@ source_lang  = en
 type         = STRUCTURED_JSON
 minimum_perc = 0
 
+[o:getodk:p:web_forms:r:strings]
```
latin-panda, line 11: no conflict because it has scope per project

packages/web-forms/package.json
```diff
@@ -32,6 +32,8 @@
     "build:demo": "vite build --mode demo --outDir dist-demo",
     "build:js": "vite build",
     "build:translations": "node bin/merge-translations.js",
+    "translations:pull": "cd ../.. && tx pull -r web_forms.strings -a -f --mode translator",
```
latin-panda, line 35: I didn't find a better way to this `cd ../..`  The options I see:
1. Move .tx/config to packages/web-forms/
2. Move these commands to a root package.json
3. Leave it like this

I'm more inclined toward option 3. It's nice to have the .tx/config with both projects and their settings in one place

latin-panda:
@garethbowen While we figure out the best way to move forward with translations, this approach helps us keep the web forms translation working smoothly with Transifex in the meantime.

packages/web-forms/package.json
garethbowen: I agree. Keep it here. Things will change when we do translation in the forms app, as well as any standardisation we want to do with central.

garethbowen (approved):

PR #1616, opened 2026-06-12
feat(#803): add source map in Sentry

### LINKED ISSUE #803: Implement error reporting

latin-panda:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,22 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v20[0-9][0-9].*'
+  workflow_dispatch:
+
+jobs:
+  sentry-sourcemaps:
+    name: Upload source maps to Sentry
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - uses: volta-cli/action@v5
+    - run: npm ci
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - uses: volta-cli/action@v5
+    - run: npm ci
+    - run: npm run build
+      env:
+        SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+        SENTRY_ORG: ${{ secrets.SENTRY_ORG }}
+        SENTRY_PROJECT: ${{ secrets.SENTRY_PROJECT }}
```
latin-panda, line 22: The `@sentry/vite-plugin` automatically reads these 3 environment variables. The Sentry DNS variable doesn't work with this plugin.

latin-panda:
@matthew-white This is the second part of the Sentry task. I'm assuming the Sentry project tracks only production, all customers, so the GitHub task triggers when a new release tag is created

latin-panda:
Closing: This approach doesn't work well. Uploading the sourcemap is more complicated and needs to be synced with the version on the server(s) to avoid misleading stack traces. And update the Vite config to produce good source maps.

PR #1617, opened 2026-06-14
fix: e2e tests should not check for new version message on first load

garethbowen:
Enketo e2e tests started failing after the [central enketo upgrade](https://github.com/getodk/central/pull/1965), probably due to changes in the way Enketo registers service workers.


lognaturel (approved):
Glad to see that’s improved, it was an odd quirk.

The fact that I didn’t at all notice this issue after merge reinforces we need to rethink when these tests are run!

PR #1623, opened 2026-06-15
Change passphrase to encryption password

### LINKED ISSUE #1079: Rename "passphrase" to "password"
We use the term "passphrase" in the context of submission encryption. That's distinguished from "password", which is used in the context of user passwords. Here are a couple of examples of text containing the word "passphrase":

> First, you will need to choose a passphrase. This passphrase will be required to decrypt your Submissions. For your privacy, the server will not remember this passphrase: only people with the passphrase will be able to decrypt and read your Submission data.

> In order to download this data, you will need to provide your passphrase. Your passphrase will be used only to decrypt your data for download, after which the server will forget it again.

@nicoleorlowski suggested using the term "password" in both contexts:

> I think we should stick with "password".. passphrase seems uncommon/unexpected?

lognaturel:
Closes https://github.com/getodk/central/issues/1079

Updated and ran tests. Search for "passphrase" and made sure it doesn't show up in code files. Compared with https://github.com/getodk/central/issues/1079#issuecomment-2956733110 and tried to simplify even further.

I considered also changing variable names and test names but I think they are fine as-is and that we can change the user-facing text independently.

This is to text only. Probably the biggest risks are that users might re-use their login password and that I may have forgotten some places, leading to inconsistencies.

https://github.com/getodk/docs/issues/2070

    I don't think this requires a changeset because it's a text-only change that most users won't notice.

### DIFF
--- a/apps/central/src/components/project/enable-encryption.vue
+++ b/apps/central/src/components/project/enable-encryption.vue
@@ -307,9 +307,9 @@ export default {
       },
       {
         "introduction": [
-          "First, you will need to choose a passphrase. This passphrase will be required to decrypt your Submissions. For your privacy, the server will not remember this passphrase: only people with the passphrase will be able to decrypt and read your Submission data.",
+          "First, choose an encryption password. The server does not store this password. Only people who know it can decrypt and read your Submission data.",
           {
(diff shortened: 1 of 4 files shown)


matthew-white (approved):
Thank you @lognaturel!

PR #1627, opened 2026-06-16
e2e-tests: ignore feature-flags console messages

alxndrsn:
Ref getodk/central#1979

Test run: https://github.com/getodk/central-frontend/actions/runs/27607386471/job/81623957207?pr=1627
Logs:
```
$ cat job-logs.txt | grep Alpha | wc -l
0
```

This message is deliberate, so seeing it in test output is 100% noise, 0% signal.

No effect.

No.

### DIFF
--- a/apps/central/src/composables/feature-flags.js
+++ b/apps/central/src/composables/feature-flags.js
@@ -58,6 +58,7 @@ export default function useFeatureFlags() {
   const { buildMode } = inject('container');
   onMounted(() => {
     if (buildMode !== 'test') {
+      // If this logging is removed, the corresponding filtering should be removed from e2e-tests/util.js
       // eslint-disable-next-line no-console
       console.log(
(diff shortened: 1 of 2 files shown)


garethbowen (commented):
I'm not so sure about this one!

I think because WF is all growed up we can remove the console statement and listener in the source because it's going to be easy to test a form in WF using UI elements.

alxndrsn:
Can we ignore the message until then?

garethbowen (approved):
Happy to ignore the message until the feature is removed. Please raise an issue to remove the console.log statement and link to this check so it's removed too!

alxndrsn:
> Please raise an issue to remove the console.log statement and link to this check so it's removed too!

I've added a comment where the logging is done.  It seems to me that the logging should remain in places as long as the "feature flag" functionality does (https://github.com/getodk/central-frontend/blob/b8c11041f773f141dcd2a0e08c722eb1457c2aea/apps/central/src/composables/feature-flags.js#L26-L41).

PR #1633, opened 2026-06-17
fix(web-forms#814): improve error messages for bad urls

### LINKED ISSUE #814: Improve error message for bad URL or form/link state
This is currently entirely handled by the host application and likely will continue to be. 

Proposal: replace all the error states below with a single full-page message (see current styling for https://github.com/getodk/web-forms/issues/813, note there is no close action):

> Unable to open form

> Please check that the link is correct. The form may no longer be available, or your access may have expired. If the problem continues, contact the person who sent you the form link.

### Invalid enketoid

Current behavior: snackbar at bottom of blank page says "No Form found with this URL, please double check."

### Invalid st query parameter

The `st` query parameter uniquely identifies a Public Access Link. Under the hood it is an auth token.

Current behavior: snackbar at bottom of blank page says "Could not authenticate with the provided credentials."

### Loading closed form or revoked Public Access Link

Current behavior: snackbar at bottom of blank page says "The authentication you provided does not have rights to perform that action."

### Context: submitting to closed form

For reference, here is the message shown when attempting to submit to a closed form:

Image

This seems ok to me. I would change "the person who asked you to fill this Form" to "the person who sent you the form link."

### Design

**Edit 2nd June 2026:** [Design ideas in Figma](https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=580-3686)

garethbowen:
Closes: getodk/web-forms#814

Mostly manual testing with various error states.

I used the Dialog because it's styled consistently with Web Forms.

It changes the errors that will be shown.

No

### DIFF
--- a/apps/forms/src/components/submission.vue
+++ b/apps/forms/src/components/submission.vue
@@ -66,7 +66,7 @@ const xform = ref<string>();
 const submissionAttachments = ref<string[] | null>(null);
 
 const loadingState = ref(true);
-const errorState = ref(false);
+const errorCode = ref<number | null>(null);
 
 const newSubmissionPath = (projectId:number, xmlFormId:string, draft:boolean) => {
   const suffix = draft ? 'draft/submissions/new' : 'submissions/new';
@@ -109,11 +109,10 @@ const setDocumentTitle = (formConfig: Form) => {
   document.title = title;
 };
 
-const fetchForm = async (): Promise<Form | undefined> => {
-  let formConfig;
+const fetchFormConfig = async (): Promise<Form> => {
   if (enketoId.value) {
     try {
-      formConfig = await getFormByEnketoId(enketoId.value, st.value);
+      return await getFormByEnketoId(enketoId.value, st.value);
     } catch(e) {
       if (e instanceof RequestError && (e.statusCode === 401.2 || e.statusCode === 403.1)) {
         // a public form with an invalid st or revoked enketoId should show as form not found
@@ -121,11 +120,15 @@ const fetchForm = async (): Promise<Form | undefined> => {
       }
       throw e;
     }
-  } else if (projectId.value && formId.value) {
-    formConfig = await getFormByFormId(projectId.value, formId.value, props.draft, st.value);
-  } else {
-    throw new RequestError('Form not found', 404);
   }
+  if (projectId.value && formId.value) {
+    return await getFormByFormId(projectId.value, formId.value, props.draft, st.value);
+  }
+  throw new RequestError('Form not found', 404);
+};
+
+const fetchForm = async (): Promise<Form | undefined> => {
+  const formConfig = await fetchFormConfig();
 
   if (formConfig.state === 'closed') {
     throw new RequestError('Form not found', 404);
@@ -147,6 +150,10 @@ const fetchForm = async (): Promise<Form | undefined> => {
     }
     webFormsEnabled.value = true;
   } else {
+    if (props.actionType === 'edit') {
+      // editing in Enketo isn't supported
+      throw new RequestError('Form not found', 404);
+    }
     if (offline.value) {
       // TODO: Update once Web Forms has support for offline
       window.location.replace(`/-/x/${formConfig.enketoId}${queryString(route.query)}`);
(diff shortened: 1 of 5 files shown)


apps/forms/src/components/submission.vue
```diff
@@ -225,19 +260,12 @@ load();
...
-}
-</style>
-
 <i18n lang="json5">
   {
     "en": {
-      "formNotFound": "No Form found with this URL, please double check."
+      "formNotFound": "Unable to open form",
+      "formNotFound.body": "Please check that the link is correct. The form may no longer be available, or your access may have expired. If the problem continues, contact the person who sent you the form link.",
+      "errorNotProblem": "Something went wrong: error code {status}."
```
latin-panda: It's not giving instructions to the user: 


It can be:

> Something went wrong; try again later. If the problem continues, contact the person who sent you the form link.
> (Error code: 500)

The part `(Error code: 500)` can be in hint font-size which is smaller than the base font-size


Indonesian: `Terjadi kesalahan; coba lagi nanti. Jika masalah berlanjut, hubungi orang yang mengirimkan tautan formulir tersebut kepada Anda.`

Spanish: `Algo salió mal; inténtelo de nuevo más tarde. Si el problema persiste, comuníquese con la persona que le envió el enlace del formulario.",`

French: `Une erreur est survenue. Veuillez réessayer plus tard. Si le problème persiste, veuillez contacter la personne qui vous a envoyé le lien vers le formulaire.`

apps/forms/src/components/submission.vue
```diff
@@ -107,13 +108,28 @@ const setDocumentTitle = (formConfig: Form) => {
...
+  if (projectId.value && formId.value) {
+    return await getFormByFormId(projectId.value, formId.value, props.draft, st.value);
+  }
+  throw new RequestError('Form not found', 404);
+};
+
+const fetchForm = async (): Promise<Form | undefined> => {
+  const formConfig = await fetchFormConfig();
+
+  if (formConfig.state === 'closed') {
```
latin-panda: Should a project admin who can read and update submissions be allowed to edit a submission on a closed form?

latin-panda (changes requested):
Nice! I have a couple of comments below

apps/forms/src/components/submission.vue
garethbowen: Ok... I spent some more time understanding project verbs and the whole `hasAccess` function which I ported over from the previous code verbatim. Now I realise I shouldn't check the formconfig.state here because the hasAccess function was already doing it and using the verbs to allow specific roles access. So I've removed this and cleaned up where the hasAccess is called, and now it returns a 404 too.

apps/forms/src/components/submission.vue
garethbowen: Yes much better, thanks!

Instead of using that one liner, I broke it in to 3: the header, the body, and the code. Would you mind checking the translations to make sure they still make sense?

apps/forms/src/components/submission.vue
```diff
@@ -247,25 +283,50 @@ load();
...
+    "errorNotProblem": "Etwas ging schief",
+    "errorNotProblem.status": "Fehlercode {status}",
   },
   "es": {
-    "formNotFound": "No se ha encontrado ningún formulario con esta URL, vuelva a comprobarlo."
+    "formNotFound": "No se puede abrir el formulario",
+    "formNotFound.body": "Por favor, verifique que el enlace es correcto. Es posible que el formulario ya no esté disponible o que su acceso haya expirado. Si el problema persiste, comuníquese con la persona que le envió el enlace del formulario.",
+    "errorNotProblem": "Algo salió mal",
+    "errorNotProblem.body": "Inténtelo de nuevo más tarde. Si el problema persiste, comuníquese con la persona que le envió el enlace del formulario.",
+    "errorNotProblem.status": "Error de código: {status}",
```
latin-panda:
```suggestion
    "errorNotProblem.status": "Código de error: {status}",
```

latin-panda (approved):
It looks good, and I have tested again. No issues. 
One small fix below, and it's good to ship! 🚤

PR #1635, opened 2026-06-17
fix(web-forms#538): handle errors loading forms

### LINKED ISSUE #538: Loading errors not always handled (Safari + Lockdown Mode)
### Expected Behaviour

When there is an error loading a form, this would be communicated to the user somehow.

### Observed Behaviour

On iOS Safari with Lockdown Mode enabled, form loads perpetually.

### Reproduction Steps

1. enable Lockdown Mode
2. browse to https://staging.getodk.cloud/projects/179/forms/SelectOneFromMapTestForm/preview?webforms=true

### Screenshots or Videos

<kbd></kbd>

### Environment

- Browser version: Safari
- OS: iOS 18.7.1
- Web Forms version: ?

### Additional Context

It's unclear how/if it's possible to recreate this without an iOS device at the moment.

Filed https://github.com/microsoft/playwright/issues/38094 in the hope that one day bugs like this can be caught earlier.

garethbowen:
Closes getodk/web-forms#538

Manually tested in multiple browsers with wasm disabled. I tried to find other features that would fail but I think I need an older browser.

This has to be handled outside of the main JS bundle because it happens when the bundle fails to load for whatever reason.

When errors occur the loading spinner disappears and the error is shown. It should not impact users who haven't encountered errors.

No.

### DIFF
--- a/apps/forms/eslint.config.js
+++ b/apps/forms/eslint.config.js
@@ -57,7 +57,7 @@ export default defineConfig(
 			'web-forms/dist-demo/**/*',
 			'web-forms/bin/**/*',
 			'xforms-engine/api-docs/**/*',
-			'**/vendor'
+			'**/vendor',
 		],
 	},
 
--- a/apps/forms/eslint.config.js
+++ b/apps/forms/eslint.config.js
@@ -58,6 +58,7 @@ export default defineConfig(
 			'web-forms/bin/**/*',
 			'xforms-engine/api-docs/**/*',
 			'**/vendor',
+			'src/init.js'
 		],
 	},
 
--- a/apps/forms/eslint.config.js
+++ b/apps/forms/eslint.config.js
@@ -57,8 +57,7 @@ export default defineConfig(
 			'web-forms/dist-demo/**/*',
 			'web-forms/bin/**/*',
 			'xforms-engine/api-docs/**/*',
-			'**/vendor',
-			'src/init.js'
+			'**/vendor'
 		],
 	},
 
@@ -250,6 +249,8 @@ export default defineConfig(
 			reportUnusedDisableDirectives: true,
 		},
 
+		ignores: ['src/init.js'],
+
 		plugins: {
 			'no-only-tests': noOnlyTestsPlugin, // eslint-disable-line @typescript-eslint/no-unsafe-assignment
 		},
@@ -377,6 +378,22 @@ export default defineConfig(
 		rules: {
 			'@typescript-eslint/triple-slash-reference': 'off',
 		},
+	},
+
+
+	{
+		files: ['src/init.js'],
+		rules: eslint.configs.recommended.rules,
+		languageOptions: {
+			ecmaVersion: 5,
+			sourceType: 'script',
+			globals: {
+				document: 'readonly',
+				window: 'readonly',
+				console: 'readonly',
+				globalThis: 'readonly'
+			}
+		}
 	},
 
 	{
--- a/apps/forms/index.html
+++ b/apps/forms/index.html
@@ -43,22 +43,26 @@
       }
     </style>
     <script>
-      const handle = (message) => {
+      // NB use legacy JavaScript to support older browsers
+      var __handleInitializationErrors = function(event) {
         if (!document.getElementById('loading')) {
           // do not intercept errors after the app has loaded
           return;
         }
-        console.error('FATAL ERROR: loading forms app');
+        var message = event.message
+          || event.reason && event.reason.message
+          || 'An unknown error occurred.';
+        console.error('FATAL ERROR while loading forms app:', message);
         document.getElementById('loading').style.display = 'none';
         document.getElementById('loading-error').style.display = 'block';
-        document.getElementById('error-message').innerText = message ?? 'An unknown error occurred';
+        document.getElementById('error-message').innerText = message;
       };
-      window.addEventListener('error', event => {
-        handle(event.message);
-      });
-      window.addEventListener('unhandledrejection', event => {
-        handle(event.reason && event.reason.message);
-      });
+      var __deregisterInitializationErrorHandling = function() {
+        window.removeEventListener('error', __handleInitializationErrors, { capture: true, once: true });
+        window.removeEventListener('unhandledrejection', __handleInitializationErrors, { once: true });
+      };
+      window.addEventListener('error', __handleInitializationErrors, { capture: true, once: true });
+      window.addEventListener('unhandledrejection', __handleInitializationErrors, { once: true });
     </script>
   </head>
   <body>
@@ -72,7 +76,7 @@
     <div id="forms-app">
       <div id="loading-error">
         <h1>Error loading forms application</h1>
-        <p>Please try updating to the latest version of your browser or contact support.</p>
+        <p>Please try reloading the page, updating your browser to the latest version, or contacting support.</p>
         <p id="error-message"></p>
         <p>Powered by <a href="https://getodk.org">ODK</a>.</p>
       </div>
--- a/apps/forms/index.html
+++ b/apps/forms/index.html
@@ -9,6 +9,7 @@
     <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png"/>
     <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png"/>
     <link rel="manifest" href="/site.webmanifest"/>
+    <script src="/apps/forms/src/init.js"></script>
     <style>
       .loading-container {
         text-align: center;
@@ -42,28 +43,6 @@
         font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
       }
     </style>
-    <script>
-      // NB use legacy JavaScript to support older browsers
-      var __handleInitializationErrors = function(event) {
-        if (!document.getElementById('loading')) {
-          // do not intercept errors after the app has loaded
-          return;
-        }
-        var message = event.message
-          || event.reason && event.reason.message
-          || 'An unknown error occurred.';
-        console.error('FATAL ERROR while loading forms app:', message);
-        document.getElementById('loading').style.display = 'none';
-        document.getElementById('loading-error').style.display = 'block';
-        document.getElementById('error-message').innerText = message;
-      };
-      var __deregisterInitializationErrorHandling = function() {
-        window.removeEventListener('error', __handleInitializationErrors, { capture: true, once: true });
-        window.removeEventListener('unhandledrejection', __handleInitializationErrors, { once: true });
-      };
-      window.addEventListener('error', __handleInitializationErrors, { capture: true, once: true });
-      window.addEventListener('unhandledrejection', __handleInitializationErrors, { once: true });
-    </script>
   </head>
   <body>
     <noscript>
--- a/apps/forms/index.html
+++ b/apps/forms/index.html
@@ -9,7 +9,7 @@
     <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png"/>
     <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png"/>
     <link rel="manifest" href="/site.webmanifest"/>
-    <script src="/apps/forms/src/init.js"></script>
+    <script src="/apps/forms/src/init.js" defer></script>
     <style>
       .loading-container {
         text-align: center;
--- a/apps/forms/index.html
+++ b/apps/forms/index.html
@@ -34,7 +34,32 @@
           transform: rotate(360deg);
         }
       }
+      #loading-error {
+        display: none;
+        margin: 20px auto;
+        max-width: 800px;
+        padding: 1rem;
+        font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+      }
     </style>
+    <script>
+      const handle = (message) => {
+        if (!document.getElementById('loading')) {
+          // do not intercept errors after the app has loaded
+          return;
+        }
+        console.error('FATAL ERROR: loading forms app');
+        document.getElementById('loading').style.display = 'none';
+        document.getElementById('loading-error').style.display = 'block';
+        document.getElementById('error-message').innerText = message;
+      };
+      window.addEventListener('error', event => {
+        handle(event.message);
+      });
+      window.addEventListener('unhandledrejection', event => {
+        handle(event.reason && event.reason.message);
+      });
+    </script>
   </head>
   <body>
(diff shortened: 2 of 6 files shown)


apps/forms/index.html
```diff
@@ -32,15 +32,46 @@
...
+        font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+      }
     </style>
+    <script>
+      const handle = (message) => {
+        if (!document.getElementById('loading')) {
+          // do not intercept errors after the app has loaded
+          return;
+        }
+        console.error('FATAL ERROR: loading forms app');
```
alxndrsn: Would it be helpful to log `event.error` here?  I imagine anyone looking at the console would appreciate it.

apps/forms/index.html
```diff
@@ -32,15 +32,46 @@
           transform: rotate(360deg);
         }
       }
+      #loading-error {
+        display: none;
+        margin: 20px auto;
+        max-width: 800px;
+        padding: 1rem;
+        font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
```
alxndrsn, line 43: Why does this message need its own special font?  Can't it inherit whatever the standard styles have defined?

apps/forms/index.html
```diff
@@ -32,15 +32,46 @@
...
+        }
+        console.error('FATAL ERROR: loading forms app');
+        document.getElementById('loading').style.display = 'none';
+        document.getElementById('loading-error').style.display = 'block';
+        document.getElementById('error-message').innerText = message;
+      };
+      window.addEventListener('error', event => {
+        handle(event.message);
+      });
+      window.addEventListener('unhandledrejection', event => {
```
alxndrsn: Does this occur in form loading?  Hopefully these errors can be logged and resolved when they do happen.

apps/forms/index.html
```diff
@@ -32,15 +32,46 @@
...
+        handle(event.reason && event.reason.message);
```
alxndrsn:
```suggestion
        handle(event.reason?.message);
```

apps/forms/index.html
```diff
@@ -32,15 +32,46 @@
...
+    <script>
+      const handle = (message) => {
+        if (!document.getElementById('loading')) {
+          // do not intercept errors after the app has loaded
+          return;
+        }
+        console.error('FATAL ERROR: loading forms app');
+        document.getElementById('loading').style.display = 'none';
+        document.getElementById('loading-error').style.display = 'block';
+        document.getElementById('error-message').innerText = message;
```
alxndrsn: Per `unhandledrejection`, it looks like message may be `undefined` sometimes.


```suggestion
        document.getElementById('error-message').innerText = message ?? 'An unknown error occurred';
```

apps/forms/index.html
```diff
@@ -32,15 +32,46 @@
...
     <noscript>
       <p>Please enable JavaScript to continue.</p>
       <p>Powered by <a href="https://getodk.org">ODK</a>.</p>
     </noscript>
     <div id="forms-app">
-      <div class="loading-container">
+      <div id="loading-error">
+        <h1>Error loading forms application</h1>
+        <p>Please try updating to the latest version of your browser or contact support.</p>
+        <p id="error-message"></p>
```
alxndrsn, line 59: Maybe this is an overly-generic `id` to be reserving DOM-wide?

```suggestion
        <p id="loading-error-message"></p>
```

alxndrsn (commented):
This looks like it will work, brutally!

A few queries:

1. Will caught errors be reported to Sentry?
2. Would there be any benefit in removing the event listeners either after they've fired, or after loading is completed?
3. Should there be handling for network errors, e.g. if `forms.js` fails to load?

apps/forms/index.html
garethbowen: There is no stylesheet. The css is loaded if/when the app boots up which doesn't happen in this case.

apps/forms/index.html
garethbowen: Unless the event is explicitly stopped it bubbles up and gets logged as per usual, so if I logged it it would end up in the console twice. I stayed with the default logging so it captures all the usual goodness.

apps/forms/index.html
garethbowen: Done.

apps/forms/index.html
garethbowen: I didn't manage to trigger this one it's just an overabundance of caution. Once the app is actually loaded then none of this will do anything.

apps/forms/index.html
garethbowen: I was trying to be backwards compatible in case older browsers end up here.

apps/forms/index.html
garethbowen: Well... the entire "forms-app" div content gets replaced when the app loads so it's less likely to conflict.

apps/forms/index.html
alxndrsn: :+1: are these consistent with the fonts used elsewhere?  From a quick look it seems like this matches `apps/central` whereas web-forms uses `Roboto, sans-serif`.

apps/forms/index.html
alxndrsn: :+1: makes sense.

apps/forms/index.html
garethbowen: I don't think Roboto is included in browsers, and because the bundle hasn't worked, it won't be included as an asset. Anything 'sans-serif' is good, I just wanted to get away from the default serif font!

garethbowen:
> Will caught errors be reported to Sentry?

No almost certainly not, because the sentry code is in the bundle that failed to load.

> Would there be any benefit in removing the event listeners either after they've fired, or after loading is completed?

I thought about that and yes I think there would be if only to reduce complexity. I'm not quite sure how to trigger this state though. Any ideas? I could have some JS in the bundle that invokes it which would obviously only fire if the bundle succeeds?

> Should there be handling for network errors, e.g. if forms.js fails to load?

Hmmm. Could add an `onerror` attribute to the script tag. Should it notify or just retry I wonder?

apps/forms/index.html
alxndrsn: 𝐓𝐡𝐚𝐭 𝐦𝐚𝐤𝐞𝐬 𝐬𝐞𝐧𝐬𝐞.

alxndrsn:
> No almost certainly not, because the sentry code is in the bundle that failed to load.

Sentry should load and initialise as early as possible - hopefully before any WASM issues are hit.  So there's a reasonable chance it's available as `window.Sentry`.  But equally it may already have global `error` and `unhandledrejection` listeners defined, so maybe not something that this code should be handling, given https://github.com/getodk/central-frontend/pull/1635/changes#r3426085957.

apps/forms/index.html
alxndrsn: I think it would be good to add a comment at the end of this function (or the actual event handlers) to state that this is intentional, e.g.

```
// allow the event to propagate to other handlers, e.g. Sentry
```

alxndrsn:
> I thought about that and yes I think there would be if only to reduce complexity. I'm not quite sure how to trigger this state though. Any ideas? I could have some JS in the bundle that invokes it which would obviously only fire if the bundle succeeds?

Some options:

* remove both event listeners on first execution of `handle()`
* add `{ once:true }` to `addEventListener()` calls
* remove both event listeners in `if (!document.getElementById('loading')) {`

alxndrsn:
> Hmmm. Could add an `onerror` attribute to the script tag. Should it notify or just retry I wonder?

I'd just `handle()` that one as well, so that the user sees an error instead of endless loading.

It might suggest an update or alternative to the error message, e.g.

> Please try [reloading](http://example.com), or updating to the latest version of your browser or contact support.

garethbowen:
> add { once:true } to addEventListener() calls

Done.

> update or alternative to the error message

Done.

> Should there be handling for network errors, e.g. if forms.js fails to load?

Added capturing to the error handler which works for exactly this case. Tested by blocking the forms.js in the browser.

I've also added a call when the vue app is finished mounting to remove both of these event handlers.

garethbowen:
I made a few changes here, mostly because the inline script was breaking csp. My options were...

1. Use a nonce, but this should be generated on the server at runtime and included in the csp per request, which would be a big change to our csp generation.
2. Use a hash, which could be generated at build time and hardcoded into the csp, but I worry that if even one character changed in the script then the csp would block it and our testing wouldn't catch it. I thought about adding a test, in either central or frontend, which would compare the two but it feels like introducing another circular dependency.
3. Extract the JS into a separate file, make it as small as reasonably possible, and load it in the head.

I went with 3.

garethbowen:
@alxndrsn I think this is ready for review again.

apps/forms/eslint.config.js
```diff
@@ -58,6 +58,7 @@ export default defineConfig(
 			'web-forms/bin/**/*',
 			'xforms-engine/api-docs/**/*',
 			'**/vendor',
+			'src/init.js'
```
alxndrsn: Maybe this _should_ be linted, but with ES4(?) rules?

apps/forms/index.html
```diff
@@ -9,6 +9,7 @@
     <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png"/>
     <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png"/>
     <link rel="manifest" href="/site.webmanifest"/>
+    <script src="/apps/forms/src/init.js"></script>
```
alxndrsn: Does this work IRL?  It _looks_ like there's a potential race condition - because this script is loaded in `<head>`, `__handle()` may fire before the spinner is on the screen.

<details>
<summary>worked example</summary>

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    ...
    <script src="./init.js"></script>
    <script src="./missing.js"></script>
    <style>
      ...
    </style>
  </head>
  <body>
    ...
  </body>
</html>
```


```js
// NB use legacy JavaScript to support older browsers
console.log('init.js loading');
var __handle = function(e) {
  console.log('__handle() executing...');
  console.log('i am __handle()');
  var spinner = document.getElementById('loading-spinner');
  if (!spinner) return console.log('__handle() did not find spinner');
  var error = document.getElementById('loading-error');
  var message = document.getElementById('error-message');
  var msg = e.message || (e.reason && e.reason.message) || e;
  console.error('FATAL ERROR while loading forms app:', msg);
  spinner.style.display = 'none';
  if (error) error.style.display = 'block';
  if (message) message.innerText = msg;
};
globalThis.__deregisterInitializationErrorHandling = function() {
  window.removeEventListener('error', __handle, { capture: true });
  window.removeEventListener('unhandledrejection', __handle);
};
window.addEventListener('error', __handle, { capture: true, once: true });
window.addEventListener('unhandledrejection', __handle, { once: true });
console.log('init.js loaded');
```


<kbd>


Image

</kbd>

</details>

That would seem to be OK if the next JS load is at the end of `<body>`, but checking a prod web-forms build (e.g. https://staging.getodk.cloud/projects/202/forms/geofence_example/submissions/uuid%3A7db8e734-3135-4140-ae8b-c5420d14f82a/edit), it seems like the forms JS `<script>` is moved into `<head>` at build-time.

<details>
<summary>screenshot</summary>
<kbd>

Image

</kbd>
</details>

The most robust method may be to register the error handler as early as possible, and defer the DOM update, e.g.

```js
if (document.readyState === "loading") {
  // Loading hasn't finished yet
  document.addEventListener("DOMContentLoaded", doSomething);
} else {
  // `DOMContentLoaded` has already fired
  doSomething();
}
```

https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete

alxndrsn (commented):
It's a shame that the CSP would prevent the script from being included inline.

Option 2a would be to relax the CSP header enough to allow an additional CSP to be included via `<meta>` tag, and either calculated at build-time, or verified as a lint/testing step, e.g.

```html
<meta
  http-equiv="Content-Security-Policy" 
  content="default-src 'self'; script-src 'self' 'sha256-$script-hash';"
>
```

apps/forms/index.html
garethbowen: Nice! I added the `defer` attribute so the script is downloaded in parallel but executed immediately after the DOM is parsed.

apps/forms/eslint.config.js
garethbowen: Done (ES5)

garethbowen:
> Option 2a would be to relax the CSP header enough to allow an additional CSP to be included via <meta> tag, and either calculated at build-time, or verified as a lint/testing step,

Yeah that would work too. I'm comfortable with the external JS approach because it should still load quick enough, and avoids the infrastructure requirement for hashing.

apps/forms/eslint.config.js
```diff
@@ -57,7 +57,7 @@ export default defineConfig(
 			'web-forms/dist-demo/**/*',
 			'web-forms/bin/**/*',
 			'xforms-engine/api-docs/**/*',
-			'**/vendor',
```
alxndrsn, line 60: looks unrelated; revert?

garethbowen: Done

apps/forms/index.html
alxndrsn: 🤷  fair enough

apps/forms/src/init.js
```diff
@@ -0,0 +1,20 @@
...
+  if (!spinner) return;
+  var error = document.getElementById('loading-error');
+  var message = document.getElementById('error-message');
+  var msg = e.message || (e.reason && e.reason.message) || 'An unknown error occurred.';
+  console.error('FATAL ERROR while loading forms app:', msg);
+  spinner.style.display = 'none';
+  if (error) error.style.display = 'block';
+  if (message) message.innerText = msg;
+};
+if (globalThis) {
```
alxndrsn, line 13: Is `globalThis` a safe constant to use when trying to be backward-compatible?  I appreciate the code is safe due to this `if()`, but why not use `window`?

apps/forms/index.html
alxndrsn: Looks good :+1:

apps/forms/src/init.js
garethbowen: Because then the TS code needs to use window and it complains. Trying to get both eslint configs happy is impossible. I could just ignore it but I felt ok with this because the only browsers that are going to be deregistering are going to be the modern ones anyway.

alxndrsn (approved):

PR #1636, opened 2026-06-17
fix: aligns dev nginx config with production equivalent

garethbowen:
Closes #

Executed `npm run dev:build` and ensured forms were accessible.

The best possible solution would be to use the same file in both as much as possible, but this is better than letting them get more out of sync than they already are.

None.

None.


alxndrsn (approved):
This helps improve on https://github.com/getodk/central/issues/1601.  Definitely an improvement.

PR #1638, opened 2026-06-17
fix: output main js files to assets dir so they get cached

garethbowen:
Closes #

Executed locally in a prod like environment and saw the cache headers were as expected.

One option was to change the nginx config to cache more paths, but it's complicated enough without additional changes.

Files are now cached as expected.

No.

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -98,9 +98,9 @@ export default defineConfig(({ mode }) => ({
         entryFileNames: (chunkInfo) => {
           const moduleId = chunkInfo.facadeModuleId;
           if (moduleId && moduleId.includes('/apps/forms/')) {
-            return 'forms/[name]-[hash].js';
+            return 'assets/forms/[name]-[hash].js';
           }
-          return 'central/[name]-[hash].js';
+          return 'assets/central/[name]-[hash].js';
         }
       },
     },


vite.config.js
```diff
@@ -98,9 +98,9 @@ export default defineConfig(({ mode }) => ({
         entryFileNames: (chunkInfo) => {
           const moduleId = chunkInfo.facadeModuleId;
           if (moduleId && moduleId.includes('/apps/forms/')) {
-            return 'forms/[name]-[hash].js';
+            return 'assets/forms/[name]-[hash].js';
           }
-          return 'central/[name]-[hash].js';
+          return 'assets/central/[name]-[hash].js';
```
alxndrsn, line 103: Now the location has changed, does vite/rollup update the HTML/JS references to these scripts automatically?

alxndrsn:
When did the output dir change from `/assets` to `/`?

alxndrsn (commented):
Comments above.

vite.config.js
garethbowen: Yes. I checked that by looking at the source, and verifying that it worked in dev, _and_ then the e2e tests passed.

garethbowen:
> When did the output dir change from /assets to /?

The whole function there was introduced to stop vite bundling forms and central into one mega JS asset, so it was done in the "separate forms app" change that was merged a few days ago.

alxndrsn (approved):

PR #1641, opened 2026-06-18
fix(web-forms#827): Fix map not loading on first load in iOS Safari

### LINKED ISSUE #827: Maps appearance is unable to load on Iphone devices
### Expected Behaviour

Maps appearance should load without any issues on iPhone devices

### Observed Behaviour

**Update note!**

I verified once again on test after the server update and I was able to reproduce it there.
So my guess is that this issue was introduced here: https://github.com/getodk/web-forms/issues/535

It happens only if the form is loaded for the first time.
If I'll choose Fill out again at the end of the form then the maps are loaded correctly.
The issue is not reproducing on Android devices or on web

> **It is only in staging instance! I was not able to reproduce it in test** -> **it reproduces on test after server update**

### Reproduction Steps

**Steps:**
1. Open All question types form on iPhone
2. Scroll down to the map question type
3. Geopoint type with maps appearance, Geotrace type no appearance and Geoshape type with no appearance are not loaded. 

### Screenshots or Videos

Image

### Environment

- Browser version: Safari
- OS: iOS 26.5
- Web Forms version: Web Forms version: v0.24.0

latin-panda:
Closes https://github.com/getodk/web-forms/issues/827

iOS Safari throws on multiple dynamic imports

Issue before fix: only one map loads:
 
https://github.com/user-attachments/assets/b03083a2-6466-42f0-b220-9c6409a3fa3b

After the fix: 

https://github.com/user-attachments/assets/16c845c6-3eff-48aa-b35b-b91fe61f09fd

### DIFF
--- a/.changeset/sour-houses-wink.md
+++ b/.changeset/sour-houses-wink.md
@@ -1,0 +1,6 @@
+---
+"@getodk/web-forms": patch
+"@getodk/forms": patch
+---
+
+Serialise MapBlock dynamic import to fix map not loading on first load in iOS Safari
--- a/packages/web-forms/src/components/common/map/AsyncMap.vue
+++ b/packages/web-forms/src/components/common/map/AsyncMap.vue
@@ -6,21 +6,12 @@
  */
 import { createFeatureCollectionAndProps } from '@/components/common/map/geojson-parsers.ts';
 import type { Mode, SingleFeatureType } from '@/components/common/map/getModeConfig.ts';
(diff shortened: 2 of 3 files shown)


latin-panda:
@garethbowen could you please review? I've included videos to demonstrate the issue.

garethbowen (approved):
Yikes.

This feels like a landmine we could easily step on in future. I wonder why this doesn't fail on webkit.

PR #1643, opened 2026-06-18
fix(web-forms#824): reset dompurify config when validating link href attributes

### LINKED ISSUE #824: Hyperlink in form does not always work
### Expected Behaviour

Hyperlinks should move user to the correct page.

### Observed Behaviour

After clicking on the hyperlink I'm moved to the new tab but the URL is empty.

### Reproduction Steps

Form (XML or XLSX file): [3-notes-with-markdown (1).xml.txt](https://github.com/user-attachments/files/28880452/3-notes-with-markdown.1.xml.txt)

[MarkdownExamples(1).xml.txt](https://github.com/user-attachments/files/28880467/MarkdownExamples.1.xml.txt)

Steps:
1.Open link https://test.getodk.cloud/projects/478/forms/MarkdownExamples/submissions
2. Click New submission 
3. Scroll down until you notice the "Clickable link" 
4. Click on the hyperlink -> New empty tab will be opened

### Environment

- Browser version: Firefox 136.0 (64-bit)
- OS: Ubuntu 20.04
- Web Forms version: v0.24.0

### Additional Context

Markdown Examples form:
```
<label>[clickable
                    link](https://nafundi.com/blog/posts/customize-odk-forms-with-fonts-colors-sizes-and-links/)
                </label>
```
vs the example in other form without that issue:
`[get odk](<output value="/data/group/url" />)`

garethbowen:
Closes getodk/web-forms#824

DOMPurify stores the config that you last used, and applies it future checks. So, if the link runs first, DOMPurify uses default settings. If it runs second, it uses our settings which disallow links.

The solution here is to use two configs and set the right one before calling DOMPurify.

We could allow links in the default config but that would also allow HTML links which we don't want to do for security reasons.

None.

None.

### DIFF
--- a/packages/web-forms/src/lib/format/markdown.ts
+++ b/packages/web-forms/src/lib/format/markdown.ts
@@ -38,6 +38,8 @@ export const getStylePropertyMap = (node: ParentMarkdownNode): StyleValue | unde
 export const getUrl = (node: ParentMarkdownNode): string | undefined => {
   if (node.elementName === 'a') {
     const url = (node as AnchorMarkdownNode).url;
+    // validate that the URL isn't a javascript prompt
+    DOMPurify.setConfig({ ALLOWED_TAGS: ['a'], ALLOWED_ATTR: ['href'] });
     if (!DOMPurify.isValidAttribute('a', 'href', url)) {
       return 'about:blank';
     }
--- a/packages/web-forms/src/lib/format/markdown.ts
+++ b/packages/web-forms/src/lib/format/markdown.ts
@@ -48,5 +48,6 @@ export const getUrl = (node: ParentMarkdownNode): string | undefined => {
 };
 
 export const purify = (node: HtmlMarkdownNode): string => {
-  return DOMPurify.sanitize(node.unsafeHtml, DOM_PURIFY_SETTINGS);
+  DOMPurify.setConfig(DOM_PURIFY_SETTINGS);
+  return DOMPurify.sanitize(node.unsafeHtml);
 };
--- a/.changeset/two-paws-wish.md
+++ b/.changeset/two-paws-wish.md
@@ -1,0 +1,5 @@
+---
+"@getodk/web-forms": patch
+---
+
+Fixed markdown sanitization to reliably work for links


packages/web-forms/src/lib/format/markdown.ts
```diff
@@ -38,6 +38,8 @@ export const getStylePropertyMap = (node: ParentMarkdownNode): StyleValue | unde
 export const getUrl = (node: ParentMarkdownNode): string | undefined => {
   if (node.elementName === 'a') {
     const url = (node as AnchorMarkdownNode).url;
+    // validate that the URL isn't a javascript prompt
+    DOMPurify.setConfig({ ALLOWED_TAGS: ['a'], ALLOWED_ATTR: ['href'] });
```
latin-panda, line 42: I ran out of time to test this today, but I'm thinking of this scenario:

```
[click me](javascript%3Aalert(document.cookie)) 
```

alxndrsn:
> DOMPurify stores the config that you last used, and applies it future checks. So, if the link runs first, DOMPurify uses default settings. If it runs second, it uses our settings which disallow links.

Is this expected behaviour in DOMPurify?  If not, it may be worth reporting.

packages/web-forms/src/lib/format/markdown.ts
garethbowen: Looking at the code it uses an allow list for the protocol, and only allows: ftp(s), http(s), mailto, tel, callto, cid, xmpp

Any other prefix should fail.

garethbowen:
> Is this expected behaviour in DOMPurify?

I can't find any actual API docs so I can't know for sure, but the code makes it look intentional. If you call `setConfig` then a flag is set and it is used even if a specific config is provided: https://github.com/cure53/DOMPurify/blob/26006ca3a6414ef4300d84c3ee23fa45dc050839/src/purify.ts#L2351

packages/web-forms/src/lib/format/markdown.ts
latin-panda: It works fine! When tapping a harmful URL, it opens a new empty tab and malicious code doesn't execute.

latin-panda (approved):
LGTM! Tested with both safe and malicious links.

PR #1645, opened 2026-06-18
fix: show spinner until web-form component is done

garethbowen:
Closes #

Testing

The spinner being inline in the HTML means it loads asap.
Only having one spinner means it doesn't jump when the vue bundle loads.
But that does mean we have to make sure every possible route through the app remembers to hide it.

Just visual.

No.

### DIFF
--- a/apps/forms/index.html
+++ b/apps/forms/index.html
@@ -12,7 +12,12 @@
     <style>
       .loading-container {
         text-align: center;
-        padding-top: 35vh;
+        position: absolute;
+        left: 0;
+        right: 0;
+        top: 0;
+        bottom: 0;
+        padding-top: 30vh;
       }
       .spinner {
         width: 60px;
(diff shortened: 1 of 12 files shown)


apps/forms/index.html
```diff
@@ -12,7 +12,12 @@
     <style>
       .loading-container {
         text-align: center;
-        padding-top: 35vh;
+        position: absolute;
+        left: 0;
+        right: 0;
+        top: 0;
+        bottom: 0;
```
latin-panda: Now there is a short way to write this. You can use [inset: 0](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset)

apps/forms/src/components/submission.vue
latin-panda, line 200: There's another loading variable here that could use `hideSpinner`. It might be more helpful to `watch` the `loadingState` and react to when it turns false.

apps/forms/src/components/preview.vue
latin-panda, line 38: Same here

latin-panda (approved):
Tested in Chrome and Safari, works as expected! Thanks!

A couple of minor comments below but approving now to unblock

apps/forms/index.html
garethbowen: Done. It'll look bad on browsers before 2021 but people on browsers that old must be used to things looking bad...

apps/forms/src/components/submission.vue
garethbowen: Unfortunately no... In the submission and preview components we only want to hide the spinner once the web-forms component is finished. If we hide the spinner here, then the spinner would disappear and then the component would do a bunch of work before rendering the form which leaves a white screen for seconds.

So I hide the spinner here if and only if the component fails on some error. If it succeeds then we rely on the enketo-iframe and web-form components to hide the spinner.

I think the confusion comes from too many components having control over the spinner state. One way to fix this would be to have the enketo and web-forms components also emitting when they were done so the event bubbles up to the submission and preview components which then hide the spinner. But I felt this was overkill...

apps/forms/src/components/preview.vue
garethbowen: As above.

PR #1646, opened 2026-06-19
fix(web-forms#815): catch uncaught errors

### LINKED ISSUE #815: Missing function fails silently
[invalid_function.xlsx](https://github.com/user-attachments/files/28164148/invalid_function.xlsx)

There's an error in the console but nothing that's user-facing. I think this should fail loudly because otherwise the issue may go unnoticed until data analysis time which might be way too late.

garethbowen:
Closes getodk/web-forms#815

Catch any errors that propagate up to the web-forms vue component and display an undismisable dialog just like a form loading error. The error continues to propagate so sentry and console still get the error.

Over time it would be good to catch specific errors and handle in a more user friendly way, however we still need a catch all so errors are silently ignored.

Forms that previously failed silently will now block user access. If this causes too many disruptions for users we should reconsider but at least this will shine a light on the issue!

No.

### DIFF
--- a/.changeset/frank-wings-attend.md
+++ b/.changeset/frank-wings-attend.md
@@ -1,0 +1,5 @@
+---
+"@getodk/web-forms": patch
+---
+
+Catch undefined errors and show a user facing dialog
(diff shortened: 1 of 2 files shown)


latin-panda (approved):
Works as described, ready to merge!

PR #1647, opened 2026-06-19
Fix(web-forms#830): set value with now() into a date field returning empty string

### LINKED ISSUE #830: Trying to set a date field to now() doesn't work
### Expected Behaviour

Using `setvalue` to set the value of a `date` field to `now()` should default to today's date.

### Observed Behaviour

Using `setvalue` to set the value of a `date` field to `now()` doesn't do anything (it evaluates to empty string).

### Reproduction Steps

Form (XML or XLSX file): 

[webforms_now_date.xml](https://github.com/user-attachments/files/29032261/webforms_now_date.xml)

Steps:
1. Load the form

### Environment

- Browser version:
- OS:
- Web Forms version:

### Additional Context

Switching the field to `dateTime` solves the issue.
First reported on the forum: https://forum.getodk.org/t/webforms-behaviour-notes/58272/4

latin-panda:
Closes https://github.com/getodk/web-forms/issues/830


latin-panda:
@garethbowen this small PR is ready to review :)

packages/xforms-engine/test/integration/bind-types.test.ts
```diff
@@ -722,6 +715,16 @@ describe('Data (<bind type>) type support', () => {
           expectedAsObject: Temporal.PlainDate.from('2025-12-21'),
           expectedAsText: '2025-12-21',
         },
+        {
+          expression: '2025-03-07T14:30:00-08:00',
+          expectedAsObject: Temporal.PlainDate.from('2025-03-07'),
+          expectedAsText: '2025-03-07',
```
lognaturel, line 721: This is only guaranteed to be the case if the tests run in UTC-8:00. I seem to remember this is set somewhere, right?

packages/xforms-engine/test/integration/actions-events/set-value-action.test.ts
```diff
@@ -706,6 +706,33 @@ describe('setvalue action', () => {
...
+          ),
+          body(
+            input('/data/source', setvalue('xforms-value-changed', '/data/destination', 'now()'))
+          )
+        )
+      );
+
+      scenario.answer('/data/source', 'trigger');
+
+      const now = new Date();
```
lognaturel: Do the tests globally set a fixed time? If not, this test could possibly fail if run right at the midnight boundary: `now()` in the form could be day1, then by the time this line is reached it could be day2. I think for tests that call `now()` or `today()` in forms we should use a frozen clock if we can.

lognaturel: Maybe there's already a suite with date/datetime tests like this? It's really verifying that if an iso datetime value with a time component is stored in a date field, the date component in the local timezone is retained. setvalue is just one of the several mechanisms through which this could happen.

And thanks, I'm porting these straight to JavaRosa which has a related problem, almost the inverse! If you include a datetime like now() in a label, the time component is not included.

garethbowen: +1 for deterministic tests!

Instead of freezing the time I presume this code would work if setting the value of a date to a hardcoded datetime, so it may be easier just to hardcode a full datetime instead of using `now()`. This would also allow for testing of specific timezones too.

packages/xforms-engine/test/integration/bind-types.test.ts
garethbowen: It's set as an environment variable in a few places, eg: https://github.com/getodk/central-frontend/blob/ab71c5e18435ec32535999b4b17684577e7c9f22/.github/workflows/wf-ci.yml#L15

garethbowen (changes requested):
Nice! Great to tick off some more TODOs.

One change requested for making the tests more deterministic.

packages/xforms-engine/test/integration/actions-events/set-value-action.test.ts
latin-panda: Fixed!

latin-panda:
@garethbowen I've addressed the feedback; please have another look. Thanks!

garethbowen (approved):
Nice!

PR #1649, opened 2026-06-19
fix(web-forms#803): deliver Sentry DSN at runtime via client-config.json

### LINKED ISSUE #803: Implement error reporting

latin-panda:

Tested that both apps send to Sentry:

CI passes, and local testing connecting with a project in Sentry: 

I prefer environment variables (a build-time solution). But this is the best option we have to fit Central's architecture, and it's been discussed with the team to use client-config.json file to get the DNS. 

It doesn't impact features from Central or Web Forms. But now it loads more KB for the Sentry script to track performance and errors. 

Already updated in central's readme file


latin-panda:
@alxndrsn could you please review? This is the other half of the work 

[PR in Central to pass DNS ](https://github.com/getodk/central/pull/2005)

apps/forms/src/main.ts
```diff
@@ -12,8 +12,25 @@ import initSentry from './utils/sentry';
...
+    console.error('Failed to fetch /client-config.json', error);
+    return {};
+  }
+};
+
+const clientConfig = await fetchClientConfig();
+
 const app = createApp(Forms as Component);
-initSentry(app, 'web-forms');
+initSentry(app, 'web-forms', clientConfig.sentryDsn);
```
alxndrsn, line 33: Is it expected that web-forms and central-frontend share the same DSN?

bin/check-bundle-size.js
```diff
@@ -62,6 +62,7 @@ function isTooBig({ path, size }) {
     case 'web-form.js':    return size > 1_000_000; // Matches web-form-renderer.js
     case 'MapBlock.js':    return size >   600_000; // A Web Forms' feature bundle
     case 'geojson-map.js': return size >   500_000;
+    case 'browserTracingIntegration.js': return size > 215_000; // Sentry bundled since DSN is runtime
```
alxndrsn: I don't understand this comment.

apps/central/src/main.js
```diff
@@ -26,5 +26,15 @@ import './jquery';
 import './bootstrap';
 
 const app = createApp(App);
-initSentry(app, 'central-frontend');
-app.use(createContainer()).directive('tooltip', vTooltip).mount('#app');
+const container = createContainer();
```
alxndrsn: Why has the order of `initSentry(...)`/`createContainer()` switched?

apps/forms/src/main.ts
latin-panda: Yes, that's my understanding from the [team discussion ](https://getodk.slack.com/archives/C051L6AKG4F/p1780676691482789?thread_ts=1780597625.941729&cid=C051L6AKG4F)

Web Forms and Central Front-end pass a tag, for example, the second parameter in this call:

apps/central/src/main.js
latin-panda: It's a leftover from a previous test I was doing. It doesn't affect anything since it's an empty Vue container. But I've moved it to L40, which makes more sense. Thanks!

apps/forms/src/main.ts
latin-panda: The tag mentioned before, it's visible here:

Screenshot 2026-06-22 at 1 20 58 PM

bin/check-bundle-size.js
latin-panda: Now that the DSN is passed at runtime (fetched from `/client-config.json`), the bundler can't tell whether Sentry's init path will run, so it has to bundle the whole thing. That's why the chunk now appears. 

I've updated the comment to just indicate it's from Sentry and shared between apps.

latin-panda:
@alxndrsn this is ready for another review! :)

bin/check-bundle-size.js
alxndrsn: :+1: makes sense now, thanks!

apps/central/src/main.js
```diff
@@ -26,5 +26,13 @@ import './jquery';
...
 const app = createApp(App);
-initSentry(app, 'central-frontend');
+
+const { sentryDsn } = await fetch('/client-config.json')
+  .then(r => r.json())
+  .catch((error) => {
+    // eslint-disable-next-line no-console
+    console.error('Failed to fetch /client-config.json', error);
+    return {};
+  });
```
alxndrsn, line 36: If I've understood correctly, `client-config.json` will now be loaded twice:

1. here, and
2. at https://github.com/getodk/central-frontend/blob/45052e9c7bcdd156e2ea833abfd23e69089a042c/apps/central/src/router.js#L115

Is that the case?  If so, it's probably not necessary in the long-term.  If it's not planned to resolve in this PR, is it possible to create a new Issue to track this?

latin-panda: Yes, it's a duplicate. I've created an issue here: https://github.com/getodk/central/issues/2007. Primarily to better understand where/how it's used in Central and to run targeted tests.

alxndrsn: Great, thanks :+1:

alxndrsn (approved):

PR #1652, opened 2026-06-22
fix(web-forms#818): detect whitespace in csv column headers

### LINKED ISSUE #818: Detect and fail early when csv column names don't match usage
### Expected Behaviour

When using `select_*_from_file` the csv file must have column names that exactly match the expected values. By default these are `name` and `label`, which can be overridden using additional parameters. When they don't match exactly web forms should give a clear and useful error so it's corrected early.

### Observed Behaviour

The form loads but behaves strangely, for example, selecting one option selects them all because they all have a null value. This is particularly hard to debug when it's a trailing space in the column name.

### Reproduction Steps

Use the form in this forum post: https://forum.getodk.org/t/error-message-on-using-csv-as-an-external-list/58214

### Environment

- Browser version: all
- OS: all
- Web Forms version: latest

### Additional Context

- Reported on the forum: https://forum.getodk.org/t/error-message-on-using-csv-as-an-external-list/58214
- Enketo handles the trailing whitespace gracefully.
- Collect throws an error when loading.

garethbowen:
Closes getodk/web-forms#818

Tests added, manual testing.

Forms that previously loaded will now fail immediately, but we expect that they were buggy, so this fails earlier and clearer.

### DIFF
--- a/packages/xforms-engine/src/parse/model/SecondaryInstance/sources/external-instance-csv-parser.ts
+++ b/packages/xforms-engine/src/parse/model/SecondaryInstance/sources/external-instance-csv-parser.ts
@@ -43,8 +43,8 @@ const rejectNullCharacters = (resourceURL: JRResourceURL, cell: string) => {
   }
 };
 
-const rejectWhitespaceCharacters = (resourceURL: JRResourceURL, cell: string) => {
-  if (/\s/.exec(cell)) {
+const rejectWrappingWhitespaceCharacters = (resourceURL: JRResourceURL, cell: string) => {
+  if (/^\s|\s$/.exec(cell)) {
     throw new ErrorProductionDesignPendingError(
       `Failed to parse CSV ${resourceURL.href}: whitespace character`
     );
@@ -80,7 +80,7 @@ const assertCSVRow: AssertCSVRow = (resourceURL: JRResourceURL, columns) => {
         `Failed to parse CSV ${resourceURL.href}: invalid column at ${index}`
       );
     }
-    rejectWhitespaceCharacters(resourceURL, column);
+    rejectWrappingWhitespaceCharacters(resourceURL, column);
     rejectNullCharacters(resourceURL, column);
   }
 };
--- a/packages/xforms-engine/src/parse/model/SecondaryInstance/sources/external-instance-csv-parser.ts
+++ b/packages/xforms-engine/src/parse/model/SecondaryInstance/sources/external-instance-csv-parser.ts
@@ -43,6 +43,14 @@ const rejectNullCharacters = (resourceURL: JRResourceURL, cell: string) => {
   }
 };
 
+const rejectWhitespaceCharacters = (resourceURL: JRResourceURL, cell: string) => {
+  if (/\s/.exec(cell)) {
+    throw new ErrorProductionDesignPendingError(
+      `Failed to parse CSV ${resourceURL.href}: whitespace character`
+    );
+  }
+};
+
 const stripTrailingEmptyCells = (columns: CSVRow, row: CSVRow): CSVRow => {
   const result = row.slice();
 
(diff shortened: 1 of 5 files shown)


packages/xforms-engine/src/parse/model/SecondaryInstance/sources/external-instance-csv-parser.ts
```diff
@@ -72,7 +80,7 @@ const assertCSVRow: AssertCSVRow = (resourceURL: JRResourceURL, columns) => {
         `Failed to parse CSV ${resourceURL.href}: invalid column at ${index}`
       );
     }
-
+    rejectWrappingWhitespaceCharacters(resourceURL, column);
```
latin-panda: It's really useful to know which column failed, like in L80. For example: `Failed to parse cities.csv: whitespace character at column 9.`
```suggestion
    rejectWrappingWhitespaceCharacters(resourceURL, column, index);
```

packages/xforms-engine/test/parse/model/SecondaryInstance/sources/external-instance-csv-parser.test.ts
```diff
@@ -37,6 +37,37 @@ describe('external-instance-csv-parser', () => {
...
+    );
+  });
+
+  it('errors when given trailing whitespace character in header', () => {
+    expect(() => parseItems(url, 'foo,baz ')).to.throw(
+      'Failed to parse CSV jr://csv/mock.csv: whitespace character'
+    );
+  });
+
+  it('passes when given inner whitespace character', () => {
```
latin-panda, line 52: Could you please add a test case where the cell value has leading or trailing spaces? Just ensure that the error is thrown only with the header. Thanks!

latin-panda: For values, I get an error, but I think it should trim and keep working. 
Image

csv, see ` Option B`:
```
name,label
opt_a,Option A 
opt_b, Option B
```

latin-panda (changes requested):
I tested for missing columns and whitespace in the header and values. I left a couple of suggestions below.

packages/xforms-engine/src/parse/model/SecondaryInstance/sources/external-instance-csv-parser.ts
garethbowen: Done. Error messages now include row and column where relevant. Have a look at the unit tests to see if the error messages are better now.

packages/xforms-engine/test/parse/model/SecondaryInstance/sources/external-instance-csv-parser.test.ts
garethbowen: Riiiight. For non-header rows I think we should just leave the whitespace as is. I don't think it's required to be a valid reference so it can be any string. I've modified the code to only check for whitespace in the header row.

lognaturel: Consistent with what we said for values at https://github.com/getodk/xforms-spec/issues/351: "Processors MUST preserve whitespace in values exactly as represented in the source document."

packages/xforms-engine/src/parse/model/SecondaryInstance/sources/external-instance-csv-parser.ts
```diff
@@ -30,15 +35,36 @@ interface CSVExternalSecondaryInstanceItemColumn {
 
 type CSVExternalSecondaryInstanceItem = readonly CSVExternalSecondaryInstanceItemColumn[];
 
+class CSVExternalSecondaryInstanceValidationError extends ErrorProductionDesignPendingError {
```
latin-panda: Please consider moving this to the errors folder.

latin-panda (approved):
I tested several cases and found no regressions. Just one comment below, but approving to unblock.

packages/xforms-engine/src/parse/model/SecondaryInstance/sources/external-instance-csv-parser.ts
garethbowen: Done.

latin-panda: Thank you LGTM!

PR #1653, opened 2026-06-22
DOMPurify: clear persistent config before sanitisation

alxndrsn:
By design, DOMPurify falls back to a global "persistent" configuration if set with `DOMPurify.setConfig(...)`.  This means that 3rd-party calls to `DOMPurify.setConfig()` can break expectations about content filtering.

See:

* https://github.com/cure53/DOMPurify#persistent-configuration
* https://github.com/cure53/DOMPurify/issues/1134

Related:

* https://github.com/getodk/central-frontend/pull/1643
* https://github.com/enketo/enketo/pull/1541

Added a set of tests that fail without the change to application code, but pass with it.

This behvaiour is surprising, and the only sure approach to avoiding it seems to be the call to `clearConfig()`.

No effect.

No.

### DIFF
--- a/apps/central/test/components/markdown/view.spec.js
+++ b/apps/central/test/components/markdown/view.spec.js
@@ -1,3 +1,5 @@
+import DOMPurify from 'dompurify';
+
 import MarkdownView from '../../../src/components/markdown/view.vue';
 
 import { mount } from '../../util/lifecycle';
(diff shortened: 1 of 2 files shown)


apps/central/test/components/markdown/view.spec.js
```diff
@@ -9,45 +11,55 @@ const mountComponent = (raw) => mount(MarkdownView, {
...
 describe('MarkdownView', () => {
-  it('renders basic text', () => {
-    const component = mountComponent('some text');
-    const { outerHTML } = component.get('div').element;
-    outerHTML.should.equal('<div class="markdown-view"><p>some text</p>\n</div>');
-  });
+  [
+    ['clean slate',                () => { DOMPurify.clearConfig(); DOMPurify.removeAllHooks(); }], // eslint-disable-line no-multi-spaces
+    ['polluted persistent config', () => DOMPurify.setConfig({ ALLOWED_TAGS: [], ALLOWED_ATTR: [] })],
+    ['stray hooks',                () => DOMPurify.addHook('uponSanitizeAttribute', (_, hookEvent) => hookEvent.forceKeepAttr = true)], // eslint-disable-line no-multi-spaces
```
garethbowen: It's not clear to me how and why `forceKeepAttr` works here. What about a more obvious hook, eg:


```suggestion
    ['stray hooks',                () => DOMPurify.addHook('uponSanitizeAttribute', () => throw new Error('should have been unhooked'))], // eslint-disable-line no-multi-spaces
```

garethbowen (approved):
Nice!

I'll put this on my list of things to put in the common package so frontend and web-forms can both use safe versions.

apps/central/test/components/markdown/view.spec.js
alxndrsn: :+1: I think both actually demonstrate interesting properties

alxndrsn: Added.

PR #1656, opened 2026-06-22
web-forms: DOMPurify: remove hooks before use

alxndrsn:

- [x] nothing
- [ ] added new tests

DOMPurify acts as a singleton, so "hooks" added for any use will persist, perhaps beyond the expectations of developers.

Ref:

* https://github.com/cure53/DOMPurify/issues/914
* https://github.com/cure53/DOMPurify/issues/1132
* https://github.com/cure53/DOMPurify/issues/1133

No effect, hopefully.

No.


garethbowen (approved):
I 99% sure this won't do anything because there are no hooks being registered in this copy of dompurify but there's no real downside to adding this to cover the other 1%.

alxndrsn:
This will also protect against any deps which use (the same) DOMPurify.

PR #1659, opened 2026-06-23
fix(web-forms#813): improve error when required attachments fail to load

### LINKED ISSUE #813: Improve error message for missing external secondary instance
Current error:
Image

When Web Forms is used with Central, a form designer will be shown that an attachment is missing but they'll still be allowed to publish. Both Collect and Enketo will allow the form to load.

It's ok for this error to be blocking but ideally we would:
* show the filename
* Exclude technical error details because the error is understood without it.

Enketo's error:
Image

Related:
* https://github.com/getodk/web-forms/issues/202
* https://github.com/getodk/web-forms/issues/80

garethbowen:
Closes: getodk/web-forms#813

Tested locally on various errors.

Discussed in the issue.

Gives more actionable error messages.

No.

### DIFF
--- a/packages/web-forms/src/components/OdkWebForm.vue
+++ b/packages/web-forms/src/components/OdkWebForm.vue
@@ -304,11 +304,11 @@ onUnmounted(() => {
 	/>
 
 	<template v-if="state.status === 'FORM_STATE_FAILURE'">
-		<FormLoadFailureDialog severity="error" :error="state.error" />
+		<FormLoadFailureDialog :error="state.error" />
 	</template>
 
 	<template v-else-if="runtimeError">
-		<FormLoadFailureDialog severity="error" :error="runtimeError" />
+		<FormLoadFailureDialog :error="runtimeError" />
 	</template>
 
 	<div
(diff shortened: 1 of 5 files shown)


packages/xforms-engine/src/parse/model/SecondaryInstance/sources/ExternalSecondaryInstanceResource.ts
```diff
@@ -6,11 +6,22 @@ import { getResponseContentType } from '../../../../lib/resource-helpers.ts';
 import { FormAttachmentResource } from '../../../attachments/FormAttachmentResource.ts';
 import type { ExternalSecondaryInstanceSourceFormat } from './SecondaryInstanceSource.ts';
 
+class ErrorLoadingSecondaryInstanceResource extends ErrorProductionDesignPendingError {
```
latin-panda: Nice improvement! Just a tiny note: those errors are in [this folder.](https://github.com/garethbowen/central-frontend/tree/master/packages/xforms-engine/src/error)

packages/web-forms/src/components/OdkWebForm.vue
```diff
@@ -304,11 +304,11 @@ onUnmounted(() => {
 	/>
 
 	<template v-if="state.status === 'FORM_STATE_FAILURE'">
-		<FormLoadFailureDialog severity="error" :error="state.error" />
+		<FormLoadFailureDialog :error="state.error" />
```
latin-panda, line 307: ✨

packages/xforms-engine/src/parse/model/SecondaryInstance/sources/ExternalSecondaryInstanceResource.ts
garethbowen: Done.

latin-panda (approved):
It looks good! Thanks!

There seems to be a test failing in the GitHub Actions

garethbowen:
@latin-panda I'm 99% sure that test is to do with the change to make web-forms default. It's failing on all my builds now. I'm working on a fix...

PR #1661, opened 2026-06-23
fix(web-forms#828): preserve map state during user edits

### LINKED ISSUE #828: Geopoint map is showing "point saved" without any pin on the map.
### Expected Behaviour

If there is no pin in the map there should not be information that the point is saved even if the geopoint is prefilled

### Observed Behaviour

Issue from the forum: https://forum.getodk.org/t/webforms-behaviour-notes/58272
A dynamic default for a geopoint doesn't show a pin on the map when calculated in but does show ":white_check_mark: Point Saved" at the bottom.

### Reproduction Steps

Steps:
1. Navigate to the project https://staging.getodk.cloud/projects/237
2. Open form WF bug 6 - geopoint dynamic default in Web Forms
3. In Web Forms Pick a city to prefill the geopoint -> Point saved note will appear

### Screenshots or Videos

Image

### Environment

- Browser version: Firefox 136.0 (64-bit)
- OS: Ubuntu 20.04
- Web Forms version: v0.24.0

latin-panda:
Closes https://github.com/getodk/web-forms/issues/828

https://github.com/user-attachments/assets/2748cb3f-5ccc-42eb-8a58-5f40c445b9cb

The previous fix, combined with coordinates defaulting to zero to match Collect, caused points to disappear and phantom points to appear in incorrect locations during user edits or when adding new points ([see video](https://jam.dev/c/ee836d74-5199-428a-bc95-07db4ddcf42b)).

The engine reformats every value, and the map component watches the prop for changes (needed for dynamic defaults and setvalue). So the value that comes back never equals what was emitted, and the rebuild destroys in-flight user edits.

There are 2 fixes here:
- Formats coordinates sent to the engine (integers padded with .0, segments separated without spaces) so the value the map emits matches the value the engine sends back. There's room for improvement here. For example, the engine could expose its codec functions, but that doesn't fit how the engine is currently structured. It's something to explore as part of another ticket to unblock the release fix.                                                             
- Only redraws the map when the incoming value differs from what's already drawn.

It should not introduce a new behavior, but fixes to match expected behavior.

No

### DIFF
--- a/packages/web-forms/src/components/common/map/useMapBlock.ts
+++ b/packages/web-forms/src/components/common/map/useMapBlock.ts
@@ -356,9 +356,16 @@ export function useMapBlock(config: MapBlockConfig, events: MapBlockEvents) {
 
   /**
    * Places a newly built feature onto the empty layer and marks it saved.
+   * Skips when the incoming geometry's ODK value already matches what's on the map,
+   * So user edits coming back through the engine don't rebuild the feature.
    */
   const loadAndSaveSingleFeature = (feature: Feature | undefined) => {
-    if (!mapInstance || currentMode.capabilities.canLoadMultiFeatures || !feature) {
+    if (
+      !mapInstance ||
+      currentMode.capabilities.canLoadMultiFeatures ||
+      !feature ||
+      feature.get(ODK_VALUE_PROPERTY) === mapFeatures?.getSavedFeatureValue()
+    ) {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


latin-panda:
@garethbowen This is a new report from the QA Team. I'm going to create a ticket to centralize the value construction, more in the PR description.

packages/web-forms/src/components/common/map/useMapBlock.ts
```diff
@@ -356,9 +356,16 @@ export function useMapBlock(config: MapBlockConfig, events: MapBlockEvents) {
 
   /**
    * Places a newly built feature onto the empty layer and marks it saved.
+   * Skips when the incoming geometry's ODK value already matches what's on the map,
+   * So user edits coming back through the engine don't rebuild the feature.
```
garethbowen, line 360: I'm a little worried about the engine firing unnecessary events even if we do ignore them here.

Now that we are formatting user edits before setting them in the engine, this should not happen any more, right?

So is this additional check just being extra defensive or is it actually firing an update with nothing changed?

garethbowen (approved):
Approved so it's not blocking code freeze.

One comment inline.

packages/web-forms/src/components/common/map/useMapBlock.ts
latin-panda: Yes, it's defensive. Also, I feel it gives more clarity to the code.

PR #1662, opened 2026-06-23
chore: adds workflow to publish Web Forms packages

### LINKED ISSUE #821: Update central release process to publish web-forms packages to npm
### Description

Now that web-forms is merged into central-frontend we need to update the release process so when central is published, the web-forms packages are also published to npm.

- [x] Port over the [release github action](https://github.com/getodk/web-forms/blob/main/.github/workflows/release.yml) so it's triggered when central creates a git tag or GH release
- [x]  Update the [release checklist](https://docs.google.com/document/d/1Bqg2LfqshyRpVI2R71PQVxMSRLhTwPL37n0IRjQfvOA/edit?tab=t.0) to run `npm run version` to bump the package versions, and find a way to revert apps/central which does not follow the same versioning scheme as web-forms
- [x] Update the release checklist with guidance (or scripting) to build the release notes from the combined changelogs

latin-panda:
Resolves one item from https://github.com/getodk/web-forms/issues/821

### DIFF
--- a/.changeset/wide-ties-sniff.md
+++ b/.changeset/wide-ties-sniff.md
@@ -5,4 +5,4 @@
 "@getodk/xpath": major
 ---
 
-First stable release"
+First stable release
--- a/.changeset/wide-ties-sniff.md
+++ b/.changeset/wide-ties-sniff.md
@@ -1,0 +1,8 @@
+---
+"@getodk/common": major
+"@getodk/web-forms": major
+"@getodk/xforms-engine": major
+"@getodk/xpath": major
+---
+
+First stable release"
--- a/.github/workflows/wf-publish.yml
+++ b/.github/workflows/wf-publish.yml
@@ -1,0 +1,47 @@
+name: Publish Web Forms packages
+
+on:
+  push:
+    tags:
+      - 'v[0-9]*.[0-9]*.[0-9]*'
+      - '!v*-*' # Excludes pre-releases
+
+permissions:
+  id-token: write # Required for OIDC
+  contents: write # Required for pushing package version tags
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Set up npm registry
+        uses: actions/setup-node@v6
+        with:
+          node-version: '24.14.1'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Cache dependencies
+        uses: actions/cache@v5
+        with:
+          path: |
+            node_modules
+            **/node_modules
+          key: install-24.14.1-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/*.yml', 'packages/*/package.json', 'packages/*/package-lock.json') }}
+          restore-keys: |
+            install-24.14.1-
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build packages
+        run: npm run build
+
+      - name: Publish packages
+        run: npm run changeset publish
+
+      - name: Push package version tags
+        run: git push --follow-tags


.github/workflows/wf-publish.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build packages
+        run: npm run build
+
+      - name: Publish packages
+        run: npm run changeset publish
+
+      - name: Push package version tags
```
latin-panda, line 46: The `changeset publish` creates tags automatically 
https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#publish

latin-panda:
@garethbowen What do you think about this workflow?

garethbowen (commented):
Ideally this would also run changeset version which would update the version numbers and the release notes automatically. Because it already has repo commit rights it feels like a natural fit. The thing we have to be careful about is not bumping the version of the apps, because they will stay on 0.1.0. What do you think about executing this just on the web-forms packages?

Also, I think now's a good time to bump to 1.0.0 for all wf packages. How about including a major changelog in this PR? Then whether we bump version manually or automatically it'll be ready to go!

latin-panda:
> Ideally this would also run changeset version which would update the version numbers and the release notes automatically. 

I really like keeping the approval step since it's low effort but gives us a chance to fix a note. We can try setting it up so it automatically opens a PR with the version bumps and release notes for us to review. 

>  The thing we have to be careful about is not bumping the version of the apps, because they will stay on 0.1.0. What do you think about executing this just on the web-forms packages?

I ran some experiments, and it looks like `changeset` actually requires a version bump to generate the release notes. 

Looking at the recent PRs, the `/apps/central` is adding changesets ([example](https://github.com/getodk/central-frontend/commit/6947cd169967a5f939ad0f85ad6b902e30b14c32#diff-6ed18df14ec6cf8baf923d6d7030b688ddadbf2c9b9edca66bd8649536f7b0ab)), it's a good thing! 

The main `/central` repo can link to the more technical changelogs generated in `/apps` and `/packages`. Meanwhile, the user-friendly, readable release notes (with images, etc.) will still live on the forum. Because of how changeset works, this is the standard way to track changelogs over time, and the root `package.json` will still stay untouched at 0.1.0. Does that flow sound good?

> Also, I think now's a good time to bump to 1.0.0 for all wf packages. How about including a major changelog in this PR? Then whether we bump version manually or automatically it'll be ready to go!

I'd feel much more comfortable holding off on the 1.0.0 bump until QA finishes their release testing for the Central side. Even though they aren't doing a full regression on Web Forms again, the Central test cases that involve WF are a great smoke test. I want to ensure we don't have to immediately patch 1.0.0 if something unexpected comes up! Once they finish, we can definitely push forward with the major bump.

---

Right now, the changeset is a bit messy. For example, it throws because the scenario package is not found. I'm going to fix it a bit in a separate PR.

garethbowen:
> We can try setting it up so it automatically opens a PR with the version bumps and release notes for us to review.

That works for me!

> I ran some experiments, and it looks like changeset actually requires a version bump to generate the release notes.

Yeah. We could revert the package.json in apps/* to keep the version the same but then the release notes wouldn't make much sense. I'm comfortable with that but may be worth checking with the team.

> I'd feel much more comfortable holding off on the 1.0.0 bump until QA finishes their release testing for the Central side. 

I just mean, add the changeset which does the major bump, don't actually do the release! Then when we do the release in a week or two the packages will automatically get the 1.0.0 bump.

latin-panda:
> I just mean, add the changeset which does the major bump, don't actually do the release! Then when we do the release in a week or two the packages will automatically get the 1.0.0 bump.

Sure! I've pushed the changeset. I understood differently at first 😅

.changeset/wide-ties-sniff.md
```diff
@@ -0,0 +1,8 @@
+---
+"@getodk/common": major
+"@getodk/web-forms": major
+"@getodk/xforms-engine": major
+"@getodk/xpath": major
+---
+
+First stable release"
```
garethbowen: Typo here..


```suggestion
First stable release
```

garethbowen (approved):
One typo in the changeset but otherwise good to go!

PR #1666, opened 2026-06-25
Fixes: Translation keys for events related to custom properties

sadiqkhoja:


apps/central/src/components/audit/row.vue
```diff
@@ -137,6 +137,7 @@ export default {
     // having a "category" indicated by the first part. For example, the
     // category of project.create is project.
     category() {
+      if (this.audit.action.startsWith('actor_property.')) return 'project';
```
matthew-white: Instead of changing the logic behind `category`, could you try changing these config objects instead:

- `typeByFirstPart`
  - I think `actor_property: 'resource.project'` should work.
- `acteeSpeciesByCategory`
  - I think you can do `acteeSpeciesByCategory.actor_property = acteeSpeciesByCategory.project;`

apps/central/src/locales/en.json5
```diff
@@ -160,7 +166,8 @@
         "assignment_create": "Give Access",
         "assignment_delete": "Remove Access",
         "session_end": "@:audit.action.default.session_end",
-        "delete": "@:audit.action.default.delete"
+        "delete": "@:audit.action.default.delete",
+        "property_set": "Set Property"
```
matthew-white: Instead of translating "Set Property" twice (once for app users and once for public links), you can:

- Define it once in `audit.action.default.property_set`
- Then reference it here (and for `field_key`), similar to how `delete` works

apps/central/src/components/audit/filters/action.vue
matthew-white, line 1: How about adding `field_key.property.set` and `public_link.property.set` as options?

matthew-white (approved):

PR #1667, opened 2026-06-26
chore: npm audit fix

garethbowen:
```
npm audit fix
npm audit fix --workspaces
```

- Checked the updates and they're mostly small number changes and/or in devdependencies. I did not check the release notes for each.
- CI
- manual smoke test, though I certainly didn't hit every page

...

None.

No.


matthew-white (approved):
Thank you!

PR #1669, opened 2026-06-26
Fixes: show form upload error inline

### LINKED ISSUE #1897: Make sure form upload issues are close to the upload control
Now that the new form upload control is inline rather than in a modal, some errors that have been displayed in toasts are now far from where they would be expected:

* Upload something that's not an XLSForm:

Image

* Hit a Validate error:

Image

* Hit a pyxform error:

Image

One possible way to address this would be to move the errors inline: https://github.com/getodk/central/issues/1080

sadiqkhoja:
Closes https://github.com/getodk/central/issues/1897

### DIFF
--- a/apps/central/src/components/form/upload.vue
+++ b/apps/central/src/components/form/upload.vue
@@ -14,6 +14,22 @@ except according to the terms contained in the LICENSE file.
 definition for an existing form -->
 <template>
   <div id="form-upload">
+    <div class="introduction">
+      <p>
+        <template v-if="!draft">{{ $t('introduction[0].create') }}</template>
+        <template v-else>{{ $t('introduction[0].update') }}</template>
+        <sentence-separator/>
+        <i18n-t keypath="introduction[1].full">
+          <template #tools>
+            <doc-link to="form-tools/">{{ $t('introduction[1].tools') }}</doc-link>
+          </template>
+        </i18n-t>
+      </p>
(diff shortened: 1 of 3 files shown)


apps/central/src/components/form/upload.vue
```diff
@@ -14,6 +14,22 @@ except according to the terms contained in the LICENSE file.
 definition for an existing form -->
 <template>
   <div id="form-upload">
+    <div class="introduction">
```
sadiqkhoja, line 17: move introduction above warnings and errors div, which looks better to my eyes. Open to revert back

apps/central/test/components/form/upload.spec.js
```diff
@@ -39,6 +40,26 @@ const upload = async (component, file = xlsForm()) => {
...
+          if (message != null) errorDiv.text().should.stringMatch(message);
+        } else {
+          _super.call(this, type, message);
+        }
+      };
+    });
+  });
+  afterAll(() => {
+    Assertion.overwriteMethod('alert', () => originalAlert);
+  });
```
sadiqkhoja, line 61: primarily did it for `testStandardButton` but there are other tests which uses this assertion in this file. Another option was to add additional parameter in `testStandardButton` to modify alert assertion

matthew-white: Didn't know you could do this! Seems like a nice solution for this scenario. :+1:

matthew-white (approved):

PR #1670, opened 2026-06-26
chore: adds workflow to automate release and version bump

latin-panda:
Resolves one item from https://github.com/getodk/web-forms/issues/821

You can see the PR it generates in my fork: https://github.com/latin-panda/test-central-frontend/pull/1/ 

Manual trigger, mostly because of the PR step. After this, the developer can push the release tag and trigger the other workflow, wf-publish.yml.

N/A

I'll update the release steps after approval.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,29 @@
+name: Prepare release
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  prepare-release:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        ref: master
+    - uses: actions/setup-node@v6
+      with:
+        node-version-file: package.json
+        cache: 'npm'
+    - run: npm ci
+    - name: Open release PR
+      uses: changesets/action@v1
+      with:
+        version: npm run version
+        title: 'chore: version bump and release notes'
+        commit: 'chore: version bump and release notes'
+      env:
+        GITHUB_TOKEN: ${{ github.token }}


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,29 @@
...
+        cache: 'npm'
+    - run: npm ci
+    - name: Open release PR
+      uses: changesets/action@v1
+      with:
+        version: npm run version
+        title: 'chore: version bump and release notes'
+        commit: 'chore: version bump and release notes'
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
```
latin-panda, line 29: We could try to generate links to the changelogs, but it's a bit hacky/fragile, I'd rather keep the workflow simple.

Still, I'm adding the option here for suggestion:

```suggestion
      env:
        GITHUB_TOKEN: ${{ github.token }}
      run: |
          git fetch origin changeset-release/master
          git checkout changeset-release/master
          node -e '
            const base = "https://github.com/getodk/central-frontend/blob/master";
            const lines = ["apps/central", "apps/forms", "packages/web-forms", "packages/xforms-engine", "packages/xpath"]
              .map(p => `* [${p}](${base}/${p}/CHANGELOG.md#${require(`./${p}/package.json`).version.replaceAll(".", "")})`);
            console.log("```\n" + lines.join("\n") + "\n```");
          ' | gh pr edit ${{ steps.changesets.outputs.pullRequestNumber }} --body-file -

```

latin-panda:
@garethbowen small PR :) 
I've tested in my fork, link in the PR details.

garethbowen (changes requested):
I don't think this action simplifies the current process enough to be an improvement. 

If I understand correctly the release process will change from...

- Run `npm run version` in `central-frontend`.
- Commit the changes on a new branch

To...

- Execute this GH action (link)

Is that right?

If so, I prefer the existing way because I can review the diff in the command line before creating the PR (to do things link adding the links to changelogs), and I don't have to context switch from command line to GH and wait. IMO it's also easier to maintain and extend in the command line, for example. if you wanted to add your code for generating links to the changelogs IMO it would be less hacky/fragile if it were done in a node script. In other words, doing this as an npm command would give us a better place to continue tweaking and adding more automation to the build process.

Things that would make me prefer the GH action approach would be if it were triggered automatically somehow, or if it were automating more steps.

latin-panda:
Yes, I agree. The addition was straightforward since the changeset already has an action that automates PR creation. I'll close it for now, but we can reopen it later if the process becomes more complex or if it's needed again.

PR #1672, opened 2026-06-26
Feature: upload form immediate without upload button

### LINKED ISSUE #2003: Immediatelly upload form draft upon drag/select (remove upload/cancel buttons)
[Slack](https://getodk.slack.com/archives/C01TKJ7TG7J/p1781605996724389)

Form draft upload is now inline with the form draft page. Currently there are upload and cancel buttons to the right. The upload button must be pressed for the form to be uploaded. It's easy for a user to miss the upload button and then be confused as to why they have published the same form version again. This is similar to the issue we're already aware of with form attachments and users forgetting to confirm to upload.

We want to replace the upload component to be more like the settings logo/hero image uploads where the file is uploaded immediately and shows a status indicator inline.

sadiqkhoja:
Closes https://github.com/getodk/central/issues/2003

### DIFF
--- a/apps/central/src/components/form/upload.vue
+++ b/apps/central/src/components/form/upload.vue
@@ -95,7 +95,7 @@ definition for an existing form -->
         <template #chooseOne>
           <!-- eslint-disable-next-line vuejs-accessibility/form-control-has-label -->
           <input v-show="false" ref="input" type="file" accept=".xls,.xlsx,.xml"
-            @change="afterInputChange">
+            @change="afterFileSelection($event.target.files[0])">
           <button type="button" class="btn btn-primary"
             :aria-disabled="awaitingResponse" @click="$refs.input.click()">
             <span class="icon-folder-open"></span>{{ $t('dropZone.chooseOne') }}
@@ -106,17 +106,6 @@ definition for an existing form -->
         {{ file != null ? file.name : '' }}
       </div>
     </file-drop-zone>
-    <div v-if="file != null" class="actions">
-      <button type="button" class="btn btn-link" :aria-disabled="awaitingResponse"
-        @click="clear()">
-        {{ $t('action.cancel') }}
-      </button>
-      <button id="upload-button" type="button"
-        class="btn btn-primary" :aria-disabled="awaitingResponse"
-        @click="upload(false)">
-        {{ $t('action.upload') }} <spinner :state="awaitingResponse"/>
-      </button>
-    </div>
   </div>
 </template>
 
@@ -176,10 +165,7 @@ export default {
       this.error = null;
       this.file = file;
       this.warnings = null;
-    },
-    afterInputChange(event) {
-      this.afterFileSelection(event.target.files[0]);
-      this.$refs.input.value = '';
+      this.upload(false);
     },
     upload(ignoreWarnings) {
       // Try to read the file, error means file is modified or deleted or moved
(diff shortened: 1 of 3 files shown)


apps/central/src/components/form/upload.vue
```diff
@@ -172,10 +161,7 @@ export default {
       this.redAlert.hide();
       this.file = file;
       this.warnings = null;
-    },
-    afterInputChange(event) {
-      this.afterFileSelection(event.target.files[0]);
-      this.$refs.input.value = '';
```
sadiqkhoja, line 182: I don't fully understand why we have `this.$refs.input.value = '';`

apps/central/test/components/form/upload.spec.js
```diff
@@ -78,46 +77,26 @@ describe('FormUpload', () => {
...
-        await setFiles(component.get('input'), [xlsForm()]);
-        component.get('#form-upload-filename').text().should.equal('my_form.xlsx');
-      });
-
-      it('resets the input', async () => {
-        const component = mount(FormUpload, mountOptions());
-        await setFiles(component.get('input'), [xlsForm()]);
-        component.get('input').element.value.should.equal('');
-      });
-    });
```
sadiqkhoja, line 113: removed this; related to `this.$refs.input.value = '';`

apps/central/test/components/form/upload.spec.js
```diff
@@ -78,46 +77,26 @@ describe('FormUpload', () => {
...
-      component.find('.actions').exists().should.be.false;
-    });
-
-    it('renders actions after a file is selected', async () => {
-      const component = mount(FormUpload, mountOptions());
-      component.find('.actions').exists().should.be.false;
-      await setFiles(component.get('input'), [xlsForm()]);
-      component.find('.actions').exists().should.be.true;
+      it('shows the filename', () =>
+        mockHttp()
```
sadiqkhoja, line 102: selecting file now makes a request

apps/central/src/components/form/upload.vue
matthew-white: Maybe it's not needed. I think it's there to clear the file from the input. Otherwise the input will hold on to the file reference. I don't totally remember, but I think it may have some user-facing effect. I definitely added it for a reason. What happens if:

1. You select a file
2. The request fails
3. You go to select a different file
    - Is the previous file still selected? (Though maybe that's OK. 🤔)

apps/central/src/components/form/upload.vue
```diff
@@ -103,17 +103,6 @@ definition for an existing form -->
...
     </file-drop-zone>
-    <div v-if="file != null" class="actions">
-      <button type="button" class="btn btn-link" :aria-disabled="awaitingResponse"
-        @click="clear()">
-        {{ $t('action.cancel') }}
-      </button>
-      <button id="upload-button" type="button"
-        class="btn btn-primary" :aria-disabled="awaitingResponse"
-        @click="upload(false)">
-        {{ $t('action.upload') }} <spinner :state="awaitingResponse"/>
```
matthew-white, line 117: I wonder if we should add a spinner somewhere else in case the request takes a long time. Like adding it next to or within the drop zone. We do something like that in the `ConfigLoginFileSelect` component.

matthew-white: Or actually I guess we do show a spinner inside "choose one"

matthew-white (approved):

apps/central/src/components/form/upload.vue
sadiqkhoja: added it back. Still not really important but when you:

- select a file using choose file button
- request fails
- drop a different file (input will still have old value)

PR #1673, opened 2026-06-29
chore(web-forms#820): clean up common package

### LINKED ISSUE #820: Clean up the "common" package
### Description

When web-forms was merged into central-frontend one of the recommendations that I defered was to clean up the /packages/common. Specifically it was felt the name wasn't accurate because the contents were only used by web-forms packages.

In addition, much of common isn't used in more than one package anyway, which will be even more so once https://github.com/getodk/web-forms/issues/807 is complete. For example, demo forms which were only used by the demo app, and scenario fixtures which are only needed for the xforms-engine integration tests.

- [x] Move, inline, or delete code that doesn't belong in common
- [x] If there's anything left, and it's still only intended for web-forms, rename the package to web-forms-common (or similar)

garethbowen:
Closes getodk/web-forms#820

CI
Local dev

The demo app has been moved to a new repo: https://github.com/getodk/web-forms-demo
These files are being moved into a different repo in this PR: https://github.com/getodk/sample-forms/pull/7
This means these files are no longer needed here except for where they are used in tests. Where they were used by xforms-engine integration tests I have moved them into the xforms-engine package. Where they were used by e2e tests I've moved them into the web-forms package.
I have created the bare minimum app for testing purposes, and removed utility files that were doing things like scanning for forms.
This is much simpler and easier to maintain and therefore more appropriate for the now much smaller number of files being managed.

It shouldn't.

No.


garethbowen:
Converted to draft because this should not be merged until after the 2026.2 release.

packages/web-forms/tests/fixtures/all-question-types-v2024091201-3.xml
latin-panda, line 1: I haven't checked the other forms, but this one has additional attachments we'll likely need for testing.

Can you please add them to the /attachments folder? They use different file extensions intentionally. 

Image
Image
Image
Image
Image
Image
Image
Image
Image

packages/web-forms/e2e/app/app.vue
latin-panda, line 1: Is this only used for e2e tests? Maybe it's better to move it inside /e2e for now.

latin-panda (changes requested):
Wow, so many files were deleted! Just a couple of suggestions below.

packages/web-forms/tests/fixtures/all-question-types-v2024091201-3.xml
garethbowen: While the form does reference these files there is no visual or functional test that checks that they load so I left them out intentionally. I'd rather not add files which aren't used or useful.

> ... we'll likely need for testing.

Do you mean they'll be needed in the future? Are you thinking manual or automatic testing?

latin-panda: I mean automated tests. The display of all select appearances is quite complex and took a lot of effort to get right, and we should have e2e tests; we just haven't added them yet. 😣

garethbowen: This would be a screenshot/visual e2e test I assume? I don't know how you would test this functionally...

garethbowen: Ok.. I've added screenshot tests for 4 of the image select types and added them back as attachments. I extracted the screenshot code from the maps control so it can be reused, so it's worth reviewing.

packages/web-forms/e2e/app/app.vue
garethbowen: Done.

latin-panda (approved):
Thank you for adding the new test coverage for select labels! ✨

PR #1674, opened 2026-06-29
Update publishing notes text

lognaturel:
I'm not sure that "publishing notes" would be meaningful to users.

I tried to steer people towards what we want them to use the field for. If we were prompting only on edits, we could say something like "Describe what changed" but given that it's also shown on first publish, that doesn't work. Using the word "version" gives a hint that the text ends up on the Versions tab.

Text-only.

Added to https://github.com/getodk/docs/issues/2054


sadiqkhoja (approved):

lognaturel:
@nicoleorlowski merged so we can see it in action and you can see if further updates are needed.

PR #1675, opened 2026-06-29
Show suggestions in case of entity.error

### LINKED ISSUE #1867: Recommend action when Entity processing fails
Slack thread: https://getodk.slack.com/archives/C0259F2544A/p1777907654703019

When Entity processing fails, it's not clear whether any action can be taken. The error looks like: 

Image

We could add generic guidance such as: "Check the form definition and submission for issues. After making corrections, edit the submission and submit to retry processing."

Thematically linked to:
* https://github.com/getodk/central/issues/699
* https://github.com/getodk/central/issues/688
* https://github.com/getodk/central/issues/1259

matthew-white:
Closes getodk/central#1867.

I updated tests. I also tried it locally.

I'll leave a couple of comments about particular lines.

### DIFF
--- a/apps/central/src/components/submission/feed-entry.vue
+++ b/apps/central/src/components/submission/feed-entry.vue
@@ -96,7 +96,10 @@ except according to the terms contained in the LICENSE file.
     <template #body>
       <markdown-view v-if="comment != null" :raw-markdown="comment"/>
       <template v-else-if="entry.action === 'entity.error'">
-        <div class="entity-error-message">{{ entry.details.problem?.problemDetails?.reason ?? entry.details.errorMessage ?? '' }}</div>
+        <div class="entity-error-message">
+          <p v-if="entityError != null">{{ entityError }}</p>
+          <p>{{ $t('entityErrorInstructions') }}</p>
+        </div>
       </template>
       <template v-else-if="entryDiffs != null">
         <submission-diff-item v-for="(diff, index) in entryDiffs" :key="index" :entry="diff"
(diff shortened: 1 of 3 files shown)


apps/central/src/components/submission/feed-entry.vue
```diff
@@ -96,7 +96,10 @@ except according to the terms contained in the LICENSE file.
     <template #body>
       <markdown-view v-if="comment != null" :raw-markdown="comment"/>
       <template v-else-if="entry.action === 'entity.error'">
-        <div class="entity-error-message">{{ entry.details.problem?.problemDetails?.reason ?? entry.details.errorMessage ?? '' }}</div>
+        <div class="entity-error-message">
+          <p v-if="entityError != null">{{ entityError }}</p>
+          <p>{{ $t('entityErrorInstructions') }}</p>
```
matthew-white, line 101: The error message can potentially be quite long, so I thought it was best to put the generic guidance on its own line/paragraph. That way it's less likely to be missed.

apps/central/src/components/submission/feed-entry.vue
```diff
@@ -96,7 +96,10 @@ except according to the terms contained in the LICENSE file.
     <template #body>
       <markdown-view v-if="comment != null" :raw-markdown="comment"/>
       <template v-else-if="entry.action === 'entity.error'">
-        <div class="entity-error-message">{{ entry.details.problem?.problemDetails?.reason ?? entry.details.errorMessage ?? '' }}</div>
+        <div class="entity-error-message">
+          <p v-if="entityError != null">{{ entityError }}</p>
```
matthew-white, line 100: I don't think it's ever expected that `entityError` will be `null`. Before in that case, we would just show an empty `<div>`. Now we'll skip this first paragraph and just show the generic guidance.

apps/central/src/components/submission/feed-entry.vue
matthew-white: > The error message can potentially be quite long

Especially if there are multiple entities in the submission with errors.

lognaturel (approved):
Thank you! Two paragraphs is a good idea.

PR #1676, opened 2026-06-29
Update What's New modal for v2026.2

### LINKED ISSUE #1847: Update What's New modal for v2026.2
@nicoleorlowski, do we want a What's New modal for v2026.2?

matthew-white:
This PR updates the What's New modal (closes getodk/central#1847):

- Updates the text of the modal (title + body)
- Updates who can see the modal
  - In v2026.1, we restricted Data Collector users from seeing the modal. However, IIRC we actually intended to show the modal to everyone for that release. For v2026.2, I think it's similarly OK to show the modal to everyone.
  - I've refactored the related code to make it easy to limit the modal to a particular audience if we prefer (e.g., just project managers and sitewide administrators for .2).
- Does *not* update the image (**UPDATE:** now it does update the modal)
  - I didn't see that there was a new image yet in Figma, so I'm thinking I'll update that in a follow-up PR. I'll either wait to close the issue until I update the image, or I'll file a follow-up issue about the image. I don't think the image has a big bearing on the risk of regression. I can check the image myself manually/locally.

Updated tests.

I'll add a few comments about particular lines.


apps/central/test/components/account/login.spec.js
```diff
@@ -160,7 +160,11 @@ describe('AccountLogin', () => {
   });
 
   it('shows an info alert if the password is too short', () => {
-    testData.extendedUsers.createPast(1, { email: 'test@email.com', role: 'none' });
+    testData.extendedUsers.createPast(1, {
+      email: 'test@email.com',
+      role: 'none',
+      preferences: { site: { whatsNewDismissed: '2026.2' }, projects: {} }
```
matthew-white, line 166: If the What's New modal is shown, then this toast (the info alert) is not shown, as showing a modal usually hides the toast. I didn't realize that was happening, but I think it's OK behavior in this case. But to keep this test operational, I updated it to mark the user as having seen the What's New modal already. That way they will continue to see the toast.

apps/central/test/components/whats-new.spec.js
```diff
@@ -62,7 +62,7 @@ describe('WhatsNew modal', () => {
   });
 
   describe('does not show modal', () => {
-    it('does not show modal to data collector', async () => {
+    it.skip('does not show modal to data collector', async () => {
```
matthew-white, line 65: I'm thinking that we can `.skip` this test for releases where it's not relevant (where we want to show the modal to everyone).

apps/central/test/components/account/login.spec.js
matthew-white: This change is only needed now because we're showing the modal to everyone, including those with `role: 'none'`.

matthew-white:
Oh I see the image in Figma now, I can add that to this PR.

lognaturel (approved):

lognaturel:
Working on final wording and image in Figma!

lognaturel:
Ready: https://www.figma.com/design/ULpCuTKkIys6Y99YGyMcrF/Central?node-id=0-1&p=f&t=cNJfnF9yOjdmnyBf-0 there's been a small change to copy to make the link to custom properties clear.

lognaturel:
@matthew-white My apologies, I made one more copy change. Image is fine as-is though.

PR #1678, opened 2026-06-29
chore(#1929): improve gh ci build order

### LINKED ISSUE #1929: Reconcile central-frontend github build workflow
Post the web-forms merge the CI build is quite complicated and not as efficient as it could be. The current workflow looks like this...

```mermaid
graph TD;
    static-checks-->e2e-tests;
    unit-tests-->e2e-tests;
    build-->e2e-tests;
    build-wf-->wf-tests;
    wf-tests-->wf-e2e-functional;
    wf-tests-->wf-e2e-visual;
```

The proposed improvement would look more like this...

```mermaid
graph TD;
    static-checks-->unit-tests;
    static-checks-->wf-tests;
    unit-tests-->e2e-tests;
    wf-tests-->wf-e2e-functional;
    wf-tests-->wf-e2e-visual;
```

In this version...
- The `static-checks` now does more with building, linting, typescript compiling, checking bundle size, etc
- That means `build` and `build-wf` can be removed altogether
- Once `static-checks` has passed cache the dist directories so the unit and integration tests don't need to rebuild

garethbowen:
Closes getodk/central#1929

This combines frontend build steps into one tree rather than two, and does more work in the pre-flight check.

Don't merge until...
- [x] 2026.2 is tagged.
- [ ] The required build steps are updated to match the new names

Run CI
Compared execution times
Compared execution steps

The biggest change here is probably loading more work into the static-checks step. This means it takes longer, but this is mostly due to the work being moved up front, and with the caching the overall build is marginally quicker. It also means none of the heavier steps (eg: e2e) are triggered if a static check fails so will mean fewer wasted build steps.

None.

None.


.github/workflows/tests.yml
```diff
@@ -10,15 +10,51 @@ on:
 
 jobs:
   static-checks:
-    timeout-minutes: 2
+    name: 'Static checks'
```
garethbowen: Arguably this is not the right name any more.

.github/workflows/wf-ci.yml
```diff
@@ -242,17 +241,5 @@ jobs:
...
-        run: 'npm run test-browser:${{ matrix.browser }} -w=packages/web-forms || npm run test-browser:${{ matrix.browser }} -w=packages/web-forms'
-
       - if: ${{ matrix.target == 'Web' && matrix.browser != 'webkit' }}
         run: 'npm run test-browser:${{ matrix.browser }} -w=packages/web-forms'
-
-      - if: ${{ matrix.node-version == '24.16.0' && matrix.target == 'Node' }}
-        uses: actions/upload-artifact@v7
-        with:
-          name: dist-demo
-          path: packages/web-forms/dist-demo
```
garethbowen: dist-demo is now being published by the web-forms-demo repo.

.github/workflows/tests.yml
```diff
@@ -10,93 +10,119 @@ on:
...
+    - test-forms-app
     timeout-minutes: 20
     runs-on: ubuntu-latest
     steps:
-
-    # This one weird trick speeds up every build!
-    # see: https://github.com/getodk/central-backend/pull/1642
-    # see: https://github.com/actions/runner/issues/4030
-    - run: sudo apt-get remove --purge man-db
-
```
garethbowen: Not needed any more, right?

.github/workflows/wf-ci.yml
```diff
@@ -242,17 +255,5 @@ jobs:
       - if: ${{ matrix.target == 'Web' }}
         run: 'npx playwright install ${{ matrix.browser }} --with-deps'
 
-      # TODO: hopefully temporary! Attempt to mitigate flakiness in webkit by
-      # running twice when first run fails. At least in local testing, this has
-      # been near (if not at) 100% success.
-      - if: ${{ matrix.target == 'Web' && matrix.browser == 'webkit' }}
-        run: 'npm run test-browser:${{ matrix.browser }} -w=packages/web-forms || npm run test-browser:${{ matrix.browser }} -w=packages/web-forms'
-
```
garethbowen: Experiment to see if this is still needed. So far, so good.

garethbowen:
@alxndrsn Ready for review, but marked as draft because it shouldn't be merged until after the 2026.2 release.

.github/workflows/wf-ci.yml
```diff
@@ -44,24 +45,27 @@ jobs:
...
         with:
           path: |
             node_modules
             **/node_modules
-          key: install-${{ matrix.node-version }}-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/*.yml', 'packages/*/package.json', 'packages/*/package-lock.json') }}
-          fail-on-cache-miss: true
+          key: install-v${{ matrix.node-version }}-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/*.yml', 'packages/*/package.json', 'packages/*/package-lock.json') }}
 
-      - uses: 'actions/cache@v5'
+      - uses: 'actions/cache@v6'
```
alxndrsn: Can we upgrade this version separately, or are there functional changes required related to this PR?

.github/workflows/tests.yml
alxndrsn: Separate PR?

.github/workflows/wf-ci.yml
```diff
@@ -44,24 +45,27 @@ jobs:
...
         id: cache-build
         with:
           path: |
+            dist
             packages/*/dist
-            packages/web-forms/dist-demo
-          key: build-${{ matrix.node-version }}-${{ github.sha }}
+          key: build-${{ github.sha }}
```
alxndrsn, line 61: Should it be obvious why the `key` has changed here?

.github/workflows/wf-ci.yml
```diff
@@ -75,8 +79,9 @@ jobs:
         run: 'npm run test-browser:${{ matrix.browser }} -w packages/common'
 
   xforms-engine:
-    name: '@getodk/xforms-engine'
+    name: 'xforms-engine'
```
alxndrsn, line 81: Are these names changes functional or cosmetic?

.github/workflows/wf-ci.yml
```diff
@@ -215,24 +226,26 @@ jobs:
...
         id: cache-install
         with:
           path: |
             node_modules
             **/node_modules
-          key: install-${{ matrix.node-version }}-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/*.yml', 'packages/*/package.json', 'packages/*/package-lock.json') }}
-          fail-on-cache-miss: true
-
-      - uses: 'actions/cache@v5'
+          key: install-v${{ matrix.node-version }}-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/*.yml', 'packages/*/package.json', 'packages/*/package-lock.json') }}
```
alxndrsn, line 231: Has this key changed?  Or is the diff confused by the change in `actions/cache` version?  Is it just the addition of `v`?

.github/workflows/wf-ci.yml
```diff
@@ -215,24 +226,26 @@ jobs:
...
+            dist
             packages/*/dist
-            packages/web-forms/dist-demo
-          key: build-${{ matrix.node-version }}-${{ github.sha }}
+          key: build-${{ github.sha }}
           fail-on-cache-miss: true
 
+      - name: 'npm ci'
+        if: steps.cache-install.outputs.cache-hit != 'true'
+        run: npm ci
```
alxndrsn, line 243: Neat - I'm surprised how long this takes when there's nothing to update.

.github/workflows/wf-ci.yml
alxndrsn: Can we make the change separately?

If looking to gain confidence that flakiness is resolved, I think the best way of testing flakiness like this is to run an aggressive build matrix.

.github/workflows/wf-ci.yml
```diff
@@ -101,24 +106,26 @@ jobs:
...
+          key: install-v${{ matrix.node-version }}-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/*.yml', 'packages/*/package.json', 'packages/*/package-lock.json') }}
+      - uses: 'actions/cache@v6'
         id: cache-build
         with:
           path: |
+            dist
             packages/*/dist
-            packages/web-forms/dist-demo
-          key: build-${{ matrix.node-version }}-${{ github.sha }}
+          key: build-${{ github.sha }}
```
alxndrsn, line 120: How come this `key` changes?

.github/workflows/tests.yml
```diff
@@ -10,93 +10,119 @@ on:
...
+        path: |
+          dist
+          packages/*/dist
+        key: build-${{ github.sha }}
+        fail-on-cache-miss: true
+    - run: npx playwright install chromium --with-deps
+    - run: npm run test -w=@getodk/forms
+
   e2e-tests:
+    name: 'Test e2e'
```
alxndrsn: I'm not 100% confident with where this sits in the build tree vs how it's named.  Depending on what it's testing, I think it should either:

1. be renamed to make it clear it's specific to central+forms-app, e.g. "Test /apps/* e2e", or
2. be moved to being the final step of the matrix, dependent on all of:
    * test central app
    * test forms app
    * all the web-forms e2e tests

alxndrsn (commented):
Generally looking good.

How has it cut the "unit" tests from >6 minutes to <3?

<3 <3 <3

.github/workflows/wf-ci.yml
garethbowen: This was prompted by building the dist cache in the static-checks step. Because static-checks only runs with the latest node version (no matrix) if this job was run in a matrix (which it soon will be again) then the other node version run would cache miss. Because I've chosen to fail on cache miss this would fail the job. I like failing on cache miss here because otherwise you have to include the build logic here too just in case the cache missed so this is easier to manage.

However this is an interesting topic... Theoretically if we added Node 22 to the matrix then the old build would build the dist _using_ Node 22 and then test it by running in Node 22. This new cache will build the dist in Node 24 and test it by running Node 22. My assertion is that the new way is correct, because the Node 24 dist is what we publish so it's more production like for people including our WF packages dist in their Node 22 apps.

.github/workflows/wf-ci.yml
garethbowen: Mostly cosmetic. The actions build graph truncates these names heavily so if they all start with `@getodk` it's not very helpful.

They are referenced in the "required builds" config in GH so those will have to be updated before this gets merged, but that was going to be the case anyway, given that other steps have changed.

.github/workflows/wf-ci.yml
garethbowen: It should be just the addition of the `v`. The problem arose because the way that static-checks gets the node version (from the package.json) is inconsistent with the way these steps get the node version (from the matrix). This is potentially fragile if the node version in the package.json is updated (ie: to the next minor) without making the identical change to the matrix. If that happens then the cache will always miss and we'll waste some time doing `npm ci` again, like what we do with the other node versions in the matrix.

There is an argument to removing the node_modules cache... it doesn't save us all that much time.

.github/workflows/wf-ci.yml
garethbowen: Well... it does need to actually download all the packages to node_modules. That takes time.

.github/workflows/wf-ci.yml
garethbowen: Covered in https://github.com/getodk/central-frontend/pull/1678#discussion_r3503233246

garethbowen:
> How has it cut the "unit" tests from >6 minutes to <3?

Don't celebrate too soon... it moved those 3 minutes to the static checks. That's basically the build time, which is now done in static checks so that the checks can also check bundle size and types, but also it means we can cache the built result so unit tests just test and don't need to build build.

.github/workflows/tests.yml
garethbowen: Done: https://github.com/getodk/central-frontend/pull/1687

.github/workflows/wf-ci.yml
garethbowen: On it...

.github/workflows/tests.yml
garethbowen: I don't like option 2 because it would significantly slow down the actual build time by chaining the long tasks together.

I've gone with option 1. It feels weird now because it feels like the web-forms e2e tests aren't actually e2e in that they don't use production apps, just a fake one.

Now I'm thinking all frontend e2e tests should be colocated against the whole central stack. They could still be in separate suites so you could run them in different CI jobs and locally.

I'll leave this convo for another day, in part, because it could be that the e2e tests don't belong in frontend anyway!

.github/workflows/wf-ci.yml
garethbowen: Reverted.

garethbowen: Done.

.github/workflows/wf-ci.yml
garethbowen: Reverted.

.github/workflows/wf-ci.yml
alxndrsn: Yeah `node_modules` caching has always seemed to have disappointing results IME - fetching the cache often takes a similar time to rebuilding the dir from `npm` or `yarn` 🤷

.github/workflows/wf-ci.yml
alxndrsn: Surely it doesn't need to do anything when `node_modules` has been restored from cache.

Checking locally with `npm ci >/dev/null 2>&1 && time (npm ci >/dev/null 2>&1)`:

frontend:

```
real	0m8.515s
user	0m11.149s
sys	0m7.595s
```

backend:

```
real	0m6.925s
user	0m8.084s
sys	0m6.527s
```

That `odk-central-backend` also takes this long suggests that it's not even the result of frontend's complex workspaces.

.github/workflows/wf-ci.yml
```diff
@@ -132,8 +139,9 @@ jobs:
           npm run integration-test-browser:${{ matrix.browser }} -w=packages/xforms-engine
 
   xpath:
-    name: '@getodk/xpath'
+    name: 'xpath'
     runs-on: 'ubuntu-latest'
+    timeout-minutes: 2
```
alxndrsn: Was this missing previously?

garethbowen: Yes we didn't have build timeouts in the web-forms packages at all. So occasionally due to some (usually environmental) flake, they would run for an hour or so.

garethbowen: I've set the timeout comfortably high (ie: 50%-100% higher than the actual run) in the hope that it doesn't fail unless something has gone catastrophically wrong.

This can be tweaked over time.

.github/workflows/wf-ci.yml
garethbowen: I'm not sure if there's a question there, so... I think this is resolved?

.github/workflows/wf-ci.yml
garethbowen: I'm ambivalent... what do you think? Keep the cache for a slight speed improvement, or remove it for reliability and maintainability?

alxndrsn: I think orthogonal to this PR, so we don't have to think about it here?

.github/workflows/tests.yml
alxndrsn: Could this be:

```
static-checks -> build -> ...
```

Or rename to `check-and-build`?

.github/workflows/tests.yml
```diff
@@ -10,84 +10,116 @@ on:
...
+        key: build-${{ github.sha }}
+    - name: Upload Release Artefact
+      if: github.ref_type == 'tag'
+      env:
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        tar -cvzf dist-${{ github.ref_name }}.tar.gz dist/
+        echo "Uploading dist bundle..."
+        gh release upload ${{ github.ref_name }} dist-${{ github.ref_name }}.tar.gz
+        echo "Upload complete."
```
alxndrsn, line 53: I feel like I've already asked this, but:

Does this mean that release artefacts are uploaded before tests have passed?

I'd expect uploading of release artefacts to be the ultimate step in the pipeline.

.github/workflows/wf-ci.yml
alxndrsn: :+1: Any chance of a separate PR?  If the change is unrelated to changing the build order.  Also I think adding timeouts could be merged immediately, rather than waiting until after the pending release.

garethbowen: Some (not all) of the build lengths have changed due to the caching changes so it might mean changing it twice, once before this PR and then once more in this PR.

.github/workflows/tests.yml
garethbowen: Yes I would have expected that too, but this is consistent with when the artefact is currently uploaded: https://github.com/getodk/central-frontend/blob/1940cc987f39717507eb6a1662c60d77ad1663f2/.github/workflows/tests.yml#L53

I guess do a git blame and ask the author why it's done then? ;)

I would agree with moving to the final build step perhaps after this PR is merged?

alxndrsn: :laughing: 

But it was done that way originally because:

1. e2e-test outcome is not entirely a signal that this repo is ok (e.g. you could tag a passing `central-frontend` build as a release, but a recent change to `central-backend` repo might break e2e-tests, which would mean your build artefact then fails to generate unexpectedly)
2. someone added web-forms to this repo without considering the impact on the artefact build :¬p

garethbowen: Touché!

+1 to fixing the mistake whosever's it was.

Should I make the change in this PR, in a PR before this one, or a PR after this one?

garethbowen:
@alxndrsn I've made a few changes since you last reviewed...

1. I attempted to do linting before building in a few different ways but the TS linting we do leans itself towards doing a full build first. One easy change that might help is to lint _just_ the frontend app before build, then run a full lint afterwards. With more finesse we could modify the after build lint to just the TS apps/packages but at this point the frontend lint results will be cached so it should run very quickly, and running a full lint gives me more confidence that everything is getting checked, whereas I always worry cherry picking will silently miss something.
2. I updated the wf-publish workflow which was also using the node_modules cache. Firstly because the key needed to be updated to match the new key, but then I decided to remove the cache usage altogether. It runs infrequently so the `npm ci` is quick enough, and it's more maintainable this way.

.github/workflows/tests.yml
alxndrsn: I'd say this PR, but not if it means protracted discussions of where e2e-tests fit in the tree 😁

.github/workflows/wf-ci.yml
alxndrsn: Does that change happen in this PR, or did it change previously in `master`, and is being fixed in this PR?

.github/workflows/wf-publish.yml
```diff
@@ -21,18 +21,8 @@ jobs:
...
-      - name: Cache dependencies
-        uses: actions/cache@v5
-        with:
-          path: |
-            node_modules
-            **/node_modules
-          key: install-24.16.0-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/*.yml', 'packages/*/package.json', 'packages/*/package-lock.json') }}
-          restore-keys: |
-            install-24.16.0-
+          node-version-file: package.json
```
alxndrsn: Does this change relate to matrixes, and build vs runtime versions?

.github/workflows/wf-ci.yml
alxndrsn: :+1: I think that would clarify the intent of the changes, and make it easier to track what's changing.

.github/workflows/tests.yml
```diff
@@ -10,84 +10,116 @@ on:
...
+    name: 'Static checks'
+    timeout-minutes: 5
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v6
+    - uses: actions/setup-node@v6
+      id: setup-node
+      with:
+        node-version-file: package.json
+        cache: 'npm'
```
alxndrsn, line 26: This can be deferred until after `./bin/check-images.sh`.  Should mean:

* faster failures for non-node checks
* more logical grouping with `npm ci`

.github/workflows/tests.yml
```diff
@@ -10,84 +10,116 @@ on:
...
     - run: sudo apt-get install shellcheck
     - run: ./bin/check-scripts.sh
     - run: ./bin/check-for-large-files.sh
     - run: ./bin/check-images.sh
-  unit-tests:
+    - run: npm ci
+    - run: npm run build
+    - run: npm run lint
+    - run: ./bin/check-bundle-size.js
+    - uses: 'actions/cache/save@v6'
```
alxndrsn: Presumably there's no point in doing this directly after `npm ci`, as the cache won't get used if the other tests don't pass...

.github/workflows/tests.yml
```diff
@@ -10,84 +10,116 @@ on:
...
-    - uses: browser-actions/setup-chrome@v2
+    - uses: 'actions/cache@v5'
+      id: cache-install
+      with:
+        path: |
+          node_modules
+          **/node_modules
+        key: install-${{ steps.setup-node.outputs.node-version }}-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/*.yml', 'packages/*/package.json', 'packages/*/package-lock.json') }}
+        fail-on-cache-miss: true
+    - uses: 'actions/cache@v5'
```
alxndrsn, line 76: Is there a reason these aren't explicitly `actions/cache/restore`?

.github/workflows/tests.yml
alxndrsn: Is it ok to mix different versions of `actions/cache`?

alxndrsn (commented):
Trying to understand a bit better what's going on, and answer a few of my queries, I had a play around:

https://github.com/garethbowen/central-frontend/compare/1929-streamline-build...alxndrsn:odk-central-frontend:alx-1929-streamline-build-rebased?expand=1

Perhaps there's something useful there?

I think extracting the cache restores really cleans things up, and makes the diff vs `master` clearer:

https://github.com/getodk/central-frontend/compare/master...alxndrsn:odk-central-frontend:alx-1929-streamline-build?expand=1

Seems like it's going to pass: https://github.com/alxndrsn/odk-central-frontend/actions/runs/28873106295

.github/workflows/tests.yml
garethbowen: > Could this be: static-checks -> build -> ...

Yes. It does slow things down a little to have to trigger another job (spin up, checkout, download cache, etc), so I'd prefer not to. 

> Or rename to check-and-build?

Done.

.github/workflows/wf-ci.yml
garethbowen: It changed previously and is being fixed. I've pulled this change out into a separate PR.

.github/workflows/wf-ci.yml
garethbowen: Ok. Removed.

.github/workflows/tests.yml
garethbowen: Happy to do it as part of this PR. I don't think we need to move the e2e-tests. If I just create a new step that depends on e2e-tests AND the web-forms e2e tests does that work for you?

.github/workflows/wf-publish.yml
garethbowen: Nope... It was changed originally to keep the cache key consistent because it was missing 100% before, but now that I decided to remove the cache usage altogether it's not necessary. Removed.

.github/workflows/tests.yml
garethbowen: Done.

garethbowen:
> Trying to understand a bit better what's going on, and answer a few of my queries, I had a play around:

Yes that's nice, though should be a separate PR.

.github/workflows/tests.yml
garethbowen: Technically it could be used because caches aren't just used for this instance of the build. But publishing the cache when the lint fails seems like a bad idea.

.github/workflows/tests.yml
garethbowen: I didn't want to change anything that wasn't strictly necessary.

.github/workflows/tests.yml
garethbowen: I mean, I'm sure it's not recommended. They used to be consistent. I've reverted this to v5 for now.

.github/workflows/tests.yml
alxndrsn: > If I just create a new step that depends on e2e-tests AND the web-forms e2e tests does that work for you?

No, this is the discussion I want to avoid :grin:  `e2e-tests` is non-deterministic, especially if building older branches... which would be necessary for any point/patch release of an older `central-frontend` version.

Let's keep the current location in this PR, and discuss moving it separately.

.github/workflows/tests.yml
alxndrsn: This is a new job, so I don't think it would be changing anything?  I'd think the only thing that should be updating the cache is the `check-and-build` job, and allowing any other job to do it is a risk.  Or am I over-thinking it?

.github/workflows/wf-ci.yml
alxndrsn: > They are referenced in the "required builds" config in GH so those will have to be updated before this gets merged

I'm surprised that the `name` prop is used in preference to the job key/ID!  I'd prefer the simplicity & consistency of removing the `name` fields altogether, but not a big deal.

alxndrsn (approved):
If you're happy with `actions/cache` vs `actions/cache/restore` use, then I think this is good to merge.

.github/workflows/tests.yml
garethbowen: The existing jobs all use `cache` and I like the consistency. Maybe a subsequent PR would be to switch them all over to restore.

garethbowen:
@lognaturel No hurry... this is ready to merge but we need to update the GH "expected checks". Can you update and then I'll merge?

lognaturel:
I have updated it it to 
Screenshot 2026-08-10 at 9 29 08 AM

Of course this means that the change is active now and code not yet using the code in this PR will have weird behavior!

I think I'm missing the WF package checks. When I try to add those I see:

Screenshot 2026-08-10 at 9 30 32 AM

Do I need to add each of those variants manually? Is there a way to make it so we could just add "Test web-forms packages" once?

garethbowen:
> Do I need to add each of those variants manually? Is there a way to make it so we could just add "Test web-forms packages" once?

Discussed in Slack - we decided to require the "test web-forms e2e" steps, which only run if the package steps pass.

PR #1680, opened 2026-06-29
transifex:destructure: don't warn for variable enclosed in guillemets

matthew-white:
It's very common in Transifex for translators not to add spaces around variables. destructure.js logs a warning in that case so that the developer knows about it and can choose to correct it in Transifex. We're about to have our first translations with guillemets («these»), but right now the destructure.js warning doesn't understand them. It logs a warning when it shouldn't:

```
destructuring fr
warning: component.FieldKeyEdit.title: variable without separator.
warning: component.PublicLinkEdit.title: variable without separator.
```

This PR adds guillements to the punctuation regex in destructure.js.

When I make this change, I no longer see the warning.

I followed the existing pattern of listing punctuation characters in a regex.

### DIFF
--- a/apps/central/bin/util/transifex.js
+++ b/apps/central/bin/util/transifex.js
@@ -998,7 +998,7 @@ const validateTranslation = (locale) => ({ source, translated, path }) => {
       logThenThrow({ source, translated }, 'unexpected linked locale message');
 
     if (locales[locale].warnVariableSeparator) {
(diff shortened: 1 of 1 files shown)


lognaturel (approved):
Thank you for being supportive of guillemets! I'm very surprised we haven't used them before and will check in with some of our francophone collaborators.

PR #1681, opened 2026-06-29
Update translations before v2026.2 regression testing

matthew-white:
This PR makes progress on getodk/central#1843, updating translations to the latest prior to regression testing.

Only translations should have changed, nothing else.

Tests continue to pass. We're also about to undergo regression testing.


matthew-white:
The order of a bunch of translations/keys changed due to a change I made to destructure.js to fix a separate issue. I'm looking into that now.

**UPDATE:** I was able to revert the order except in one case (see below). I'll file an issue about this soon, then create a PR from my local change to destructure.js.

apps/central/src/locales/de.json
```diff
@@ -172,14 +172,12 @@
...
     "hidePreview": "Vorschau ausblenden",
-    "newPreview": "Neue Vorschau",
     "expand": "Vergrössern",
     "collapse": "Verkleinern",
     "selectRow": "Zeile auswählen",
     "reset": "Zurücksetzen",
-    "reviewParallelUpdates": "Parallele Updates überprüfen",
-    "tryAgain": "Nochmals versuchen",
     "upload": "Hochladen",
+    "reviewParallelUpdates": "Parallele Updates überprüfen",
```
matthew-white, line 180: I made a change to destructure.js that reorders `upload` before `reviewParallelUpdates`. I'll file an issue with more details about that. Only the order should have changed, not the content of the translation.

matthew-white (commented):
Comment about translation order to facilitate async code review

lognaturel (approved):
Only translations have been updated 🎉

PR #1684, opened 2026-06-30
fix(web-forms#673): Display lat and lon truncated to 5 decimals

### LINKED ISSUE #673: Latitude in Advanced settings adding more decimals

latin-panda:
Closes https://github.com/getodk/web-forms/issues/673

Tested with the All question form and [2-all-possible-notes.xml](https://github.com/user-attachments/files/29513791/2-all-possible-notes.xml)

This [focuses on fixing the display only](https://github.com/getodk/web-forms/issues/673#issuecomment-3974473326) except for inputs, since they change the value and the value should be preserved.

Less visual noise 

No


latin-panda:
@garethbowen this is a draft because of code freezing but it's ready for review

packages/web-forms/src/components/common/GeolocationFormattedValue.vue
```diff
@@ -9,10 +9,12 @@ import type {
 	GeotraceNoteValue,
 	GeoshapeNoteValue,
 } from '@getodk/xforms-engine';
+import { truncateDecimals } from '@/lib/format/truncate-decimals.ts';
 import { TRANSLATE } from '@/lib/constants/injection-keys.ts';
 import type { Translate } from '@/lib/locale/useLocale.ts';
 import { computed, inject } from 'vue';
-import { truncateDecimals } from '@/lib/format/truncate-decimals.ts';
+
+const COORDINATE_DECIMALS = 5;
```
garethbowen: Having this const defined in two places has a little smell to it. I think the ideal solution would be to have a util class with `formatLatitude` or similar, which means there's no risk of the output diverging in future.

This could be done as a function on `truncate-decimals.ts`, or alternatively a custom number format in vue-i18n.

garethbowen (approved):
Nice!

One suggested code improvement inline, but approved in case you disagree! :)

packages/web-forms/src/components/common/GeolocationFormattedValue.vue
latin-panda: Fixed! Thanks

garethbowen (approved):
Thanks!

Because this isn't urgent I'll leave it for you to merge when you get back.

PR #1686, opened 2026-06-30
fix(web-forms#842): redirect to login for auth error on non-public forms

### LINKED ISSUE #842: When a web form link is requested without auth, should redirect to Central login
Identified at https://github.com/getodk/web-forms/issues/837#issuecomment-4844358055

It can be handy to save a bookmark or link directly to a form that requires auth. Originally we had `/-/<id>` links that could be bookmarked and always redirected to login if no auth was provided. When we introduced `/f/<id>` links, those did not redirect to login. Recently we have made it so that `/-/<id>` no longer redirects to login.

The `/-/<id>` structure is most likely to be still linked to and saved as bookmarks so I think we should make sure that this structure redirects to login. Ideally `/f/<id>` would as well.

As of Central frontend https://github.com/getodk/central-frontend/commit/9512dd8db8b418df7b450cf805d9250ae110f10c:
- ✅ Going to https://staging.getodk.cloud/projects/232/forms/pack_parcels/submissions/new without a valid session redirects to login
- ❌ Going to https://staging.getodk.cloud/f/foDO89bGf9mbQhErdwh4jSWvQL4ECr3 without a valid session does NOT redirect to login
- ❌ Going to https://staging.getodk.cloud/-/foDO89bGf9mbQhErdwh4jSWvQL4ECr3 without a valid session does NOT redirect to login

In Central v2026.1:
- ✅ Going to https://production.getodk.cloud/projects/3/forms/pack_parcels/submissions/new without a valid session redirects to login
- ❌ Going to https://production.getodk.cloud/f/LfMy1dprcpInBd2umqR3rhlGghe2vJE without a valid session does NOT redirect to login
- ✅ Going to https://production.getodk.cloud/-/LfMy1dprcpInBd2umqR3rhlGghe2vJE without a valid session does redirect to login

garethbowen:
Closes getodk/web-forms#842

Manual and e2e testing

Requests without `st` aren't public forms so a session is required, so treat these requests differently than ones with `st`.

This change should only impact on requests using enketoId without an `st` token, so the risks are low, and now well tested.

No

### DIFF
--- a/.changeset/orange-nails-cheat.md
+++ b/.changeset/orange-nails-cheat.md
@@ -1,0 +1,5 @@
+---
+"@getodk/forms": patch
+---
+
+Redirect to login page when auth issues for non-public form
--- a/apps/forms/src/components/submission.vue
+++ b/apps/forms/src/components/submission.vue
@@ -114,7 +114,7 @@ const fetchFormConfig = async (): Promise<Form> => {
     try {
       return await getFormByEnketoId(enketoId.value, st.value);
     } catch(e) {
-      if (e instanceof RequestError && (e.statusCode === 401.2 || e.statusCode === 403.1)) {
+      if (st.value && e instanceof RequestError && (e.statusCode === 401.2 || e.statusCode === 403.1)) {
         // a public form with an invalid st or revoked enketoId should show as form not found
         throw new RequestError('Form not found', 404);
       }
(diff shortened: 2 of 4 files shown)


lognaturel (approved):
Hurray for a small change and new tests! 🎉

PR #1687, opened 2026-07-01
chore: remove the build performance hack which is no longer needed

garethbowen:

CI

?

none

none

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -91,12 +91,6 @@ jobs:
     timeout-minutes: 20
     runs-on: ubuntu-latest
     steps:
-
-    # This one weird trick speeds up every build!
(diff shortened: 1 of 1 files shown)


alxndrsn:
How do we know it's no longer needed?

garethbowen:
Going off this comment: https://github.com/actions/runner/issues/4030#issuecomment-3551022523

Then tracing the build log we see:

```
2026-07-01T05:27:58.6369740Z Processing triggers for man-db (2.12.0-4build2) ...
2026-07-01T05:27:58.6393305Z Not building database; man-db/auto-update is not 'true'.
```

It's not building the db based on some upstream setting. As long as they don't regress we don't need to also do this trick.

alxndrsn (approved):

PR #1688, opened 2026-07-01
chore: remove duplicated test run

garethbowen:

The initial commit runs the build 30 times in serial - if any of them fail the whole build would fail. It didn't fail. The subsequent change removes the last 29 runs leaving just one.

### DIFF
--- a/.github/workflows/wf-ci.yml
+++ b/.github/workflows/wf-ci.yml
@@ -243,37 +243,7 @@ jobs:
         run: 'npx playwright install ${{ matrix.browser }} --with-deps'
 
       - if: ${{ matrix.target == 'Web' && matrix.browser == 'webkit' }}
-        run: |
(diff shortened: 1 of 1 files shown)


alxndrsn:
> The initial commit runs the build 30 times in serial - if any of them fail the whole build would fail. It didn't fail. The subsequent change removes the last 29 runs leaving just one.

I'm struggling to find the relevant build jobs; can you share links?

<kbd>
Screenshot_2026-07-01_09-26-15
</kbd>

garethbowen:
In my personal fork: https://github.com/garethbowen/central-frontend/actions/runs/28495603472/job/84461298842

I just duplicated the run command so it's all stacked up in the 15 minute long test job.

garethbowen:
NB: I also realised last night that I now have two branches which do the same thing... so I've merged the branches for simplicity.

alxndrsn (approved):

PR #1689, opened 2026-07-01
fix(web-forms#837): prevent layout shift when scrollbar appears

### LINKED ISSUE #837: Make scrollbar gutter stable to avoid page jitter when scrollbar appears
### Description
Currently when a form loads with very little relevant content and then more content becomes visible the form contents shift horizontally.

Firefox 150.0.3

[scroll_gutter.xlsx](https://github.com/user-attachments/files/29517362/scroll_gutter.xlsx)

https://github.com/user-attachments/assets/5b4543ee-b9ae-4c1e-9beb-0c604a9edcf5

`scrollbar-gutter: stable;` fixes it.

---

### Checklist
  - [ ] No
  - [ ] No
  - [ ] No
  - [ ] No
---

### User Stories

- **As a** [user type],
  **I want** [functionality],
  **so that** [benefit].

---

### Design

- **Figma**: [Insert URL]

---

### Dependencies

-

---

### Additional Notes

-

---

### Related Issues

-

latin-panda:
Closes https://github.com/getodk/web-forms/issues/837

The macOS overlay scrollbars don't take layout space, so no visual change there, but Windows/Linux consistently reserve about 15px, so this scroll setting prevents the shift.

Smooth experience when content becomes relevant.


latin-panda:
@lognaturel This is still a draft because of the current code freeze. If you have a chance, could you please try this fix?

latin-panda:
@garethbowen When you get a chance, could you try this fix? I wonder about the display in Linux

garethbowen (changes requested):
On my browsers and Ubuntu in general scrollbars are hidden and overlay the content so this isn't an issue with the default settings.

I changed my settings to always show the scrollbar so I could reproduce it and the fix does work.

However, and I suspect this is an issue with other OSes too, now I get a permanent white bar down the right hand side of the page. This is because the background color is only set on the .odk-form element. The quick fix is to set the background color on the html in the reset.scss however it's not very friendly for a vue component to change css on the html and body elements. Arguably it's not friendly to set the scroll gutter either...

I think the correct fix is to move most if not all of the reset.scss into the forms app. We no longer need to reset anything from central, so potentially a lot can be deleted. Then we have full control over scroll bars, html background colors, and everything else without messing with any other apps that include the web-forms vue component.

latin-panda:
@garethbowen thank you! I've fixed the blank bar.
I've created a [separate issue for reset.scss](https://github.com/getodk/web-forms/issues/853) mainly because it requires more testing effort.

garethbowen (approved):
Looks great. On mobile it does nothing because scrollbars aren't shown. On linux with scrollbars off (default) it does nothing. On linux with scrollbars on the box is slightly off center because of the extra right padding, but it's not very noticeable, and the only fix would be to use `both-edges` which would add padding to both sides. I don't think it's worth it for a very small visual change on a small set of devices.

PR #1690, opened 2026-07-02
change the textcolor of error message in FormUpload

sadiqkhoja:
image

PR #1691, opened 2026-07-02
Update styling for entity.error

matthew-white:
This PR addresses the QA comments at https://github.com/getodk/central/issues/1867#issuecomment-4832277812:

1. Remove the `max-width` on `<p>` when rendering an `entity.error` in the submission activity feed
2. Change the background color

Re: background color, I'm actually not 100% sure what works best there. Nicole mentioned "Reusing the styling from the entity upload designs", but looking at the Figma file, I'm not totally sure which background color she's referring to. Part of the Figma file does show errors with a red background color, so I ran with that. I also made the text color red to match what we did in #1690. Here's what it looks like:

I tried it out locally. We don't usually test styles.

### DIFF
--- a/apps/central/src/components/submission/feed-entry.vue
+++ b/apps/central/src/components/submission/feed-entry.vue
@@ -218,7 +218,15 @@ export default {
 
 .submission-feed-entry {
   .entity-error-message{
-    margin: 10px;
+    background-color: $color-danger-light;
+    color: $color-danger;
+    padding: 10px;
+
+    p {
+      max-width: none;
+
+      &:last-child { margin-bottom: 0; }
+    }
   }
 
   .icon-cloud-upload, .icon-comment, .icon-trash, .icon-recycle, .icon-clock-o {


apps/central/src/components/submission/feed-entry.vue
```diff
@@ -218,7 +218,15 @@ export default {
 
 .submission-feed-entry {
   .entity-error-message{
-    margin: 10px;
+    background-color: $color-danger-light;
+    color: $color-danger;
+    padding: 10px;
```
matthew-white, line 223: Replaced `margin` with `padding` so the background color is shown edge-to-edge in the element.

apps/central/src/components/submission/feed-entry.vue
```diff
@@ -218,7 +218,15 @@ export default {
...
+
+    p {
+      max-width: none;
+
+      &:last-child { margin-bottom: 0; }
```
matthew-white, line 228: After replacing `margin` with `padding`, there's no more margin collapse, so I needed to remove the margin on the second/last `<p>`. I could have put negative margin on the `<div>` instead, but this feels a little cleaner to me.

matthew-white (commented):
Notes to facilitate async code review

matthew-white:
Just looking for a single checkmark for the PR, but I tagged both @sadiqkhoja and @lognaturel in case either has time for a quick look.

lognaturel (approved):
Looks eye-catching and attractive, thank you!

PR #1694, opened 2026-07-03
Bump version for v2026.2 release

matthew-white:
I ran `npm run version`. As described in getodk/central#2037: "This consumes the `.changeset/` files, bumps package versions, and updates each `CHANGELOG.md`."

Not a code change, but we'll check CI.

I'm planning to make additional changes to the Central changelog. I can do that either in a follow-up commit or a follow-up PR.

### DIFF
--- a/apps/central/CHANGELOG.md
+++ b/apps/central/CHANGELOG.md
@@ -1,6 +1,6 @@
 # @getodk/central-frontend
 
-## 0.2.0
+## 2026.2.0
 
 ### Minor Changes
 
--- a/apps/central/CHANGELOG.md
+++ b/apps/central/CHANGELOG.md
@@ -1,0 +1,19 @@
+# @getodk/central-frontend
+
+## 0.2.0
+
+### Minor Changes
+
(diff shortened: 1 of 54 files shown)


apps/central/CHANGELOG.md
```diff
@@ -0,0 +1,19 @@
+# @getodk/central-frontend
+
+## 0.2.0
```
matthew-white: I should make this v2026.2, right?

apps/central/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@getodk/central-frontend",
-  "version": "0.1.0",
+  "version": "0.2.0",
```
matthew-white: Are we supposed to commit this? Or should I revert?

matthew-white (commented):
Questions about version numbers

apps/central/CHANGELOG.md
lognaturel: Yes, that would make sense.

lognaturel (approved):

PR #1697, opened 2026-07-05
Date ranges: use half-open interval

### LINKED ISSUE #1323: Time range (mis-)expression
Related: [Slack discussion](https://getodk.slack.com/archives/C01TKJ7TG7J/p1753008465407679)

In the submission table view date filter, and perhaps other places, time ranges meant to express a particular day or day-range, don't. In the OData filter they miss the last millisecond of the day, because instead of, say, `lt 2025-07-21T00:00:00+05:30`
they use `lte 2025-07-20T23:59:59.999+05:30` which looks like an attempt to express the last instant of a day (but you fundamentally can't express the last instance of the day as there are always infinitely many later ones still, even though the storage system might not be able to encode them. But it must be mentioned that in this case the storage system *can* encode later instants than `23:59:59.999` — PostgreSQL has nanosecond timestamps for the `timestamptz` type).

The less problematic (and widely used!) way of expressing an interval is start-inclusive, end-exclusive.

alxndrsn:
Time is continuous.

Closes getodk/central#1323

Updated tests.

No other approaches considered.

Risk that it would break filters.

`lt` is defined in the odata docs: https://learn.microsoft.com/en-us/azure/search/search-query-odata-comparison-operators (I'm not sure if these are the canonical docs).

Related tests can be seen in `odk-central-backend` at:

* https://github.com/getodk/central-backend/blob/bee93414a2215bd73d7f371113c6640c66d90c95/test/integration/api/geodata.js#L644
* https://github.com/getodk/central-backend/blob/bee93414a2215bd73d7f371113c6640c66d90c95/test/integration/api/geodata.js#L656
* https://github.com/getodk/central-backend/blob/bee93414a2215bd73d7f371113c6640c66d90c95/test/integration/api/geodata.js#L759
* https://github.com/getodk/central-backend/blob/bee93414a2215bd73d7f371113c6640c66d90c95/test/integration/api/geodata.js#L771
* https://github.com/getodk/central-backend/blob/bee93414a2215bd73d7f371113c6640c66d90c95/test/integration/api/odata.js#L1353
* https://github.com/getodk/central-backend/blob/bee93414a2215bd73d7f371113c6640c66d90c95/test/unit/data/odata-filter.js#L25
* https://github.com/getodk/central-backend/blob/bee93414a2215bd73d7f371113c6640c66d90c95/test/unit/data/odata-filter.js#L50
* https://github.com/getodk/central-backend/blob/bee93414a2215bd73d7f371113c6640c66d90c95/test/unit/data/odata-filter.js#L184

No.


matthew-white:
The changes here look reasonable to me. :+1:

As I mentioned at https://github.com/getodk/central/issues/1323#issuecomment-4900422195, I feel like it'd be slightly awkward to make this change in `SubmissionList` and `EntityList`, but not `AuditList`. The same questions would remain for `AuditList` around time being continuous, plus we'd have an inconsistent approach in the codebase.

That said, I'm happy to get this merged if it feels like progress. Do you generally agree with Wicher's perspective that it's confusing to use the last millisecond of the day instead of the first millisecond of the next day?

alxndrsn:
> Do you generally agree with Wicher's perspective that it's confusing to use the last millisecond of the day instead of the first millisecond of the next day?

Yes - I agree that:

1. on paper it's incorrect, even if in practice it works
2. **1** makes for a confusing situation!

I think fixing in `SubmissionList` and `EntityList` is a good start.  If `AuditList` has the same problem, but implemented in a different way, I'm happy for that to be fixed separately.

.github/workflows/tests.yml
matthew-white, line 1: I think the removals here can be reverted.

matthew-white (approved):
Happy to get this merged!

.github/workflows/tests.yml
alxndrsn: yes, thanks!

PR #1699, opened 2026-07-07
fix(web-forms#846): setvalue updates in an unexpected order

### LINKED ISSUE #846: Order of execution of setvalue actions is unexpected
### Expected Behaviour

When multiple setvalue actions are defined on the same element I would expect them to execute in the order they are defined. This gives the form designer control over the order of execution. This matters in some cases where the setvalue value calculation from one of the actions depends on the value from another action being set.

### Observed Behaviour

setvalue actions are executed in a seemingly random but deterministic order, that is, before loading the xform it's not clear which order they will be performed in, but it is the same every time.

### Reproduction Steps

Form (XML or XLSX file):

```xml
    <h:head>
        <h:title>Multiple set value events on the same element</h:title>
        <model>
            <instance>
                <data id="multiple-events">
                    <meta>
                        <instanceID/>
                    </meta>
                    <my-value-1>1</my-value-1>
                    <my-value-2>1</my-value-2>
                    <my-value-3>1</my-value-3>
                    <my-value-4>1</my-value-4>
                    <my-value-5>1</my-value-5>
                </data>
            </instance>

            <bind nodeset="/data/my-value-1" type="integer" />
            <bind nodeset="/data/my-value-2" type="integer" />
            <bind nodeset="/data/my-value-3" type="integer" />
            <bind nodeset="/data/my-value-4" type="integer" />
            <bind nodeset="/data/my-value-5" type="integer" />

        </model>
    </h:head>
    <h:body>
        <input ref="/data/my-value-1">
            <setvalue event="xforms-value-changed" ref="/data/my-value-2" value="/data/my-value-1 + 1" />
            <setvalue event="xforms-value-changed" ref="/data/my-value-3" value="/data/my-value-2 + 1" />
            <setvalue event="xforms-value-changed" ref="/data/my-value-4" value="/data/my-value-3 + 1" />
            <setvalue event="xforms-value-changed" ref="/data/my-value-5" value="/data/my-value-4 + 1" />
        </input>
        <input ref="/data/my-value-2" />
        <input ref="/data/my-value-3" />
        <input ref="/data/my-value-4" />
        <input ref="/data/my-value-5" />
    </h:body>
</h:html>
```

Steps:
1. Load the above form
2. Change the value of the first input
3. The other values should be updated to 2, 3, 4, 5 respectively

### Screenshots or Videos

https://github.com/user-attachments/assets/659d8b84-4b14-4003-9e6e-dd7d53053c30

### Environment

- Browser version: Any
- OS: Any
- Web Forms version: 1.0.0

### Additional Context

Initially raised via a support request so this is a real world issue.

garethbowen:
Closes getodk/web-forms#846

Added integration test, and manually testing with the user provided form.

This enforces the order of execution is the same as the order the elements are defined in the xml. This means the order ultimately in the form designer's control.

Another option was to try and figure out which values were dependant on other fields to make sure the events are executed in the correct order. This was technically more challenging.

Forms will now behave consistently with Collect and in a more intuitive order.

No.


garethbowen:
@latin-panda This is ready for review. No hurry.

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -235,41 +236,58 @@ const resolveAndSetValueChanged = (
...
+    return;
+  }
+  const listener: ValueChangedEventListener = { context, ref, action, setRelevantValue };
+  const listeners = context.rootDocument.getValueChangedEventListeners();
+  if (listeners.has(source)) {
+    listeners.get(source)!.push(listener);
+  } else {
+    listeners.set(source, [listener]);
+    registerValueChangedListener(context, source);
+  }
```
latin-panda: Where does it remove the listener from the map?

I'm thinking of the case where setValue is inside a repeat and the repeat instance is deleted. In that case, it needs to free that piece of memory.  This is more noticeable in specific cases; I recently saw a client form for parcels with a lookup field that hides and shows tons of repeats, super slow.

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -235,41 +236,58 @@ const resolveAndSetValueChanged = (
...
-  context: ValueContext,
-  setRelevantValue: SimpleAtomicStateSetter<string>,
-  action: ActionDefinition
-): void => {
-  const { source, ref } = bindToRepeatInstance(context, action);
-  if (!source) {
-    // No element to listen to
-    return;
+const updateValueChangedRefs = (listeners: ValueChangedEventListener[]) => {
+  for (const listener of listeners) {
```
latin-panda: Tiny:
This can use `listeners.forEach(({context, ref, action, setRelevantValue}) => {  `

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -235,41 +236,58 @@ const resolveAndSetValueChanged = (
...
+    // No element to listen to
+    return;
+  }
+  const listener: ValueChangedEventListener = { context, ref, action, setRelevantValue };
+  const listeners = context.rootDocument.getValueChangedEventListeners();
+  if (listeners.has(source)) {
+    listeners.get(source)!.push(listener);
+  } else {
+    listeners.set(source, [listener]);
+    registerValueChangedListener(context, source);
```
latin-panda: Could you add a test for the case where multiple setvalues share a source but one target is inside a repeat that gets deleted? 
```
<input ref="/data/trigger">
  <setvalue event="xforms-value-changed" ref="/data/rep[1]/y"   value="/data/trigger"/> <-- registers first
  <setvalue event="xforms-value-changed" ref="/data/survivor"   value="/data/trigger"/> <-- registers second
</input>
<repeat nodeset="/data/rep">
  <input ref="/data/rep/y"/>
</repeat>
```

To confirm that `<setvalue event="xforms-value-changed" ref="/data/survivor"   value="/data/trigger"/>` still updates. 

```
<data id="xyz">
  <rep>  <-- registered first bc walked in document order
    <y/>
  </rep>
  <survivor/>  <-- registered second & appends to same source's listener list??
  <trigger>0</trigger>
</data>
``

**Review by latin-panda — CHANGES_REQUESTED** [review]
This requires some testing. Once approved, it can be forwarded to QA to catch any issues I might have missed.

File: packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts [hunk 4]
```diff
@@ -235,41 +236,58 @@ const resolveAndSetValueChanged = (
...
+  for (const listener of listeners) {
+    const { context, ref, action, setRelevantValue } = listener;
+    if (referencesCurrentNode(context, ref)) {
+      // Only update if value has changed
+      if (action.element.nodeName === SET_GEOPOINT_LOCAL_NAME) {
+        getGeopointValue(context, (point) => {
+          setRelevantValue(point);
+        });
+      } else {
+        resolveAndSetValueChanged(context, setRelevantValue, action.computation.expression);
```
latin-panda: In this [forum post](https://forum.getodk.org/t/collect-2026-3-0-default-image-inline-preview-missing-take-choose-disabled-crash-on-null-annotation/58380/4?u=jenniferq), the user is trying to use `setvalue` to build a media URL and pass it to `annotate` so the image loads as the default value.

Form:
[default-image-crash.xml](https://github.com/user-attachments/files/30239706/default-image-crash.xml)

We have tested default values in `annotate`, but it would be good to test this case where it uses `setvalue` to build a media URL and check if it fetches the image.

garethbowen:
@latin-panda Unfortunately this required a significant rewrite... but I think it also fixed other bugs around repeats and actions, so could be a win?

The 1.0.0 implementation used the setValue of the destination node which is nice. But I think it was impossible to clean up reactive signals when repeats were removed.

This flips the script, so we use the getValue of the source element and then find the destination node and set the value, which required a new function to be added. This means when the repeat element is cleaned up the listeners get automatically cleaned up too.

There are drawbacks to this approach too, for example, I'm not over the moon about adding the new set function. But the overall approach feels more deterministic than the 1.0.0 version and cleaner than the previous version in this PR. Let me know what you think!

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
garethbowen: I haven't looked into this yet - I'll have a look tomorrow.

latin-panda:
Update: I haven't finished reviewing. I'm still testing cases and thinking about whether we're missing something. I'll continue tomorrow :)

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
garethbowen: I think this is a separate issue so I've split it out here: https://github.com/getodk/web-forms/issues/858

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -311,8 +260,46 @@ const dispatchAction = (
       performActionComputation(context, setValue, action);
     }
   }
-  if (action.events.includes(XFORM_EVENT.xformsValueChanged)) {
-    createValueChangedCalculation(context, setValue, action);
+};
+
+const registerValueChangedActions = (context: ValueContext, getValue: Accessor<string>) => {
+  if (context.valueChangedActions?.length) {
+    for (const action of context.valueChangedActions) {
```
latin-panda: Looking to reduce the complexity of this function to make it easier to read and understand. 

```suggestion
context.valueChangedActions?.forEach((action) -> {
```

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -311,8 +260,46 @@ const dispatchAction = (
...
+      let previous = getValue();
+
+      createComputed(() => {
+        const ref = bindRefToRepeatInstance(context, action.ref);
+        const destinationNodes = context.evaluator.evaluateNodes(ref, {
+          contextNode: context.contextNode,
+        });
+        if (destinationNodes.length) {
```
latin-panda:
```suggestion
        if (!destinationNodes.length) {
          return;
       }
```

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -311,8 +260,46 @@ const dispatchAction = (
...
+          const destinationNode = destinationNodes[0];
+          if (
+            (destinationNode instanceof ValueNode || destinationNode instanceof Attribute) &&
+            destinationNode.isAttached() &&
+            context.isAttached()
+          ) {
+            const sourceValue = getValue();
+
+            if (sourceValue && sourceValue !== previous) {
+              if (referencesCurrentNode(destinationNode, ref)) {
```
latin-panda:
```suggestion
            if (sourceValue && sourceValue !== previous && referencesCurrentNode(destinationNode, ref)) {
```

latin-panda: Here is a good place to fix the empty string issue I posted below

packages/xforms-engine/test/integration/actions-events/set-value-action.test.ts
```diff
@@ -863,6 +1013,23 @@ describe('setvalue action', () => {
...
+          ),
+          body(input('/data/element', setvalue('xforms-value-changed', '/data/element/@attr', '7')))
+        )
+      );
+      expect(scenario.attributeOf('/data/element', 'attr').getValue()).toBe('');
+      scenario.next('/data/element');
+      scenario.answer('trigger');
+      expect(scenario.attributeOf('/data/element', 'attr').getValue()).toBe('7');
+    });
   });
```
latin-panda, line 1064: There is a regression when triggering the action to empty the input. I've verified it with this test: 

```js
  it('is triggered when the input is cleared', async () => {
    const scenario = await Scenario.init(
      'xforms-value-changed on clear',
      html(
        head(
          title('Value changed on clear'),
          model(
            mainInstance(t('data id="cleared-source"', t('source'), t('counter', '0'))),
            bind('/data/counter').type('int')
          )
        ),
        body(
          input('/data/source',
            setvalue('xforms-value-changed', '/data/counter', '/data/counter + 1')
          )
        )
      )
    );

    expect(scenario.answerOf('/data/counter')).toEqualAnswer(intAnswer(0));

    scenario.answer('/data/source', 'abc');
    expect(scenario.answerOf('/data/counter')).toEqualAnswer(intAnswer(1));

    scenario.answer('/data/source', '');
    expect(scenario.answerOf('/data/source').getValue()).toBe('');

    expect(scenario.answerOf('/data/counter')).toEqualAnswer(intAnswer(2));
  });
```

latin-panda (changes requested):
I finished testing. I found one regression and added small suggestions to make the code flatter

packages/xforms-engine/test/integration/actions-events/set-value-action.test.ts
garethbowen: Thanks! Fixed, and test added.

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
garethbowen: Done.

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
garethbowen: Done.

garethbowen:
@latin-panda I've fixed that regression (thanks!) and done the refactoring you suggested. I've also done a couple of other small refactors which have been bothering me. Let me know what you think!

latin-panda (approved):
Nice improvements! After retesting, I don't have any other concerns, and it's ready to merge!

PR #1702, opened 2026-07-08
fix(web-forms#847): fix styling of media controls in repeats

### LINKED ISSUE #847: Styling breaks for image question in nested repeat
### Expected Behaviour

Gray shading goes all the way across the container.

### Observed Behaviour

Image

### Reproduction Steps

Load a form with an image question inside a nested repeat such as [annotate_rpt.xlsx](https://github.com/user-attachments/files/29725058/annotate_rpt.xlsx)

### Environment

- Browser version: Firefox 150.0.3
- OS: macOS
- Web Forms version: 1.0

### Additional Context

I assume this affects all question types that have the drag and drop component but haven't confirmed.

garethbowen:
Closes getodk/web-forms#847

This was caused by using primevue panels for both repeats and file uploads so the styling for one also impact the other.

Tested a range of forms

It's a minimal solution. I think we should think about our approach to styling, for example, considering adding our own classes instead of using the primevue classes. This would help here because we could add a different class for FormPanel than UploadControl which would give us more control as to what styles we're using.

Unintended changes are possible...

No.

### DIFF
--- a/packages/web-forms/src/components/form-elements/upload/UploadControl.vue
+++ b/packages/web-forms/src/components/form-elements/upload/UploadControl.vue
@@ -339,12 +339,14 @@ onUnmounted(() => {
 	border-radius: var(--odk-radius);
 
 	:deep(.upload-control-header) {
-		border-radius: var(--odk-radius) var(--odk-radius) 0 0;
 		background: var(--odk-light-background-color);
 		justify-content: flex-start;
 		gap: var(--odk-spacing-xl);
 		border-bottom: 1px solid var(--odk-border-color);
 		padding: 12px var(--odk-spacing-xl);
+		display: flex;
+		align-items: center;
+		width: 100%;
 	}
 
 	:deep(.upload-control-content) {
--- a/packages/web-forms/src/components/form-elements/upload/UploadControl.vue
+++ b/packages/web-forms/src/components/form-elements/upload/UploadControl.vue
@@ -9,7 +9,6 @@ import type { Translate } from '@/lib/locale/useLocale.ts';
 import type { UploadNode } from '@getodk/xforms-engine';
 import Button from 'primevue/button';
 import Message from 'primevue/message';
-import Panel from 'primevue/panel';
 import { computed, inject, onUnmounted, ref, watchEffect } from 'vue';
 import DeleteConfirmDialog from './DeleteConfirmDialog.vue';
 import UploadAudioHeader from './UploadAudioHeader.vue';
@@ -246,8 +245,8 @@ onUnmounted(() => {
 <template>
 	<ControlText :question="question" />
 
-	<Panel :class="{ 'mobile-only-header': isBlankCanvas }">
-		<template #header>
+	<div class="upload-control" :class="{ 'mobile-only-header': isBlankCanvas }">
+		<div class="upload-control-header">
 			<template v-if="mediaType === 'image'">
 				<UploadImageHeader
 					:question="question"
@@ -267,8 +266,8 @@ onUnmounted(() => {
 			<template v-else>
 				<UploadFileHeader :question="question" :accept="accept" :is-disabled="isDisabled" @change="onChange" />
 			</template>
-		</template>
-		<template #default>
+		</div>
+		<div class="upload-control-content">
 			<div class="drag-and-drop" :class="{ 'disabled': isDisabled, 'canvas-mode': !!canvasMode && showImagePreview }" @drop.prevent.stop="onDrop" @dragover.prevent>
 				<div v-if="canvasEmptyPlaceholder.length && !question.currentState.value" class="canvas-empty-placeholder">
 					{{ canvasEmptyPlaceholder }}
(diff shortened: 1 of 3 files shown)


latin-panda (changes requested):
I think the underlying issue is that UploadControl shouldn't be using `<Panel>` in the first place. In concept it's a form control, not a region. It doesn't use any of [Panel's behavior](https://primevue.dev/panel#api) (no toggle, no collapse, no icons).

Switching to a plain `<div class="upload-control">` (like other form question types do) would remove the need for the `>` combinators, the padding reset, and would protect us from future PrimeVue changes. It'd also scope the testing of this fix better.

I think this is a small refactor that can be done here; one file, mostly renaming and adding a couple of CSS lines. Can we go with the refactor?

garethbowen:
@latin-panda I've done the refactor.

Now the problem is we're reusing styles for panels and uploads because, while you're right that the functionality is different, the styling is the same.

I don't like this duplication. One solution would be to use mixins but you still end up with duplication in the dist, which is why my preferred option (above) is to use a common class for everything that should be styled the same which applies ODK styling over the top of the primevue panel styling.

Are you happy with what's there now, or should I go further?

packages/web-forms/src/components/form-elements/upload/UploadControl.vue
```diff
@@ -333,25 +331,29 @@ onUnmounted(() => {
 <style scoped lang="scss">
 @use '../../../assets/styles/breakpoints' as odk;
 
-.p-panel {
+.upload-control {
 	background: var(--odk-base-background-color);
 	box-shadow: none;
```
latin-panda: This can be removed. The div has no issues with shadows.

packages/web-forms/src/components/form-elements/upload/UploadControl.vue
```diff
@@ -333,25 +331,29 @@ onUnmounted(() => {
...
 	border: 1px solid var(--odk-border-color);
-  overflow: hidden;
+	overflow: hidden;
+	border-radius: var(--odk-radius);
 
-	:deep(.p-panel-header) {
-		border-radius: var(--odk-radius) var(--odk-radius) 0 0;
+	:deep(.upload-control-header) {
```
latin-panda:
```suggestion
	.upload-control-header {
```

All `:deep` below should be removed.

packages/web-forms/src/components/form-elements/upload/UploadControl.vue
```diff
@@ -333,25 +331,29 @@ onUnmounted(() => {
...
+		padding: 12px var(--odk-spacing-xl);
+		display: flex;
+		align-items: center;
+		width: 100%;
 	}
 
-	:deep(.p-panel-content-container .p-panel-content) {
+	:deep(.upload-control-content) {
 		padding: 0;
 	}
```
latin-panda: All these should be removed

packages/web-forms/src/components/form-elements/upload/UploadControl.vue
```diff
@@ -419,16 +421,16 @@ onUnmounted(() => {
...
 @include odk.sm-constrained {
-	.p-panel :deep(.p-panel-header) {
+	.upload-control :deep(.upload-control-header) {
 		flex-direction: column;
 		justify-content: center;
 
-		.p-panel-header-actions {
+		.upload-control-header-actions {
 			display: none;
 		}
```
latin-panda: No element has this class; it should be removed.

packages/web-forms/src/components/form-elements/upload/UploadControl.vue
```diff
@@ -333,25 +331,29 @@ onUnmounted(() => {
...
 	border: 1px solid var(--odk-border-color);
-  overflow: hidden;
+	overflow: hidden;
+	border-radius: var(--odk-radius);
 
-	:deep(.p-panel-header) {
-		border-radius: var(--odk-radius) var(--odk-radius) 0 0;
+	:deep(.upload-control-header) {
 		background: var(--odk-light-background-color);
 		justify-content: flex-start;
```
latin-panda: This can also be removed since it's the default value and no longer overrides any component
```suggestion

```

latin-panda (changes requested):
Looking good! Just a few minor suggestions below. After that, I'll do another round of testing in Safari and Android

packages/web-forms/src/components/form-elements/upload/UploadControl.vue
garethbowen: Done.

packages/web-forms/src/components/form-elements/upload/UploadControl.vue
garethbowen: Done.

packages/web-forms/src/components/form-elements/upload/UploadControl.vue
garethbowen: Done.

packages/web-forms/src/components/form-elements/upload/UploadControl.vue
garethbowen: Done.

packages/web-forms/src/components/form-elements/upload/UploadControl.vue
garethbowen: Done.

latin-panda (approved):
It's ready to merge! I tested in Safari and Android, and it looks good.

PR #1703, opened 2026-07-09
Redirect user navigating to parent route

### LINKED ISSUE #1850: [frontend] /account/ crashes

matthew-white:
Closes getodk/central#1850. The core issue is that routes are expected to have route meta fields. routes.js has a section describing those meta fields, as well as a set of default values. However, in this case, the user is navigating directly to a parent route (/account) instead of a bottom-level route (e.g., /account/edit). We only define route meta fields on bottom-level routes. This PR checks for navigation to a parent route and redirects the user elsewhere.

I wrote a new test that fails without the changes to router.js.

I made this change after seeing getodk/central#1850 and #1536. #1536 is focused on `validateData` in particular, which makes sense, since that's what the error message mentions. However, I think the potential issues with missing meta fields extends beyond `validateData`. I'm making the check here one of the first `beforeEach()` navigation guards, because many of the navigation guards that follow assume that `to.meta` is well-defined and has default values.

### DIFF
--- a/apps/central/src/router.js
+++ b/apps/central/src/router.js
@@ -59,6 +59,12 @@ router.afterEach(unlessFailure(to => {
   //////////////////////////////////////////////////////////////////////////////
   // REDIRECTS
 
+  // All bottom-level routes in routes.js should have route meta fields. Many of
+  // the navigation guards below assume that `to` has well-defined meta fields.
+  // Here, we check that `to` has meta fields. If it doesn't, that means the
+  // user is trying to navigate to a parent route, which isn't expected.
+  router.beforeEach(to => (Object.keys(to.meta).length === 0 ? '/' : true));
+
   // If a route is nested, its relative path is '', and that path is an alias,
   // then we redirect to the canonical path. That turned out to be easier than
   // using the `redirect` option of Vue Router.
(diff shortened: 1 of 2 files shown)


matthew-white:
Just to keep things linked, I also added some related thoughts at https://github.com/getodk/central-frontend/pull/1536#issuecomment-4921717391.

alxndrsn (approved):

PR #1705, opened 2026-07-13
Remove attachment confirmation modal if any files match

### LINKED ISSUE #1720: Improve form attachment upload flow
On the Edit Form page, @lognaturel has found that it's possible to select new form attachments, then publish the form draft without realizing that the new attachments haven't been uploaded.

Part of the reason behind that is that the "Looks good" button can end up being far away from the form attachments section, appearing in the bottom-right corner of the page. A user might not even notice the "Looks good" button.

@lognaturel suggested a couple of ideas:

1. Move the "Looks good" button closer to the form attachments section
2. In the publish modal, warn if new form attachments have been selected, but not uploaded

I've been thinking about form attachments as well in the context of https://github.com/getodk/central/issues/1291#issuecomment-3215537292. In that case, there can be unexpected behavior if a form attachment is modified in-between when the user selects the file and when the user uploads it. If we make it easier to upload a file right away, then that issue is less likely to emerge.

@nicoleorlowski, what do you think? Is there a way to improve this flow?

ktuite:
Closes https://github.com/getodk/central/issues/1720

This removes the confirmation step when you drag multiple attachments onto a form draft. As long as some of the filenames match the expected attachments, the matching files will be uploaded automatically. Only if none of the files match will there be a popup explaining how nothing could be uploaded. 

**Removed**: 
(dragging b.jpg and c.jpg vs dragging b.jpg, c.jpg, other.jpg)
Screenshot 2026-07-13 at 11 55 01 AM Screenshot 2026-07-13 at 11 55 13 AM

**Kept**:
(dragging other.jpg, other2.jpg)
Screenshot 2026-07-13 at 12 31 31 PM

The popup is now simpler, and the table for rendering the attachments is also a bit simpler in that it wont highlight rows of pending attachments. 

I've tried it out and updated the tests. I had to make some changes to the tests
- remove sections about testing the highlighting on the table for pending uploads
- change the tests that automatically uploaded attachment data to include the POST requests and chain the requests together instead of using async/await 

This is a pretty complex bit of code but I think I cut out the confirmation step in an acceptable way. 

Will help users not to forget to upload their attachments if dragging multiple attachments at once. 

If the users have a mix of matching/non-matching filenames, they'll get less feedback about the non-matching filenames. But they still get feedback (via rows highlighted after upload) on which files did successfully upload. 

I don't think so.

### DIFF
--- a/apps/central/src/components/form-attachment/list.vue
+++ b/apps/central/src/components/form-attachment/list.vue
@@ -29,8 +29,8 @@ except according to the terms contained in the LICENSE file.
       :count-of-files-over-drop-zone="countOfFilesOverDropZone"
       :dragover-attachment="dragoverAttachment"
       :planned-uploads="plannedUploads" :unmatched-files="unmatchedFiles"
-      :name-mismatch="nameMismatch" :upload-status="uploadStatus"
-      @confirm="uploadFiles" @cancel="cancelUploads"/>
+      :upload-status="uploadStatus"
+      @cancel="cancelUploads"/>
 
     <form-attachment-upload-files v-bind="uploadFilesModal"
       @hide="uploadFilesModal.hide()" @select="afterFileInputSelection"/>
@@ -207,8 +207,13 @@ export default {
         else
           this.unmatchedFiles.push(file);
       }
+
+      // Automatically upload planned files without confirmation
+      if (this.plannedUploads.length > 0)
+        this.uploadFiles();
+
       // With the changes to this.plannedUploads and this.unmatchedFiles, the
-      // popup will show in the next tick.
+      // popup will show in the next tick if there are no plannedUploads, only unmatched files.
     },
     // cancelUploads() cancels the uploads before they start, after files have
     // been selected. (It does not cancel an upload in progress.)
--- a/apps/central/src/components/form-attachment/list.vue
+++ b/apps/central/src/components/form-attachment/list.vue
@@ -28,15 +28,15 @@ except according to the terms contained in the LICENSE file.
     <form-attachment-popups
       :count-of-files-over-drop-zone="countOfFilesOverDropZone"
       :dragover-attachment="dragoverAttachment"
-      :planned-uploads="plannedUploads" :unmatched-files="unmatchedFiles"
+      :unmatched-files="unmatchedFiles"
       :upload-status="uploadStatus"
-      @cancel="cancelUploads"/>
+      @cancel="clearPendingFiles"/>
 
     <form-attachment-upload-files v-bind="uploadFilesModal"
       @hide="uploadFilesModal.hide()" @select="afterFileInputSelection"/>
     <form-attachment-name-mismatch v-bind="nameMismatch"
       :planned-uploads="plannedUploads" @hide="nameMismatch.hide()"
-      @confirm="uploadFiles" @cancel="cancelUploads"/>
+      @confirm="uploadFiles" @cancel="clearPendingFiles"/>
     <form-attachment-link-dataset v-bind="linkDatasetModal"
       @hide="linkDatasetModal.hide()" @success="afterLinkDataset"/>
   </file-drop-zone>
@@ -87,9 +87,11 @@ export default {
            drop or using the file input) and that are reset once the uploads
            have started or been canceled
            - plannedUploads. An array of file/attachment pairs, indicating which
-             file to upload for which attachment. plannedUploads is used to
-             confirm or cancel the uploads, then to start them, but it is not
-             used during the uploads themselves.
+             file to upload for which attachment. In most flows, plannedUploads
+             is populated and immediately consumed by uploadFiles(). The
+             exception is when a file is dropped onto a mismatched row: in that
+             case plannedUploads holds the pending upload while the name-mismatch
+             modal awaits confirmation.
            - unmatchedFiles. An array of the files whose names did not match any
              attachment's.
         3. Properties set once the uploads have started and reset once they have
@@ -165,7 +167,7 @@ export default {
           ? this.draftAttachments.get(tr.dataset.name)
           : null;
       }
-      this.cancelUploads();
+      this.clearPendingFiles();
       this.updatedAttachments.clear();
     },
     dragleave(_, leftDropZone) {
@@ -208,16 +210,16 @@ export default {
           this.unmatchedFiles.push(file);
       }
 
-      // Automatically upload planned files without confirmation
+      // Automatically upload planned files without confirmation.
+      // If there are no planned files, the popup will show in the next tick for unmatched files.
       if (this.plannedUploads.length > 0)
         this.uploadFiles();
-
-      // With the changes to this.plannedUploads and this.unmatchedFiles, the
-      // popup will show in the next tick if there are no plannedUploads, only unmatched files.
     },
-    // cancelUploads() cancels the uploads before they start, after files have
-    // been selected. (It does not cancel an upload in progress.)
-    cancelUploads() {
+    // clearPendingFiles() clears any pending file selection before an upload
+    // has started. It is called both when the user cancels a name mismatch
+    // (clearing plannedUploads) and when the user dismisses the unmatched-files
+    // popup (clearing unmatchedFiles), so it always clears both.
+    clearPendingFiles() {
       if (this.plannedUploads.length !== 0) this.plannedUploads = [];
       if (this.unmatchedFiles.length !== 0) this.unmatchedFiles = [];
     },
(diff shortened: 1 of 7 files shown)


apps/central/test/components/form-attachment/list.spec.js
```diff
@@ -130,69 +130,6 @@ describe('FormAttachmentList', () => {
...
-        });
-        await select(component, blankFiles(['a', 'b', 'd']));
-        const labels = component.findAllComponents(FormAttachmentRow)
-          .map(row => row.get('.label'));
-        labels[0].should.be.visible();
-        labels[0].text().should.equal('Replace');
-        labels[1].should.be.visible();
-        labels[1].text().should.equal('Upload');
-        labels[2].should.be.hidden();
-      });
```
matthew-white, line 209: Seeing these tests go away, I'm wondering, are we able to remove this label logic from the `FormAttachmentRow` component now? It'd be nice to simplify that if possible.

apps/central/src/components/form-attachment/list.vue
matthew-white, line 1: I searched this file to see how `unmatchedFiles` was used, and I saw `cancelUploads()`. `cancelUploads()` is now only relevant in the no-matching-file case, right? If so, what do you think about removing this line from the method?

```js
if (this.plannedUploads.length !== 0) this.plannedUploads = [];
```

Maybe we rename the method to `cancelUnmatched()` or `clearUnmatched()`?

Since it's a complex component, I think it's nice to make it clear how/where each data property can be modified.

apps/central/src/components/form-attachment/list.vue
```diff
@@ -207,8 +207,13 @@ export default {
...
           this.unmatchedFiles.push(file);
       }
+
+      // Automatically upload planned files without confirmation
+      if (this.plannedUploads.length > 0)
+        this.uploadFiles();
+
       // With the changes to this.plannedUploads and this.unmatchedFiles, the
-      // popup will show in the next tick.
+      // popup will show in the next tick if there are no plannedUploads, only unmatched files.
```
matthew-white: What do you think about merging this comment with the one above? I feel like this one is slightly confusing now. I'm thinking of something like:

> Automatically upload planned files without confirmation.
> If there are no planned files, the popup will show in the next tick for unmatched files.

apps/central/src/components/form-attachment/list.vue
matthew-white, line 1: Do the comments underneath `data()` describing the data flow need to be updated? (I suspect so, but I haven't looked carefully.)

apps/central/src/components/form-attachment/list.vue
```diff
@@ -207,8 +207,13 @@ export default {
         else
           this.unmatchedFiles.push(file);
       }
+
+      // Automatically upload planned files without confirmation
+      if (this.plannedUploads.length > 0)
+        this.uploadFiles();
```
matthew-white: In this case, what do you think about also clearing `unmatchedFiles`? We don't need to hold onto those files anymore, and clearing them may simplify the data flow.

matthew-white (commented):
Leaving some questions/comments, but this looks great so far. It's a complex component for sure!

apps/central/test/components/form-attachment/list.spec.js
matthew-white: Similarly, is it possible to remove some of the highlighting logic from `FormAttachmentRow` now?

apps/central/src/components/form-attachment/list.vue
ktuite: `cancelUploads()` (now renamed to `clearPendingFiles()`) is used in a couple more places with the name mismatch modal (for a single file), the popup (when all files are mismatched), and something with resetting the drag enter, so it still needs the `plannedUploads` line.

ktuite: updated!

apps/central/src/components/form-attachment/list.vue
ktuite: `unmatchedFiles` are cleared as part of `uploadFiles()` so I don't think it's necessary to add a clearing step here. And when the files aren't uploaded, the `unmatchedFiles` are used in the next modal/popup.

apps/central/src/components/form-attachment/list.vue
ktuite: comments updated and improved

apps/central/test/components/form-attachment/list.spec.js
ktuite: This sounded like a good idea at first, but the highlighting and labeling table logic is still used for
- single file upload (like when you drag something onto a table and then the name mismatch modal shows up)
- just dragging a set of files over the table 

There might be a way to simplify it a _little_ bit, but it would require untangling and understanding the rest of possible paths and I don't know if that's worth it. I think it's kinda nice UX here!

I may change my mind in the future if/when we have entity media attachments... like when there's a need to upload a bunch of files for an entity list and that wants to reuse some of this UI.

ktuite (commented):
Responding to comments

apps/central/src/components/form-attachment/list.vue
matthew-white: OK, I see that the name mismatch modal in particular does need the `plannedUploads` line. :+1: I don't think the other cases need it, but one is enough.

apps/central/src/components/form-attachment/list.vue
matthew-white: > `unmatchedFiles` are cleared as part of `uploadFiles()` so I don't think it's necessary to add a clearing step here.

Maybe not necessary, but I do feel like it'd improve the data flow. `uploadFiles()` doesn't use `unmatchedFiles` at all. So `matchFilesToAttachments()` ends up setting a data property, then calling another method that doesn't use the data property and just immediately resets it.

> when the files aren't uploaded, the `unmatchedFiles` are used in the next modal/popup.

Yes, we definitely need it set in that case. One code pattern that might make everything more explicit is to use a different variable than `this.unmatchedFiles`. Idea: rather than setting `this.plannedUploads` and `this.unmatchedFiles`, the for-loop could set `this.plannedUploads` and a local variable (not a data property). Then only set the data property to that variable if that's needed for the popup. Something like:

```js
this.plannedUploads = [];
const unmatched = [];
for (let i = 0; i < files.length; i += 1) {
  // ...
}

if (this.plannedUploads.length > 0)
  // Automatically upload planned files without confirmation.
  this.uploadFiles();
else
  // Setting this data property will show the popup in the next tick.
  this.unmatchedFiles = unmatched;
```

apps/central/src/components/form-attachment/list.vue
matthew-white: Thinking more about it, I can't think of a case that now needs to clear both sets of files:

- Name mismatch modal (for a single file): Needs to clear `plannedUploads`, but not `unmatchedFiles` (which should be empty).
- The popup (when all files are mismatched): Needs to clear `unmatchedFiles`, but not `plannedUploads` (which should be empty).
- Drag enter: Needs to clear `unmatchedFiles`, but not `plannedUploads` (which should be empty).

Whereas before, it really was necessary to clear both sets of files.

I think it'd make the data flow more explicit to use two different methods instead, one for each set of files. Then it'd be clearer when/how each set of files is mutated.

I'm also happy to leave it as-is, just wanted to throw out this idea.

apps/central/test/components/form-attachment/list.spec.js
matthew-white: Ah that's a good point that some highlighting/labeling is still used. :+1:

I think we should remove the following line. If `plannedUploads` is set, that now means that the request is sent immediately (it's in progress), and an overlay is shown over the table. So there's no reason to do the highlighting/labeling in that case. If we remove this line, then we can stop passing `plannedUploads` to `FormAttachmentTable` and `FormAttachmentRow`, simplifying the data flow.

https://github.com/getodk/central-frontend/blob/f6b6bbfd08aeb40654d6d86c13f145d3b17e5f5d/apps/central/src/components/form-attachment/row.vue#L113-L114

matthew-white: > If `plannedUploads` is set, that now means that the request is sent immediately (it's in progress), and an overlay is shown over the table.

Or I guess another case is that the name mismatch modal is shown over the table (the request isn't in progress yet). But also in that case, I don't think we need the highlighting/labeling. In both cases, there's an overlay over the table.

matthew-white:
I like the new commit! I've left a few replies in the threads above. I think there are a couple more opportunities to improve the code and make this complex data flow clearer for our future selves.

matthew-white (approved):

PR #1704, opened 2026-07-13
e2e-tests: throw when something bad is logged

alxndrsn:
Would have caught:

* https://github.com/getodk/central/issues/2065

Working on this also uncovered other bugs:

* https://github.com/getodk/central/issues/2069
* https://github.com/getodk/central/issues/2070

#### What has been done to verify that this works as intended?

Ran in CI a bunch.

#### Why is this the best possible solution? Were any other approaches considered?

I tried a few different methods of failing tests:

1. throwing inside the event listener
2. modifying the `testInfo` object
3. asserting in the test fixture after `use()`

3 seems like the cleanest solution while still maintaining `auto:true`, although it does include the confusing logging `1 error was not a part of any test, see above for details` when there's a failure due to unexpected logging:

```
Running 70 tests using 1 worker

  ✓   1 [setup] › global.setup.js:77:1 › create new project (1.2s)
[chromium|console.error] http://central-test.localhost/:17:0
    message: Loading the image 'http://central-test.localhost/favicon.ico' violates the following Content Security Policy directive: "img-src data: https:". The action has been blocked.
[chromium|console.error] http://central-test.localhost/:18:0
    message: Loading the image 'http://central-test.localhost/apple-touch-icon.png' violates the following Content Security Policy directive: "img-src data: https:". The action has been blocked.
[chromium|console.error] http://central-test.localhost/:19:0
    message: Loading the image 'http://central-test.localhost/favicon-32x32.png' violates the following Content Security Policy directive: "img-src data: https:". The action has been blocked.
[chromium|console.error] http://central-test.localhost/:20:0
    message: Loading the image 'http://central-test.localhost/favicon-16x16.png' violates the following Content Security Policy directive: "img-src data: https:". The action has been blocked.
[chromium|console.error] http://central-test.localhost/projects/2/forms/chromium_wf_draft/submissions/new:6:0
    message: Loading the image 'http://central-test.localhost/favicon.ico' violates the following Content Security Policy directive: "img-src blob: data: https:". The action has been blocked.
[chromium|console.error] http://central-test.localhost/projects/2/forms/chromium_wf_draft/submissions/new:7:0
    message: Loading the image 'http://central-test.localhost/apple-touch-icon.png' violates the following Content Security Policy directive: "img-src blob: data: https:". The action has been blocked.
[chromium|console.error] http://central-test.localhost/projects/2/forms/chromium_wf_draft/submissions/new:8:0
[...]

PR #1706, opened 2026-07-14
e2e-tests/util: clarify message-related var names

alxndrsn:

- [x] ci

* these variables currently have very similar, ambiguous names
* the scope of the variables will increase greatly in https://github.com/getodk/central-frontend/pull/1704

No impact - just a var rename.

No.

### DIFF
--- a/e2e-tests/util.js
+++ b/e2e-tests/util.js
@@ -22,30 +22,30 @@ const test = testBase.extend({
   // See: https://playwright.dev/docs/test-fixtures#adding-global-beforeeachaftereach-hooks
   browserConsoleToTestStdout: [
     async ({ browserName, page }, use) => {
-      page.on('console', async msg => {
-        const { url, line, column } = msg.location();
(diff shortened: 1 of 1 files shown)


e2e-tests/util.js
```diff
@@ -22,30 +22,30 @@ const test = testBase.extend({
   // See: https://playwright.dev/docs/test-fixtures#adding-global-beforeeachaftereach-hooks
   browserConsoleToTestStdout: [
     async ({ browserName, page }, use) => {
-      page.on('console', async msg => {
-        const { url, line, column } = msg.location();
+      page.on('console', async consoleMsg => {
```
alxndrsn, line 25: `consoleMsg` is an instance of playwright's `ConsoleMessage` class (https://playwright.dev/docs/api/class-consolemessage)

matthew-white (approved):

latin-panda (approved):
Thanks!

PR #1708, opened 2026-07-14
fix(web-forms#848): use current form definition's version when editing a submission

### LINKED ISSUE #848: Edited submissions carry the form version from the original submission, not the current one
When a submission is edited against a newer form version, the resulting XML retains the version attribute from the originally-submitted XML rather than the version of the form definition the edit was actually performed against.

### Expected Behaviour

Submission carries the latest form version when modified. 

### Observed Behaviour

Downstream consumers (Central) see an outdated version and can't identify which form definition the edit was based on. For instance, if v1 of a form has a single upload field, a submission is made, then the form is edited to include an additional upload field, Central doesn't recognize that it should update the value for the new field. This leads to errors because the submission references an old version number. [Forum post](https://forum.getodk.org/t/unable-to-update-images-when-editing-with-collect/58334)

### Reproduction Steps

Steps:
  1. Publish form v1, submit a filled instance against it.
  2. Publish form v2 (adds new fields).
  3. Edit the v1 submission. Web Forms correctly renders it against v2 (new fields appear).
  4. Save the edit. The resulting XML still shows version="1" on the root element.

### Screenshots or Videos

[See video](https://getodk.slack.com/files/U088FHC94HZ/F0BH31ZJ3J5/screen_recording_2026-07-14_at_12.49.17___am.mov)

### Environment

- Browser version: All
- OS: All 
- Web Forms version: 1.0.0

latin-panda:
Closes https://github.com/getodk/web-forms/issues/848

 When editing a submission originally filled against an older form version, the edited instance retained the old `version` attribute on the root element. This PR makes root's `version` always resolve from the current form definition instead of the source XML.

Tested using these forms:

[Form v1](https://github.com/user-attachments/files/30002919/form-v1.xml)

[Form v2](https://github.com/user-attachments/files/30002900/form-with-new-field.xml)

 - The `buildAttributes` treats source XML as authoritative for all attributes on all nodes, falling back to the template only when the source lacks the attribute.
- That rule is correct for element attributes that hold user data, but wrong for root's `version`, which is form metadata. It describes the form definition the submission is filled against, not something the user edited.
- Rather than changing the source-wins rule globally (which would break user-data attributes), the fix narrowly identifies form-metadata attributes and resolves them from the template. The predicate is scoped to root's `version`, the only mutable root-level metadata attribute in ODK XForms; `id` is stable across versions, and non-root attributes remain user data.

Unblocks edition of submissions when the form definition changes. 

No

### DIFF
--- a/.changeset/salty-worlds-stick.md
+++ b/.changeset/salty-worlds-stick.md
@@ -1,0 +1,5 @@
+---
+"@getodk/xforms-engine": patch
+---
+
+Use current form version when editing a submission.
--- a/packages/xforms-engine/src/instance/attachments/buildAttributes.ts
+++ b/packages/xforms-engine/src/instance/attachments/buildAttributes.ts
@@ -1,12 +1,16 @@
 import type { StaticAttribute } from '../../integration/xpath/static-dom/StaticAttribute.ts';
 import type { StaticDocument } from '../../integration/xpath/static-dom/StaticDocument.ts';
 import type { StaticElement } from '../../integration/xpath/static-dom/StaticElement.ts';
+import type { AttributeDefinition } from '../../parse/model/AttributeDefinition.ts';
 import { Attribute } from '../Attribute';
 import type { AnyNode } from '../hierarchy.ts';
 import type { InputControl } from '../InputControl.ts';
 import type { ModelValue } from '../ModelValue.ts';
 import type { Note } from '../Note.ts';
 import type { RangeControl } from '../RangeControl.ts';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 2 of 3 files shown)


latin-panda:
@sadiqkhoja Could you please review the fix? It's small. I added details in the PR description. Please let me know if you have any questions!

sadiqkhoja (approved):
Looks good.

Only question that I have is why `buildAttributes.ts` lives under attachments directory? when it buildAttribute is a generic function to be called by anynode

latin-panda:
> Only question that I have is why buildAttributes.ts lives under attachments directory? when it buildAttribute is a generic function to be called by anynode

Good catch! If I remember correctly, it started with something about attachments and later got generalized, but we probably forgot to move it to a better folder. I'll open a separate PR for that. Thanks!

PR #1709, opened 2026-07-14
fix(web-forms#849): decreasing repeat count doesn't remove repeat instances

### LINKED ISSUE #849: Web Forms deletes repeat instances when jr:count is decreased
### Expected Behaviour

Per the ODK XForms docs, decreasing a `jr:count` value must not delete existing repeat instances, to prevent data-loss. Web Forms currently removes the excess instances (and any user-entered data), matching Enketo but diverging from Collect and the spec. If the count is later restored, the previously entered content is gone.

### Observed Behaviour

Repeat instances are gone when decreasing the repeat count. [Forum post](https://forum.getodk.org/t/decreasing-repeat-count-in-a-submission-removes-instances-in-webforms/58372).

### Reproduction Steps

Using the [repeat_count_decrement.xml](https://github.com/user-attachments/files/30007692/repeat_count_decrement.xml): 
- Enter 3, 
- Fill each repeat_input, 
- Lower to 1, 
- Then raise back to 3
- The last two instances come back blank instead of restored.

### Environment

- Browser version: all supported
- OS: all supported
- Web Forms version: 1.0.0

latin-panda:
Closes https://github.com/getodk/web-forms/issues/849

I tested with all these forms:
- [repeat_count_decrement.xml](https://github.com/user-attachments/files/30019808/repeat_count_decrement.xml)
- [repeat sample forms](https://github.com/getodk/sample-forms/tree/master/xml-examples/repeats)

When a `jr:count` expression decreased at runtime, the engine called `removeChildren`, silently deleting user-entered data in the trimmed instances. This was wrong: a count expression going down is not the user's intent to discard their work.

The fix removes the `removeChildren` path entirely. On form creation, the reactive count loop still seeds the correct number of instances. On edit submission, instances are seeded from the saved state first, so the live count expression can only ever grow the set. 

The FIXED/DYNAMIC control-type distinction is removed because the policy is now uniform: the count controls the floor, not a ceiling that triggers deletion.

Note that if the count is set to 10 and then changed to 5, but the user hasn't filled in the fields in any of the repeats yet, the form still keeps 10 repeats. This is the same behavior as Collect and saves processing by not checking which repeats are fully empty. 

Maybe some edge scenario about fixed / dynamic count in repeats and edit submissions; I tried to test carefully but will still send it to the QA team.

The docs were already describing the behavior we implemented in this PR, which is what Collect uses.


latin-panda:
@garethbowen, Could you please review when you have a chance? No rush!

packages/xforms-engine/test/integration/repeat.test.ts
```diff
@@ -2791,24 +2791,39 @@ describe('jr:count', () => {
...
           updatedCount = randomCount();
-        } while (updatedCount === previousCount);
+        } while (updatedCount === retainedCount);
 
         scenario.answer('/data/rep-count', updatedCount);
 
-        expect(scenario.countRepeatInstancesOf('/data/rep')).not.toBe(previousCount);
-        expect(scenario.countRepeatInstancesOf('/data/rep')).toBe(updatedCount);
+        // Decreasing jr:count never deletes repeat instances.
+        const expectedCount = Math.max(updatedCount, retainedCount);
```
garethbowen: I don't like using random numbers in tests because it means the result is non-deterministic. For example, in this test some of the time updatedCount will be greater than retainedCount and sometimes it won't be. With your code change it's possible that this test has never actually tested the decreasing jr:count.

Can we replace randomCount with two tests with hardcoded counts and deterministic evaluation?

packages/xforms-engine/src/instance/repeat/RepeatRangeControlled.ts
```diff
@@ -39,117 +32,69 @@ export class RepeatRangeControlled
...
+    if (Number.isNaN(currentCount) || currentCount <= previousCount) {
+      return previousCount;
     }
 
-    const droppedCount = repeatInstanceNodes.length - initialCount;
+    const delta = currentCount - previousCount;
+    this.addChildren(
+      previousCount - 1,
+      this.resolveInstanceNodes(savedNodes, template, previousCount, delta)
+    );
```
garethbowen: I find this slightly more readable


```suggestion
    const instanceNodes = this.resolveInstanceNodes(savedNodes, template, previousCount, delta);
    this.addChildren(previousCount - 1, instanceNodes);
```

garethbowen: Also looking into how `addChildren` is used, I think this API can be improved, for example if it defaulted to appending which is how it's almost always used, rather than passing in an index. Anyway I'll leave this as an optional extra refactor...

garethbowen (changes requested):
Great work.

I'd like to remove the randomness in the test, but the other comment is just a nit.

packages/xforms-engine/src/instance/repeat/RepeatRangeControlled.ts
latin-panda: Done!

packages/xforms-engine/test/integration/repeat.test.ts
latin-panda: Done!

latin-panda:
@garethbowen I've resolved the feedback, thank you!

garethbowen (approved):
Nice work.

PR #1710, opened 2026-07-15
Bootstrap/remove dropdown

sadiqkhoja:

Added tests and manual testing

Tried to keep css changes to minimum.

No

No

### DIFF
--- a/apps/central/src/components/dropdown.vue
+++ b/apps/central/src/components/dropdown.vue
@@ -1,15 +1,17 @@
 <template>
-  <component :is="tag" ref="container" :class="{ open: isOpen }">
+  <component :is="tag" ref="container" class="dropdown" :class="{ open: isOpen }">
     <slot name="toggle" :toggle="toggle" :is-open="isOpen" :attrs="triggerAttrs"></slot>
-    <ul v-show="isOpen" ref="menu" class="dropdown-menu" :style="menuStyle">
+    <ul ref="menu" class="dropdown-menu" :style="menuStyle">
       <slot name="menu"></slot>
     </ul>
   </component>
 </template>
 
 <script setup>
-import { computed, onBeforeUnmount, onMounted, ref } from 'vue';
+import { computed, ref } from 'vue';
 import { computePosition } from '@floating-ui/dom';
+
+import useEventListener from '../composables/event-listener';
 
 defineOptions({
   name: 'Dropdown'
@@ -89,21 +91,7 @@ const handleEscape = (event) => {
   }
 };
 
-const handleResize = () => {
-  if (isOpen.value) {
-    hide();
-  }
-};
-
-onMounted(() => {
-  document.addEventListener('click', handleClickOutside, true);
-  document.addEventListener('keydown', handleEscape);
-  window.addEventListener('resize', handleResize);
-});
-
-onBeforeUnmount(() => {
-  document.removeEventListener('click', handleClickOutside, true);
-  document.removeEventListener('keydown', handleEscape);
-  window.removeEventListener('resize', handleResize);
-});
+useEventListener(document, 'click', handleClickOutside, true);
+useEventListener(document, 'keydown', handleEscape);
+useEventListener(window, 'resize', hide);
 </script>
--- a/apps/central/src/components/dropdown.vue
+++ b/apps/central/src/components/dropdown.vue
@@ -1,0 +1,109 @@
+<template>
+  <component :is="tag" ref="container" :class="{ open: isOpen }">
+    <slot name="toggle" :toggle="toggle" :is-open="isOpen" :attrs="triggerAttrs"></slot>
+    <ul v-show="isOpen" ref="menu" class="dropdown-menu" :style="menuStyle">
+      <slot name="menu"></slot>
+    </ul>
+  </component>
+</template>
+
+<script setup>
+import { computed, onBeforeUnmount, onMounted, ref } from 'vue';
+import { computePosition } from '@floating-ui/dom';
+
+defineOptions({
+  name: 'Dropdown'
+});
+
+const props = defineProps({
+  tag: {
+    type: String,
+    default: 'div'
+  },
+  placement: {
+    type: String,
+    default: 'bottom-end'
+  }
+});
+
+const container = ref(null);
+const menu = ref(null);
+const isOpen = ref(false);
+const menuStyle = ref({});
+
+const triggerAttrs = computed(() => ({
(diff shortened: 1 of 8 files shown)


apps/central/src/components/dropdown.vue
```diff
@@ -0,0 +1,109 @@
+<template>
+  <component :is="tag" ref="container" :class="{ open: isOpen }">
```
sadiqkhoja: `class='dropdown'`

apps/central/src/components/dropdown.vue
```diff
@@ -0,0 +1,109 @@
...
+    <slot name="toggle" :toggle="toggle" :is-open="isOpen" :attrs="triggerAttrs"></slot>
+    <ul v-show="isOpen" ref="menu" class="dropdown-menu" :style="menuStyle">
```
sadiqkhoja: can remove it `v-show="isOpen"`

apps/central/src/components/dropdown.vue
```diff
@@ -0,0 +1,109 @@
...
+const handleResize = () => {
+  if (isOpen.value) {
+    hide();
+  }
+};
+
+onMounted(() => {
+  document.addEventListener('click', handleClickOutside, true);
+  document.addEventListener('keydown', handleEscape);
+  window.addEventListener('resize', handleResize);
```
sadiqkhoja: `window.addEventListener('resize', hide);`

apps/central/src/components/dropdown.vue
```diff
@@ -0,0 +1,109 @@
...
+  }
+};
+
+const handleResize = () => {
+  if (isOpen.value) {
+    hide();
+  }
+};
+
+onMounted(() => {
```
sadiqkhoja: useEventListener

apps/central/src/components/navbar/actions.vue
```diff
@@ -15,14 +15,16 @@ except according to the terms contained in the LICENSE file.
...
   </li>
-  <li v-else id="navbar-actions" class="dropdown">
-    <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button"
-      aria-haspopup="true" aria-expanded="false">
-      <span class="icon-user-circle-o"></span>
-      <span v-tooltip.text>{{ currentUser.displayName }}</span>
-      <span class="caret"></span>
-    </a>
-    <ul class="dropdown-menu">
+  <Dropdown v-else id="navbar-actions" tag="li" class="dropdown">
```
matthew-white: I think `class="dropdown"` can be removed here since that's now in the component. (This comment also applies to other files in this PR.)

matthew-white: Should we capitalize `<Dropdown>`? We don't usually capitalize component names, but that's mostly for historical reasons. I see that it's not capitalized (it's `<dropdown>`) in the `Infonav` component below. I would lean toward lowercase everywhere.

apps/central/src/components/dropdown.vue
matthew-white, line 1: Can we also now remove the Bootstrap dropdown plugin from src/bootstrap.js? Or even remove that file altogether now?

apps/central/src/components/dropdown.vue
```diff
@@ -0,0 +1,97 @@
...
+    show();
+  }
+};
+
+const handleClickOutside = (event) => {
+  if (!isOpen.value) return;
+
+  // If clicked on trigger, the toggle handler will manage the state
+  const clickedOnTrigger = container.value?.contains(event.target) &&
+    !menu.value?.contains(event.target);
```
matthew-white: Above, you used `container.value.firstElementChild` to access the trigger element. I feel like it'd be a little clearer to use the same approach here.

One case to consider: if the `Dropdown` has enough padding, it's conceivable that the user could click on the `Dropdown` itself without clicking either the trigger or the menu. In that case, I think the dropdown menu should be hidden: that should be considered a click outside.

```suggestion
  const clickedOnTrigger = container.value?.firstElementChild?.contains(event.target);
```

apps/central/src/components/dropdown.vue
```diff
@@ -0,0 +1,97 @@
...
+
+const container = ref(null);
+const menu = ref(null);
+const isOpen = ref(false);
+const menuStyle = ref({});
+
+const triggerAttrs = computed(() => ({
+  'aria-haspopup': 'true',
+  'aria-expanded': String(isOpen.value)
+}));
```
matthew-white, line 40: Is there any chance we could use this mechanism to automatically add `class="dropdown-toggle"` to the toggle/trigger element? It'd be nice if the `toggle` slot could just be a simple `<a>` and not have the additional requirement of needing a specially named class. No strong feelings here, just curious whether this would work and whether you think it'd be an improvement.

```js
class: 'dropdown-toggle'
```

apps/central/test/components/dropdown.spec.js
```diff
@@ -0,0 +1,39 @@
+import TestDropdownMenu from '../util/components/dropdown-menu.vue';
+
+import { mount } from '../util/lifecycle';
+
+describe('Dropdown', () => {
+  it('opens when toggle is clicked', async () => {
+    const component = mount(TestDropdownMenu, { attachTo: document.body });
```
matthew-white: Is it necessary to `attachTo: document.body`? It's OK if so, I'm more so just curious about why it's needed.

apps/central/test/components/dropdown.spec.js
matthew-white, line 1: Would you mind adding a test about `attrs` as well, since that's a new (but very cool) pattern for us?

apps/central/src/components/dropdown.vue
matthew-white: Another idea is that you could use `.querySelector('.dropdown-toggle')` to identify the trigger element.

matthew-white (approved):
Thank you, @sadiqkhoja! Very excited to bid farewell to another Bootstrap plugin. 🎉

apps/central/src/components/dropdown.vue
sadiqkhoja: there are other components which are using bootstrap dropdown. I'll remove bootstrap.js once follow up PRs are merged.

apps/central/test/components/dropdown.spec.js
sadiqkhoja: not necessary for testing open menu, so removing. Necessary for the tests for close because we rely on event listener on the document.

PR #1711, opened 2026-07-15
Bootstrap/remove collapse

sadiqkhoja:
Part of https://github.com/getodk/central/issues/1278

Manual verification in comparison with staging to confirm navbar looks same.

Overriding bootstrap class was an option to disable responsiveness but found it too cumbersome. Using flexbox with our own classes made the styling simpler.

none

No

### DIFF
--- a/apps/central/src/components/navbar.vue
+++ b/apps/central/src/components/navbar.vue
@@ -174,10 +174,6 @@ $border-height: 3px;
 <i18n lang="json5">
 {
   "en": {
-    "action": {
-      // Used by screen readers to describe the button used to show or hide the navigation bar on small screens ("hamburger menu").
-      "toggle": "Toggle navigation"
-    },
     "analyticsNotice": "Help improve Central!"
   }
 }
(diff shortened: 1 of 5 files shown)


apps/central/src/components/navbar.vue
```diff
@@ -10,35 +10,23 @@ including this file, may be copied, modified, propagated, or distributed
...
-          <button type="button" class="navbar-toggle collapsed"
-            data-toggle="collapse" data-target=".navbar-collapse"
-            aria-expanded="false">
-            <span class="sr-only">{{ $t('action.toggle') }}</span>
-            <span class="navbar-icon-bar"></span>
-            <span class="navbar-icon-bar"></span>
-            <span class="navbar-icon-bar"></span>
-          </button>
+  <div id="navbar">
+    <nav>
```
sadiqkhoja, line 14: we can get rid of top `div`

apps/central/src/components/navbar.vue
sadiqkhoja: remove it

matthew-white (approved):

sadiqkhoja (commented):
comments from the interactive code review

apps/central/src/components/navbar.vue
sadiqkhoja: can't remove the root element because parent passes `v-show` to navbar and a warning is shown `Runtime directive used on component with non-element root node. The directives will not function as intended. 
  at <Navbar> `

PR #1713, opened 2026-07-17
chore(web-forms): move buildAttributes to instance root

latin-panda:

CI passes

Moves `buildAttributes.ts` out of `instance/attachments/` (which is for file upload attachments) and into `instance/` next to `Attribute.ts`, where it belongs. It's a generic utility used by all node types, not attachment-specific.

No

No


latin-panda:
@garethbowen This is a small refactor. It was [spotted by Sadiq](https://github.com/getodk/central-frontend/pull/1708#pullrequestreview-4696692044) while reviewing another of my PRs.

garethbowen (approved):
Thanks!

PR #1716, opened 2026-07-17
Bootstrap/update form def button

sadiqkhoja:

Follow-up of https://github.com/getodk/central-frontend/pull/1710

Test and manual verification

Just using the newly created component. And added `triggerId` using vue's `useId()`.

None

No

### DIFF
--- a/apps/central/src/components/dropdown.vue
+++ b/apps/central/src/components/dropdown.vue
@@ -1,14 +1,14 @@
 <template>
   <component :is="tag" ref="container" class="dropdown" :class="{ open: isOpen }">
     <slot name="toggle" :toggle="toggle" :is-open="isOpen" :attrs="triggerAttrs"></slot>
-    <ul ref="menu" class="dropdown-menu" :style="menuStyle">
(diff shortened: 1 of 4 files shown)


apps/central/src/components/dropdown.vue
```diff
@@ -28,12 +28,15 @@ const props = defineProps({
   }
 });
 
+const triggerId = `dropdown-${useId()}`;
```
sadiqkhoja, line 31: prefix is not necessarily required because useId scope is global. Add it for the html readability.

apps/central/src/components/dropdown.vue
```diff
@@ -28,12 +28,15 @@ const props = defineProps({
...
+
 const container = ref(null);
 const menu = ref(null);
 const isOpen = ref(false);
 const menuStyle = ref({});
 
 const triggerAttrs = computed(() => ({
+  id: triggerId,
```
matthew-white, line 39: This pattern makes a lot of sense to me. :+1: I guess it means that the `toggle` slot can't specify its own custom `id` for the toggle element. That seems OK though.

matthew-white (approved):

PR #1717, opened 2026-07-17
Show file size in form attachment table

### LINKED ISSUE #1895: Add file size to list of form attachments
[Figma](https://www.figma.com/design/ULpCuTKkIys6Y99YGyMcrF/Central?node-id=5213-1888&t=R8lk7DtruyNbdlJs-0)

As a project manager, I want to see each attachment's file size in a human-readable format (e.g. 14 KB, 821 B) so that I can validate that I've uploaded the right file(s) and to get a sense of how long form download will take for my users.

ktuite:
Closes https://github.com/getodk/central/issues/1895
Backend to add `size` field to form attachments:
- https://github.com/getodk/central-backend/pull/1877

Screenshot 2026-07-17 at 3 25 31 PM

Trying it and tests. 

Just pipes through size data now available in backend.

More helpful to users. 

No.

### DIFF
--- a/.changeset/neat-insects-wear.md
+++ b/.changeset/neat-insects-wear.md
@@ -1,0 +1,5 @@
+---
+"@getodk/central-frontend": patch
+---
+
+Show file size in form attachment table
--- a/.changeset/neat-insects-wear.md
+++ b/.changeset/neat-insects-wear.md
@@ -2,4 +2,4 @@
 "@getodk/central-frontend": patch
 ---
 
-Show file size in form attachment table
+Show file size in form attachment table (issue central#1895)
--- a/apps/central/src/components/form-attachment/row.vue
+++ b/apps/central/src/components/form-attachment/row.vue
@@ -76,14 +76,10 @@ except according to the terms contained in the LICENSE file.
 import DateTime from '../date-time.vue';
 
 import { apiPaths } from '../../util/request';
+import { useI18nUtils } from '../../util/i18n';
 import { useRequestData } from '../../request-data';
 
-const formatSize = (bytes) => {
-  if (bytes == null) return null;
-  if (bytes < 1024) return `${bytes} B`;
-  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} KB`;
-  return `${(bytes / (1024 * 1024)).toFixed(1)} MB`;
-};
+
 
 export default {
   name: 'FormAttachmentRow',
@@ -116,6 +112,7 @@ export default {
     // The component assumes that this data will exist when the component is
     // created.
     const { form } = useRequestData();
+    const { formatSize } = useI18nUtils();
     return { form, formatSize };
   },
   computed: {
--- a/apps/central/src/components/form-attachment/row.vue
+++ b/apps/central/src/components/form-attachment/row.vue
@@ -48,6 +48,11 @@ except according to the terms contained in the LICENSE file.
         </span>
       </template>
     </td>
+    <td class="form-attachment-list-size">
+      <template v-if="attachment.blobExists && attachment.size != null">
+        {{ formatSize(attachment.size) }}
+      </template>
+    </td>
     <td class="form-attachment-list-action">
       <div>
         <template v-if="attachment.datasetExists">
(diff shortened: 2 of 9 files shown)


apps/central/src/components/form-attachment/row.vue
matthew-white, line 1: What should be shown for entity lists? The Figma file on the issue does show entity lists. I vaguely remember it being discussed that in the case of an entity list, we should show the number of entities rather than the file size of the entities.csv (since the file size is not easy to compute).

apps/central/src/components/form-attachment/row.vue
```diff
@@ -73,6 +78,13 @@ import DateTime from '../date-time.vue';
 import { apiPaths } from '../../util/request';
 import { useRequestData } from '../../request-data';
 
+const formatSize = (bytes) => {
+  if (bytes == null) return null;
+  if (bytes < 1024) return `${bytes} B`;
+  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} KB`;
+  return `${(bytes / (1024 * 1024)).toFixed(1)} MB`;
```
matthew-white: Could you investigate what the best approach is in terms of i18n? I'm reading that other languages use different abbreviations, e.g., French uses ko for kilobyte ("o" for "octet"). I think you can leverage `Intl.NumberFormat` here (via Vue I18n `n()`/`$n()`).

**Step 1: Add new number formats to src/i18n.js**

Something like:

```js
for (const unit of ['byte', 'kilobyte', 'megabyte'])
  numberFormats[unit] = { style: 'unit', unit, unitDisplay: 'short', maximumFractionDigits: 1 };
```

**Step 2: Reference those new formats here in `formatSize()`**

Something like the following. You'll need to make `formatSize()` a method in order to use `this.$n()`.

```suggestion
  if (bytes < 1024) return this.$n(bytes, 'byte');
  if (bytes < 1024 * 1024) return this.$n(bytes / 1024, 'kilobyte');
  return this.$n(bytes / (1024 * 1024), 'megabyte');
```

apps/central/src/components/form-attachment/row.vue
```diff
@@ -73,6 +78,13 @@ import DateTime from '../date-time.vue';
 import { apiPaths } from '../../util/request';
 import { useRequestData } from '../../request-data';
 
+const formatSize = (bytes) => {
```
matthew-white: My intuition is that we'll want to use this function in other places as well. What do you think about adding it to src/util/i18n.js (the same file as e.g. `formatRange()`)? Then it could be easily unit-tested.

Also an idea we could come back to a future PR.

apps/central/src/components/form-attachment/table.vue
```diff
@@ -76,6 +77,8 @@ const dsHashset = computed(() =>
 
   th:first-child { width: 125px; }
   th:nth-child(2) { width: 250px; }
+  th:nth-child(3) { width: 200px; }
+  th:nth-child(4) { width: 100px; }
```
matthew-white: If we want the table to be 100% width, and we're using `table-layout: fixed`, I think we need to leave one column width unspecified (which is interpreted as "use the remaining width"). I.e., I think only one width should be added here, not two.

apps/central/src/components/form-attachment/row.vue
matthew-white: Not necessarily suggesting that you show the number of entities, but it might be worth checking with Nicole about the expected behavior.

ktuite: Checking with @nicoleorlowski! My vague recollection from discussion is to not show anything special for entity lists, but that might have come from pressure to try to squeeze this into a previous release.

apps/central/src/components/form-attachment/row.vue
ktuite: Tried something like this and it works! Also moved it out to the shared i18n utils.

apps/central/src/components/form-attachment/row.vue
ktuite: Nicole says let's not add this for entity lists right now.

.changeset/neat-insects-wear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@getodk/central-frontend": patch
+---
+
+Show file size in form attachment table
```
matthew-white: Would you mind referencing the GitHub issue here? I think we should try to reference it when possible when using changesets.

matthew-white (approved):

matthew-white:
Love the unit tests of `formatSize()` by the way!

PR #1718, opened 2026-07-22
fix(web-forms#856): trim whitespace from attachment filename when editing submissions

### LINKED ISSUE #856: Attachment doesn't load when editing a submission with whitespace in the filename
### Expected Behaviour

When editing an old submission in Web Forms, all attachments load correctly, regardless of how the XML was serialized. These submissions might have been created with an old Enketo version or via Central's API, which doesn't enforce strict attachment name validation (Web Forms doesn't produce attachment names with spaces)

### Observed Behaviour

Attachments fail to load when the submission XML contains surrounding whitespace in an attachment field's text content (e.g., `photo.png\n        ` instead of `photo.png`). This was observed while reviewing submissions from a user who sent them privately.  

### Reproduction Steps

This can be reproduced by editing a submission whose XML was submitted directly via Central API.

Form (XML or XLSX file): Any form with an image or file attachment field.

[upload-images.xml](https://github.com/user-attachments/files/30273346/image-edit-test.xml)

  Steps:
  1. Submit a form with an attachment via the API, with the attachment field as:
     
```
     <?xml version="1.0"?>
  <data id="image-edit-test">
    <name>test 33 44</name>
    <photo>1784022917619.jpg
    </photo>
    <new-photo>1784022920734.jpg     </new-photo>
    <meta>
      <instanceID>uuid:NEW-UUID</instanceID> <----- Change
      <deprecatedID>uuid:PREVIOUS-UUID</deprecatedID>
    </meta> 
  </data>

```

```sh
// POST submission XML file
curl -X POST http://localhost:8383/v1/projects/[PROJECT ID]/forms/[FORM ID]/submissions -H 'Authorization: Bearer TOKEN' -H "Content-Type: application/xml" --data-binary @[PATH TO SUBMISSION FILE]/test-submission.xml
// Upload the attachment files:
  curl -X POST 'http://localhost:8383/v1/projects/[PROJECT ID]/forms/[FORM ID]/submissions/uuid:[NEW-UUID]/attachments/1784022917619.jpg' -H 'Authorization: Bearer TOKEN' -H "Content-Type: image/jpeg"
  --data-binary @[PATH TO FILE]/any-image.jpg
```

  2. Open the submission for editing in web forms.
  3. Observe that the attachment field appears empty.

### Environment

- Browser version: All
- OS: All
- Web Forms version: 1.0.0

latin-panda:
Closes https://github.com/getodk/web-forms/issues/856

Added unit tests and manual testing.

Before:

After fix:

Trailing spaces can be difficult to notice, so the match is relaxed slightly to still display the attachment. For other mismatches where the filename cannot be determined or the file failed to load, an error message appears, and it's fixed here: https://github.com/getodk/central-frontend/pull/1719

Attachments are visible when the filename in the submission includes trailing spaces. 

No

### DIFF
--- a/.changeset/real-news-dream.md
+++ b/.changeset/real-news-dream.md
@@ -1,0 +1,5 @@
+---
+"@getodk/xforms-engine": patch
+---
+
+Trim whitespace from attachment filename when editing submissions
--- a/packages/xforms-engine/src/instance/attachments/InstanceAttachmentsState.ts
+++ b/packages/xforms-engine/src/instance/attachments/InstanceAttachmentsState.ts
@@ -34,7 +34,7 @@ export class InstanceAttachmentsState extends Map<InstanceAttachmentContext, Ins
       return null;
     }
 
-    const { value } = instanceNode;
+    const value = instanceNode.value.trim();
     const sourceFile = this.sourceAttachments?.get(value) ?? null;
     if (sourceFile != null) {
       return sourceFile;
@@ -51,7 +51,7 @@ export class InstanceAttachmentsState extends Map<InstanceAttachmentContext, Ins
 
   retryFileValue(instanceNode: StaticLeafElement | null) {
     if (instanceNode !== null) {
-      this.sourceAttachments?.retry(instanceNode.value);
+      this.sourceAttachments?.retry(instanceNode.value.trim());
     }
   }
 }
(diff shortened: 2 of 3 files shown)


latin-panda:
@garethbowen I found this yesterday while reviewing the user submission they sent me privately.

garethbowen (approved):
I think this works...

I was a little concerned because it is technically valid (but not recommended) to have filenames with leading or trailing whitespace. While web-forms generates the file name so this isn't possible, other clients may not, so editing could be an issue. To test this I modified the web-forms filename generation to include a leading whitespace on the main branch. The XML that was submitted included the whitespace, but somewhere in central this was trimmed off, so the expected filename was missing the leading whitespace.

Editing is a bit of a special case... if this modification had change the filename in any other way the result would be the same. But because xml formatting makes it more likely the filename will have leading and trailing whitespace this seems a reasonable patch.

PR #1719, opened 2026-07-22
fix(web-forms#857): show error indicator when submission attachment cannot be loaded

### LINKED ISSUE #857: Attachment field shows empty when file referenced in submission XML cannot be loaded
### Description
  
When editing a submission, if an attachment filename is present in the instance XML but the file cannot be resolved (e.g. the attachment was never uploaded, or was deleted from the server), the upload field silently shows the "drag and drop" placeholder as if no file was ever selected. There is no indication to the user that the field had a value that couldn't be loaded.

### Expected

 An error indicator is shown, consistent with how other attachment load failures are displayed.
  
### Observed

The field appears blank with no error or message.

### Reproduction Steps

Form (XML or XLSX file):

[upload-images.xml](https://github.com/user-attachments/files/30273346/image-edit-test.xml)

Steps:
1. Submit a new submission 
2. Block the request to fetch the attachment using the Developer Tools' network tab
3. Observe that the field is empty instead of showing an error

### Environment

- Browser version: All
- OS: All
- Web Forms version: 1.0.0

latin-panda:
Closes https://github.com/getodk/web-forms/issues/857

Unit test updated and manual test

Before:

After fix:

Not found error

Network error

If the filename exists but the file itself isn't found, show an error.

Clearer outcome when editing submissions, instead of mistakenly believing the question didn't have an answer. 

No

### DIFF
--- a/.changeset/dull-humans-join.md
+++ b/.changeset/dull-humans-join.md
@@ -1,0 +1,5 @@
+---
+"@getodk/xforms-engine": patch
+---
+
+Show error indicator when submission attachment cannot be loaded
--- a/packages/web-forms/src/components/form-elements/upload/UploadControl.vue
+++ b/packages/web-forms/src/components/form-elements/upload/UploadControl.vue
@@ -310,6 +310,9 @@ onUnmounted(() => {
 				<Message v-else-if="fileError" severity="error" :closable="true" @close="fileError = null">
 					{{ fileError }}
 				</Message>
+				<Message v-else-if="loadingError === 'not-found'" severity="error" :closable="false">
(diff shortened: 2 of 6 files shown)


latin-panda:
@garethbowen a follow up from https://github.com/getodk/central-frontend/pull/1718

garethbowen (approved):
Approving because this is a definite improvement.

If you felt like taking this further, there is a slight improvement that could be made to the error message shown to the user. In this case the error is not recoverable so there's no point showing the retry icon. Furthermore the message could be strengthened - essentially the form xml is inconsistent with the attachments that central expects, which can only be resolved by following up with an admin.

PR #1721, opened 2026-07-22
Use dropdown component in UserRow

sadiqkhoja:

Follow-up of https://github.com/getodk/central-frontend/pull/1710

Test and manual verification

Just using the newly created component

None

No


apps/central/src/components/user/row.vue
```diff
@@ -35,14 +35,14 @@ except according to the terms contained in the LICENSE file.
...
-        <button :id="actionsButtonId" type="button"
-          class="btn btn-default dropdown-toggle" data-toggle="dropdown"
-          aria-haspopup="true" aria-expanded="false">
-          <span class="icon-cog"></span><span class="caret"></span>
-        </button>
-        <ul class="dropdown-menu dropdown-menu-right"
-          :aria-labelledby="actionsButtonId">
+      <dropdown placement="bottom-end">
+        <template #toggle="{ toggle, attrs }">
+          <button type="button" class="btn btn-default dropdown-toggle"
```
matthew-white: No need for `dropdown-toggle` class here I think.

apps/central/src/components/user/row.vue
```diff
@@ -35,14 +35,14 @@ except according to the terms contained in the LICENSE file.
...
     </td>
     <td>
-      <div class="dropdown">
-        <button :id="actionsButtonId" type="button"
-          class="btn btn-default dropdown-toggle" data-toggle="dropdown"
-          aria-haspopup="true" aria-expanded="false">
-          <span class="icon-cog"></span><span class="caret"></span>
-        </button>
-        <ul class="dropdown-menu dropdown-menu-right"
-          :aria-labelledby="actionsButtonId">
```
matthew-white, line 45: Do we use `aria-labelledby` in the `Dropdown` component? Should we? (If we don't and should, that could come in a follow-up PR.)

matthew-white (approved):

apps/central/src/components/user/row.vue
matthew-white: Also kind of unexpected that this used `aria-labelledby` to point to the button, yet the button itself didn't have any text content or `aria-label`.

matthew-white: > Do we use `aria-labelledby` in the `Dropdown` component? Should we?

Ah I see you've added that as part of #1716. :+1:

> Also kind of unexpected that &hellip; the button itself didn't have any text content or `aria-label`.

This does feel like an issue. Maybe it's something to circle back to at some point. I think all toggle buttons should have either text or an `aria-label`. But also feels low-priority as a preexisting issue.

One idea is that `Dropdown` could pass an `aria-label` to the slot as part of its `attrs`. Though then the slot couldn't specify its own `aria-label`.

sadiqkhoja: I have added `:aria-label="$t('header.actions')"` for this button. I think parent of the `dropdown` component would be right place to decide label or aria-label.

PR #1723, opened 2026-07-23
fix(web-forms#852): remove sentry exceptions on non-response

### LINKED ISSUE #852: Unhandled exception when fetching submission/attachments on Apple devices
## Details

Sentry flagged an uncaught `TypeError: Load failed` in `web-form-renderer` when fetching a submission or attachment from Central while opening a submission's edit view. This is observed on Apple devices, in both Safari and Firefox. Likely in the `fetch(..)` calls within apps/forms.

https://getodk.sentry.io/issues/7606810471/events/22891f9b1f164095b32892c4e4b0df29/

https://getodk.sentry.io/issues/7601091138/events/abdd891edf8d474d96256bb2a2d1608e/

https://getodk.sentry.io/issues/7612192878/events/9df704b566ff4b77bd0fb7a044de3499/

garethbowen:
Closes getodk/web-forms#852

I'm not sure this is the best but submitting it to see what you think. This error is thrown on ios when the request just doesn't respond. This can happen if the user just goes offline, and in this case we needn't bother sentry.

If I'm right about where the error originated from (and I think I am) then it is being caught and handled and displayed properly to the user.

None.

None.

### DIFF
--- a/.changeset/young-trams-cheer.md
+++ b/.changeset/young-trams-cheer.md
@@ -1,0 +1,5 @@
+---
+"@getodk/forms": patch
+---
+
+Reduces sentry warnings for requests that don't respond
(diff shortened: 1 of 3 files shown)


latin-panda (approved):
Yes, it looks good to me

PR #1724, opened 2026-07-23
fix(web-forms#854): escape greater than symbols in markdown

### LINKED ISSUE #854: Select options fail to render when label starts with `>` character
### Expected Behaviour

It should render the label just like Collect does: 

### Reproduction Steps

Form (XML or XLSX file): 

[select-with-gt-lt.xml](https://github.com/user-attachments/files/30235394/select-with-gt-lt.xml)

Steps:
1. Load form
2. Start a new submission
3. See missing label

### Environment

- Browser version: any
- OS: any 
- Web Forms version: 1.0.0

garethbowen:
Closes getodk/web-forms#854

Tests added
Loaded the form in the issue

Blockquotes are not supported by web-forms at all as per the spec: https://docs.getodk.org/form-styling/
mdast was discovering the blockquote but the engine markdown formatting was discarding it because it didn't know what to do with it.
By escaping before submitting to mdast the `>` character is preserved.

Before nothing showed so I expect nobody was relying on this behaviour.

No.

### DIFF
--- a/.changeset/every-pears-sip.md
+++ b/.changeset/every-pears-sip.md
@@ -1,0 +1,6 @@
+---
+"@getodk/xforms-engine": patch
+"@getodk/web-forms": patch
+---
+
+Escapes greater than symbols in markdown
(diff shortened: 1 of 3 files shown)


latin-panda (approved):
It works well for the simple case. Other cases, like no space `>text` or `>>`, are not considered, and I think that's okay for now. Ready to merge!

PR #1726, opened 2026-07-24
chore: add unit test coverage for forms api utility

garethbowen:

### DIFF
--- a/apps/forms/test/utils/api.spec.ts
+++ b/apps/forms/test/utils/api.spec.ts
@@ -1,0 +1,214 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { getProject, getFormByFormId, getFormByEnketoId, getFormXml, getSubmissionAttachmentNames } from '../../src/utils/api';
+
+describe('Test api utility', () => {
+
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  describe('getSubmissionAttachmentNames', () => {
+
+    it('responds with an empty array when no attachments', async () => {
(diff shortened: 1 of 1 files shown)


apps/forms/test/utils/api.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+        json: () => Promise.resolve([{ name: 'a', exists: false },{ name: 'b', exists: true }]),
+      }));
+      const actual = await getSubmissionAttachmentNames(5, 'simple', 'abc');
+      expect(actual).toEqual(['b']);
+      expect(fetch).toHaveBeenCalledTimes(1);
+      expect(fetch).toHaveBeenCalledWith('/v1/projects/5/forms/simple/submissions/abc/attachments');
+    });
+
+    it('handles central errors', async () => {
+      const expected = { message: 'Not found', code: '404' };
```
latin-panda: The code is a number as per [this type](https://github.com/getodk/central-frontend/blob/8ab73a3b673393fccb236f46ca4947f462f204a2/apps/forms/src/utils/api.ts#L18)
```suggestion
      const expected = { message: 'Not found', code: 404 };
```

latin-panda: Same fix in L90, L148, L176 and L205

latin-panda (approved):
Nice addition! Thank you 💛

apps/forms/test/utils/api.spec.ts
garethbowen: Thanks. Fixed.

PR #1727, opened 2026-07-24
fix(central#2088): disable Sentry's trace headers

### LINKED ISSUE #859: Media files do not load in WebForms on test server
#### Problem description
Media files do not load in WebForms, additionaly when trying to open form in WebForms with geojson file attached there is "NetworkError when attempting to fetch resource" error.

Issue only occurs on test server
#### URL of the page
https://test.getodk.cloud/projects/513/forms/build_mediatest_1536186301/submissions
1. Open URL linked above
2. Tap on new submission
3. Check media files
#### Screenshot

Image

#### Expected behavior
Media should load correctly in WebForms
#### Central version shown in version.txt
versions:
ee0e4207ab6fb628247f4d5b5b576325782a5a05 (v2026.2.0)
 0000000000000000000000000000000000000000 client (v2026.2.0)
 bee93414a2215bd73d7f371113c6640c66d90c95 server (v2026.2.0)
#### Browser version
Firefox 152.0.0

latin-panda:
Closes https://github.com/getodk/web-forms/issues/859

Added an e2e test; however, we still need to test it on our test servers. 

[Reading: trace propagation](https://docs.sentry.io/platforms/javascript/tracing/distributed-tracing/#how-to-use-distributed-tracing)
[Disabling trace](https://docs.sentry.io/platforms/javascript/tracing/distributed-tracing/#disabling-distributed-tracing)

Skipping attachments from Sentry tracing headers because they redirect to S3, which doesn't allow those headers.

We keep the `browserTracingIntegration` because it's a client-side performance monitoring tool (page load and navigation timings) that's still useful. 

Stops blocking S3 requests for those who have configured Sentry.

No


apps/central/src/util/sentry.js
```diff
@@ -12,6 +12,8 @@ const initSentry = (app, source, dsn) => {
     // Captures performance timing for 1 in 5 page loads.
     // Could be extracted to VITE_SENTRY_TRACES_SAMPLE_RATE in the future.
     tracesSampleRate: 0.2,
+    // Disable trace headers: they break CORS on requests that redirect to S3.
+    tracePropagationTargets: [],
```
alxndrsn: What are these headers for?  Would it be better to disable specifically for requests to s3?

alxndrsn (changes requested):
Any chance of a test?

apps/central/src/util/sentry.js
latin-panda: This is for end-to-end tracing. The back-end also needs to be configured to provide useful insights. I think we need to define the goal and enable it carefully, with more testing. 🤔

latin-panda:
@alxndrsn I've added the test. If the headers are added, the error will appear like this:

apps/forms/src/utils/sentry.ts
```diff
@@ -12,9 +12,11 @@ const initSentry = (app: App, source: string, dsn: string | undefined) => {
...
+    // Disable trace headers: they break CORS on requests that redirect to S3.
+    tracePropagationTargets: [],
   });
 
   setTag('source', source);
 };
 
-export default initSentry;
\ No newline at end of file
+export default initSentry;
```
alxndrsn, line 23: It would be nice if PRs don't change random whitespace.  PR to catch this when linting at https://github.com/getodk/central-frontend/pull/1728.

e2e-tests/tests/sentry-headers.spec.js
```diff
@@ -0,0 +1,40 @@
...
+
+test.describe('Sentry request headers', () => {
+  test('does not attach sentry-trace or baggage header to requests', async ({ allowedLogs, page }) => {
+    allowedLogs.push((msg, text) => text.includes('401') && msg.location().url.includes('/v1/'));
+
+    await page.route('**/client-config.json', r => r.fulfill({
+      contentType: 'application/json',
+      body: JSON.stringify({ sentryDsn: 'https://public@o0.ingest.sentry.io/0' }),
+    }));
+    await page.route('**/*.ingest.sentry.io/**', r => r.fulfill({ status: 200, contentType: 'application/json', body: '{}' }));
```
alxndrsn: It's surprising to me that `**/*.ingest.sentry.io/` is for matching a host, whereas `**/v1/**` is for matching a path.  Could this line be something more specific, like:

```suggestion
    await page.route('https://*.sentry.io/**', r => r.fulfill({ status: 200, contentType: 'application/json', body: '{}' }));
```

?

e2e-tests/tests/sentry-headers.spec.js
```diff
@@ -0,0 +1,40 @@
...
+      v1RequestCount++;
+      const { 'sentry-trace': trace, baggage } = request.headers();
+      if (trace || baggage) {
+        violations.push({ url, 'sentry-trace': trace, baggage });
+      }
+    });
+
+    await page.goto(appUrl);
+    await expect(page.getByRole('heading', { name: 'Welcome to ODK Central' })).toBeVisible();
+    await page.waitForLoadState('networkidle');
```
alxndrsn: Should it be obvious what's triggering a call to Sentry here?

apps/central/src/util/sentry.js
alxndrsn: I think tracing is helpful already for contextualising errors, e.g. https://getodk.sentry.io/explore/traces/trace/e6f4f433ec244108ac3c0bd2cbf42422/

Would it be possible for these headers to be included in general, but not for s3 URLs?

apps/forms/src/utils/sentry.ts
```diff
@@ -12,9 +12,11 @@ const initSentry = (app: App, source: string, dsn: string | undefined) => {
     integrations: [browserTracingIntegration()],
     // Captures performance timing for 1 in 5 page loads.
     tracesSampleRate: 0.2,
+    // Disable trace headers: they break CORS on requests that redirect to S3.
+    tracePropagationTargets: [],
```
alxndrsn: (as above)

apps/central/src/util/sentry.js
alxndrsn: Additionally, IIUC the bug affects forms only, so maybe no changes are required in `apps/central`?

e2e-tests/tests/sentry-headers.spec.js
latin-panda: I added a new test with comments. Please let me know if it's clearer.

latin-panda:
@alxndrsn I applied your feedback, and it's ready for another look.

apps/forms/test/utils/sentry.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+  beforeAll(() => {
+    initSentry(createApp({ render: () => null }), 'web-forms-test', 'https://fake@sentry.io/1');
+  });
+
+  afterAll(async () => {
+    await getClient()?.close();
+    fetchSpy.mockRestore();
+  });
+
+  it('should omits Sentry headers on attachment requests', async () => {
```
alxndrsn:
```suggestion
  it('should omit Sentry headers on attachment requests', async () => {
```

typo

apps/forms/src/utils/sentry.ts
```diff
@@ -12,6 +12,8 @@ const initSentry = (app: App, source: string, dsn: string | undefined) => {
     integrations: [browserTracingIntegration()],
     // Captures performance timing for 1 in 5 page loads.
     tracesSampleRate: 0.2,
+    // Attachment downloads redirect to S3, where Sentry headers cause CORS errors.
+    tracePropagationTargets: [/^\/(?!.*\/attachments\/)/],
```
alxndrsn: A few queries:

* do we have to skip all calls to `/attachments/` because the headers will also be included in the redirected requests to s3?
* would this also match a form with `xmlFormId` of `attachments`?
* are there any other paths to blobs which are referenced from web-forms?  e.g. submission attachments(?)

An alternative might be to strip these headers from the `Location` header for blob redirects.

e2e-tests/tests/sentry-headers.spec.js
alxndrsn: :+1: much :+1:

apps/forms/src/utils/sentry.ts
latin-panda: > do we have to skip all calls to /attachments/ because the headers will also be included in the redirected requests to s3?  

Yes. It reuses the original request headers on the redirected request. And the client can't know in advance whether a given attachment request will redirect to S3. 

> would this also match a form with xmlFormId of attachments?

Yes. So you mean like `/v1/projects/1/forms/attachments/submissions/[submission-id]`, right? Sentry won't attach trace headers, so the request works fine; we just lose the frontend/backend trace link for that form's submission calls. I think that's still okay, to keep it safe and simple.

> are there any other paths to blobs which are referenced from web-forms? e.g. submission attachments(?)
  
Web Forms relies on apps/forms to make the fetch calls (both form attachments and submission attachments during edit), so the same fix covers it. Those requests are the responsibility of the hosting app, not the library (package/web-forms).
  
> An alternative might be to strip these headers from the Location header for blob redirects.

Sentry adds the headers on the client at the JS layer, before native fetch runs. The server has no channel to tell the browser "drop these on the follow-up". An alternative solution could be adding the `sentry-trace` and `baggage` to S3's `Access-Control-Allow-Headers` instead of fixing it here. But I think the PR solution is the simplest, so system admins don't need to remember to add those headers when configuring Sentry.

apps/forms/test/utils/sentry.spec.ts
latin-panda: Fixed!

latin-panda:
@alxndrsn I've fixed the typo and answered your questions :)

apps/forms/src/utils/sentry.ts
alxndrsn: > I think the PR solution is the simplest, so system admins don't need to remember to add those headers when configuring Sentry.

:+1: Additionally, the Sentry headers may cause CORS to be triggered for any requests to other (sub-)domains.


> > would this also match a form with xmlFormId of attachments?
>
> ... I think that's still okay, to keep it safe and simple.

If it's complicated to handle attachments more specifically, then I agree that it's good to get a broader fix out.  But I think over-filtering may be surprising down the line.  If that's the path taken, then a new Issue to make the filter more specific to attachments would be good.


> > are there any other paths to blobs which are referenced from web-forms? e.g. submission attachments(?)
> 
> Web Forms relies on apps/forms to make the fetch calls (both form attachments and submission attachments during edit), so the same fix covers it. Those requests are the responsibility of the hosting app, not the library (package/web-forms).

I don't understand this explanation, but it looks like both submission attachments and form attachments include `/attachment/` in their API paths, so should both be covered by this fix :+1:  Including a test case for each would be helpful to prevent future regression.

I've confirmed this issue does not affect form XLS(X) download, which is the only blob-response I can see from the API which doesn't include `/attachment/` in its path.

latin-panda:
@alxndrsn, this is ready for another look. Thank you!

apps/forms/test/utils/sentry.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+          draft: false,
+          webformsEnabled: true
+        },
+        actionType: 'new',
+      },
+    });
+    await flushPromises();
+
+    const calls = fetchSpy.mock.calls as FetchCall[];
+    const attachmentCalls = calls.find(([input]) => {
```
alxndrsn:
```suggestion
    const attachmentCall = calls.find(([input]) => {
```
Variable name implies plural, but `find()` returns max 1?

apps/forms/test/utils/sentry.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+      return getFetchUrl(input).includes('/attachments/') && getFetchUrl(input).includes('cities');
+    });
+    expect(attachmentCalls, 'expected the attachment fetch triggered by mount').toBeDefined();
+
+    const headers = readSentryHeaders(attachmentCalls!);
+    expect(headers['sentry-trace']).toBeNull();
+    expect(headers.baggage).toBeNull();
```
alxndrsn, line 105: Consider: 

```suggestion
    expect(headers['sentry-trace']).toBeNull();
    expect(headers['baggage']).toBeNull();
```

consistency -> easier reading

alxndrsn (approved):
Looks great.

Personally i'd inline `urlsWithTraceHeaders` & `urlsWithoutTraceHeaders` so that they're more closely associated with their uses.

Please open/link an Issue for tightening the `tracePropagationTargets` config.

apps/forms/test/utils/sentry.spec.ts
latin-panda: Fixed!

apps/forms/test/utils/sentry.spec.ts
latin-panda: I had that but lint complains
`124:22  error  ["baggage"] is better written in dot notation  @typescript-eslint/dot-notation`

alxndrsn: I'd say that's unhelpful linting :shrug:

PR #1714, opened 2026-07-17
chore: bump safe dependencies

latin-panda:
#### What has been done to verify that this works as intended?

smoke test

#### Why is this the best possible solution? Were any other approaches considered?

- Bumps in-range dependencies across all workspaces. No breaking changes, no major version jumps. 
Updating early to give us time to detect issues before release. 

- Updates package-lock.json with correct version that matches apps/central/package.json and apps/forms/package.json

#### How does this change impact users? Describe intentional behavior changes from code updates. What are the regression risks?

No

#### Does this change require updates to user documentation? If so, please file an issue [here](https://github.com/getodk/docs/issues/new) and include the link below.

No

PR #1728, opened 2026-07-25
eslint: enable rule: eol-last

alxndrsn:
Also fix violation.

- [x] tested locally
- [x] ci

Repeated whitespace changes add noise to PRs and make tracking changes to code over time more difficult, for no real gain.

No effect.

No.


latin-panda (approved):
Thank you!

apps/forms/eslint.config.js
```diff
@@ -316,6 +316,7 @@ export default defineConfig(
 					allowInterfaces: 'with-single-extends',
 				},
 			],
+      'eol-last': 'error',
```
alxndrsn: Looks like some enforcement of tabs vs spaces is also required...

package-lock.json
```diff
@@ -61,7 +61,7 @@
     },
     "apps/central": {
       "name": "@getodk/central-frontend",
-      "version": "0.1.0",
+      "version": "2026.2.0",
```
alxndrsn: Why has this file changed?

alxndrsn: Reverted.

apps/forms/eslint.config.js
alxndrsn: Created issue at https://github.com/getodk/central/issues/2090.

package-lock.json
latin-panda: IMO, the change is correct because it matches package.json and is probably a miss from the last release. But it's okay; I have a PR that proposes an update to this file too

https://github.com/getodk/central-frontend/pull/1714

alxndrsn: :+1: The change might be correct, but it's entirely unrelated to this PR.

PR #1730, opened 2026-07-28
feat(web-forms#464): prefill default values from url parameter

### LINKED ISSUE #464: Pre-filling using URL parameters
### Description
Support pre-filling form fields using URL parameters with XPath syntax, such as `&d[/data/uuid]=foo`, to match the functionality already available in Enketo. This feature would allow seamless pre-population of form fields (e.g., a text field named uuid) via URL parameters.

Form: 

[pre-fill using url parameter.xlsx](https://github.com/user-attachments/files/21468598/pre-fill.using.url.parameter.xlsx)

---

### Checklist
  - [x] No
  - [ ] No
  - [x] No
  - [ ] No
---

### User Stories

- **As a** [user type],
  **I want** [functionality],
  **so that** [benefit].

---

### Design

- **Figma**: [Insert URL]

---

### Dependencies

-

---

### Additional Notes

- Likely, Central needs to pass these parameters to Web Forms.
- Forum post describing the feature: [How can I specify default values in a web form link?](https://forum.getodk.org/t/how-can-i-specify-default-values-in-a-web-form-link/42847)
- Forum post reporting the missing feature: [Pre-filling using URL parameters does not work in ODK Web Forms](https://forum.getodk.org/t/pre-filling-using-url-parameters-does-not-work-in-odk-web-forms/55939)

---

### Related Issues

-

garethbowen:
Closes getodk/web-forms#464

Testing. Test plan to come...

One other option was to create an XML doc and bind that in to the form (a bit like when editing) but because there are multiple ways to set the value it's important to get the order right. This felt clean.

There are risks around malicious URLs allowing injection into the form, or modifying protected fields.

Yes. https://github.com/getodk/docs/issues/2080

### DIFF
--- a/apps/forms/src/components/preview.vue
+++ b/apps/forms/src/components/preview.vue
@@ -20,7 +20,7 @@ const formId = computed(() => encodeURIComponent(route.params.xmlFormId as strin
 const useWebForms = route.query.webforms === 'true';
 const loadingState = ref(true);
 const errorState = ref(false);
-const defaultParameters = computed(() => getDefaultParameters(route.query));
+const defaultParameters = computed(() => getDefaultParameters(route.query, 'preview'));
 const webFormsEnabled = ref(true);
 const form = ref<Form>();
 const xform = ref<string>();
--- a/apps/forms/src/components/preview.vue
+++ b/apps/forms/src/components/preview.vue
@@ -3,6 +3,7 @@ import { computed, ref, defineAsyncComponent } from 'vue';
 import { useRoute } from 'vue-router';
 import { getFormByFormId, getFormXml, RequestError, type Form } from '../utils/api.ts';
 import { hideSpinner } from '../utils/spinner.ts';
+import { getDefaultParameters } from '../utils/default-parameters.ts';
 
 const props = defineProps({
   draft: Boolean,
@@ -19,6 +20,7 @@ const formId = computed(() => encodeURIComponent(route.params.xmlFormId as strin
 const useWebForms = route.query.webforms === 'true';
 const loadingState = ref(true);
 const errorState = ref(false);
+const defaultParameters = computed(() => getDefaultParameters(route.query));
 const webFormsEnabled = ref(true);
 const form = ref<Form>();
 const xform = ref<string>();
@@ -59,7 +61,7 @@ fetchForm();
       {{ $t('formNotFound') }}
     </div>
     <template v-else-if="webFormsEnabled">
-      <WebFormRenderer :form="form!" :xform="xform!" action-type="preview"/>
+      <WebFormRenderer :form="form!" :xform="xform!" action-type="preview" :default-parameters="defaultParameters" />
     </template>
     <template v-else>
       <EnketoIframe :form="form!" :enketo-id="form!.enketoId" action-type="preview"/>
--- a/apps/forms/src/components/submission.vue
+++ b/apps/forms/src/components/submission.vue
@@ -14,6 +14,7 @@ import {
 } from '../utils/api.ts';
 import Dialog from 'primevue/dialog';
 import { hideSpinner } from '../utils/spinner.ts';
+import { getDefaultParameters } from '../utils/default-parameters.ts';
 
 const props = defineProps({
   draft: Boolean,
@@ -58,20 +59,7 @@ const enketoId = computed(() => route.params.enketoId ? encodeURIComponent(route
 const st = computed(() => route.query.st ? route.query.st as string : null);
 const useWebForms = computed(() => route.query.webforms === 'true');
 const offline = computed(() => route.params.offline === 'offline');
-const defaultParameters = computed(() => {
-  // TODO pull this out and unit test it
-  const DEFAULT_PARAMETERS_REGEX = /^d\[(.+)\]$/;
-  const result: Record<string, string> = {};
-  Object.entries(route.query).forEach(([key, value]) => {
-    if (typeof value === 'string') {
-      const path = DEFAULT_PARAMETERS_REGEX.exec(key)?.[1];
-      if (path) {
-        result[path] = value;
-      }
-    }
-  });
-  return result;
-});
+const defaultParameters = computed(() => getDefaultParameters(route.query));
 const webFormsEnabled = ref(true);
 
 const form = ref<Form>();
--- a/apps/forms/src/components/submission.vue
+++ b/apps/forms/src/components/submission.vue
@@ -59,7 +59,7 @@ const enketoId = computed(() => route.params.enketoId ? encodeURIComponent(route
 const st = computed(() => route.query.st ? route.query.st as string : null);
 const useWebForms = computed(() => route.query.webforms === 'true');
 const offline = computed(() => route.params.offline === 'offline');
-const defaultParameters = computed(() => getDefaultParameters(route.query));
+const defaultParameters = computed(() => getDefaultParameters(route.query, props.actionType));
 const webFormsEnabled = ref(true);
 
 const form = ref<Form>();
--- a/apps/forms/src/components/submission.vue
+++ b/apps/forms/src/components/submission.vue
@@ -58,13 +58,16 @@ const enketoId = computed(() => route.params.enketoId ? encodeURIComponent(route
 const st = computed(() => route.query.st ? route.query.st as string : null);
 const useWebForms = computed(() => route.query.webforms === 'true');
 const offline = computed(() => route.params.offline === 'offline');
-const prefillParameters = computed(() => {
-  const PREFILL_PARAMETERS_REGEX = /^d\[(.+)\]$/;
-  const result = {};
+const defaultParameters = computed(() => {
+  // TODO pull this out and unit test it
+  const DEFAULT_PARAMETERS_REGEX = /^d\[(.+)\]$/;
+  const result: Record<string, string> = {};
   Object.entries(route.query).forEach(([key, value]) => {
-    const path = PREFILL_PARAMETERS_REGEX.exec(key)?.[1];
-    if (path) {
-      result[path] = value;
+    if (typeof value === 'string') {
+      const path = DEFAULT_PARAMETERS_REGEX.exec(key)?.[1];
+      if (path) {
+        result[path] = value;
+      }
     }
   });
   return result;
@@ -265,7 +268,7 @@ load();
         :instance-id="instanceId"
         :action-type="props.actionType ?? 'new'"
         :submission-attachments="submissionAttachments"
-        :prefill-parameters="prefillParameters"
+        :default-parameters="defaultParameters"
         :st="st"
       />
     </template>
--- a/apps/forms/src/components/submission.vue
+++ b/apps/forms/src/components/submission.vue
@@ -58,6 +58,17 @@ const enketoId = computed(() => route.params.enketoId ? encodeURIComponent(route
 const st = computed(() => route.query.st ? route.query.st as string : null);
 const useWebForms = computed(() => route.query.webforms === 'true');
 const offline = computed(() => route.params.offline === 'offline');
+const prefillParameters = computed(() => {
+  const PREFILL_PARAMETERS_REGEX = /^d\[(.+)\]$/;
+  const result = {};
+  Object.entries(route.query).forEach(([key, value]) => {
+    const path = PREFILL_PARAMETERS_REGEX.exec(key)?.[1];
+    if (path) {
+      result[path] = value;
+    }
+  });
+  return result;
+});
 const webFormsEnabled = ref(true);
 
 const form = ref<Form>();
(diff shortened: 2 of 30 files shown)


packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -33,9 +34,11 @@ const isEditInitialLoad = (context: ValueContext) => {
 };
 
 const getInitialValue = (context: ValueContext): string => {
-  const sourceNode = context.instanceNode ?? context.definition.template;
-
-  return context.decodeInstanceValue(sourceNode.value);
+  const value =
```
latin-panda, line 37: This prefill via URL parameters should not override submissions on edit. If not there already, could you please add a test that covers this scenario?

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -33,9 +34,11 @@ const isEditInitialLoad = (context: ValueContext) => {
...
+    getInstanceDefaultValue(context.instanceConfig.instanceDefaults, context) ??
+    context.instanceNode?.value ??
+    context.definition.template?.value;
+  return context.decodeInstanceValue(value);
```
latin-panda, line 41: Could you add a test post submission to confirm that the form state is reset and that it prefills using URL parameters?

apps/forms/src/components/submission.vue
```diff
@@ -58,6 +59,7 @@ const enketoId = computed(() => route.params.enketoId ? encodeURIComponent(route
 const st = computed(() => route.query.st ? route.query.st as string : null);
 const useWebForms = computed(() => route.query.webforms === 'true');
 const offline = computed(() => route.params.offline === 'offline');
+const defaultParameters = computed(() => getDefaultParameters(route.query));
```
latin-panda: When in edit submission mode, it shouldn't pass these default parameters.

apps/forms/src/components/preview.vue
```diff
@@ -59,7 +61,7 @@ fetchForm();
       {{ $t('formNotFound') }}
     </div>
     <template v-else-if="webFormsEnabled">
-      <WebFormRenderer :form="form!" :xform="xform!" action-type="preview"/>
+      <WebFormRenderer :form="form!" :xform="xform!" action-type="preview" :default-parameters="defaultParameters" />
```
latin-panda, line 64: We kinda call this feature "pre-fill from parameters", which could be an alternative name instead of `:default-parameters`

apps/forms/test/utils/default-parameters.test.ts
```diff
@@ -0,0 +1,57 @@
...
+      'd[/root/path/that/is/deep]': 'c'
+    };
+    expect(getDefaultParameters(given)).toEqual({
+      '/root/path': 'a',
+      'shortform': 'b',
+      '/root/path/that/is/deep': 'c',
+    });
+  });
+
+  it('returns values with special characters', () => {
```
latin-panda, line 63: What does the regex match if someone sends something like `d[a][b] `?

packages/xforms-engine/src/lib/instance-defaults.ts
```diff
@@ -0,0 +1,42 @@
...
+  return !!PROTECTED_META_FIELDS.find((field) => {
+    return refWithoutRoot === `${OPENROSA_XFORMS_PREFIX}:meta/${OPENROSA_XFORMS_PREFIX}:${field}`;
+  });
+};
+
+export const getInstanceDefaultValue = (
+  defaults: InstanceDefaults,
+  context: ValueContext
+): string | undefined => {
+  if (!defaults || context.contextNode.nodeType === 'attribute') {
```
latin-panda: According to the parameter definition, `defaults` is never undefined or null. So this IF statement can be simplified, or the parameter definition should be `InstanceDefaults | undefined`.

packages/xforms-engine/src/lib/instance-defaults.ts
```diff
@@ -0,0 +1,42 @@
...
+    return;
+  }
+  const ref = context.contextReference();
+  const positionOfSecondSlash = ref.indexOf('/', 1);
+  const refWithoutRoot = ref.slice(positionOfSecondSlash + 1);
+  if (isProtectedMetaProperty(refWithoutRoot)) {
+    return;
+  }
+  return defaults[ref] ?? (refWithoutRoot && defaults[refWithoutRoot]) ?? undefined;
```
latin-panda, line 42: I understand that keys must match the full XPath paths, right?
So the second case is not valid because the root is necessary, isn't it?

packages/xforms-engine/src/lib/instance-defaults.ts
```diff
@@ -0,0 +1,42 @@
+import { OPENROSA_XFORMS_PREFIX } from '@getodk/common/constants/xmlns';
+import type { InstanceValueContext } from '../instance/internal-api/InstanceValueContext';
+import type { AttributeContext } from '../instance/internal-api/AttributeContext';
+import type { InstanceDefaults } from '../client';
+
+type ValueContext = AttributeContext | InstanceValueContext;
+
+const PROTECTED_META_FIELDS = [
```
latin-panda, line 8: If not already there, could you add tests for these cases?
- Check that the fields with preloaded values are not overridden.  For example, have the deviceID,  pass a URL default for it. And the test asserts the final value is the preload value, not the URL value.
- Now the other way, have instanceName (with no preloaded value), pass a URL default. The test asserts the URL default is rejected.

packages/xforms-engine/src/lib/instance-defaults.ts
```diff
@@ -0,0 +1,42 @@
...
+  'deviceID',
+  'deprecatedID',
+  'email',
+  'phoneNumber',
+  'audit',
+];
+
+const isProtectedMetaProperty = (refWithoutRoot: string): boolean => {
+  return !!PROTECTED_META_FIELDS.find((field) => {
+    return refWithoutRoot === `${OPENROSA_XFORMS_PREFIX}:meta/${OPENROSA_XFORMS_PREFIX}:${field}`;
```
latin-panda: We also support meta without a prefix, right?

packages/xforms-engine/src/entrypoints/FormInstance.ts
```diff
@@ -42,6 +42,7 @@ export class FormInstance<
       clientStateFactory: instanceConfig.stateFactory ?? identity,
       computeAttachmentName: instanceConfig.instanceAttachments?.fileNameFactory ?? (() => null),
       preloadProperties: instanceConfig.preloadProperties ?? {},
+      instanceDefaults: instanceConfig.instanceDefaults ?? {},
```
latin-panda, line 45: If it's defined here and coming from InstanceConfig, is it still necessary to have `instanceDefaults` in packages/xforms-engine/src/client/form/LoadForm.ts?

packages/xforms-engine/src/client/form/FormInstanceConfig.ts
```diff
@@ -16,6 +16,8 @@ export interface PreloadProperties {
   readonly phoneNumber?: string;
 }
 
+export interface InstanceDefaults extends Record<string, string> {}
```
latin-panda, line 25: The `Record<string, string>` doesn't provide enough information about the expected key shapes. Is there a way to make it clearer? If explaining it with types is hard, a comment will help a lot.

latin-panda (changes requested):
Great work! This feature needs a lot of testing due to many potential scenarios. It's worth investing in unit test coverage to prevent future regressions. 

I haven't completed my review yet, I still need to consider the security, repeats and edge cases. Looking forward to seeing the test plan!

apps/forms/src/components/preview.vue
garethbowen: I struggled with naming and added a comment to the issue with my rationale for where this ended up: https://github.com/getodk/web-forms/issues/464#issuecomment-5097357558

While we have been calling it "pre-fill", in Enketo it's always "default", and the param is `d` which is short for "default" so I wanted to stay consistent there. I'm definitely open to better naming if you have thoughts!

packages/xforms-engine/src/lib/instance-defaults.ts
garethbowen: There's a little known and undocumented feature in Enketo where you can leave it off because 99% of the time you're binding into the primary instance so it saves a bit in the URL. So both `d[/root/name]` and `d[name]` are equivalent. I've added this to the related docs PR and there are tests which make sure it works.

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
garethbowen: Thanks! Added test, failed, fixed. I decided to just null out the defaults in the engines editInstance function long before it gets here.

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
garethbowen: Done.

apps/forms/src/components/submission.vue
garethbowen: It's now ignored at the engine level but I've also added a check here so we don't waste the time checking for params that are going to be ignored.

apps/forms/test/utils/default-parameters.test.ts
garethbowen: Vue parses that to: `{ "d[a][b]": "123" }`
Then the regex matches and sets the default params to: `{ "a][b": "123" }`

Nothing complains but it's never going to match a ref so it's a no-op.

packages/xforms-engine/src/lib/instance-defaults.ts
garethbowen: Updated this to check if defaults has any keys so we can short circuit the code if it'll never match (which is the 99% case).

packages/xforms-engine/src/lib/instance-defaults.ts
garethbowen: I think this is covered by the "instance-defaults" unit test which checks that checking for the default for any of these fields returns undefined, which means the preload works how it always did.

packages/xforms-engine/src/lib/instance-defaults.ts
garethbowen: True! I've added a test for the various combinations of namespaces.

packages/xforms-engine/src/entrypoints/FormInstance.ts
garethbowen: No you're right. I think that was left over from a previous idea. Removed now. Thanks!

packages/xforms-engine/src/client/form/FormInstanceConfig.ts
garethbowen: Because the key can be either the full xpath or the short form it's difficult to type. I've added a comment. Let me know if that explains it sufficiently!

e2e-tests/tests/default-parameters.spec.js
```diff
@@ -0,0 +1,50 @@
...
+  backendClient = new BackendClient(playwright, `${testInfo.project.name}_wf`, projectId);
+  await backendClient.alwaysHideModal();
+  simpleForm = await backendClient.createForm();
+});
+
+test.afterAll(async () => {
+  await backendClient.dispose();
+});
+
+test.describe('Default Parameters', () => {
```
latin-panda, line 21: Could you also assert the preview URL and public links?

packages/xforms-engine/test/integration/instance-defaults.test.ts
```diff
@@ -0,0 +1,183 @@
...
+      scenario.createNewRepeat({ assertCurrentReference: '/root/repeat' });
+      expect(scenario.answerOf('/root/repeat[1]/child')).toEqualAnswer(stringAnswer(''));
+      expect(scenario.answerOf('/root/repeat[2]/child')).toEqualAnswer(stringAnswer('gregory'));
+      expect(scenario.answerOf('/root/repeat[3]/child')).toEqualAnswer(stringAnswer(''));
+    });
+
+    it('ignored instance defaults when editing', async () => {
+      const scenario = await Scenario.init('Bind defaults', formDefinition);
+      scenario.answer('/root/name', 'original');
+      const instanceDefaults = {
```
latin-panda, line 115: Could you please also add a datetime field and another field with an invalid value, such as an invalid geopoint or date, so the codecs work with the prefill feature and help catch future regressions?

latin-panda (approved):
Excellent! Just a couple of suggestions below. I also checked other scenarios, including repeats, security, memory leaks, and performance.

e2e-tests/tests/default-parameters.spec.js
garethbowen: I don't think the preview URL warrants a full e2e test because I don't think it'll ever be used. The public link is good though, as this is the primary use case. Added.

packages/xforms-engine/test/integration/instance-defaults.test.ts
garethbowen: Good idea. Added.

PR #1731, opened 2026-07-28
eslint: enforce indentation (tabs or spaces)

alxndrsn:
Also fix violations.

This PR aims to introduce rules which most closely reflect the current code, and minimise changes.

This omits applying any indent rule in `packages/eslint.config.js`, as there is extensive usage of both tabs & spaces in the `packages` directory.

Ref https://github.com/getodk/central/issues/2090

- [x] ran locally
- [x] ci

Because consistency is good, and minimising accidental & incidental whitespace changes is helpful in keeping PRs focussed.

No effect - this isn't Python.

No.


package.json
```diff
@@ -9,7 +9,9 @@
     "build": "turbo run build",
     "build:web": "vite build",
     "changeset": "changeset",
-    "eslint": "eslint bin/ e2e-tests/ *.js -c bin/eslint.config.js --cache --max-warnings 0",
+    "eslint": "npm-run-all -p eslint:bin eslint:e2e-tests",
+    "eslint:bin": "eslint bin/ *.js -c bin/eslint.config.js --cache --max-warnings 0",
+    "eslint:e2e-tests": "eslint e2e-tests/ *.js -c e2e-tests/eslint.config.js --cache --max-warnings 0",
```
latin-panda: These two use `--cache`. How does `eslintcache` work when running `npm-run-all -p eslint:bin eslint:e2e-tests` (L12)? Are the caches appended or do they override each other? Maybe it needs a cache file for each of these (e.g. `.eslintcache.bin` and `.eslintcache.e2e`)

apps/forms/eslint.config.js
```diff
@@ -40,375 +40,376 @@ const vueFileExtensions = [...typeScriptFileExtensions, 'vue'];
...
+
+      // See explanation of typescript-eslint recommendation above
+      'no-undef': 'off',
+
+      // Consistent with all other source code
+      'vue/html-indent': ['error', 'tab'],
+      'vue/html-comment-indent': ['error', 'tab'],
+      'vue/script-indent': [
+        'error',
+        'tab',
```
latin-panda, line 359: It'd be good to update this for consistency. This refers to the TypeScript inside the Vue components.

bin/eslint.config.js
```diff
@@ -9,11 +9,12 @@ export default defineConfig(
...
       },
     },
-		extends: [
-			eslint.configs.recommended,
+    extends: [
+      eslint.configs.recommended,
     ],
     rules: {
       'eol-last': 'error',
+      'indent': [ 'error', 2, { SwitchCase:1, MemberExpression:2 } ],
```
latin-panda, line 17: The other configs in apps/central and apps/forms don't use `MemberExpression:2`. Should it be added?

latin-panda (commented):
Good contribution! I’ve included a few comments below.

package.json
alxndrsn: Great question!

apps/forms/eslint.config.js
alxndrsn: With this PR, I'm aiming for minimal changes to source files, and consistency within them, rather than cross-repo consistency and a big changeset.  With that in mind, should this change?

bin/eslint.config.js
alxndrsn: No - that would require significant changes to indentation across the project.  The aim of this PR is to introduce rules which most closely reflect the current code.  I'll update the PR description to reflect this.

package.json
latin-panda: Is the cache appending, or does it need separate cache files?

alxndrsn: :facepalm: forgot to check!

alxndrsn: > Is the cache appending

Maybe?  But even if it does, it's unpredictable running parallel tasks which update the same file.  I've added the `--cache-location` args to each `eslint` invocation so that they have their own separate cache files.

See: https://eslint.org/docs/latest/use/command-line-interface#--cache-location

latin-panda (approved):
Thank you!

PR #1729, opened 2026-07-27
Bootstrap/remove bootstrap file

sadiqkhoja:
Addresses part of https://github.com/getodk/central/issues/1278

Dependent on https://github.com/getodk/central-frontend/pull/1720

#### What has been done to verify that this works as intended?

Manual verification

#### Why is this the best possible solution? Were any other approaches considered?

Just deleted the bootstrap.js as all dependent components of `dropdown` have been upgraded.

#### How does this change impact users? Describe intentional behavior changes from code updates. What are the regression risks?

None

#### Does this change require updates to user documentation? If so, please file an issue [here](https://github.com/getodk/docs/issues/new) and include the link below.

No

PR #1732, opened 2026-07-28
Bootstrap/remove bootstrap modal

sadiqkhoja:

Dependent on https://github.com/getodk/central-frontend/pull/1729

Test sand manual verification

- Removed the separate overlay element, as it simplified our code. Bootstrap likely had it to support animations or multiple simultaneous modals.
- Removed the `resize` emit since it was not used anywhere.
- Skipped Bootstrap's `handleUpdate` method, which only exists to prevent a slight leftward shift of the modal when the scrollbar appears. The [source code](https://github.com/twbs/bootstrap/blob/68b0d231a13201eb14acd3dc84e51543d16e5f7e/js/modal.js#L295) for this looks overly complex for such a minor UI adjustment. On modern OSes (at least macOS and Windows), scrollbar is shown above the content only when scrolling, so this fix isn't really necessary. If needed later, I'd just nudge the modal slightly left to compensate for the scrollbar.
- I also considered removing everything related to `has-scroll` to simplify the modal component further, but decided against it. For the full-size modal, my plan is to make the modal body scrollable instead of the `.modal` element itself. I'll open an issue with more details on this.

None

No

### DIFF
--- a/apps/central/test/components/dataset/property/new.spec.js
+++ b/apps/central/test/components/dataset/property/new.spec.js
@@ -1,3 +1,5 @@
+import { nextTick } from 'vue';
+
 import DatasetPropertyNew from '../../../../src/components/dataset/property/new.vue';
 import DatasetProperties from '../../../../src/components/dataset/overview/dataset-properties.vue';
 
@@ -50,8 +52,9 @@ describe('DatasetPropertyNew', () => {
       hide: '.btn-link'
     }));
 
-  it('focuses the input', () => {
+  it('focuses the input', async () => {
     const modal = mount(DatasetPropertyNew, mountOptions({ attachTo: document.body }));
+    await nextTick();
     modal.get('input').should.be.focused();
   });
 
--- a/apps/central/test/components/entity/update.spec.js
+++ b/apps/central/test/components/entity/update.spec.js
@@ -1,3 +1,5 @@
+import { nextTick } from 'vue';
+
 import EntityUpdate from '../../../src/components/entity/update.vue';
 import EntityUpdateRow from '../../../src/components/entity/update/row.vue';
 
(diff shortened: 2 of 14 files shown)


apps/central/test/components/dataset/property/new.spec.js
```diff
@@ -50,8 +52,9 @@ describe('DatasetPropertyNew', () => {
       hide: '.btn-link'
     }));
 
-  it('focuses the input', () => {
+  it('focuses the input', async () => {
     const modal = mount(DatasetPropertyNew, mountOptions({ attachTo: document.body }));
+    await nextTick();
```
sadiqkhoja, line 57: understand this is really required if state is true initially.

apps/central/test/components/entity/update.spec.js
```diff
@@ -83,7 +86,8 @@ describe('EntityUpdate', () => {
     const textareas = modal.findAll('textarea');
     textareas.map(({ element }) => element.style.height).should.eql(['', '']);
     await modal.setProps({ state: true });
-    await modal.vm.$nextTick();
+    await nextTick();
+    await nextTick();
```
sadiqkhoja, line 90: remove one?

matthew-white (approved):

apps/central/test/components/dataset/property/new.spec.js
sadiqkhoja: it is required because emit.shown is emitted after onMounted event and inside `nextTick()`

apps/central/test/components/entity/update.spec.js
sadiqkhoja: no, because resizing of textarea happens after @shown event and inside `nextTick()`

PR #1733, opened 2026-07-28
Bootstrap/remove jquery

### LINKED ISSUE #1278: Replace Bootstrap plugins
We use several Bootstrap 3 plugins. It'd be nice to remove them, for a few reasons:

- The plugins require jQuery, which is one of our heaviest dependencies. We have removed all other uses of jQuery.
- The plugins don't always play well with Vue. For example, our `Popover` component has significant limitations due to its use of the Bootstrap popover plugin.
- Bootstrap 3 has reached EOL. I doubt that it'd be simple to migrate to Bootstrap 4 or 5. We also don't really want to continue using Bootstrap from a design perspective.

These are the Bootstrap plugins that we currently use:

- [ ] `collapse`
  - This is just needed for the `Navbar` component I think.
  - Possibly it won't be needed anymore once we implement getodk/central#777
- [ ] `popover` + `tooltip`
  - These are needed for the `Popover` component.
  - We use Floating UI for our tooltips, and I'd like to replace these plugins with Floating UI.
- [ ] `modal`
  - This is needed for the `Modal` component.
  - I doubt this plugin is doing that much, so I think we should just implement the modal functionality ourselves. Our `Modal` component does a lot beyond what the plugin does, so I think it'd be better if we could control everything related to it rather than using an external component.
- [ ] `dropdown`
  - Unlike some of the other plugins, this isn't used in a single generic component. Instead, it's used throughout Frontend, mostly via `data` attributes.
  - I think the main thing that the plugin's doing is positioning the dropdown menu once the dropdown has been opened. We could probably use Floating UI for that.
- [ ] `transition`
  - This is imported, but I'm actually not sure that we use it for anything.

sadiqkhoja:
Closes https://github.com/getodk/central/issues/1278

All tests pass. Manual smoke testing.

Just removal of jquery

None

No


apps/central/package.json
```diff
@@ -18,7 +18,6 @@
     "bootstrap": "~3",
     "dompurify": "^3.4.11",
     "flatpickr": "~4.6",
-    "jquery": "~3",
```
sadiqkhoja, line 21: update the package.lock.json

matthew-white (approved):

PR #1734, opened 2026-07-30
package-lock: add missing peer dep: solid-js

alxndrsn:
No idea why, but when I run `npm i` locally, `solid-js` gets added as a peer dependency of `packages/xforms-engine`.

- [x] ci

I don't know what's going on, but it stops my local working directory from being constantly dirty.

No impact on users.

No.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22943,6 +22943,9 @@
         "node": "^24.16.0",
         "npm": "11"
       },
+      "peerDependencies": {
+        "solid-js": "^1.9.7"
+      },
       "peerDependenciesMeta": {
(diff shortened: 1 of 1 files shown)


garethbowen (approved):
Pending build.

This is fine.

I don't know why this got out of sync but I worry it's going to be something like different npm versions so it may oscillate...

alxndrsn:
> I don't know why this got out of sync but I worry it's going to be something like different npm versions so it may oscillate...

I'm using the same version of `npm` as CI (`11.13.0`), and I think you have a different version but still see this change.  So I think it's safe to merge.

PR #1735, opened 2026-07-30
xforms-engine: remove duplicate solid-js dependency

alxndrsn:
It was previously listed as both a dependency, and a peer dependency.

Before that, it was listed as both a **dev** dependency, and a peer dependency, which made more sense.  When it was moved from dev dependency to full dependency, the `peerDependencies` entry became redundant.

- [x] ci

No other approaches were considered.

No effect.

No.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22942,14 +22942,6 @@
       "engines": {
         "node": "^24.16.0",
         "npm": "11"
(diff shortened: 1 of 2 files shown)


garethbowen (approved):
Thanks!

PR #1737, opened 2026-07-30
fix(web-forms#846): setvalue order wrong on second update

### LINKED ISSUE #846: Order of execution of setvalue actions is unexpected

garethbowen:
Closes getodk/web-forms#846

Integration test was modified to set the value a second time. Initially this failed but after patch it passed.
Also checked the QA form and everything is behaving as expected.

Now there is one value changed listener that iterates over the setvalue elements, which puts us in control of the execution order rather than relying on solid to do it.

None.

No.

### DIFF
--- a/packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
+++ b/packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
@@ -267,9 +267,13 @@ const registerSetValueActions = (
 };
 
 const registerValueChangedActions = (context: ValueContext, getValue: Accessor<string>) => {
-  context.valueChangedActions?.forEach((action) => {
-    let previous = getValue();
-    createComputed(() => {
+  if (!context.valueChangedActions?.length) {
+    return;
+  }
+  let previous: string;
+  createComputed(() => {
+    const sourceValue = getValue();
+    context.valueChangedActions.forEach((action) => {
       const ref = bindRefToRepeatInstance(context, action.ref);
       const destinationNodes = context.evaluator.evaluateNodes(ref, {
         contextNode: context.contextNode,
@@ -283,8 +287,6 @@ const registerValueChangedActions = (context: ValueContext, getValue: Accessor<s
         destinationNode.isAttached() &&
         context.isAttached()
       ) {
-        const sourceValue = getValue();
-
         if (
           previous !== undefined &&
           previous !== sourceValue &&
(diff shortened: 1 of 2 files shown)


latin-panda (approved):
LGTM!

PR #1740, opened 2026-08-02
fix(web-forms#862): catch error accessing disabled localstorage

### LINKED ISSUE #862: Gracefully handle localStorage disabled
### Expected Behaviour

Forms should still be able to be completed even when localStorage is blocked.

### Observed Behaviour

An uncaught error is thrown. Reported in sentry.

### Reproduction Steps

Form (XML or XLSX file): any

Steps:
1. Disable localstorage: https://stackoverflow.com/questions/6600754/how-can-i-browse-with-localstorage-disabled
2. Load any form

### Environment

- Browser version: Firefox
- OS: Linux
- Web Form s version: 1.0.0

garethbowen:
Closes getodk/web-forms#862

Testing loading form with localstorage enabled and then disabled.
I checked for other references to localstorage and all seem to be safe.

It's the way.

None.

None.

### DIFF
--- a/.changeset/cute-showers-switch.md
+++ b/.changeset/cute-showers-switch.md
@@ -1,0 +1,5 @@
+---
+"@getodk/forms": patch
+---
+
+Fixed an uncaught promise rejection reading from localStorage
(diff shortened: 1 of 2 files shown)


latin-panda (approved):
Looks good!

PR #1742, opened 2026-08-05
feat(web-forms#306): implement last-saved virtual secondary instance

garethbowen:
getodk/web-forms#306

Testing.

Discussion regarding a few options, and why I chose this one, on the issue: https://github.com/getodk/web-forms/issues/306#issuecomment-5171900195

Forms which have the last-saved defined will start using it.
Data will be stored in idb, max 1 submission per form.
There is no way for the end user to opt out, other than to disable idb, but the form designer can opt out by removing the last-saved instance.

It's already documented.


latin-panda:
I'm still reviewing and testing. I'll continue tomorrow :)

apps/forms/src/utils/last-saved.ts
```diff
@@ -0,0 +1,33 @@
+import { get, set, del } from 'idb-keyval';
+
+const getIdbKey = (projectId: number, xmlFormId: string) => {
+  return `last-saved-${projectId}-${xmlFormId}`;
+};
+
+export const getLastSaved = async (projectId: number, xmlFormId: string): Promise<string | undefined> => {
+  const key = getIdbKey(projectId, xmlFormId);
+  try {
+    return get(key);
```
latin-panda: If it doesn't await and throws, it won't reach the catch, and Sentry will report it. Better add `await`, the function already has `async` anyways. 
```suggestion
    return await get(key);
```

Same feedback for `setLastSaved` and `deleteLastSaved`

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -246,14 +249,22 @@ const handleSubmit = async (
...
     // Status is not ready when Form is not valid and in that case submit button will be disabled,
     // hence this branch should never execute.
     return;
   }
   initializeSubmissionState(data as unknown as SubmissionData, clearFormCallback);
   await submitData();
+
+  if (!isEdit.value) {
+    if (hasLastSaved) {
+      setLastSaved(props.form.projectId, props.form.xmlFormId, submissionData.instanceFile);
```
latin-panda: If the submission (submitData) fails, it shows a modal but doesn't throw an error. It's best to check if it was successful before setting the lastSaved

packages/xforms-engine/src/parse/model/SecondaryInstance/SecondaryInstancesDefinition.ts
```diff
@@ -35,29 +35,34 @@ export interface SecondaryInstanceRootDefinition extends StaticElement {
   getAttributeValue(localName: string): string | null;
 }
 
+const createLastSavedInstance = (
+  domElement: DOMSecondaryInstanceElement,
+  instanceId: string,
+  resourceURL: JRResourceURL,
+  xml: string | undefined
+) => {
+  if (xml) {
```
latin-panda, line 44: If the data from "last saved" is malformed, should it be ignored, or should something be said to the user?

packages/xforms-engine/src/entrypoints/FormResult/BaseInstantiableFormResult.ts
```diff
@@ -63,10 +62,14 @@ export abstract class BaseInstantiableFormResult<
       });
     };
 
-    this.resetInstance = (instanceConfig: FormInstanceConfig = {}) => {
+    this.resetInstance = (instanceConfig: FormInstanceConfig = {}, lastSavedXml?: string) => {
```
latin-panda: It seems a bit odd to have lastSaved in a reset function.
I was thinking something along these lines, for example:

```
form.setLastSaved(xml);                                                                                                                                                                                    
const instance = form.resetInstance(config);
```

packages/xforms-engine/test/integration/smoketests/form-init.test.ts
```diff
@@ -216,46 +216,46 @@ describe('Form initialization smoke tests', () => {
...
-      // DataInstance lastSaved = formDef.getNonMainInstance("last-saved");
-      // AbstractTreeElement root = lastSaved.getRoot();
-      // AbstractTreeElement item = root
-      // 		.getChild("head", 0)
-      // 		.getChild("model", 0)
-      // 		.getChild("instance", 0)
-      // 		.getChild("data", 0)
-      // 		.getChild("item", 0);
-      // assertEquals("Foo", item.getValue().getDisplayText());
+    it('parses last saved instance with filled form', async () => {
```
latin-panda, line 226: Could you add a test in which editing a submission doesn't pull values from "last saved"?

apps/forms/src/utils/last-saved.ts
```diff
@@ -0,0 +1,33 @@
+import { get, set, del } from 'idb-keyval';
+
+const getIdbKey = (projectId: number, xmlFormId: string) => {
+  return `last-saved-${projectId}-${xmlFormId}`;
```
latin-panda, line 4: Can forms from a public link (anonymous) use "last-saved"? 🤔

apps/forms/src/utils/last-saved.ts
```diff
@@ -0,0 +1,33 @@
...
+  } catch {
+    return;
+  }
+};
+
+export const setLastSaved = async (projectId: number, xmlFormId: string, file: File): Promise<void> => {
+  const key = getIdbKey(projectId, xmlFormId);
+  try {
+    const xml = await file.text();
+    return set(key, xml);
```
latin-panda: Should data saved in IDB always have an expiration date? Maybe if it's too old, it could be more annoying or confusing than helpful.

latin-panda (approved):
I finished testing, and it looks good. I have some questions below and minor suggestions.

latin-panda (changes requested):
😅 Sorry, I picked the wrong option before.
Please let me know when it's good to review again

apps/forms/src/utils/last-saved.ts
garethbowen: Good spotting. Fixed.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Good thinking. Done.

packages/xforms-engine/src/parse/model/SecondaryInstance/SecondaryInstancesDefinition.ts
garethbowen: Nice. I've caught any error and decided to just swallow it and return a blank instance. I don't want to block the user, and allowing them to submit again allows the app to overwrite the bad idb value and fix the problem.

packages/xforms-engine/src/entrypoints/FormResult/BaseInstantiableFormResult.ts
garethbowen: I see what you mean. I'm not really excited about setting it on the result either. I've put it on the config - let me know what you think.

packages/xforms-engine/test/integration/smoketests/form-init.test.ts
garethbowen: If I understand what you mean, that's the next test (line 240) in this file?

apps/forms/src/utils/last-saved.ts
garethbowen: The implementation works. Whether or not we should allow it is another question. I think until we support only filling in the form once there's no reason not to allow last-saved, is there?

apps/forms/src/utils/last-saved.ts
garethbowen: I think it's too context dependant - for some forms the last-saved should expire quickly and other times slowly.

I had a look at the Collect codebase and I can't see anywhere where they check the date or delete the file after a certain time, so this is consistent at least...

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -241,6 +241,16 @@ const webFormLoaded = () => {
   hideSpinner();
 };
 
+const updateLastSaved = (hasLastSaved) => {
+  if (!isEdit.value && submissionResult.primaryInstanceResult.success) {
```
latin-panda: Draft form data is saved in IndexedDB. When publishing the form and opening a new submission, it's prefilled with the draft data, which was surprising because I didn't immediately remember I had submitted a draft. 

The indexedDB shouldn't store draft form data, or at least not mix it with published form data.

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -241,6 +241,16 @@ const webFormLoaded = () => {
...
+    if (hasLastSaved) {
+      setLastSaved(props.form.projectId, props.form.xmlFormId, submissionData.instanceFile);
+    } else {
+      deleteLastSaved(props.form.projectId, props.form.xmlFormId);
+    }
```
latin-panda, line 252: When `hasLastSaved` is false, what is the intention of this deletion? Most forms don't declare a last-saved instance at all, so this fires on almost every submission and deletes a key that was never written. Is it cleaning up after a form version that dropped last-saved? 

Could you add a comment in the code?

apps/forms/src/utils/last-saved.ts
latin-panda: I'm considering how to handle cases when a new form version is introduced. For instance, if there's a last-saved record in the indexedDB with a select option that no longer exists, or an image upload with an unsupported extension (e.g. the form now only supports png). 

Should the last-saved feature track the form version to discard the entire last-saved state, or let the engine handle incorrect answers (empty fields, or show constraint messages if configured)?

packages/xforms-engine/src/client/form/FormInstanceConfig.ts
```diff
@@ -51,4 +51,6 @@ export interface FormInstanceConfig {
   readonly instanceDefaults?: InstanceDefaults;
 
   readonly geolocationProvider?: GeolocationProvider;
+
+  readonly lastSavedXml?: string | undefined;
```
latin-panda, line 57: This interface is used in 4 methods: createInstance, resetInstance, editInstance, and restoreInstance. However, `lastSavedXml` is only used in `resetInstance`, so that's not immediately clear to the developer when reading this interface. Especially since it's not relevant for editInstance.

A clearer way would be to create a new interface that extends this one, for example:
```
 export interface ResetFormInstanceConfig extends FormInstanceConfig {                                                                                                                              
    readonly lastSavedXml?: string | undefined;                                                                                                                                                      
  }  
```

packages/xforms-engine/src/parse/model/SecondaryInstance/SecondaryInstancesDefinition.ts
```diff
@@ -35,29 +35,38 @@ export interface SecondaryInstanceRootDefinition extends StaticElement {
...
+    try {
+      const resource = ExternalSecondaryInstanceResource.loadXml(instanceId, resourceURL, xml);
+      return new XMLExternalSecondaryInstanceSource(domElement, resource);
```
latin-panda: The parseDefinition() called in a couple of places in this file can still throw; it needs a try/catch too 
```suggestion
      const source = new XMLExternalSecondaryInstanceSource(domElement, resource);
       source.parseDefinition();   
       return source;
```

latin-panda (changes requested):
@garethbowen I did another full review, and left some comments below:

apps/forms/src/utils/last-saved.ts
garethbowen: Because it generally fails silently I like trying to bind regardless of version. This way if the version made a backwards compatible change it will just work. One possible downside was if the nature of the field had changed (ie: "do you have cats?" to "do you have dogs?"), but that would break any existing submissions too so I hope nobody would do that.

I think it's ok to ignore the version but let me know what you think.

apps/forms/src/components/web-form-renderer.vue
garethbowen: >  Is it cleaning up after a form version that dropped last-saved?

That's exactly the idea. Comment added.

I know it's a long shot but given how hard it is for end users to clear their data I thought it made sense to be proactive here.

packages/xforms-engine/src/client/form/FormInstanceConfig.ts
garethbowen: Nice. Done.

packages/xforms-engine/src/parse/model/SecondaryInstance/SecondaryInstancesDefinition.ts
garethbowen: Nice. I didn't like parsing the definition twice, so I've refactored it.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Fixed.

garethbowen: I also updated the code that fetches the last saved so it doesn't fetch for drafts or edits.

latin-panda (approved):
I did a smoke test and found no issues. It's ready!

We noticed last-saved doesn't indicate data may be stored for months, as it has no expiration date and can't be removed in Central UI (it could contain personal data). @lognaturel suggested adding a warning to the docs to clarify its permanence. Could you update the docs with this warning please?

garethbowen:
PR raised for docs addition: https://github.com/getodk/docs/pull/2087

PR #1743, opened 2026-08-05
async-route: show load error in console

alxndrsn:
For the end-user, there is already a toast displayed when a load error occurs, reading:

> The page you requested could not be loaded. Please refresh the page and try again.

Without this logging, there is no indication to a developer what the error actually is.

<kbd>

Screenshot_2026-08-05_08-54-33

</kbd>

- [x] ran locally
- [x] ci

If bundle size or prod behaviour are sacrosanct, the logging could be filtered from the final bundle somehow?

Most users won't encounter these errors, and if they do then they're unlikely to notice the logging.

No.

### DIFF
--- a/apps/central/src/components/async-route.vue
+++ b/apps/central/src/components/async-route.vue
@@ -111,8 +111,9 @@ export default {
             this.component = markRaw(m.default);
           }
         })
-        .catch(() => {
+        .catch(err => {
           if (!canceled) {
+            console.error('async-route', 'loadError:', err);
             // It would be ideal to show a more informative error message.
             // However, the error seems to provide limited information. For
             // example, if there is a 404 because Frontend was rebuilt, the
--- a/apps/central/src/components/async-route.vue
+++ b/apps/central/src/components/async-route.vue
@@ -113,7 +113,7 @@ export default {
         })
         .catch(err => {
           if (!canceled) {
-            console.error('async-route', 'loadError:', err); // eslint-disable-line no-console
+            logger.error('async-route', 'loadError:', err);
 
             // It would be ideal to show a more informative error message.
             // However, the error seems to provide limited information. For
--- a/apps/central/src/components/async-route.vue
+++ b/apps/central/src/components/async-route.vue
@@ -32,7 +32,7 @@ import { noop } from '../util/util';
 export default {
   name: 'AsyncRoute',
   components: { Loading, PageBody },
-  inject: ['alert', 'location'],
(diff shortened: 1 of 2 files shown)


apps/central/src/components/async-route.vue
```diff
@@ -111,8 +111,10 @@ export default {
             this.component = markRaw(m.default);
           }
         })
-        .catch(() => {
+        .catch(err => {
           if (!canceled) {
+            console.error('async-route', 'loadError:', err); // eslint-disable-line no-console
```
matthew-white:
```suggestion
            logger.error('async-route', 'loadError:', err);
```

Tests will flag unexpected console errors and may fail. What we usually do in a case like this is use `logger` from the `container` object, which suppresses such logging during testing.

matthew-white:
Adding this logging sounds reasonable to me. :+1: We do similar logging when there's a normal request error (e.g., `useRequest()` calls `logAxiosError()`).

That said, I wouldn't expect to see these errors during local development. The main case I've seen it is when a server has been upgraded.

apps/central/src/components/async-route.vue
matthew-white: Sorry, it's an Options API component, so it'd need to be `this.logger.error()`. You can specify `logger` under `inject`.

alxndrsn:
> That said, I wouldn't expect to see these errors during local development.

Can be triggered with a broken import:

```patch
$ git diff
diff --git a/apps/central/src/components/user/edit/password.vue b/apps/central/src/components/user/edit/password.vue
index 7da343aa6..c95f52d7e 100644
--- a/apps/central/src/components/user/edit/password.vue
+++ b/apps/central/src/components/user/edit/password.vue
@@ -40,7 +40,7 @@ except according to the terms contained in the LICENSE file.
 
 <script>
 import FormGroup from '../../form-group.vue';
-import Spinner from '../../spinner.vue';
+import Spanner from '../../spinner.vue';
 
 import useRequest from '../../../composables/request';
 import { apiPaths } from '../../../util/request';
```

matthew-white:
Ooh good point. OK maybe I'm totally misremembering and this error does come up. 🤔

matthew-white:
Just wanted to make sure you saw the follow-up to my code suggestion. My first suggestion wasn't quite right. 😅

apps/central/src/components/async-route.vue
alxndrsn: :+1: fixed

matthew-white (approved):

PR #1744, opened 2026-08-05
user/edit/password: check haveibeenpwned.com

### LINKED ISSUE #2150: Check haveibeenpwned when setting a password
Whenever a user sets a password in Central, check haveibeenpwned for the appearance of the password in a previous security breach.

- When the user sets their password for the first time
- When the user changes their password on their account page
- Probably also when the user sets an encryption password?

For now, we are considering this functionality only for Central Frontend. If a user sets a password using the API or CLI, Central Backend will not check haveibeenpwned.

Related:

- getodk/central-frontend#1744
- #2106

alxndrsn:
Closes https://github.com/getodk/central/issues/2150

<kbd>
Screenshot_2026-08-12_10-15-24
</kbd>

- [x] tested manually
- [x] ci
- [ ] new tests

* request to haveibeenpwned.com API could be done server-side, but it introduces some issues with 
* include a timeout in the `fetch()` call?
* manually cache layer could be removed; maybe haveibeenpwned.com API already offers HTTP caching which would be adequate?

* help users choose better passwords
* might make changing password slower or more frustrating, especially if haveibeenpwned API is slow or down

I don't think so.


apps/central/src/components/user/edit/password.vue
```diff
@@ -62,15 +74,17 @@ export default {
...
     };
   },
   methods: {
     validate() {
       this.tooShort = false;
       this.mismatch = false;
+      this.pwned = false;
 
-      if (this.newPassword.length < 10) {
+      if (this.newPassword.length < 1) {
```
alxndrsn: revert

apps/central/src/components/user/edit/password.vue
```diff
@@ -86,26 +105,45 @@ export default {
...
   }
 };
 </script>
 
 <style lang="scss">
 #user-edit-password input[autocomplete="username"] { display: none; }
+.collapsible-error {
+  display: grid;
+  grid-template-rows: 1fr;
+  color: #de2a11;
```
matthew-white: How about using one of the Sass variables like `$color-danger`? You'd need to `@import` the `variables` file.

apps/central/src/components/user/edit/password.vue
matthew-white, line 1: Will some of these changes be needed in the `AccountClaim` component as well? That's the other place where users enter a new password. Or are you thinking to do that in a follow-up PR?

apps/central/src/components/user/edit/password.vue
```diff
@@ -24,7 +24,18 @@ except according to the terms contained in the LICENSE file.
           autocomplete="current-password"/>
         <form-group id="user-edit-password-new-password" v-model="newPassword"
           type="password" :placeholder="$t('field.newPassword')" required
-          :has-error="tooShort || mismatch" autocomplete="new-password"/>
+          :has-error="tooShort || mismatch || pwned" autocomplete="new-password">
+          <template #after>
+            <transition name="collapse">
+              <div v-if="pwned" class="collapsible-error">
+                <div class="collapsible-inner">
+                  <p>This password has previously been included in a breach.</p>
```
matthew-white: Needs i18n/translation:

1. Add this English text under `<i18n>` below (under `en`).
2. Use the message key here, e.g., `<p>{{ $t('includedInBreach') }}</p>`

apps/central/src/components/user/edit/password.vue
```diff
@@ -24,7 +24,18 @@ except according to the terms contained in the LICENSE file.
           autocomplete="current-password"/>
         <form-group id="user-edit-password-new-password" v-model="newPassword"
           type="password" :placeholder="$t('field.newPassword')" required
-          :has-error="tooShort || mismatch" autocomplete="new-password"/>
+          :has-error="tooShort || mismatch || pwned" autocomplete="new-password">
+          <template #after>
+            <transition name="collapse">
+              <div v-if="pwned" class="collapsible-error">
```
matthew-white, line 30: If this `collapsible-error` class will be used elsewhere (in `AccountClaim`?), how about either:

1. Making a little component out of it
    - The component would use a Vue slot
2. Adding the class to app.scss

apps/central/src/components/user/edit/password.vue
```diff
@@ -24,7 +24,18 @@ except according to the terms contained in the LICENSE file.
...
+                <div class="collapsible-inner">
+                  <p>This password has previously been included in a breach.</p>
+                  <p>For more information, see <a href="https://haveibeenpwned.com/Passwords" target="_blank" rel="noopener noreferrer">here</a>.</p>
```
matthew-white: As long as "click here" is OK (instead of "see here"), you could reuse the existing i18n message `moreInfo.clickHere` (defined in en.json5). Because the text contains markup (a link), it needs to use Vue I18n "component interpolation".

```suggestion
                  <i18n-t tag="p" keypath="moreInfo.clickHere.full">
                    <template #clickHere>
                      <a href="https://haveibeenpwned.com/Passwords" target="_blank" rel="noopener noreferrer">{{ $t('moreInfo.clickHere.clickHere') }}</a>
                    </template>
                  </i18n-t>
```

apps/central/src/components/user/edit/password.vue
```diff
@@ -62,13 +74,20 @@ export default {
...
       tooShort: false,
       confirm: '',
-      mismatch: false
+      mismatch: false,
+      pwned: false,
     };
   },
+  watch: {
+    newPassword() {
+      this.pwned = false;
```
matthew-white, line 87: It sounds reasonable to me to reset `this.pwned` as soon as `this.newPassword` changes. At the same time, we don't do that for `this.tooShort` or `this.mismatch`: we only reset those in the `validate()` method. I think it'd be good to be consistent between the three. How about:

1. Removing this watcher for now and only resetting `this.pwned` in `validate()`
2. Maybe a follow-up PR adding a watcher that resets `this.pwned` along with `this.tooShort` and `this.mismatch`

apps/central/src/components/user/edit/password.vue
```diff
@@ -86,26 +105,45 @@ export default {
...
-      })
-        .then(() => {
-          this.alert.success(this.$t('alert.success'));
 
-          // The Chrome password manager does not realize that the form was
-          // submitted. Should we navigate to a different page so that it does?
-        })
-        .catch(noop);
+      (async () => {
+        const isPwned = await checkPasswordPwnage(this.request, this.newPassword);
```
matthew-white: I'm a big fan of `async`/`await`, but I'm less in favor of mixing `async`/`await` with `.then()`/`.catch()`. We already have a promise chain below with `.then()`/`.catch()`, so I think it'd be better just to make it all one promise chain. Something like:

```js
checkPasswordPwnage(this.request, this.newPassword)
  .then(isPwned => {
    if (isPwned) {
      this.pwned = true;
      return Promise.resolve();
    } else {
      const data = { old: this.oldPassword, new: this.newPassword };
      return this.request({ /* ... */ })
        .then(() => { /* ... */ })
        .catch(noop); // or maybe this .catch() could go at the very bottom?
    }
  });
```

matthew-white: FWIW I've struggled using `async`/`await` in Frontend to send requests. That's why there's so much `.then()`/`.catch()` in the codebase even though I like `async`/`await`. The issue is that ultimately we do need to catch the error. I've thought about creating a wrapper function to take an `async` function and automatically catch the error. Suppressing errors isn't usually best practice haha, but it'd be helpful for cases like here where we know there's already automatic error handling.

```js
const noopCatch = (f) => {
  try {
    await f();
  } catch (error) {}
};
```

If we had that, we could do something like this here:

```js
await noopCatch(async () => {
  const isPwned = await checkPasswordPwnage(this.request, this.newPassword);
  if (isPwned) {
    this.pwned = true;
  } else {
    const data = { old: this.oldPassword, new: this.newPassword };
    await this.request({ /* ... */ });
    this.alert.success(this.$t('alert.success'));
  }
});
```

matthew-white: I'm noticing the use of `this.newPassword` here and below, but I'm realizing that the `<form-group>` isn't disabled during the request. So users could theoretically change `this.newPassword` between the pwned check and the request to Backend. We don't usually disable form fields during requests, but I think that'd be a reasonable choice here in order to prevent this unlikely but awkward possibility. Alternatively, you could save `this.newPassword` to a local variable at the start of `validate()`, then use that local variable instead of the potentially changing `this.newPassword`.

apps/central/src/util/password.js
```diff
@@ -0,0 +1,33 @@
...
+  return hashArray
+    .map(b => b.toString(16).padStart(2, '0'))
+    .join('')
+    .toUpperCase();
+}
+
+async function getSuffixesFor(request, prefix) {
+  try {
+    const url = `https://api.pwnedpasswords.com/range/${prefix}`;
+    const res = await request({ url });
```
matthew-white: Usually we automatically show a toast after an error response. That's not needed here, so you can specify `alert: false`:

```suggestion
    const res = await request({ url, alert: false });
```

apps/central/test/components/user/edit/password.spec.js
matthew-white, line 1: How about adding a test for the case where the pwned response is an error? Usually we use `.respondWithProblem()` for error responses, but since this isn't a response from Backend (it isn't a Backend Problem), you can use the more generic `.respond()`:

```js
.respond(() => ({ status: 500 }))
```

matthew-white, line 1: How about adding a test for the case where the password has been pwned? The test could assert that:

```js
const formGroups = component.findAllComponents(FormGroup);
formGroups.length.should.equal(3);
const formGroup = formGroups[1];
formGroup.props().hasError.should.be.true;
formGroup.find('.collapsible-error').exists().should.be.true;
```

apps/central/test/components/user/edit/password.spec.js
```diff
@@ -27,6 +27,36 @@ const submit =
       : (!tooShort ? 'testPasswordZ' : 'z'));
     return component.get('#user-edit-password form').trigger('submit');
   };
+const haveIbeenPwnedRequest = password => {
```
matthew-white, line 30: If `haveIbeenPwnedRequest()` and `haveIbeenPwnedResponse()` will eventually be needed for test/components/account/claim.spec.js as well, then maybe it'd be good to put them somewhere reusable. Maybe in a new test/util/password.js.

matthew-white (commented):
It's looking nice! I've added several comments, but a lot of them are just hints for working with the Frontend codebase.

apps/central/src/components/user/edit/password.vue
alxndrsn: > We don't usually disable form fields during requests

Why not?

apps/central/src/components/user/edit/password.vue
alxndrsn: > At the same time, we don't do that for `this.tooShort` or `this.mismatch`

I'd like this to change.  Would that be OK?  If so, would you prefer a prior to watch these values?

apps/central/src/components/user/edit/password.vue
alxndrsn: https://github.com/getodk/central/issues/2106 should mean that the collapsible error is an implementation of the `<new-password-input>` component.

apps/central/src/components/user/edit/password.vue
alxndrsn: https://github.com/getodk/central/issues/2106 should mean that the collapsible error is an implementation of the `<new-password-input>` component.

apps/central/src/components/user/edit/password.vue
matthew-white: IIRC issa had a preference along those lines. The idea was to not lock down parts of the UI if we don't need to. If the user wants to edit those fields during the request for some reason (perhaps anticipating an error response), we don't restrict them from doing so.

In terms of code complexity, it's probably a wash. It's a little extra to disable fields during requests, but usually it's just one or two attributes or props. But this is the downside of not disabling them, that we have to consider the possibility of the data property changing during the request.

I don't have a strong preference either way. Maybe it's something we could ask Nicole or the larger team on Slack.

apps/central/src/components/user/edit/password.vue
matthew-white: Changing that sounds good to me. :+1: I'm happy for that to happen in a prior PR or a follow-up PR or this PR. Mostly I just want `this.pwned` to behave in a similar way as `this.tooShort` and `this.mismatch`.

matthew-white: > would you prefer a prior to watch these values?

I'm thinking we only need one watcher, on `newPassword`, where the watcher can reset all of `this.pwned`, `this.tooShort`, and `this.mismatch`.

alxndrsn: Oops, looks like there's a word missing.  Should have read:

> would you prefer a prior **PR** to watch these values?

matthew-white: A prior PR sounds good to me. :+1: But I'm also happy for it to happen in this PR or in a follow-up PR. Basically whatever's easiest as long as it's done in time for the release.

apps/central/src/components/user/edit/password.vue
matthew-white: That sounds good to me for work on the `<new-password-input>` component to come after this PR is merged. Anything that's added to this component can be extracted later.

apps/central/test/components/user/edit/password.spec.js
matthew-white: This is also something we can come back to as part of getodk/central#2106.

apps/central/src/components/user/edit/password.vue
alxndrsn: Fixed.

.github/workflows/tests.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         cache: 'npm'
     - run: npm ci
     - run: npm run build
-    - run: npm run lint
+    - run: npm run lint -w=@getodk/central-frontend
```
alxndrsn: revert

.github/workflows/tests.yml
```diff
@@ -65,85 +65,3 @@ jobs:
...
-    name: 'Test web-forms packages'
-    needs:
-    - check-and-build
-    uses: ./.github/workflows/wf-ci.yml
-
-  wf-e2e-tests:
-    name: 'Test web-forms e2e'
-    needs:
-    - wf-tests
-    uses: ./.github/workflows/wf-ci-e2e.yml
```
alxndrsn: revert

apps/central/src/components/user/edit/password.vue
alxndrsn: There's a weird race condition here - `awaitingResponse` is only set `true` once a request has started, but actually:

* there are 2 separate requests, so it may become true and then false and then true and then false again during a normal submission, and
* any async work done before the first request starts can delay setting of `awaitingResponse` for an unknown period

alxndrsn: Replaced derived `awaitingResponse` with manual `submitInProgress`.

apps/central/src/components/user/edit/password.vue
alxndrsn: Now the whole form is disabled while it submits.

apps/central/test/components/user/edit/password.spec.js
```diff
@@ -27,8 +27,42 @@ const submit =
...
+          '1C31795B5ECF960907E0E9CA1B3863B0762:999999',
+        ];
+      default: throw new Error(`No haveibeenpwned API response defined for password '${password}'`);
+    }
+  })();
+
+  return () => hashes.join('\r\n');
+};
+
+describe.only('UserEditPassword', () => {
```
alxndrsn, line 65: revert

apps/central/src/components/user/edit/password.vue
```diff
@@ -18,20 +18,37 @@ except according to the terms contained in the LICENSE file.
...
+                  </div>
+                </div>
+              </transition>
+            </template>
+          </form-group>
+          <form-group id="user-edit-password-confirm" v-model="confirm"
+            type="password" :placeholder="$t('field.passwordConfirm')" required
+            :has-error="mismatch" autocomplete="new-password"/>
+          <button type="submit" class="btn btn-primary"
+            :aria-disabled="awaitingResponse">
```
alxndrsn, line 48: This is now functionally pointless.  It should be removed, but this will require changes to the test for "normal button" behaviour.

PR #1747, opened 2026-08-05
ci/wf-e2e: add timeouts

alxndrsn:
This job took 42 minutes: https://github.com/getodk/central-frontend/actions/runs/30985056489/job/92238449922?pr=1743

Normal run is < 3 minutes: https://github.com/getodk/central-frontend/actions/runs/30971906331/job/92198578466

Timeouts on these jobs would have been helpful.

Specific cause looks like it was https://github.com/getodk/central/issues/2060.

- [x] ci

* could add timeouts in all the other places they're missing, but this PR at least improves on the status quo
* timeouts chosen are 2x the most recent `master` builds for each job, hopefully allowing for some normal deviation

No change - just CI.

No.


latin-panda (approved):
Thank you!

PR #1749, opened 2026-08-05
Table redesign/zebra rows

sadiqkhoja:
Towards https://github.com/getodk/central/issues/1604

image

Manual verification

Just added zebra stripe style to the table freeze and simple tables. For Form Access table removed the redundant css. 

I'll add rounded border on table in subsequent PR.

Form list table has different background color for review state column, I'll adjust that with the new state badges.

No

No

### DIFF
--- a/apps/central/src/components/project/home-block.vue
+++ b/apps/central/src/components/project/home-block.vue
@@ -209,12 +209,20 @@ export default {
       border-bottom-left-radius: 5px;
     }
 
-    tr:nth-child(3n + 2 of .project-form-row) {
-      background: #eee;
+    tr:nth-child(even of .project-form-row) {
+      background: $color-even-row;
     }
 
-    tr:nth-child(3n + 2 of .project-dataset-row) {
-      background: #eee;
+    tr:nth-child(odd of .project-form-row) {
+      background: $color-odd-row;
+    }
+
+    tr:nth-child(even of .project-dataset-row) {
+      background: $color-even-row;;
+    }
+
+    tr:nth-child(odd of .project-dataset-row) {
+      background: $color-odd-row;
     }
 
     .col-icon {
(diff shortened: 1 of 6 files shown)


apps/central/src/components/table/freeze.vue
```diff
@@ -163,6 +163,22 @@ defineExpose({ getRowPair });
...
 
+// Zebra striping
+.table-freeze-frozen,
+.table-freeze-scrolling {
+  thead tr {
+    background-color: #fff;
+  }
+
+  tbody tr:nth-child(even) {
+    background-color: #F6F6F6;
```
sadiqkhoja: there should be a variable or create one.

sadiqkhoja (commented):
let's talk to Nicole first + change the simple table as well.

apps/central/src/components/table/freeze.vue
```diff
@@ -161,6 +161,22 @@ defineExpose({ getRowPair });
...
+
+  tbody tr:nth-child(even) {
+    background-color: $color-even-row;
+  }
+
+  tbody tr:nth-child(odd) {
+    background-color: $color-odd-row;
+  }
+}
+
```
sadiqkhoja: this is really needed here since we have added zebra stripping on all tables in app.scss?

apps/central/src/components/project/home-block.vue
```diff
@@ -209,12 +209,20 @@ export default {
...
     }
 
-    tr:nth-child(3n + 2 of .project-dataset-row) {
-      background: #eee;
+    tr:nth-child(odd of .project-form-row) {
+      background: $color-odd-row;
+    }
+
+    tr:nth-child(even of .project-dataset-row) {
+      background: $color-even-row;;
```
sadiqkhoja:
```suggestion
      background: $color-even-row;
```

matthew-white (approved):

PR #1750, opened 2026-08-06
fix(web-forms#864): performance for selects on large entity lists

### LINKED ISSUE #864: Performance issue when using select from a large entity list
When a form has a cascading select which filters a large (eg: 16k) entity list, it locks the browser up. Even once it's loaded, interacting with the dropdown is very laggy.

Raised on [the forum](https://forum.getodk.org/t/web-form-ui-freezing-when-select-from-a-large-entity-list/58440) and via support request.

Also [discussed in Slack](https://getodk.slack.com/archives/C051L6AKG4F/p1785904908669079).

garethbowen:
Closes getodk/web-forms#864

There's a slight regression where short select lists are now shown with whitespace at the bottom. I think it's unavoidable when lazy loading.


packages/web-forms/src/components/common/MultiselectDropdown.vue
```diff
@@ -14,19 +15,24 @@ interface MultiselectDropdownProps {
...
+				search: option.label.asString,
+				loaded: true
+			};
 		}
-
 		return {
 			value: option.value,
-			label: option.label.formatted,
-			search: option.label.asString,
+			loaded: false
```
garethbowen: This is a placeholder that's swapped out when the page is loaded.

packages/web-forms/src/components/common/MultiselectDropdown.vue
```diff
@@ -35,6 +41,40 @@ const selectValues = (values: readonly string[]) => {
...
+	if (isJSDOM) {
+		return;
+	}
+	return {
+		lazy: true,
+		onLazyLoad: handleLazyLoad,
+		itemSize: DEFAULT_PRIMEVUE_ITEM_HEIGHT,
+		showLoader: true
+	};
+});
```
garethbowen: Unfortunately JSDOM completely breaks primevue's expectations, so to get the tests to pass I have to disable the lazy loading.

packages/xforms-engine/src/instance/SelectControl.ts
```diff
@@ -214,10 +214,7 @@ export class SelectControl
   getValueOption(value: string): SelectItem | null {
     // Note: this method is a client-facing convenience API for reading state,
     // so it **MUST** read from client-reactive state!
-    const valueOption = this.currentState.valueOptions.find((item) => {
-      return item.value === value;
-    });
-
+    const valueOption = this.mapOptionsByValue().get(value);
```
garethbowen, line 215: This seems to work, and is far faster than scanning through each of the valueOptions. The comment above is concerning though and makes me think I might have missed something.

packages/web-forms/src/components/common/MultiselectDropdown.vue
```diff
@@ -14,19 +15,24 @@ interface MultiselectDropdownProps {
...
+const INITIAL_PAGE_SIZE = 20;
+const DEFAULT_PRIMEVUE_ITEM_HEIGHT = 38;
+
 defineEmits(['update:modelValue', 'change']);
 
 const options = computed(() => {
-	return props.question.currentState.valueOptions.map((option) => {
-		const label = props.question.getValueOption(option.value);
-		if (label == null) {
-			throw new Error(`Failed to find option for value: ${option.value}`);
```
garethbowen, line 23: I don't know why we needed this, but it massively slowed down the rendering, because it was doing a n^2 lookup. Do we still need this validation? Is there a better way to do it than on the client every time?

garethbowen:
NB: multiselect is deprecated, and should be replaced with select. I've put that change off because it would be too disruptive at this point.

packages/web-forms/src/components/common/MultiselectDropdown.vue
latin-panda: The labels for options with index +20 won't be reactive anymore, right? So if a label with media takes time to load, or if a calc changes some values, is it refreshed properly? 

What about these scenarios but with autocomplete feature? Finding an option with index +20 and media request is slow.

packages/web-forms/src/components/common/SearchableDropdown.vue
```diff
@@ -14,31 +15,69 @@ interface SearchableDropdownProps {
...
+const DEFAULT_PRIMEVUE_ITEM_HEIGHT = 38;
+
 defineEmits(['update:modelValue', 'change']);
 
 const options = computed(() => {
-	return props.question.currentState.valueOptions.map((option) => {
-		const label = props.question.getValueOption(option.value);
-		if (label == null) {
-			throw new Error(`Failed to find option for value: ${option.value}`);
+	return props.question.currentState.valueOptions.map((option, i) => {
```
latin-panda: Is it possible to extract some of this to /lib to reuse code in both components?

latin-panda (commented):
I haven't had a chance to set up my environment for testing this yet. The code looks good so far—just a few questions below while I get my environment ready.

packages/web-forms/src/components/common/SearchableDropdown.vue
garethbowen: I was tempted to do that, but as this is likely going to be released as a patch I didn't want to risk destabalising anything else. After realising mutliselect is deprecated and should be replaced with the select widget I think the full solution will be to merge our multiselect and searchable dropdown components into one that does both, which means extracting to lib won't be necessary any more. Issue: https://github.com/getodk/web-forms/issues/865

garethbowen:
> What about these scenarios but with autocomplete feature? Finding an option with index +20 and media request is slow.

In actual fact my changes just broke autocomplete altogether - anything that wasn't "loaded" was never returned. So I added back the `search` property and now it works.

Unfortunately there was another problem which is that the `onLazyLoad` handler basically doesn't work for filter widgets. What was happening was when filtered, the `onLazyLoad` is called, but the `first` and `last` indexes are for the filtered list, not the whole list. That meant there was no easy way to find the option that needed to be "loaded". Ultimately I decided to remove lazy loading because it wasn't actually saving us very much (just some of the markdown processing) and was complex and fragile.

Using the virtual scroller without lazy is still a big improvement so that's probably enough for this patch.

> The labels for options with index +20 won't be reactive anymore, right? So if a label with media takes time to load, or if a calc changes some values, is it refreshed properly?

I've tested reactivity with the above changes and everything I tried works. I don't think it's possible to have media in a dropdown label is it? The code just calls the markdown component without the media component. NB: This change only impacts dropdowns, not radios, checkboxes, etc.

garethbowen:
The failing e2e test is also failing on `master`: https://github.com/getodk/central-frontend/actions/runs/31076754308/job/92537508645

So I think it can be ignored here...

packages/xforms-engine/src/instance/SelectControl.ts
```diff
@@ -214,10 +214,7 @@ export class SelectControl
   getValueOption(value: string): SelectItem | null {
     // Note: this method is a client-facing convenience API for reading state,
     // so it **MUST** read from client-reactive state!
```
latin-panda: I think we can remove this comment, it doesn't seem to help much for this case.

packages/web-forms/src/components/common/MultiselectDropdown.vue
latin-panda: Once this is merged: https://github.com/getodk/central-frontend/pull/1755, this piece of code can be deleted. It's worth waiting a bit, since the other PRs are very close to merging.

latin-panda (approved):
Nice! I don't have any other concerns. Just waiting for the other PRs to merge. Approving to unblock.

packages/xforms-engine/src/instance/SelectControl.ts
garethbowen: Done.

garethbowen:
I've removed the JSDOM code so this will fail until #1755 is merged.

latin-panda:
@garethbowen I did another pass in testing, and it works fine.
All other PRs are merged, and you can rebase. The CI should pass now :)

PR #1584, opened 2026-06-02
transifex:destructure: relax plural key expectations

alxndrsn:
Ref https://github.com/getodk/central-frontend/pull/1581#issuecomment-4599900395

~@matthew-white opened here for reference; you said you already have a fix for this issue.~

---

Closes getodk/central#1932

#### What has been done to verify that this works as intended?

- [x] Ran `transifex:destructure` with JSON keys in a different order:

##### `master`

```
$ npm run -w apps/central transifex:destructure

> @getodk/central-frontend@2026.2.0 transifex:destructure
> node bin/transifex/destructure.js

destructuring cs
{count, plural, one {Uživatel aplikace} few {Uživatelé aplikace} many {Uživatelů aplikace} other {Uživatelé aplikace}}
/home/user/workspaces/odk/frontend/apps/central/bin/util/util.js:13
  throw new Error(errorMessage);
  ^

Error: .appUser in locale "cs" expected the plural categories [one, few, many, other], but found [few, many, one, other]. Did you download the translations "to translate"?
    at logThenThrow (/home/user/workspaces/odk/frontend/apps/central/bin/util/util.js:13:9)
    at PluralForms.fromTransifex (/home/user/workspaces/odk/frontend/apps/central/bin/util/transifex.js:132:9)
    at Object.<anonymous> (/home/user/workspaces/odk/frontend/apps/central/bin/util/transifex.js:528:26)
    at JSON.parse (<anonymous>)
    at destructure (/home/user/workspaces/odk/frontend/apps/central/bin/util/transifex.js:523:44)
    at Object.<anonymous> (/home/user/workspaces/odk/frontend/apps/central/bin/transifex/destructure.js:23:22)
    at Module._compile (node:internal/modules/cjs/loader:1854:14)
    at Object..js (node:internal/modules/cjs/loader:1985:10)
    at Module.load (node:internal/modules/cjs/loader:1577:32)
    at Module._load (node:internal/modules/cjs/loader:1379:12)

Node.js v24.16.0
npm error Lifecycle script `transifex:destructure` failed with error:
npm error code 1
npm error path /home/user/workspaces/odk/frontend/apps/central
npm error workspace @getodk/central-frontend@2026.2.0
npm error location /home/user/workspaces/odk/frontend/apps/central
npm error command failed
npm error command sh -c node bin/transifex/destructure.js
```

##### this branch

Does not throw.

#### Why is this the best possible solution? Were any other approaches considered?

[...]

PR #1751, opened 2026-08-06
destructure.js: fix handling of plural categories

### LINKED ISSUE #1932: [frontend] transifex:destructure fails due to key order comparison
```
$ npm run -w apps/central transifex:destructure

> @getodk/central-frontend@0.1.0 transifex:destructure
> node bin/transifex/destructure.js

destructuring cs
{count, plural, one {Uživatel aplikace} few {Uživatelé aplikace} many {Uživatelů aplikace} other {Uživatelé aplikace}}
/odk/frontend/apps/central/bin/util/util.js:13
  throw new Error(errorMessage);
  ^

Error: Expected the plural categories [one, few, many, other], but found [few, many, one, other]. Did you download the translations "to translate"?
```

matthew-white:
This PR fixes how destructure.js handles and validates i18n plural categories, closing getodk/central#1932.

- It fixes a check that is throwing false-positive errors.
- It adds an important new check of plural categories.

I used this code in the previous release to run destructure.js. After this change, destructure.js no longer threw an error, and the resulting Vue I18n JSON generated by destructure.js matched what I expected.

I've been thinking of this as an alternative to #1584. However, after reviewing the code here, I'm realizing that the two approaches are more compatible than I thought. Specifically, I think there's a case to be made that this PR could/should be split into two. One option we have is to proceed with #1584, then add the new check in this PR separately / in a follow-up PR. I think I still prefer the approach in this PR though.

I'll also leave some additional comments about specific lines.


apps/central/bin/util/transifex.js
```diff
@@ -126,8 +126,22 @@ class PluralForms {
...
-      const expectedCategories = locales[locale].pluralCategories;
+      // First check that `categories` are in the correct order (e.g., 'one'
+      // should come before 'many'). If `categories` were in the wrong order in
+      // the Transifex JSON, they would also end up in the wrong order in the
+      // Vue I18n JSON.
+      const allCategories = ['one', 'few', 'many', 'other'];
+      const expectedOrder = allCategories.filter(category =>
+        categories.includes(category));
+      if (!equals(categories, expectedOrder))
+        logThenThrow(string, `.${key} in locale "${locale}": Expected the plural categories to be in the order [${expectedOrder.join(', ')}], but found [${categories.join(', ')}] instead.`);
```
matthew-white, line 147: This is the new check. I realized that there are actually more than two lists/orders that are relevant:

1. `categories`: The list parsed from the Transifex JSON.
2. `locales[locale].pluralCategories`: The list pulled from `Intl.PluralRules`. Just used to validate.
3. The list/order expected by Vue I18n and the `pluralizationRules` we specify to it. Even if the Transifex JSON and `Intl.PluralRules` agreed on the order, if that order doesn't match what our `pluralizationRules` expect, that would be a problem. In other words, the order isn't just something to handle and discard during validation; it's important to validate the order.

apps/central/bin/util/transifex.js
```diff
@@ -126,8 +126,22 @@ class PluralForms {
...
+
+      // Next, check that `categories` includes all the plural categories we
+      // expect. If it were missing categories (or had extra categories), our
+      // pluralizationRules probably wouldn't work.
+      const expectedCategories = locales[locale].pluralCategories
+        .sort((category1, category2) =>
+          allCategories.indexOf(category1) - allCategories.indexOf(category2));
       if (!equals(categories, expectedCategories))
```
matthew-white: Once the new check above passes, we're able to do an order-insensitive comparison. That's what #1584 does, and it's a valid approach. I think I prefer the approach here, because if there is a failure, the order of the plural categories in the error message here will be nice/normalized. That said, the `.sort()`-ing here is a little more verbose than the approach in #1584.

matthew-white (commented):
Leaving notes to facilitate async code review.

matthew-white:
Important note: destructure.js still fails after this change due to a new, separate issue involving `@transifexKey`. I'll file an issue about that, and I have a local commit that fixes it (but that isn't the most elegant).

apps/central/bin/util/transifex.js
```diff
@@ -126,8 +126,22 @@ class PluralForms {
...
+        categories.includes(category));
+      if (!equals(categories, expectedOrder))
+        logThenThrow(string, `.${key} in locale "${locale}": Expected the plural categories to be in the order [${expectedOrder.join(', ')}], but found [${categories.join(', ')}] instead.`);
+
+      // Next, check that `categories` includes all the plural categories we
+      // expect. If it were missing categories (or had extra categories), our
+      // pluralizationRules probably wouldn't work.
+      const expectedCategories = locales[locale].pluralCategories
+        .sort((category1, category2) =>
+          allCategories.indexOf(category1) - allCategories.indexOf(category2));
```
alxndrsn: Why does `pluralCategories` need sorting here?  `.sort()` mutates; could the order be fixed in the initial definition of `locales[locale].pluralCategories`?

alxndrsn (approved):
I don't really understand what's going on in this code, but:

1. anything to improve error messaging in these files is a positive
2. presumably any breakage will be found and fixed promptly

Maybe this is time for some testing around this code?

apps/central/bin/util/transifex.js
matthew-white: > Why does `pluralCategories` need sorting here?

The only reason to sort is to make the order-sensitive `equals()` comparison work. `equals()` will compare `categories` and `expectedCategories`. We know from the new check that `categories` will match a particular order (the order according to `pluralCategoryOrder`). So if we sort `expectedCategories` to match that order as well, we can use `equals()`.

> `.sort()` mutates; could the order be fixed in the initial definition of `locales[locale].pluralCategories`?

I like this idea a lot. :+1: I've implemented that in 4e11730845e81d9081bd2b9c099b1d27d2921605.

matthew-white:
> Maybe this is time for some testing around this code?

I'm eager to get this PR merged, but I think this is a reasonable suggestion. Usually, I don't feel like testing is needed for development scripts, but there's enough going on here that it would boost confidence and probably aid development if we added at least basic tests. Feel free to file an issue about adding testing.

> presumably any breakage will be found and fixed promptly

I think this is part of why I've felt comfortable going without testing for so long. Whenever translations are updated using destructure.js, I scan the diff to make sure it matches my expectation.

matthew-white:
> destructure.js still fails after this change due to a new, separate issue involving `@transifexKey`. I'll file an issue about that

Filed at getodk/central#2127.

apps/central/bin/util/transifex.js
matthew-white: > The only reason to sort is to make the order-sensitive `equals()` comparison work.

I think sorting also makes the error message a little more readable, not that that's a big deal.

PR #1754, opened 2026-08-08
apps/central: fix async route load failure on navigation

### LINKED ISSUE #2107: [frontend] test/e2e/enketo: error loading dynamically imported module
Fails almost 100% in CI.  Some examples:

```
   1) [firefox] › tests/enketo.spec.js:91:3 › Enketo › Edit submission link should be working ───────
    Error: Unexpected fatal error(s) logged: 
        0. [firefox|console.error] http://central-test.localhost/assets/central/main-bQlE2YW_.js:12:37005
        message: async-route loadError: TypeError: error loading dynamically imported module: http://central-test.localhost/assets/chunky-array-CEcAubBZ.js
```

```
  1) [firefox] › tests/enketo.spec.js:66:7 › Enketo › all old URLs should be working › shows Form using old URL - New Submission 

    Error: Unexpected fatal error(s) logged: 
        0. [firefox|console.error] http://central-test.localhost/assets/central/main-bQlE2YW_.js:12:37005
        message: async-route loadError: TypeError: error loading dynamically imported module: http://central-test.localhost/assets/home-Bk1zxsoo.js
```

```
  1) [firefox] › tests/enketo.spec.js:66:7 › Enketo › all old URLs should be working › shows Form using old URL - Preview Draft Form 

    Error: Unexpected fatal error(s) logged: 
        0. [firefox|console.error] http://central-test.localhost/assets/central/main-bQlE2YW_.js:12:37005
        message: async-route loadError: TypeError: error loading dynamically imported module: http://central-test.localhost/assets/chunky-array-CEcAubBZ.js
```

alxndrsn:
Previously, a user might be shown an "alert.danger" toast  when navigating away from web-forms.

Closes https://github.com/getodk/central/issues/2107

Fixes a failing e2e test.

The test fix is incidental.  A specific test could be written, but it would likely be hard to write, rely on a race condition, and consequently be brittle.

No effect.

No.

### DIFF
--- a/apps/central/src/components/async-route.vue
+++ b/apps/central/src/components/async-route.vue
@@ -36,7 +36,21 @@ window.addEventListener('beforeunload', () => {
   setTimeout(() => {
     setTimeout(() => {
       setTimeout(() => {
-        isUnloading = false;
+        setTimeout(() => {
+          setTimeout(() => {
+            setTimeout(() => {
+              setTimeout(() => {
+                setTimeout(() => {
+                  setTimeout(() => {
+                    setTimeout(() => {
+                      isUnloading = false;
+                    }, 0);
+                  }, 0);
+                }, 0);
+              }, 0);
+            }, 0);
+          }, 0);
+        }, 0);
       }, 0);
     }, 0);
   }, 0);
--- a/apps/central/src/components/async-route.vue
+++ b/apps/central/src/components/async-route.vue
@@ -35,7 +35,9 @@ window.addEventListener('beforeunload', () => {
   isUnloading = true;
   setTimeout(() => {
     setTimeout(() => {
-      isUnloading = false;
+      setTimeout(() => {
+        isUnloading = false;
+      }, 0);
     }, 0);
   }, 0);
 });
--- a/apps/central/src/components/async-route.vue
+++ b/apps/central/src/components/async-route.vue
@@ -29,6 +29,24 @@ import PageBody from './page/body.vue';
 import { loadAsync, loadedAsync } from '../util/load-async';
 import { noop } from '../util/util';
 
+let isUnloading;
+
+window.addEventListener('beforeunload', () => {
+  console.log('@@@@@@@@@@@@@@@ window event:', 'beforeunload');
+  window.___before_unload = true;
+  isUnloading = true;
+});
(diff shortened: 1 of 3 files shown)


apps/central/src/components/async-route.vue
```diff
@@ -29,6 +29,17 @@ import PageBody from './page/body.vue';
...
 
+let isUnloading;
+window.addEventListener('beforeunload', () => {
+  clearTimeout(isUnloading);
+
+  // Crude handling for unload event being cancelled.  Apparently there's no
+  // deterministic way to detect unload cancellation.
+  isUnloading = setTimeout(() => {
+    isUnloading = false;
+  }, 100);
```
garethbowen, line 40: What's the logic behind the setTimeout? The page is being thrown away so this should never fire, right? The only case I can think of is we also have an unsaved changes check so users can cancel the unload, but it seems unlikely that they'd get there before the async load completes as they probably need the async resource to make changes.

I don't think having the timeout will break anything but it's added cognitive complexity.

alxndrsn: > The only case I can think of is we also have an unsaved changes check so users can cancel the unload

Correct - this is the situation where `isUnloading` needs to be unset.  If that's unclear, the comment needs updating!

> they probably need the async resource to make changes.

Other ways of browsing are available, e.g. the address bar.

garethbowen: I think it'd be quite difficult to make changes, and have an async load fail, and then even if you did do all of that you then the only downside would be you wouldn't get a console warning.

But... the complexity isn't that bad either, so whatever you prefer!

garethbowen (approved):

apps/central/src/components/async-route.vue
alxndrsn: > you wouldn't get a console warning.

There's also a "danger alert" - a toast/notification displayed at the bottom of the screen.

> I think it'd be quite difficult to make changes, and have an async load fail

I think it's quite easy:

1. make changes
2. attempt to leave, but cancel the attempt when warned (without a timeout, `isUnloading` is now `true`)
3. carry on using the SPA for as long as you want
4. hit a genuine async loading error

PR #1758, opened 2026-08-11
feat(web-forms#254): adds pagination to forms

### LINKED ISSUE #254: One question per page
[Brainstorming doc](https://docs.google.com/document/d/15w-nz09krLvh0zQa1gWO-a8hp9w0tv19kW2D3d7TDrs/edit?tab=t.0)

As an enumerator, I want to see one question per page so that I can easily stay focused on my survey subject without losing my place.

As a project manager, I want to enforce one question per page to ensure respondents' focus and increase data quality.

As a project manager, I want to enforce one question per page so that Collect and Web Form users have similar experiences

## References
Usability guidelines/findings from UK government: https://designnotes.blog.gov.uk/2015/07/03/one-thing-per-page/

Enketo's default is all questions on one screen with an opt into `pages` style in form design: https://blog.enketo.org/pages/
 
## Ideas
- Could be determined in the form design (same as Enketo)
- Users may want to switch modes (need user research -- this would be new functionality)
- When viewing Web Forms on mobile it could switch to one question per page (need user research -- this would be new functionality)
- Would need to have a way to view all questions/ table of contents

latin-panda:
Closes https://github.com/getodk/web-forms/issues/254

<!-- 

-->

 - Manual testing, see forms and scenarios matrix [here](https://github.com/getodk/web-forms/issues/254#issuecomment-5257097017).
 - Added unit tests and e2e tests covering the main scenarios: field-list pages, relevance showing/hiding questions, repeats (add page, nested repeats, fixed-count), and Back/Next/Send visibility.                  
  - A performance benchmark on large forms. With 300 repeat instances (~900 questions): 
    - form load ~74 ms, 
    - page turn ~0.5 ms, 
    - jump to any page ~0.6 ms, 
    - answering a question ~0.6 ms. 
    - even with 10× slower resources, every interaction stays well. 

The engine owns pagination and exposes simple state: `currentPage`, `canGoNext`/`canGoPrevious`, and `nextPage()`/`previousPage()`. The client just renders what the engine says.                                                                                                                                                                                   
                                                                                                                                                                                                     
1. Earlier versions of this solution passed objects to the client. The engine handed the client page and node objects, and the client compared them to decide what to show. Problem: Vue and the engine (Solid) each wrap nodes in their own reactive objects, so the same node can be two different objects. Comparing them with === fails silently and those bugs were very hard to find. The API was also too big (page objects, emitted page lists, extra signals). The final version fixes both: the engine exposes a plain string id for the current page, and comparing strings works the same in any reactive system of clients or engine.                                                                                                                                      

2. A step-by-step cursor, the way other ODK form engines do it. I benchmarked it. It's faster per step, but ours is already fast everywhere it matters, and ours jumps straight to any page (relevant for navigation of questions coming next) while a cursor must walk the whole form to get there. It would also push "which questions go together on a page" onto every client and with it, the same object-comparison problem from point # 1, since the client would have to match cursor positions to nodes on its side.                                                                                                                        
                                                                                                                                                                                                     
At the end, we got one source of truth, a small API that passes plain ids instead of objects (so the object-comparison bugs cannot come back), and jumps straight to any page. 

Forms using pages + field-list appearance now show one page at a time, with Back/Next buttons and Send on the last page. The footer buttons also got icons. Forms without pagination look and behave exactly as before. 

[Yes, this is the PR](https://github.com/getodk/docs/pull/2089)


latin-panda:
@garethbowen tomorrow I'll be covering the following 
- Manual test of 2 forms from organizations that Hélène recommended.
- Send this to Nicole for feedback 
- Write the documentation for the docs site. 

In the meantime, could you please review? Thank you!

garethbowen:
@latin-panda I may not get to this today, but will be top of my todo list tomorrow.

packages/web-forms/e2e/page-objects/controls/RepeatControl.ts
```diff
@@ -3,12 +3,19 @@ import { expect, Locator, Page } from '@playwright/test';
...
   private readonly SELECTOR_INSTANCE = '.is-repeat';
   private readonly SELECTOR_HEADER = '.p-panel-header';
+  private readonly SELECTOR_ADD_BUTTON = '.button-add-instance';
   private readonly page: Page;
 
   constructor(page: Page) {
     this.page = page;
   }
 
+  private addButton(label?: string) {
```
garethbowen, line 12: It would be ideal to use getByRole rather than a css selector and filter. Can we make the label required, or would that require too much refactoring of existing tests?

packages/web-forms/e2e/test-cases/functional/pagination.test.ts
```diff
@@ -0,0 +1,332 @@
...
+    });
+
+    test('shows only the first input on load', async () => {
+      await formPage.text.expectLabels(['What is your name?']);
+      await formPage.pagination.expectNavigation({ previous: false, next: true });
+    });
+
+    test('navigates forward and backward through every page', async () => {
+      await formPage.pagination.clickNext();
+      await formPage.text.expectLabels(['What is your age?']);
```
garethbowen: The first time I saw this I didn't realise that the actual assertion is that the other questions _aren't_ being shown. This is partially due to the similarly named `expectLabel` singular, which doesn't behave the same. I think a different function name would make this clearer but naming is hard...

What about `expectLabelsMatchExactly`?

packages/web-forms/src/components/OdkWebForm.vue
```diff
@@ -523,11 +513,9 @@ onUnmounted(() => {
 				order: 3;
 			}
 
-			.footer {
+			.form-footer {
```
garethbowen: Consider moving this css into the formfooter.vue

packages/web-forms/src/lib/pagination/pagination.ts
```diff
@@ -0,0 +1,38 @@
...
+  range: RepeatRangeNode
+): range is RepeatRangeUncontrolledNode => {
+  if (range.nodeType !== 'repeat-range:uncontrolled') {
+    return false;
+  }
+  if (!isPaginated(range.root)) {
+    return true;
+  }
+  return range.currentState.pageBoundary === range.root.currentState.currentPage;
+};
```
garethbowen: I think you can call `isLeafOnCurrentPage` instead of this code?

packages/web-forms/tests/components/OdkWebForm.test.ts
```diff
@@ -454,7 +456,7 @@ describe('OdkWebForm', () => {
       await textInput.setValue(firstSubmissionInputValue);
 
       // Click submit
-      await component.get('button[aria-label="Send"]').trigger('click');
+      await component.get(SEND_BUTTON_SELECTOR).trigger('click');
```
garethbowen: Can we use getByText here instead of getting by selector? I prefer getByText because it makes sure it's testing what the user actually sees rather than an invisible class. I think it also means you can remove the `primary-action` class altogether.

packages/xforms-engine/src/client/UploadNode.ts
```diff
@@ -17,6 +18,7 @@ export interface UploadNodeState extends BaseValueNodeState<UploadValue> {
   get value(): UploadValue;
   get instanceValue(): InstanceAttachmentFileName;
   get attachmentState(): BaseInstanceAttachmentState;
+  get pageBoundary(): PageBoundary;
```
garethbowen: There's a lot of repetition here. Is there nowhere further up the hierarchy that could declare this? What NodeState's should _not_ have a page boundary? Just ModelValue?

packages/xforms-engine/src/instance/pagination/resolveFieldListPage.ts
```diff
@@ -0,0 +1,45 @@
+import type { Group } from '../Group.ts';
+import type { GeneralParentNode, RepeatRange } from '../hierarchy.ts';
+import type { RepeatInstance } from '../repeat/RepeatInstance.ts';
+
+type AncestorNode = GeneralParentNode | RepeatRange;
+type FieldListAnchor = Group | RepeatInstance;
```
garethbowen: The naming "Anchor" doesn't feel right. Why not just FieldListNode?

packages/xforms-engine/src/instance/pagination/pageSequence.ts
```diff
@@ -0,0 +1,130 @@
...
+ * An empty repeat counts as a page, so its "add" button can be reached.
+ */
+export const collectPages = (
+  children: ReadonlyArray<GeneralChildNode | RepeatInstance>
+): readonly Page[] => {
+  const pages = pageMembersOf(children).map(pageOf);
+
+  return pages.filter((page, index) => {
+    return pages.findIndex((candidate) => candidate.nodeId === page.nodeId) === index;
+  });
```
garethbowen: I think it would be slightly easier to read if it did something like this...

```suggestion
  const seenPages = new Set();
  return pages.filter(page => {
    if (seenPages.has(page.nodeId)) {
      return false;
    }
    seenPages.add(page.nodeId);
    return true;
  });
```

This also means it's a single pass through the pages so O(n) rather than O(n^2).

packages/xforms-engine/src/instance/pagination/pageSequence.ts
```diff
@@ -0,0 +1,130 @@
...
+export const nearestReachablePage = (
+  rootChildren: readonly GeneralChildNode[],
+  node: Page,
+  isReachable: (page: Page) => boolean
+): Page | null => {
+  const pages = collectPages(rootChildren);
+  const origin = findNearestRemainingAncestor(rootChildren, node);
+  const originFirstPage = origin == null ? null : (collectPages([origin])[0] ?? null);
+  const originIndex = pages.findIndex((page) => page.nodeId === originFirstPage?.nodeId);
+  const start = Math.max(0, originIndex);
```
garethbowen: I think this can be more readable, and a little more performant. For example, if origin is null, then the answer is 0, and you don't need to scan through the pages at all. Consider extracting a private function which does these three lines, but returns early when possible.

packages/xforms-engine/src/instance/pagination/pageSequence.ts
```diff
@@ -0,0 +1,130 @@
...
+  rootChildren: readonly GeneralChildNode[],
+  node: Page
+): GeneralChildNode | RepeatInstance | null => {
+  const chain: AnyChildNode[] = [];
+  let current: AnyChildNode = node;
+  while (current.parent.nodeType !== 'root') {
+    chain.push(current);
+    current = current.parent;
+  }
+  chain.push(current);
```
garethbowen: If I understand this correctly it's collecting all ancestors except for root.

If so couldn't you remove line 115 but making a change to the `while` condition?


```suggestion
  while (current.nodeType !== 'root') {
    chain.push(current);
    current = current.parent;
  }
```

packages/xforms-engine/src/instance/Root.ts
```diff
@@ -52,6 +72,11 @@ interface RootStateSpec {
 
   // Root-specific
   readonly activeLanguage: Accessor<ActiveLanguage>;
+
+  // Pagination
+  readonly currentPage: Accessor<PageBoundary | null>;
+  readonly canGoNext: Accessor<boolean>;
+  readonly canGoPrevious: Accessor<boolean>;
```
garethbowen: Naming preference... I think it'd be better for this function to talk about the current state, rather than allowed actions. ie: `hasNextPage` makes more sense to me.

packages/xforms-engine/src/instance/Root.ts
```diff
@@ -90,6 +115,11 @@ export class Root
   readonly validationState: AncestorNodeValidationState;
   readonly instanceState: InstanceState;
   readonly languages: FormLanguages;
+  readonly paginationRegistry: PaginationRegistry;
```
garethbowen: I'm not sure about the `registry` naming. Should this just be `pagination`? Or `paginationState`?

packages/xforms-engine/src/instance/Root.ts
```diff
@@ -102,13 +132,31 @@ export class Root
     });
 
     this.classes = parent.classes;
+    this.paginationRegistry = new PaginationRegistry(parent.classes.pages);
```
garethbowen: It feels weird to create a disabled pagination registry. Would it be nicer to just set paginationRegistry to null when disabled, or would that have bad flow-on effects?

packages/xforms-engine/src/instance/Root.ts
```diff
@@ -102,13 +132,31 @@ export class Root
...
+    const [getCurrentPage, setCurrentPage] = createSignal<Page | null>(null);
+    this.getCurrentPage = getCurrentPage;
+    this.setPage = setCurrentPage;
+    const currentPage: Accessor<PageBoundary | null> = () => getCurrentPage()?.nodeId ?? null;
+
+    this.getOrderedPages = this.scope.runTask(() => {
+      return createMemo(() =>
+        this.paginationRegistry.enabled ? collectPages(this.getChildren()) : []
+      );
+    });
```
garethbowen: In the case where the registry is disabled it would be better to not create a memo at all, otherwise it'll trigger and do nothing as the children change. Can we init this to a static accessor when disabled?

packages/xforms-engine/src/instance/Root.ts
```diff
@@ -140,6 +191,97 @@ export class Root
...
+        }
+
+        if (!stillListed) {
+          const recovered = nearestReachablePage(this.getChildren(), current, (page) => {
+            return this.isPageReachable(page);
+          });
+          if (recovered != null) {
+            this.setPage(recovered);
+          }
+          return;
```
garethbowen: What happens in this case? The current page is not listed, and we can't find another page to set. Will it show a blank page?

packages/xforms-engine/src/instance/Root.ts
```diff
@@ -140,6 +191,97 @@ export class Root
...
+        }
+
+        const neighbour =
+          this.findReachablePage(current, MOVE_FORWARD) ??
+          this.findReachablePage(current, MOVE_BACKWARD);
+        if (neighbour != null) {
+          this.setPage(neighbour);
+        }
+      });
+    });
```
garethbowen: This class is getting massive... Is there a clean way to split out the added pagination code into either the paginationregistry or some other file?

packages/web-forms/src/lib/pagination/pagination.ts
```diff
@@ -0,0 +1,38 @@
...
+  AnyControlNode,
+  AnyParentNode,
+  RepeatRangeNode,
+  RepeatRangeUncontrolledNode,
+  RootNode,
+} from '@getodk/xforms-engine';
+
+type ContainerNode = Exclude<AnyParentNode, RootNode>;
+
+export const isPaginated = (root: RootNode): boolean => root.classes.pages;
```
garethbowen: Move this convenience fn up to the RootNode to abstract away the name of the class.

packages/web-forms/src/lib/pagination/pagination.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+  const currentPage = leaf.root.currentState.currentPage;
+  return currentPage != null && leaf.currentState.pageBoundary === currentPage;
+};
+
+export const hasVisibleBodyNodes = (node: ContainerNode): boolean => {
+  if (isPaginated(node.root)) {
+    return node.currentState.hasBodyNodesOnCurrentPage;
+  }
+  return node.currentState.hasRelevantBodyNodes;
```
garethbowen: It feels awkward that this check is in the pagination lib and has to be called even if this form has no pagination.

I think it would be cleaner if the currentState had this `hasVisibleBodyNodes` function, and ideally didn't expose either hasRelevantBodyNodes or hasBodyNodesOnCurrentPage.

garethbowen (changes requested):
Excellent work. I love the extensive unit testing in particular.

I haven't actually tried it out yet - I'll do that next.

packages/web-forms/e2e/page-objects/controls/RepeatControl.ts
latin-panda: Done. it uses `getByRole('button', { name })` now. I kept the label optional because plain repeats use "Add" only

packages/web-forms/e2e/test-cases/functional/pagination.test.ts
latin-panda: Renamed to `expectVisibleLabelsToBe`. Something to indicate "the only visibles labels"

packages/web-forms/src/components/OdkWebForm.vue
latin-panda: Done!

packages/web-forms/src/lib/pagination/pagination.ts
latin-panda: Renamed it to `isOnCurrentPage` since a repeat range isn't a leaf

packages/web-forms/tests/components/OdkWebForm.test.ts
latin-panda: The test finds the button by its visible "Send" text now. This is vitest + vue-test-utils, not Playwright, so I built a helper to mimic Playwright's `getText`.

packages/xforms-engine/src/client/UploadNode.ts
latin-panda: Not only ModelValue, but also structural nodes like root, group, etc that have no pageBoundary.  I added a `BaseControlNodeState` declaring `pageBoundary` and now the form controls extend from that interface.

packages/xforms-engine/src/instance/pagination/resolveFieldListPage.ts
latin-panda: Yes, renamed!

packages/xforms-engine/src/instance/pagination/pageSequence.ts
latin-panda: I like it! Done

packages/xforms-engine/src/instance/pagination/pageSequence.ts
latin-panda: Done!

packages/xforms-engine/src/instance/pagination/pageSequence.ts
latin-panda: yes! Done

packages/xforms-engine/src/instance/Root.ts
latin-panda: Done! and it pairs better with `nextPage()`/`previousPage()` too

packages/xforms-engine/src/instance/Root.ts
latin-panda: I renamed the class to `Pagination` because the engine has a specific meaning for "state" as the client-facing reactive state  (`currentState`) / Solid state, and this class isn't a state in that sense.

packages/xforms-engine/src/instance/Root.ts
latin-panda: I reconsidered `null`, but every control asks pagination for its `pageBoundary`, which is non-null even in non-paginated forms (it falls back to the control's own id).

With `null`, it would need a null check in each control constructor and `pageBoundary` would become nullable for the Vue client too.

So `Pagination` owns all page decisions, including "this form has no pages". Added a short comment to make it clearer.

packages/xforms-engine/src/instance/Root.ts
latin-panda: The memo didn't actually re-run when disabled. Solid only subscribes a memo to the signals it reads, and when disabled it returns before reading any. So it ran once and never again. 

But I did it anyway as part of the split from your other comment where a disabled form now gets a static object with no reactivity in it at all.

packages/xforms-engine/src/instance/Root.ts
latin-panda: Good catch! The `currentPage` is now `null` in that case. The user sees a blank "body" with only the Send button, the same thing happens for non-paginated forms when nothing is relevant. This is an edge case and a form designer issue. As soon as any question becomes relevant again, a page is "displayed" again.
I added a test for this too.

packages/xforms-engine/src/instance/Root.ts
latin-panda: Done. The pure logic went to `pageSequence.ts` as plain functions, and the reactive part to a new `createPageNavigation.ts` (following `createChildrenState` idea). I didn't move it to `Pagination` class because that class is node-facing registration, while navigation is a client-facing matter.

packages/web-forms/src/lib/pagination/pagination.ts
latin-panda: Done!

packages/web-forms/src/lib/pagination/pagination.ts
latin-panda: Done!

latin-panda:
@garethbowen I've resolved your feedback (thank you!) and the ones Nicole agreed with.

Visual feedback from Slack:                                                                 
- Back and Next buttons: now both use the secondary style. For the hover, I kept the standard secondary hover we already use (same as the repeat Add button) to stay consistent.
- Send button: no changes for now.                                                                  
- Button max-width and label truncation: it's not related to pagination, so I'll do it in a follow-up PR.

garethbowen (approved):
Looks great. No notes. Well done!

I haven't done extensive manual testing and will leave that up to Nicole and QA and yourself.

PR #1761, opened 2026-08-12
chore(#2102): fix race condition in e2e test

### LINKED ISSUE #2102: [frontend] e2e-tests: intermittent failure: session expiry
https://github.com/alxndrsn/odk-central-frontend/actions/runs/30523965903/job/90811905790

```
  ✘  39 [chromium] › tests/web-forms.spec.js:129:3 › ODK Web Forms › allows user to relogin on session expiry during form fill (3.9s)
Error:   1) [chromium] › tests/web-forms.spec.js:129:3 › ODK Web Forms › allows user to relogin on session expiry during form fill 
    Error: expect(locator).toBeVisible() failed

    Locator: getByRole('heading', { name: 'Session expired' })
    Expected: visible
    Timeout: 2000ms
    Error: element(s) not found

    Call log:
      - Expect "toBeVisible" with timeout 2000ms
      - waiting for getByRole('heading', { name: 'Session expired' })


      144 |     await page2.getByRole('button', { name: 'send' }).click();
      145 |
    > 146 |     await expect(page2.getByRole('heading', { name: 'Session expired' })).toBeVisible();
          |                                                                           ^
      147 |
      148 |     const [page3] = await Promise.all([
      149 |       context.waitForEvent('page'),
        at /home/runner/work/odk-central-frontend/odk-central-frontend/central/client/e2e-tests/tests/web-forms.spec.js:146:75
Testing stopped early after 1 maximum allowed failures.
Error: Testing stopped early after 1 maximum allowed failures.
```

garethbowen:
Closes getodk/central#2102

My initial commit ran the test 20 times per browser and it passed.

My theory is that the first page was clicking log out but before the server could expire the session the second page was submitting the form.

The test hasn't changed in some time so I don't know why this started happening but probably the logout got slightly slower, or the web-forms submission got slightly faster. Either way, this code is more correct - just because you've clicked the button doesn't mean the server has received the request yet.

None. Test only.

No.

### DIFF
--- a/e2e-tests/tests/web-forms.spec.js
+++ b/e2e-tests/tests/web-forms.spec.js
@@ -150,7 +150,9 @@ test.describe('ODK Web Forms', () => {
     });
   });
 
-  test('allows user to relogin on session expiry during form fill', async ({ page, context }) => {
+for (let i = 0; i < 20; i++) {
+
+  test('allows user to relogin on session expiry during form fill, run: ' + i, async ({ page, context }) => {
     await login(page);
 
     const page2 = await context.newPage();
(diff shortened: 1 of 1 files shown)


alxndrsn (approved):
> My theory is that the first page was clicking log out but before the server could expire the session the second page was submitting the form.

Makes sense :+1:

PR #1762, opened 2026-08-13
fix(web-forms#870): fix rendering of outputs in ordered lists

### LINKED ISSUE #870: Output in markdown label is not being replaced
### Expected Behaviour

`output` elements in markdown should be evaluated and the result rendered into the text block.

### Observed Behaviour

The temporary output string is left behind, eg: `21. calculate, current value: `--ODK-OUTPUT-STRING-1--` `

### Reproduction Steps

Form (XML or XLSX file):

```xml
<?xml version="1.0"?>
  <h:head>
    <h:title>QA Prefill 01 - Field types</h:title>
      <instance>
        <data id="qa_prefill_01_field_types" version="2026080601">
          <t_calculate/>
          <t_calculate_show/>
          <meta>
            <instanceID/>
          </meta>
        </data>
      </instance>
      <bind nodeset="/data/t_calculate" type="string" calculate="'CALCULATED'"/>
      <bind nodeset="/data/t_calculate_show" readonly="true()" type="string"/>
    </model>
  </h:head>
  <h:body>
    <input ref="/data/t_calculate_show">
      <label> 21. calculate, current value: <output value=" /data/t_calculate "/> </label>
      <hint>XPath: /data/t_calculate - a calculation should win over the URL</hint>
    </input>
  </h:body>
</h:html>
```

Steps:
1. Load the above form
2. It should immediately show "CALCULATED" in the label.

### Screenshots or Videos

Image

### Additional Context

- Raised in https://github.com/getodk/web-forms/issues/464#issuecomment-5241023353
- UPDATE 25-Aug: Also reported on the forum: https://forum.getodk.org/t/making-reference-to-variable-names-in-odk-web-forms-not-rendering/58497

garethbowen:
Closes getodk/web-forms#870

Unit testing
Manual testing

There were more extensive fixes which I will be investigating soon (ie: dropping mdast), but this patches what's there without regression.

Should just fix a bug and nothing else.

No.

### DIFF
--- a/.changeset/brave-lights-end.md
+++ b/.changeset/brave-lights-end.md
@@ -1,0 +1,5 @@
+---
+"@getodk/xforms-engine": patch
+---
+
+Fixed a bug rendering outputs in single ordered list items
--- a/packages/xforms-engine/src/instance/text/markdownFormat.ts
+++ b/packages/xforms-engine/src/instance/text/markdownFormat.ts
@@ -213,12 +213,8 @@ function escapeEditableChunks(chunks: readonly TextChunk[]) {
 
 // mdast tends to add too many paragraphs which if left in place, puts a block level
 // element where it's not needed
-function removeSingleParagraph(nodes: MarkdownNode[]) : MarkdownNode[] {
-  if (
-    nodes.length === 1 &&
-    nodes[0]?.role === 'parent' &&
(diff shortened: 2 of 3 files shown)


latin-panda (approved):
It looks good!

PR #1764, opened 2026-08-13
destructure.js: fix @transifexKey nesting

### LINKED ISSUE #2127: destructure.js fails for complex uses of @transifexKey
Even with the changes in getodk/central-frontend#1751 (fixing #1932), destructure.js fails:

```
destructuring cs
warning: common.versionShort: variable without separator.
warning: component.ProjectFormAccessStates.introduction.2.full: variable without separator.
warning: component.ProjectFormAccessStates.introduction.3.full: variable without separator.
getodk/central-frontend2/apps/central/bin/util/transifex.js:917
        throw new Error(`No value found for ${keyPath} in locale "${this._locale}".`);
        ^

Error: No value found for component.FormUpload.action.upload in locale "cs".
    at Translations.toJSON (getodk/central-frontend2/apps/central/bin/util/transifex.js:917:15)
    at Translations.toJSON (getodk/central-frontend2/apps/central/bin/util/transifex.js:920:30)
    at JSON.stringify (<anonymous>)
    at writeTranslations (getodk/central-frontend2/apps/central/bin/util/transifex.js:1101:25)
    at Object.<anonymous> (getodk/central-frontend2/apps/central/bin/transifex/destructure.js:26:3)
    at Module._compile (node:internal/modules/cjs/loader:1554:14)
    at Object..js (node:internal/modules/cjs/loader:1706:10)
    at Module.load (node:internal/modules/cjs/loader:1289:32)
    at Function._load (node:internal/modules/cjs/loader:1108:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
```

I tracked it down to the following uses of `@transifexKey` comments:

- The `FormUpload` component specifies `@transifexKey component.FormNew` in order to move its entire block of messages from `component.FormUpload` to `component.FormNew` within the Transifex JSON. It does that in order to preserve existing translations at the old keys.
  - This use of `@transifexKey` was introduced for v2026.2, when we added the `FormUpload` component.
- en.json5 specifies `@transifexKey component.FormNew.action.upload` in order to move `action.upload` to `component.FormNew.action.upload`. Again, this is in order to preserve existing translations.
  - This use of `@transifexKey` may predate v2026.2.

The problem is that the first `@transifexKey` moves a block of messages whose new key also contains the destination of the second `@transifexKey`. destructure.js is good about identifying many `@transifexKey` conflicts, but it doesn't know how to handle this one.

More details here: https://github.com/getodk/central-frontend/pull/1764#issuecomment-5288155742

I think we have two options:

1. Update destructure.js to handle this case correctly. To do so, I think it will need to implement the `@transifexKey` comments in the correct order.
2. Don't allow this case and throw an error instead: just detect the case, don't try to handle it. If we didn't use `@transifexKey` for `action.upload`, then we could avoid this case. It'd be OK to remove that one `@transifexKey` and lose a handful of existing translations.

I'd prefer for destructure.js just to work, so I lean toward option (1). I have a fix locally along those lines that I'll push soon, but it's not very elegant.

matthew-white:
Closes getodk/central#2127.

There's some description of the problem in getodk/central#2127 and in code comments, but I plan to add more detail to this PR. It's kind of a complex case.

After this change (along with #1751), destructure.js at last runs without error. 🎉 I used this code in the previous release to run destructure.js, and the resulting Vue I18n JSON matched my expectations.

Given the complexity of the case, I'm tempted to add the first tests of our Transifex scripts. Doing so would probably help document this case.

I'll add more detail later explaining the underlying problem, after which I think the code approach will make more sense. The code itself isn't optimized, which is one reason why I've marked this PR as draft.

getodk/central#2127 proposes two main approaches: handling this case vs. detecting and disallowing it. This PR implements the former approach (handling this case), but the latter approach would probably be more straightforward. I'll probably continue to run with the former approach for now, since I've already written code for it. However, I may turn to the latter approach (just detecting this case) if I feel like it'd require fewer new tests.

### DIFF
--- a/apps/central/bin/util/transifex.js
+++ b/apps/central/bin/util/transifex.js
@@ -683,10 +683,58 @@ const rekeySource = (structured, transifexPaths) => {
 };
 
 const rekeyTranslations = (source, translated, transifexPaths) => {
-  for (const { sourcePath, transifexPath } of transifexPaths)
-    setPath(sourcePath, getPath(transifexPath, translated), translated);
+  // It's possible for one transifexPath to be a subpath of another,
+  // representing an ancestor message object. We want to rekey those shorter
+  // subpaths only after first rekeying the longer/deeper paths. Here, we
+  // reorder transifexPaths along those lines. Other than that though, we try to
+  // preserve the original order.
+  const ordered = [...transifexPaths];
+  for (let i = 0; i < ordered.length - 1;) {
+    let changed = false;
+    for (let j = i + 1; j < ordered.length; j += 1) {
+      if (startsWith(ordered[i].transifexPath, ordered[j].transifexPath) &&
+        !equals(ordered[i].transifexPath, ordered[j].transifexPath)) {
+        const ancestor = ordered[i]; // Subpath
+        const descendant = ordered[j]; // Longer/deeper path
+        ordered.splice(j, 1);
+        ordered.splice(i, 1, descendant, ancestor);
+        changed = true;
+        break;
+      }
+    }
+    if (changed)
+      i = 0; // Start over
+    else
+      i += 1;
+  }
+
+  // A single `transifexPath` can be copied to multiple `sourcePath`s. Here, we
+  // count how many times each `transifexPath` appears.
+  const pathCounts = new Map();
   for (const { transifexPath } of transifexPaths) {
-    if (!hasPath(transifexPath, source)) deletePath(transifexPath, translated);
+    const key = transifexPath.join('.');
+    pathCounts.set(key, (pathCounts.get(key) ?? 0) + 1);
+  }
+  const decrementCount = (path) => {
+    const key = path.join('.');
+    const newCount = pathCounts.get(key) - 1;
+    pathCounts.set(key, newCount);
+    return newCount;
+  };
+
+  for (const { sourcePath, transifexPath } of ordered) {
+    if (hasPath(sourcePath, translated))
+      throw new Error(`@transifexKey: attempted to copy the value at ${transifexPath.join('.')} to ${sourcePath.join('.')}, but there is already a value at ${sourcePath.join('.')}.`);
+    if (!hasPath(transifexPath, translated))
+      throw new Error(`@transifexKey: attempted to copy the value at ${transifexPath.join('.')} to ${sourcePath.join('.')}, but there is no value at ${transifexPath.join('.')}.`);
+
+    setPath(sourcePath, getPath(transifexPath, translated), translated);
+
+    // Delete the old transifexPath as soon as possible (as soon as the count is
+    // zero) in order to account for subpaths. We don't want to move a
+    // descendant message along with its ancestor message object.
+    if (!hasPath(transifexPath, source) && decrementCount(transifexPath) === 0)
+      deletePath(transifexPath, translated);
   }
 };
 


apps/central/bin/util/transifex.js
```diff
@@ -683,10 +683,58 @@ const rekeySource = (structured, transifexPaths) => {
...
+        const ancestor = ordered[i]; // Subpath
+        const descendant = ordered[j]; // Longer/deeper path
+        ordered.splice(j, 1);
+        ordered.splice(i, 1, descendant, ancestor);
+        changed = true;
+        break;
+      }
+    }
+    if (changed)
+      i = 0; // Start over
```
matthew-white, line 706: I'm not sure that we really need to start over here, so I may remove this line. I wrote this code pretty fast shortly before release. This is definitely not the most efficient sort algorithm, but I'm also pretty sure that `.sort()` can't handle this case. Given that, and given the small size of `transifexPaths`, I think code clarity is more important than sort efficiency.

apps/central/bin/util/transifex.js
```diff
@@ -683,10 +683,58 @@ const rekeySource = (structured, transifexPaths) => {
...
+  // reorder transifexPaths along those lines. Other than that though, we try to
+  // preserve the original order.
+  const ordered = [...transifexPaths];
+  for (let i = 0; i < ordered.length - 1;) {
+    let changed = false;
+    for (let j = i + 1; j < ordered.length; j += 1) {
+      if (startsWith(ordered[i].transifexPath, ordered[j].transifexPath) &&
+        !equals(ordered[i].transifexPath, ordered[j].transifexPath)) {
+        const ancestor = ordered[i]; // Subpath
+        const descendant = ordered[j]; // Longer/deeper path
```
matthew-white, line 698: I feel like I'm mixing metaphors here, referencing both subpaths and ancestors.

apps/central/bin/util/transifex.js
```diff
@@ -683,10 +683,58 @@ const rekeySource = (structured, transifexPaths) => {
...
+    const newCount = pathCounts.get(key) - 1;
+    pathCounts.set(key, newCount);
+    return newCount;
+  };
+
+  for (const { sourcePath, transifexPath } of ordered) {
+    if (hasPath(sourcePath, translated))
+      throw new Error(`@transifexKey: attempted to copy the value at ${transifexPath.join('.')} to ${sourcePath.join('.')}, but there is already a value at ${sourcePath.join('.')}.`);
+    if (!hasPath(transifexPath, translated))
+      throw new Error(`@transifexKey: attempted to copy the value at ${transifexPath.join('.')} to ${sourcePath.join('.')}, but there is no value at ${transifexPath.join('.')}.`);
```
matthew-white, line 729: These errors aren't thrown right now, but they helped me track down the problem.

matthew-white:
@alxndrsn, you may be interested in this PR, since you've been helping to get destructure.js back to a runnable state. I was thinking to tag Sadiq for code review, since he's familiar with the `@transifexKey` mechanism. He'll also be somewhat familiar with this specific case, since he wrote the `FormUpload` component. Though as I mentioned in the PR description, I'm also considering taking a different approach entirely and just detecting/disallowing this case. In any case, this PR is in a draft state and may change or be closed, but I wanted to let you know that it was up. Definitely feel free to leave any questions or comments.

apps/central/bin/util/transifex.js
```diff
@@ -683,10 +683,58 @@ const rekeySource = (structured, transifexPaths) => {
 };
 
 const rekeyTranslations = (source, translated, transifexPaths) => {
-  for (const { sourcePath, transifexPath } of transifexPaths)
-    setPath(sourcePath, getPath(transifexPath, translated), translated);
+  // It's possible for one transifexPath to be a subpath of another,
+  // representing an ancestor message object. We want to rekey those shorter
+  // subpaths only after first rekeying the longer/deeper paths. Here, we
```
matthew-white, line 688: > We want to rekey those shorter subpaths only after first rekeying the longer/deeper paths.

TODO: Explain why.

apps/central/bin/util/transifex.js
```diff
@@ -683,10 +683,58 @@ const rekeySource = (structured, transifexPaths) => {
...
+      throw new Error(`@transifexKey: attempted to copy the value at ${transifexPath.join('.')} to ${sourcePath.join('.')}, but there is already a value at ${sourcePath.join('.')}.`);
+    if (!hasPath(transifexPath, translated))
+      throw new Error(`@transifexKey: attempted to copy the value at ${transifexPath.join('.')} to ${sourcePath.join('.')}, but there is no value at ${transifexPath.join('.')}.`);
+
+    setPath(sourcePath, getPath(transifexPath, translated), translated);
+
+    // Delete the old transifexPath as soon as possible (as soon as the count is
+    // zero) in order to account for subpaths. We don't want to move a
+    // descendant message along with its ancestor message object.
+    if (!hasPath(transifexPath, source) && decrementCount(transifexPath) === 0)
```
matthew-white, line 736: TODO: It was preexisting code, but maybe let's add a code comment explaining `!hasPath(transifexPath, source)`. I think it's the difference between copying vs. moving a message. If a message is just copied, it will still exist at its original path in `source`.

matthew-white (commented):
Adding TODO comments

matthew-white:
> I'll add more detail later explaining the underlying problem

OK, here we go! The issue arises for `@transifexKey` comments involving `component.FormNew`. We used to have a `FormNew` component, but in #1535, that component was refactored and renamed to `FormUpload`.

The `FormNew` component had a bunch of translations already. By default, translations' Transifex keys are automatically derived from the component name. We didn't want to lose the translations after renaming the component, so we used `@transifexKey` to preserve the previous Transifex keys, moving the translations in `FormUpload` to the `component.FormNew` key:

https://github.com/getodk/central-frontend/blob/27b01d9247d2bef4bb4791d44cb0f1834cc43035/apps/central/src/components/form/upload.vue#L297-L346

The problem is that those translations are a large message object, and we already had individual messages that were separately moved to `component.FormNew`:

https://github.com/getodk/central-frontend/blob/27b01d9247d2bef4bb4791d44cb0f1834cc43035/apps/central/src/locales/en.json5#L234-L235

https://github.com/getodk/central-frontend/blob/27b01d9247d2bef4bb4791d44cb0f1834cc43035/apps/central/src/components/form/new-page.vue#L49-L51

What I think the problem is (what happens without this PR):

- The `transifexPaths` array passed to `rekeyTranslations()` is sorted by depth (by the path length of the `@transifexKey`)
- That means that `@transifexKey component.FormNew` is executed first, moving `component.FormNew` in strings_en.json to `component.FormUpload` for the Vue I18n JSON.
- When `@transifexKey component.FormNew.action.upload` is later executed, it doesn't actually move that message, because `component.FormNew` has already been moved to `component.FormUpload`: no message is found at `component.FormNew.action.upload`.
  - I've added a check of this case to the code.
- So there ends up being a message at `component.FormUpload.action.upload` for the Vue I18n JSON. But that leads to a structural mismatch, because the `FormUpload` component doesn't actually contain an `action.upload` message: that message is supposed to go to en.json5 . That's why the error message says `No value found for component.FormUpload.action.upload` even though no such key exists in either the Transifex JSON or the Vue I18n JSON.

What this PR does instead:

- When one `@transifexPath` is a prefixing "subpath" of another (e.g., `component.FormNew` is a subpath of `component.FormNew.action.upload`), we need to execute `@transifexPath` for the longer path before doing so for the subpath. E.g., we first need to move `component.FormNew.action.upload`, then `component.FormNew`. If we did it in the opposite order (`component.FormNew`, then `component.FormNew.action.upload`), the result would be that `component.FormNew.action.upload` would get moved along with `component.FormNew`, which doesn't work, since we'd lose track of `component.FormNew.action.upload` (as described above).
- A move is a copy + a delete. Previously, we would execute all copy operations, then execute all delete operations, but that doesn't work for this case. Even if we copy `component.FormNew.action.upload` to its correct destination in en.json5 (copying in the correct order), we also need to delete the message at `component.FormNew.action.upload` before `@transifexKey component.FormNew` is executed. Otherwise, it will be copied to the `FormUpload` component as well, leading to a structural mismatch.
  - An additional twist: a single message in the Transifex JSON can be copied to multiple paths in the Vue I18n JSON. So we can only delete a message once all the copy operations are executed, not just the first. (That's probably why previously we executed all copies, then all deletes: to ensure that the delete comes only after the last copy.) So we need to keep track of how many times each message will be copied, and delete it immediately after the last copy (before the first copy for the next key).

alxndrsn:
> @alxndrsn, you may be interested in this PR, since you've been helping to get destructure.js back to a runnable state.

I hadn't realised it was fundamentally broken!  I think it's definitely time this code got some tests.

matthew-white:
Yeah, this case introduced with #1535 isn't something that destructure.js is able to handle. 😢 We could easily resolve it just by removing `@transifexKey` for the two problematic messages in en.json5 and `FormNewPage`. But I think we should either fix it for real or add detection for this sort of case. Either way, adding tests does sound sensible.

matthew-white:
> I think it's definitely time this code got some tests.

Related: getodk/central#2151. Once we get the required testing infrastructure in place, we can add one or more tests about the specific case that this PR fixes.

PR #1765, opened 2026-08-13
fix(web-forms#872): allow setvalue actions to write to readonly fields 

### LINKED ISSUE #872: setvalue action fails to set value when target is readonly
### Expected Behaviour

A `setvalue` action triggered by `xforms-value-changed` should write to its target field even when that field's `readonly` expression is true. The `readonly` restricts user edits only and engine writes (calculates, preloads, actions) are not restricted. This matches Collect/JavaRosa, and worked in Web Forms v1.0.0. 

### Observed Behavior

The form crashes with this message:
```
> An error occurred while loading this form                                                                                                                                                         > Cannot write to readonly field: /data/item_ro1 
```

### Reproduction Steps

Form (XML or XLSX file):

[dynamic-readonly.xml](https://github.com/user-attachments/files/31036887/dynamic-readonly.xml)
[WF readonly error.xml](https://github.com/user-attachments/files/31036888/WF.readonly.error.xml)

Steps:
1. Open the form                                                                                                                                                                 
2. In "CRASH EXAMPLE1", select any item other than "Other".                                                                                                                                        
3. The form crashes with `Cannot write to readonly field: /data/item_ro1`. 

[Forum post](https://forum.getodk.org/t/webforms-behaviour-notes/58272/29)

### Environment

- Browser version:
- OS:
- Web Forms version:

latin-panda:
Closes https://github.com/getodk/web-forms/issues/872

Manual testing and added scenario tests 

The readonly only restricts user edits, so action writes should not be blocked.
This fix gives the setvalue action a way to assign a value without going through the guard that rejects writes to readonly fields, making it clear which setter is for actions and which is for client writes. 

Fix their broken forms 

No

### DIFF
--- a/packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
+++ b/packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
@@ -72,8 +72,8 @@ const createRelevantValueState = (
 };
 
 /**
- * For fields with a `readonly` bind expression, prevent downstream
- * (client/user) writes when the field is in a `readonly` state.
+ * Rejects downstream (client/user) writes while the field is in a `readonly` state.
+ * Engine writes (`calculate` computations, preloads, `setvalue` actions) never pass through this guard.
  */
 const guardDownstreamReadonlyWrites = (
   context: ValueContext,
@@ -310,7 +310,12 @@ const registerValueChangedActions = (context: ValueContext, getValue: Accessor<s
   });
 };
 
-export type InstanceValueState = SimpleAtomicState<string>;
+export interface InstanceValueState {
+  // The node's instance value. The setter rejects client writes while readonly.
+  readonly valueState: SimpleAtomicState<string>;
+  // Unguarded setter for the same value, used by `xforms-value-changed` actions writing to this node.
+  readonly setValueFromAction: SimpleAtomicStateSetter<string>;
+}
 
 /**
  * Provides a consistent interface for value nodes of any type which:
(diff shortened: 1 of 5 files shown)


packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -72,8 +72,8 @@ const createRelevantValueState = (
 };
 
 /**
- * For fields with a `readonly` bind expression, prevent downstream
- * (client/user) writes when the field is in a `readonly` state.
+ * Rejects downstream (client/user) writes while the field is in a `readonly` state.
```
latin-panda, line 75: This is a good place to start

latin-panda:
@garethbowen this is ready for review :)

packages/xforms-engine/src/lib/reactivity/createInstanceValueState.ts
```diff
@@ -341,6 +347,9 @@ export const createInstanceValueState = (context: ValueContext): InstanceValueSt
     registerValueChangedActions(context, getValue);
     registerSetValueActions(context, setValue);
 
-    return guardDownstreamReadonlyWrites(context, relevantValueState);
+    return {
+      valueState: guardDownstreamReadonlyWrites(context, relevantValueState),
+      setValueFromAction: setValue,
```
garethbowen, line 352: I'm worried that by exposing this here it will be called at the wrong time. Particularly when the function on the node is just called `setEncodedValue` - it's quite transparent that it will ignore the readonly status.

What if we add a param to setEncodedValue for `(bypassReadonly?: boolean)`. If true then it calls the `setValue` action from here, but if it's false then it calls `setValueState` like before. That way the dev has to explicitly opt in to it.

At the moment it's only called from one place where we do want to bypass readonly but that may change in future...

garethbowen: Another option would be to put `setValueState` on an internal interface so it can be "seen" within the engine, but not from the client.

garethbowen (changes requested):
Thanks, and good job on the quick turnaround.

My only concern is that it's now easy to bypass the readonly check...

latin-panda:
@garethbowen This is ready for another look

garethbowen (approved):
Thanks!

PR #1766, opened 2026-08-14
chore(#2113): fetch git history so new tags are found

### LINKED ISSUE #2113: wf-publish still fails to push the web-forms tags to the repo
```bash
🦋  Creating git tags...
🦋  New tag:  @getodk/web-forms@1.0.2
🦋  New tag:  @getodk/xforms-engine@1.0.2

...

Run git push origin --tags
  git push origin --tags

Everything up-to-date
```

I suspect it needs...

```yaml
    steps:
      - name: Check out repository
        uses: actions/checkout@v6
        with:
          fetch-depth: 0
```

At least it doesn't fail the build...

garethbowen:
Closes getodk/central#2113

Research. It's difficult to test until its executed in a GH actions environment.

It'll be a little slower to checkout because it has to fetch more, but we don't release that often.
It would be cleaner if we knew the name of the tag to explicitly push the tag name. We could try and find that by looking in the package.json. But that feels clumsy and brittle.

None.

None.

### DIFF
--- a/.github/workflows/wf-publish.yml
+++ b/.github/workflows/wf-publish.yml
@@ -17,6 +17,8 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
 
       - name: Set up npm registry
         uses: actions/setup-node@v6


matthew-white (approved):
I'm not sure I 100% understand all the details here, but it makes sense to me to give this a try. :+1:

PR #1767, opened 2026-08-14
fix(web-forms#871): remove virtual scroll on small selects

### LINKED ISSUE #871: Improve UX so there is no extra white space on selects with few items
Image

"The [performance fix](https://github.com/getodk/central-frontend/pull/1750) was to add a virtual scroller which caused the regression on the UX for short menus. I could look at only adding the virtual scroller if there are more than a handful of items..."
- @garethbowen

garethbowen:
Closes getodk/web-forms#871

Manual testing.

I was concerned about dynamically turning virtual scroll on and off, but it appears to update as quickly as it did before, and removes the whitespace as expected.

Slight UI improvement.

No.


latin-panda:
@garethbowen, is this ready for review? I'd be happy to take a look :)

garethbowen:
Yes! Thanks. I probably lost track of it while waiting for the green build....

packages/web-forms/src/components/common/MultiselectDropdown.vue
```diff
@@ -28,6 +28,13 @@ const options = computed(() => {
 	});
 });
 
+const virtualScrollerOptions = computed(() => {
+	if (props.question.currentState.valueOptions.length > 20) {
```
latin-panda, line 32: It needs a code comment explaining what it fixes. Same with the one in SeachableDropdown.vue

latin-panda (approved):
I did a smoke test. There is no longer whitespace at the end of the dropdown.
Maybe QA can explore further.

packages/web-forms/src/components/common/MultiselectDropdown.vue
garethbowen: Done.

PR #1768, opened 2026-08-14
chore(web-forms#443): silence known build warnings

### LINKED ISSUE #443: Build warning: use of eval
When running `yarn build`, I see the following warning:

```
[build:solid] ../xpath/dist/expressionParser-BUM7qFCl.js (1270:18): Use of eval in "../xpath/dist/expressionParser-BUM7qFCl.js" is strongly discouraged as it poses security risks and may cause issues with minification.
```

Is there a way to either:

1. suppress the warning, if it's inconsequential, or
2. resolve the warning?

garethbowen:
Closes getodk/web-forms#443

This does 3 things...

##### 1. Ignore the eval warning
```
@getodk/xpath:build: [build:js   ] ../../node_modules/web-tree-sitter/tree-sitter.js (1210:36): Use of eval in "../../node_modules/web-tree-sitter/tree-sitter.js" is strongly discouraged as it poses security risks and may cause issues with minification.
```

##### 2. Ignore the missing module warning
```
@getodk/xpath:build: [build:js   ] src/expressionParser.ts:17:27 - error TS2307: Cannot find module '@getodk/tree-sitter-xpath/dist/tree-sitter-xpath.wasm?url' or its corresponding type declarations.
@getodk/xpath:build: [build:js   ] 
@getodk/xpath:build: [build:js   ] 17 import xpathLanguage from '@getodk/tree-sitter-xpath/dist/tree-sitter-xpath.wasm?url';
```

Run build.

We can't actually fix these warnings so silencing them makes new warnings easier to spot.

None.

None.

### DIFF
--- a/packages/web-forms/vite.config.ts
+++ b/packages/web-forms/vite.config.ts
@@ -67,25 +67,14 @@ if (webkitFlakinessMitigations) {
 }
 
 export default defineConfig(({ mode }) => {
-  const isDev = mode === 'development';
-  const external = ['vue'];
-  const globals = { vue: 'Vue' };
-  const extraPlugins: PluginOption[] = [];
-  const lib: LibraryOptions = {
-    formats: ['es'],
-    entry: resolve(__dirname, 'src/index.ts'),
-    name: 'OdkWebForms',
-    fileName: 'index',
-  };
-
-  const versionSuffix = buildNumber && isDev ? ` - ${buildNumber}` : '';
+  const versionSuffix = buildNumber && mode === 'development' ? ` - ${buildNumber}` : '';
 
   return {
     define: {
       __WEB_FORMS_VERSION__: `"v${version}${versionSuffix}"`,
     },
     base: './',
-    plugins: [vue(), vueJsx(), cssInjectedByJsPlugin(), ...extraPlugins],
+    plugins: [vue(), vueJsx(), cssInjectedByJsPlugin()],
     resolve: {
       alias: {
         '@getodk/common': resolve(__dirname, '../common/src'),
@@ -119,15 +108,19 @@ export default defineConfig(({ mode }) => {
 
         // Per Vite docs
       },
-      lib,
+      lib: {
+        formats: ['es'],
+        entry: resolve(__dirname, 'src/index.ts'),
+        name: 'OdkWebForms',
+        fileName: 'index',
+      },
       rollupOptions: {
-        external,
+        external: ['vue'],
         onwarn(warning, warn) {
-          if (warning.code === 'EVAL') return; // ignore eval warning for tree-sitter
+          if (warning.code === 'EVAL' && warning.id?.endsWith('xforms-engine/dist/index.js')) {
+            return; // ignore eval warning for tree-sitter
+          }
           warn(warning);
-        },
-        output: {
-          globals,
         },
       },
     },
--- a/packages/web-forms/vite.config.ts
+++ b/packages/web-forms/vite.config.ts
@@ -68,7 +68,7 @@ if (webkitFlakinessMitigations) {
 
 export default defineConfig(({ mode }) => {
   const isDev = mode === 'development';
-  const external = ['fs', 'path', 'vue'];
+  const external = ['vue'];
   const globals = { vue: 'Vue' };
   const extraPlugins: PluginOption[] = [];
   const lib: LibraryOptions = {
--- a/packages/web-forms/vite.config.ts
+++ b/packages/web-forms/vite.config.ts
@@ -6,7 +6,6 @@ import { execSync } from 'node:child_process';
 import { existsSync, readFileSync } from 'node:fs';
 import { resolve } from 'node:path';
 import { fileURLToPath, URL } from 'node:url';
-import type { LibraryOptions, PluginOption } from 'vite';
 import { defineConfig } from 'vite';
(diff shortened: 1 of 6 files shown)


packages/web-forms/vite.config.ts
```diff
@@ -133,6 +122,10 @@ export default defineConfig(({ mode }) => {
       lib,
       rollupOptions: {
         external,
+        onwarn(warning, warn) {
+          if (warning.code === 'EVAL') return; // ignore eval warning for tree-sitter
```
latin-panda: What about making it more specific for tree-sitter only?

```suggestion
          if (warning.code === 'EVAL' && warning.id?.includes('web-tree-sitter')) return; // ignore eval warning for tree-sitter
```

packages/xpath/src/expressionParser.ts
```diff
@@ -14,7 +14,12 @@ import type * as ODKXPathMainEntrypoint from './index.ts';
  * - - -
  */
 
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
latin-panda: Instead of `@ts-ignore`, it's better to add type declarations for these two imports so TypeScript understands what they are whenever they are used in the project.                    

The type declaration can go here: https://github.com/getodk/central-frontend/tree/master/packages/xpath/src/global-types/internal

Something like this:                                                                                                                                                                             

```js
declare module '@getodk/tree-sitter-xpath/dist/tree-sitter-xpath.wasm?url' { 
   const url: string;                                                                                                                                                                             
   export default url;                                                                                                                                                                            
 }                                                                                                                                                                                                
                                                                                                                                                                                              
 declare module 'web-tree-sitter/tree-sitter.wasm?url' {                                                                                                                                          
   const url: string;                                                                                                                                                                             
   export default url;                                                                                                                                                                            }  
```

vite.config.js
```diff
@@ -104,6 +104,10 @@ export default defineConfig(({ mode }) => ({
           return 'assets/central/[name]-[hash].js';
         }
       },
+      onwarn(warning, warn) {
+        if (warning.code === 'EVAL') return; // ignore eval warning for tree-sitter
```
latin-panda: Same here about making it more specific

packages/xforms-engine/vite.config.ts
```diff
@@ -79,7 +79,13 @@ export default defineConfig(({ mode }) => {
           return entryName.replace(/^\.src\//, '').replace(/\.ts$/, '.js');
         },
       },
-      rollupOptions: { external },
+      rollupOptions: {
+        external,
+        onwarn(warning, warn) {
```
latin-panda, line 84: Same

packages/web-forms/vite.config.ts
```diff
@@ -67,29 +67,18 @@ if (webkitFlakinessMitigations) {
...
 export default defineConfig(({ mode }) => {
-  const isVueBundled = mode === 'demo';
   const isDev = mode === 'development';
-
-  let lib: LibraryOptions | undefined;
-  let external: string[];
-  let globals: Record<string, string>;
+  const external = ['vue'];
+  const globals = { vue: 'Vue' };
   const extraPlugins: PluginOption[] = [];
```
latin-panda: This can be removed, right? It's always empty

packages/web-forms/vite.config.ts
```diff
@@ -67,29 +67,18 @@ if (webkitFlakinessMitigations) {
...
 
 export default defineConfig(({ mode }) => {
-  const isVueBundled = mode === 'demo';
   const isDev = mode === 'development';
-
-  let lib: LibraryOptions | undefined;
-  let external: string[];
-  let globals: Record<string, string>;
+  const external = ['vue'];
+  const globals = { vue: 'Vue' };
```
latin-panda: Can we also remove this `globals = { vue: 'Vue' }` (including the `output: { globals }` below)? It does nothing, because this build produces `es` format (L75)

latin-panda (changes requested):
Great improvement! I have some suggestions below

packages/web-forms/vite.config.ts
garethbowen: Yes! Done.

packages/web-forms/vite.config.ts
garethbowen: Yes I think you're right. Done, and inlined other vars that were only used in one place.

packages/xpath/src/expressionParser.ts
garethbowen: Works for me. Done.

packages/web-forms/vite.config.ts
garethbowen: I've restricted things a bit, but unfortunately because we build everything up into dist/index we end up just ignoring all evals in the entire index bundle. I think it's an improvement.

I'm investigating changing that so we [depend on source not dist](https://github.com/getodk/web-forms/issues/875) which should improve this, so I'll look at improving it when that gets rebased.

vite.config.js
garethbowen: As above.

packages/xforms-engine/vite.config.ts
garethbowen: As above.

latin-panda (approved):
Thank you!

PR #1769, opened 2026-08-15
Remove afterEach hook outside test/index.js

matthew-white:
test/util/date-time.js contains a global `afterEach` hook. However, when I was trying to move to Vitest (getodk/central#1267), that was a problem. IIRC `afterEach` somehow wasn't in scope in that file, so an error was thrown. My solution was to move the code to an existing `afterEach` hook in test/index.js.

Honestly, I think it's better this way. I think it's a little funny to mix utility functions and Mocha hooks in the same file. It's better to do all the global Mocha setup in one place.

Tests continue to pass.

Another option would have been to define a function named something like `setupLuxon()` and pass it `afterEach` as an argument. That's what we do in test/util/i18n.js. I think I like the pattern in this PR a little better though.

### DIFF
--- a/apps/central/test/util/date-time.js
+++ b/apps/central/test/util/date-time.js
@@ -52,7 +52,7 @@ export const setLuxon = (settings) => {
   }
 };
 
-// Restore Luxon settings to their original values.
+// Restores Luxon settings to their original values after a call to setLuxon().
 export const restoreLuxon = () => {
   for (const [name, value] of originalSettings.entries())
     Settings[name] = value;
--- a/apps/central/test/util/date-time.js
+++ b/apps/central/test/util/date-time.js
@@ -4,7 +4,29 @@ import { DateTime, Settings } from 'luxon';
 export const isBefore = (isoString1, isoString2) =>
   DateTime.fromISO(isoString1) < DateTime.fromISO(isoString2);
 
-// Sets a single Luxon setting.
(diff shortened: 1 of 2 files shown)


apps/central/test/util/date-time.js
```diff
@@ -4,7 +4,29 @@ import { DateTime, Settings } from 'luxon';
...
+    .filter(s => s != null)
+    .map(s => Date.parse(s));
+  if (parsed.length === 0) return faker.date.past().toISOString();
+  const from = Math.max(...parsed);
+  let to;
+  do {
+    to = Date.now() - 1000;
+  } while (from > to);
+  return faker.date.between({ from, to }).toISOString();
+};
```
matthew-white, line 20: This function is unchanged from before. I moved it from the bottom of the file because I wanted to group the multiple Luxon-related exports in their own section of the file. There used to just be one Luxon-related export, but now there are two. I added a `LUXON SETTINGS` section to group the two exports together. These code sections are a common pattern in Frontend util files.

matthew-white (commented):
Adding a note to facilitate async code review

alxndrsn (approved):
> IIRC `afterEach` somehow wasn't in scope in that file

It looks like all the test files have the extension `.spec.js`, so it doesn't seem surprising that an `afterEach()` declared in a standard `.js` file would cause problems :+1:

PR #1771, opened 2026-08-15
Remove unneeded globals from ESLint config

matthew-white:
I had an old commit along these lines in my local branch for moving to Vitest. It's not really needed for Vitest, but I think it's a good change, so I'll go ahead and separate it out into its own PR.

This PR reduces the number of ESLint `globals`, removing `globals` that are no longer needed.

ESLint continues to pass.

I kind of wish I knew why these `globals` are no longer needed. (The reason for `$` is clear, but the others not as much.) Did something change about ESLint such that these `globals` are no longer needed? Or were they never needed in the first place?

### DIFF
--- a/apps/central/.eslintrc.js
+++ b/apps/central/.eslintrc.js
@@ -8,10 +8,6 @@ module.exports = {
     '@vue/airbnb'
   ],
   globals: {
-    $: 'readonly',
-    alert: 'readonly',
-    document: 'readonly',
-    window: 'readonly',
     defineModel: 'readonly'
   },
   rules: {


apps/central/.eslintrc.js
```diff
@@ -8,10 +8,6 @@ module.exports = {
     '@vue/airbnb'
   ],
   globals: {
-    $: 'readonly',
```
matthew-white, line 11: `$` is no longer needed, since we no longer use jQuery (#1733). 🎉

apps/central/.eslintrc.js
```diff
@@ -8,10 +8,6 @@ module.exports = {
...
-    alert: 'readonly',
-    document: 'readonly',
-    window: 'readonly',
```
matthew-white, line 14: These are standard browser globals. I'm not sure why they were needed in the `globals` list before.

matthew-white (commented):
Adding comments to facilitate async code review.

matthew-white:
Tagging both @alxndrsn and @sadiqkhoja in case either has time to take a look, but I'm only looking for a single ✅

alxndrsn (approved):

PR #1773, opened 2026-08-15
Replace Karma and Mocha with Vitest

matthew-white:
A while ago, I worked on replacing Karma (and Mocha) with Vitest (getodk/central#1267). I never finished that work, but I wanted to share what I have so far.

### State of the PR

I had these changes running locally at some point. However, that was well over a year ago. Importantly, it predated our new monorepo setup, which means that related configuration has changed since I last worked on it.

This PR is not expected to work at the moment. I'll add more details below as to why. Note that I didn't fully review the changes here. This PR is a work-in-progress.

I don't remember whether 100% of tests were passing back when I was working on this. It's very possible that they were. I seem to remember that most of them were at least. Locally, I don't see any changes to individual tests.

### Overall approach

The overall approach I took was to use Vitest browser mode. That's because it seemed most similar to Karma.

At the time, Vitest browser mode was marked as experimental. I'm pretty sure there have been breaking changes to it since I last worked on it.

### Required PRs

My original local commit was on a long branch (now quite outdated) that made a series of changes:

- Introducing Vite for the first time
  - By now, we've been using Vite for a while for local development and the production build. The changes for Vite were extracted to a separate PR.
- Replacing Should.js with Chai
  - IIRC this was needed for Vitest. Something related to the use of ES Modules in test/index.js and the fact that Should.js isn't an ES Module. But I also thought it was a good change by itself, so I extracted it to its own PR. That PR was merged a while ago.

I've rebased my local commit and also extracted other work that was on the same long branch. I've put that other work in separate PRs so that this core PR can be as slim as possible. The following PRs are required in order for the changes in this PR to work:

- #1769
- #1770
- (Maybe needed? Not sure.) #1775

### Follow-up work

I want this PR to be self-contained, i.e., to contain the minimal changes needed to move to Vitest. However, I had other changes on my local branch that can be considered in separate PRs. At least one of those PRs requires this PR to be merged first.

- #1772
  - Enhances the experience of using Vitest during local development
- getodk/central#1269
  - This is just a link to the issue, not a PR. However, the issue makes it clear what the required changes are.
  - I think this isn't strictly necessary for moving to Vitest, but it would be a nice change for the codebase.

### Package changes

I had local changes to package.json. However, that file has changed a lot in the repo since I last worked on this, so I decided not to include those changes in the PR. The packages I used back then are probably outdated by now anyway. But here's the list of packages I added locally:

```json
"vitest": "^1.6.0",
"@vitest/browser": "^1.6.0",
"webdriverio": "^8.36.1",
```

I also removed packages related to Karma, Mocha, Vue CLI, and webpack. With this PR, none of those frameworks will be needed for Central Frontend anymore.

### More details below

I'm going to leave comments below on particular files and lines in order to explain my approach and surface current open questions.

### DIFF
--- a/apps/central/test/index.js
+++ b/apps/central/test/index.js
@@ -1,36 +1,60 @@
 import sinon from 'sinon';
 import { enableAutoUnmount } from '@vue/test-utils';
-import { expect, should } from 'chai';
+import { should } from 'chai';
 
 import '../src/styles';
 
 import testData from './data';
+import { afterEachTest, beforeEachTest, setupHooks } from './setup/hooks';
 import { loadAsyncRouteComponents } from './util/load-async';
 import { mockLogin } from './util/session';
 import { setupLanguages } from './util/i18n';
 import './assertions';
-
-window.beforeAll = before; // eslint-disable-line no-undef
-window.afterAll = after; // eslint-disable-line no-undef
-window.test = it;
+import './setup/iframe';
 
 window.should = should();
-window.expect = expect;
+
+// Vitest hasn't made describe() a global yet. Once it tries to, we will run
+// setupHooks() instead. Doing so will set a describe() global that will call
+// the hooks added below.
+Object.defineProperty(window, 'describe', {
+  set(describe) {
+    delete window.describe;
+    setupHooks(describe);
+  },
+  configurable: true
+});
+
+Error.stackTraceLimit = 40;
 
 
 
 ////////////////////////////////////////////////////////////////////////////////
 // HOOKS
 
+/*
+In this section only, call these alternative hooks:
+
+  - Call beforeEachFile() instead of beforeAll().
+  - Call afterEachFile() instead of afterAll().
+  - Call beforeEachTest() instead of beforeEach().
+  - Call afterEachTest() instead of afterEach().
+
+See test/setup/hooks.js for more details if you're curious.
+*/
+
 // Even if a route is lazy-loaded, load() will need synchronous access to the
 // async components associated with the route.
 beforeAll(loadAsyncRouteComponents);
 
-enableAutoUnmount(afterEach);
-afterEach(() => {
+beforeEachTest(() => { window.resizeTo(1100, 650); });
+
+enableAutoUnmount(afterEachTest);
+afterEachTest(() => {
   const app = document.querySelector('[data-v-app]');
   if (app != null) app.parentNode.removeChild(app);
 
+  // TODO. Update selector.
   const afterScript = document.querySelector('body > script:last-of-type + *');
   if (afterScript != null) {
     console.error(document.body.innerHTML); // eslint-disable-line no-console
@@ -38,7 +62,7 @@ afterEach(() => {
   }
 });
 
-afterEach(() => {
+afterEachTest(() => {
   sinon.restore();
   window.scrollTo(0, 0);
   document.documentElement.setAttribute('lang', 'en');
@@ -47,16 +71,4 @@ afterEach(() => {
   mockLogin.reset();
   document.cookie = '';
 });
-setupLanguages(afterEach);
-
-
-
-////////////////////////////////////////////////////////////////////////////////
-// RUN TESTS
-
-// Run all tests. See the documentation for karma-webpack. We specify the files
-// here rather than in karma.conf.js, because doing so is more performant. When
-// I tried specifying the tests in karma.conf.js, I encountered an out-of-memory
-// error.
-const testsContext = require.context('.', true, /\.spec\.js$/);
-testsContext.keys().forEach(testsContext);
+setupLanguages(afterEachTest);
--- a/apps/central/test/index.js
+++ b/apps/central/test/index.js
@@ -9,7 +9,7 @@ import { afterEachTest, beforeEachTest, setupHooks } from './setup/hooks';
 import { loadAsyncRouteComponents } from './util/load-async';
 import { mockLogin } from './util/session';
 import { setupLanguages } from './util/i18n';
-import './assertions';
+import './setup/assertions';
 import './setup/iframe';
 
 window.should = should();
--- a/apps/central/test/run.sh
+++ b/apps/central/test/run.sh
@@ -8,16 +8,17 @@ cp ../../index.html ../../public/
 output=$(mktemp)
 trap 'rm -- ../../public/index.html "$output"' EXIT
 
-NODE_ENV="test" karma start karma.conf.js | tee "$output"
+vitest run "$@" | tee "$output"
 
-# Search for: warnings from console.warn(), including Vue warnings; Sass
-# warnings; and warnings from Karma.
+# Search for:
+#
+#   - Warnings from console.warn(), including Vue warnings
+#   - Sass warnings
 awk '
   BEGIN { warnings = 0 }
-  /WARN LOG:/            { ++warnings; print "WARNING: " $0 }
+  /stderr/               { ++warnings; print "WARNING: " $0 }
   /ERROR LOG:/           { ++warnings; print "WARNING: " $0 }
   /Module Warning/       { ++warnings; print "WARNING: " $0 }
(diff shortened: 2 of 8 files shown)


apps/central/vitest.config.js
matthew-white, line 1: My local commit also deleted karma.conf.js. One thing I'd want to do is review whether there have been any changes to karma.conf.js since the last time I worked on this. Maybe there's configuration in karma.conf.js that we'd want to preserve via parallel configuration in this file.

apps/central/vitest.config.js
```diff
@@ -0,0 +1,43 @@
+import { fileURLToPath } from 'node:url';
+import { defineConfig, mergeConfig } from 'vitest/config';
+
+import viteConfig from './vite.config';
```
matthew-white, line 4: My work on this predated the monorepo setup, which is why this file expects vite.config.js to exist in this directory. Not sure whether the new vitest.config.js here will also be used by the Web Forms app. If so, it should probably go somewhere else (probably the apps/ directory or the root directory of the repository).

apps/central/test/run.sh
```diff
@@ -8,16 +8,17 @@ cp ../../index.html ../../public/
 output=$(mktemp)
 trap 'rm -- ../../public/index.html "$output"' EXIT
 
-NODE_ENV="test" karma start karma.conf.js | tee "$output"
+vitest run "$@" | tee "$output"
```
matthew-white, line 11: I think `"$@"` is there so that devs can specify a filter. See the related changes to CONTRIBUTING.md about filtering tests.

apps/central/test/run.sh
```diff
@@ -8,16 +8,17 @@ cp ../../index.html ../../public/
...
+#
+#   - Warnings from console.warn(), including Vue warnings
+#   - Sass warnings
 awk '
   BEGIN { warnings = 0 }
-  /WARN LOG:/            { ++warnings; print "WARNING: " $0 }
+  /stderr/               { ++warnings; print "WARNING: " $0 }
   /ERROR LOG:/           { ++warnings; print "WARNING: " $0 }
   /Module Warning/       { ++warnings; print "WARNING: " $0 }
-  /WARN \[web-server\]:/ { ++warnings; print "WARNING: " $0 }
```
matthew-white, line 20: I understand the removal of `WARN [web-server]:`, since I'm pretty sure that's from Karma. But I'm not sure why I replaced `WARN LOG:` with `stderr`. I'm also not sure what's going on with `ERROR LOG:`. Should that line be removed now, similar to `WARN LOG:`?

apps/central/test/index.js
matthew-white, line 1: In my local commit, I moved this file to test/setup/index.js. I didn't do so here because I wanted to make the diff clearer.

apps/central/test/index.js
```diff
@@ -1,44 +1,68 @@
...
-afterEach(() => {
+beforeEachTest(() => { window.resizeTo(1100, 650); });
+
+enableAutoUnmount(afterEachTest);
+afterEachTest(() => {
   const app = document.querySelector('[data-v-app]');
   if (app != null) app.parentNode.removeChild(app);
 
+  // TODO. Update selector.
   const afterScript = document.querySelector('body > script:last-of-type + *');
```
matthew-white, line 59: This selector is specific to Karma, which is why we need to update it.

apps/central/test/setup/hooks.js
matthew-white, line 1: When I was working on this, there was something really weird going on with hooks that I'm not sure I fully understood. IIRC individual test files had access to `beforeAll`, `afterAll`, `beforeEach`, and `afterEach`, but test/index.js didn't. There was some split in the environment available to test/index.js and the environment available to individual tests. I added these hook mechanisms here so that test/index.js could continue to define global hooks. That setup worked and was running for me locally, but it definitely felt like a workaround.

apps/central/test/index.js
```diff
@@ -1,44 +1,68 @@
...
 import { setupLanguages } from './util/i18n';
 import './assertions';
-
-window.beforeAll = before; // eslint-disable-line no-undef
-window.afterAll = after; // eslint-disable-line no-undef
-window.test = it;
+import './setup/iframe';
 
 window.should = should();
-window.expect = expect;
```
matthew-white, line 19: I'm not sure why this PR removes `expect` from `window`, but keeps `should`. 🤔 But it's what I have in my local commit. Do we maybe import `expect` wherever we use it in tests (such that it doesn't need to be global)?

apps/central/test/index.js
```diff
@@ -1,44 +1,68 @@
...
+// the hooks added below.
+Object.defineProperty(window, 'describe', {
+  set(describe) {
+    delete window.describe;
+    setupHooks(describe);
+  },
+  configurable: true
+});
+
+Error.stackTraceLimit = 40;
```
matthew-white, line 29: I'm not sure why this was added. It feels like something that should go in its own PR. Maybe we can just remove it if I don't remember why it was added. Or maybe there's no harm in adding it.

apps/central/test/setup/iframe.js
```diff
@@ -0,0 +1,29 @@
...
+
+The iframe is pretty small (300px x 150px), which is actually small enough to
+cause test failures. By default, the iframe can't be resized using
+window.resizeTo(). However, it can be resized using CSS. Below,
+window.resizeTo() is overwritten to resize using CSS. testFile() will call
+window.resizeTo() to increase the size of the iframe.
+
+`npm run test` runs tests in a headless browser. On the other hand,
+`npm run test:debug` opens a browser window. This file styles the page to
+improve its look for when `npm run test:debug` is run.
```
matthew-white, line 12: `npm run test:debug` is added in #1772.

apps/central/vitest.config.js
```diff
@@ -0,0 +1,43 @@
...
+      },
+      fileParallelism: false,
+      isolate: false,
+      testerScripts: [{ src: 'test/setup/index.js', async: false }]
+    },
+    // I'm not sure whether `fileParallelism` and `isolate` are needed here
+    // given that we also specify them under `browser` above.
+    fileParallelism: false,
+    isolate: false,
+    sequence: { hooks: 'list' },
```
matthew-white, line 30: I have a vague memory that this `hooks` configuration was an important step in getting things to run.

matthew-white (commented):
Adding questions and comments

apps/central/test/run.sh
matthew-white, line 1: My local branch also made this change to package.json, which I don't really understand:

```diff
diff --git a/package.json b/package.json
index 715a4dec2..be759c7e7 100644
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "build:dev": "nf start build,nginx",
     "lint": "eslint --max-warnings 0 --cache --ext .js,.vue,.cjs src/ bin/ test/ *.js *.cjs",
     "lint:fix": "eslint --max-warnings 0 --fix --cache --ext .js,.vue,.cjs src/ bin/ test/ *.js *.cjs",
-    "test": "./test/run.sh"
+    "test": "vitest",
+    "test:run": "./test/run.sh"
   },
   "volta": {
     "node": "20.10.0"
```

Most everything still refers to `npm run test`, not `npm run test:run`, though my local branch does change CI to use `test:run`. But test/run.sh does some good things, both in CI and local development, e.g., it checks for Vue warnings and Sass warnings. I'm not sure why we would want to run `vitest` without running test/run.sh. Maybe it has something to do with filtering tests?

matthew-white: Oh maybe it's the difference between `vitest` and `vitest run`. Like does `vitest` by itself run continuously?

apps/central/test/setup/iframe.js
matthew-white: Maybe I could further reduce the size of this PR by moving these comments to #1772, along with the style changes that these comments mention that are further below in the file.

PR #1776, opened 2026-08-16
mockHttp(): include response counts in error message

matthew-white:
I made a small improvement to `mockHttp()` while working on moving from Karma to Vitest a while ago. Here I've extracted it into its own PR. This PR improves an error message in `mockHttp()`, surfacing more information that will hopefully facilitate debugging.

It's just changing an error message. The error message now includes more information (two new counts).

### DIFF
--- a/apps/central/test/util/http.js
+++ b/apps/central/test/util/http.js
@@ -854,7 +854,7 @@ class MockHttp {
       throw new Error('request without response: no response specified for request');
     } else if (this._orderedResponsesRequested < this._orderedResponses.length) {
       this._listRequestResponseLog();
-      throw new Error('response without request: not all responses were requested');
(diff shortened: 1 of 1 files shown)


apps/central/test/util/http.js
```diff
@@ -854,7 +854,7 @@ class MockHttp {
       throw new Error('request without response: no response specified for request');
     } else if (this._orderedResponsesRequested < this._orderedResponses.length) {
       this._listRequestResponseLog();
-      throw new Error('response without request: not all responses were requested');
+      throw new Error(`response without request: ${this._orderedResponses.length} required response(s) were specified, but only ${this._orderedResponsesRequested} of them were requested`);
```
matthew-white, line 857: The `else if` a couple lines above compares `this._orderedResponsesRequested` to `this._orderedResponses.length`. However, we didn't previously surface those counts in the error message. This change is to add that information to the error message.

alxndrsn (approved):
Looks helpful - thanks :+1:

PR #1777, opened 2026-08-16
Add node: prefix to import of fs

matthew-white:
I made this change locally while working on moving from Karma to Vitest. It's just a small change, but I think it's an improvement. It's not super related to Vitest though, so I've moved it into its own PR.

Instead of importing `fs`, this PR imports `node:fs`, adding the `node:` prefix.

There's at least one other file in bin/ that imports `node:fs` rather than just `fs`. I think `node:fs` is slightly preferred these days.

I checked all the files in bin/ for other places to add the `node:` prefix, but I didn't see any. I didn't check other files outside bin/.

### DIFF
--- a/apps/central/bin/transifex/destructure.js
+++ b/apps/central/bin/transifex/destructure.js
@@ -1,7 +1,7 @@
 // For a description of our Transifex workflow and how this script fits into it,
 // see CONTRIBUTING.md.
 
-const fs = require('fs');
+const fs = require('node:fs');
 
 const { destructure, readSourceMessages, rekeyTranslations, writeTranslations } = require('../util/transifex');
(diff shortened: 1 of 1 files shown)


alxndrsn (approved):

alxndrsn:
Should we consider adding an eslint rule for this, e.g. `import/enforce-node-protocol-usage`?

There's a good discussion at https://github.com/import-js/eslint-plugin-import/issues/2717.

matthew-white:
I like the idea of enforcing that rule in `central-frontend`! Would you be up for creating a PR similar to getodk/central-backend#1914? For now, I think I'll just merge this one: I'm mostly just trying to get commits off my local Vitest branch.

PR #1778, opened 2026-08-17
ci: extract action to restore node & caches

### LINKED ISSUE #2131: [frontend] ci: extract restore-cache action
Per https://github.com/getodk/central-frontend/pull/1678#pullrequestreview-4645448068, there may be scope for significantly simplifying Github Actions config by extracting a `restore-cache` custom action, and reusing it across jobs:  https://github.com/getodk/central-frontend/compare/master...alxndrsn:odk-central-frontend:alx-1929-streamline-build-rebased?expand=1.

alxndrsn:
Closes getodk/central#2131

- [x] ci
- [ ] tested publishing

* cache keys and other config are complicated, so the less copy/pasting the better
* upgrades to cache v6 (could be a separate PR / https://github.com/getodk/central-frontend/pull/1779)
* maintains custom node versions where required
* switches restore order to make failure faster

No impact - just test code.

No.


.github/actions/restore-node/action.yml
```diff
@@ -0,0 +1,43 @@
...
+    required: false
+
+runs:
+  using: 'composite'
+  steps:
+  - uses: actions/setup-node@v6
+    id: setup-node
+    with:
+      node-version-file: package.json
+      node-version: ${{ inputs.node-version }}
```
garethbowen: When `node-version` is supplied, this emits a warning.

> Both node-version and node-version-file inputs are specified, only node-version will be used

Seems like it's doing the right thing regardless.

One option to fix the warning would be to pass in `node-version-file` as an input which is more verbose, but also more explicit.

.github/actions/restore-node/action.yml
```diff
@@ -0,0 +1,43 @@
...
+      path: |
+        node_modules
+        **/node_modules
+      key: install-${{ steps.setup-node.outputs.node-version }}-${{ hashFiles('package-lock.json', '.npmrc', '.github/workflows/*.yml', 'packages/*/package.json', 'package/*/package-lock.json') }}
+      fail-on-cache-miss: ${{ inputs.node-version == '' }}
+
+  - name: 'npm ci'
+    shell: sh
+    if: steps.cache-install.outputs.cache-hit != 'true'
+    run: npm ci
```
garethbowen, line 43: How do you feel about removing and failing the build if the cache doesn't match? It means if someone changes the key in the tests.yml then we'll detect it rather than continuing.

Also, it seems strange that we're handling for a missing `cache-install` and not a missing `cache-build`.

garethbowen (commented):
A couple of suggestions inline.

.github/actions/restore-node/action.yml
alxndrsn: This PR is aiming to maintain current behaviour.  All jobs which specify `node-version` allow for `npm ci` on cache miss, e.g.:

https://github.com/getodk/central-frontend/blob/92133fbcf2e22f8d62dfe236ce2493b3027b3e09/.github/workflows/wf-ci-e2e.yml#L25-L31

https://github.com/getodk/central-frontend/blob/92133fbcf2e22f8d62dfe236ce2493b3027b3e09/.github/workflows/wf-ci-e2e.yml#L41-L43

Do you think that should be changed?

garethbowen: I think so... I can't think of a time when I wouldn't want to be notified that there was a cache miss. But I didn't spot that that was the existing behaviour, so leaving it is fine for this PR.

alxndrsn: Isn't the point that the cache is only built for one specific version of node (in `check-and-build`), but webforms runs against multiple versions of node?

.github/actions/restore-node/action.yml
alxndrsn: Warning is:

> ```
> Warning: Both node-version and node-version-file inputs are specified, only node-version will be used
> ```

alxndrsn: @garethbowen I've added a check to avoid this warning

.github/actions/restore-node/action.yml
garethbowen: Yeah. Good point.

It just feels like a bad pattern to just implicitly create the cache.

garethbowen (approved):
Thanks!

.github/actions/restore-node/action.yml
alxndrsn: > It just feels like a bad pattern to just implicitly create the cache.

I agree, but I'm not sure that's what's happening here...

That might actually be a change for `.github/workflows/wf-ci.yml`, as it's moving from `actions/cache@v6` to `actions/cache/restore@v6`.

PR #1779, opened 2026-08-17
ci: standardise cache action on v6

alxndrsn:

- [x] reviewed release notes at https://github.com/actions/cache/releases
- [x] ci

* v6 is most recent
* already used in some places, but not consistently

No impact - just tests.

No.


latin-panda (approved):
Looks good!

PR #1781, opened 2026-08-17
Create actor properties inline in app user and PAL modals

### LINKED ISSUE #1875: Inline creation of user properties from the App User / PAL create modals
Enhancement: allow project managers to create a new user property inline from the App User or Public Access Link create modals, rather than having to go to the project-level user properties management screen first.

- Add an "add new property" affordance in the create modals (per Figma)
- Creates the property at the project level and then allows assigning a value in the same flow
- Applies to both App User and Public Access Link create modals

ktuite:
Closes https://github.com/getodk/central/issues/1875

Adds the ability to create new actor properties inline within the New App User and Create Public Link modals.

**New component: `ActorPropertiesNew`**
- Inline form with a name input, Add, and Cancel button
- POSTs to `/v1/projects/:id/actor-properties` and updates `actorProperties.data` on success
- Shows a specific error on 409.3 duplicate name conflicts

**Updated `ActorPropertiesUpsert`**
- Added a header and description section with a doc link
- Added an empty state message when no properties exist
- Wrapped the table in a scrollable container to avoid modal overflow
- Shows a `+ Add Property` link (in create mode) that toggles the inline `ActorPropertiesNew` form
- Only lets new properties be added when creating a fresh app user or link, not when updating actor properties on an existing actor.

**Updated `FieldKeyNew` and `PublicLinkCreate`**
- Show `ActorPropertiesUpsert` whenever `actorProperties` data is loaded, not only when properties already exist, enabling users to add the first property from the modal

Tests and trying it out. 

This is a feature we've wanted for a while and already discussed. The code adds a pretty self-contained component to the existing modals. 

Streamlined experience without having to go to other custom properties tab. 

I don't think the docs explain exactly where the actor properties are created so I don't think they need to change.

### DIFF
--- a/.changeset/cool-beds-brake.md
+++ b/.changeset/cool-beds-brake.md
@@ -2,4 +2,4 @@
 "@getodk/central-frontend": patch
 ---
 
-feat #1875: Inline property creation for app user and PAL create modals
+feat getodk/central#1875: Inline property creation for app user and PAL create modals
--- a/.changeset/cool-beds-brake.md
+++ b/.changeset/cool-beds-brake.md
@@ -1,0 +1,5 @@
+---
+"@getodk/central-frontend": patch
+---
+
+feat #1875: Inline property creation for app user and PAL create modals
--- a/apps/central/src/components/actor-properties/new.vue
+++ b/apps/central/src/components/actor-properties/new.vue
@@ -1,21 +1,27 @@
 <template>
   <div class="actor-properties-new">
-    <form v-if="showForm" class="actor-properties-new-form" @submit.prevent="submit">
-      <form-group ref="nameGroup" v-model.trim="name"
-        :placeholder="$t('newPropertyName')" required autocomplete="off"/>
-      <div class="form-actions">
-        <button type="submit" class="btn btn-primary"
-          :aria-disabled="awaitingResponse">
-          {{ $t('action.add') }} <spinner :state="awaitingResponse"/>
-        </button>
-        <button type="button" class="btn btn-link"
-          :aria-disabled="awaitingResponse" @click="showForm = false">
-          {{ $t('action.cancel') }}
-        </button>
+    <template v-if="showForm">
+      <div class="actor-properties-header">
+        <strong>{{ $t('addProperty') }}</strong>
+        <p>{{ $t('addPropertyHint') }}</p>
       </div>
-    </form>
+      <form class="actor-properties-new-form" @submit.prevent="submit">
+        <form-group ref="nameGroup" v-model.trim="name"
+          :placeholder="$t('newPropertyName')" required autocomplete="off"/>
+        <div class="form-actions">
+          <button type="submit" class="btn btn-primary"
+            :aria-disabled="awaitingResponse">
+            {{ $t('action.add') }} <spinner :state="awaitingResponse"/>
+          </button>
+          <button type="button" class="btn btn-link"
+            :aria-disabled="awaitingResponse" @click="reset()">
+            {{ $t('action.cancel') }}
+          </button>
+        </div>
+      </form>
+    </template>
     <a v-else href="#" class="add-property-link"
-      @click.prevent="showForm = true; nextTick(() => nameGroup.focus())">
+      @click.prevent="show()">
       <span class="icon-plus-circle"></span>{{ $t('addProperty') }}
     </a>
   </div>
@@ -47,6 +53,11 @@ const nameGroup = ref(null);
 const name = ref('');
 const showForm = ref(false);
 
+const show = () => {
+  showForm.value = true;
+  nextTick(() => nameGroup.value.focus());
+};
+
 const reset = () => {
   name.value = '';
   showForm.value = false;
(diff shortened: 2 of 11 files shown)


apps/central/transifex/strings_en.json
```diff
@@ -1229,6 +1229,40 @@
         }
       }
     },
+    "ActorPropertiesNew": {
+      "addProperty": {
+        "string": "+ Add Property"
+      },
+      "newPropertyName": {
+        "string": "New property name"
```
matthew-white: It looks like we already have this string in the codebase. It'd be good to use `@transifexKey` rather than retranslating it.

apps/central/transifex/strings_en.json
```diff
@@ -1229,6 +1229,40 @@
         }
       }
     },
+    "ActorPropertiesNew": {
+      "addProperty": {
+        "string": "+ Add Property"
```
matthew-white: I feel like the `+` is usually an icon and not part of the i18n message.

apps/central/transifex/strings_en.json
```diff
@@ -1229,6 +1229,40 @@
...
+      },
+      "newPropertyName": {
+        "string": "New property name"
+      },
+      "problem": {
+        "409_3": {
+          "string": "A property already exists in this project with the name \"{propertyName}\"."
```
matthew-white: It looks like we have a similar message already in `ProjectCustomPropertiesNew`. I think it'd be good to avoid a separate translation if possible.

matthew-white (commented):
I was just curious about the i18n, so I took a quick look at that. Not sure it makes sense for me to be the primary reviewer (I'm guessing not?), but I wanted to leave these small comments about i18n.

.changeset/cool-beds-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@getodk/central-frontend": patch
+---
+
+feat #1875: Inline property creation for app user and PAL create modals
```
matthew-white: This will link back to the `central` repo from the changelog.

```suggestion
feat getodk/central#1875: Inline property creation for app user and PAL create modals
```

apps/central/transifex/strings_en.json
ktuite: We do... that one includes the word "custom" so it's a tiny bit different. It seems ok to reuse the same thing here, but in another place, I'm not sure whether to say "Add custom property" or "Add property". Will ask @nicoleorlowski. It also relates to whether or not we're getting rid of the Custom Properties tab.

matthew-white: > It also relates to whether or not we're getting rid of the Custom Properties tab.

It's possible to use `@transifexKey` to reuse translations from the Custom Properties tab even if we end up deciding to get rid of that tab / those components.

> Will ask @nicoleorlowski.

Sounds good, mostly just wanted to make sure we were considering translation reuse as an option.

apps/central/src/components/actor-properties/new.vue
```diff
@@ -0,0 +1,106 @@
...
+<i18n lang="json5">
+{
+  "en": {
+    "addProperty": "Add Property",
+    // @transifexKey component.ProjectCustomPropertiesNew.newPropertyName
+    "newPropertyName": "New property name",
+    // @transifexKey component.ProjectCustomPropertiesNew.problem
+    "problem": {
+      "409_3": "A custom property already exists in this project with the name of \"{propertyName}\"."
+
```
sadiqkhoja: nit

```suggestion
```

apps/central/src/components/actor-properties/new.vue
```diff
@@ -0,0 +1,106 @@
...
+          {{ $t('action.add') }} <spinner :state="awaitingResponse"/>
+        </button>
+        <button type="button" class="btn btn-link"
+          :aria-disabled="awaitingResponse" @click="showForm = false">
+          {{ $t('action.cancel') }}
+        </button>
+      </div>
+    </form>
+    <a v-else href="#" class="add-property-link"
+      @click.prevent="showForm = true; nextTick(() => nameGroup.focus())">
```
sadiqkhoja: maybe add a method in the `script` section instead of inline JS

apps/central/src/components/actor-properties/upsert.vue
```diff
@@ -1,25 +1,45 @@
...
+        </thead>
+        <tbody>
+          <entity-update-row v-for="{ name } of propertyDefs"
+            :key="name" ref="propertyRows" v-model="propertyValues[name]"
+            :old-value="originalValues?.[name]" :label="name"
+            :mark-value-changed="!create"/>
+        </tbody>
+      </table>
+    </div>
+    <actor-properties-new v-if="create"/>
```
sadiqkhoja: wondering if we should show it for edits as well?

sadiqkhoja (approved):

apps/central/src/components/actor-properties/upsert.vue
ktuite: Sure. We can show for edits and if that feels like too much, we can easily set it back to just showing for creates.

apps/central/transifex/strings_en.json
nicoleorlowski: no "custom"!

PR #1782, opened 2026-08-18
fix(web-forms#875): build from source not dist

### LINKED ISSUE #875: Compile src instead of bundling dist to enable HMR and reduce bundle size
### Description

Since the beginning of time, web-forms has compiled each package by bundling the dist of its dependant monorepo packages. This means,

1. the compiler can't de-dupe shared dependencies because the contents of the dist are opaque,
2. the compiler can't tree-shake dependencies - it just inlines the entire dist file, and
3. vite can't perform hot module reloading (except at the apps/forms package) so when working on one of the packages you need to manually rebuild the package every time you want to test a change.

I'm not sure why this was done, except it's good for eating our own dog food, by including the packages the same way a 3rd party product would. While this is nice, it comes at a great cost to the users of the app bundled in central, which is by far the most common method currently.

garethbowen:
Closes getodk/web-forms#875

Manual testing
CI
Checking file sizes:

1. Test localhost using: `npm run dev:build`
2. In chrome, disable cache, load a simple form, and look at how much was downloaded
3. Compare with staging

Staging: 1.9MB transferred, 5.5MB resources
This PR: 1.1MB transferred, 2.9MB resources

If my testing is correct, that's 43% less transferred.

Your mileage varies with different forms, but this PR is always smaller.

This gets HMR working so when running `npm dev` any change in any package updates the browser.
It also changes the bundling massively which makes it hard to verify that everything is still working and not bloating.

It reduces the size of the overall download so it should be a net positive.
There is a risk that because we're now not consuming the dist index.js of each package that we break users of the libraries. If this actually happens we can fix it with testing.

No.

### DIFF
--- a/bin/check-bundle-size.js
+++ b/bin/check-bundle-size.js
@@ -57,21 +57,17 @@ function isTooBig({ path, size }) {
   // TODO these have all changed drastically
   switch (simpleName) {
     case 'icomoon.svg':    return size >    60_000;
-    case 'index.js':       return size > 2_300_000;
-    case 'main.js':        return size >   400_000;
-    case 'forms.js':       return size > 2_100_000;
+    case 'forms.js':       return size > 1_400_000;
     case 'forms.css':      return size >   500_000;
-    case 'web-form.js':    return size > 1_000_000; // Matches web-form-renderer.js
-    case 'MapBlock.js':    return size >   600_000; // A Web Forms' feature bundle
-    case 'MapBlock.css':   return size >   500_000;
+    case 'web-form.js':    return size >   600_000; // Matches web-form-renderer.js
+    case 'MapBlock.css':   return size >   500_000; // A Web Forms' feature bundle
     case 'WebGLVector.js': return size >   500_000;
-    case 'geojson-map.js': return size >   500_000;
     // This is a performance tracking script from Sentry, shared between apps/central and apps/forms.
-    case 'browserTracingIntegration.js': return size > 400_000;
+    case 'browserTracingIntegration.js': return size > 350_000;
     default: // do nothing
   }
 
-  const type = extname(path).substr(1);
+  const type = extname(path).substring(1);
   switch (type) {
     case 'css':         return size > 220_000;
     case 'html':        return size >   3_000;
--- a/bin/check-bundle-size.js
+++ b/bin/check-bundle-size.js
@@ -54,16 +54,20 @@ log('File sizes look OK.');
 function isTooBig({ path, size }) {
   // Special cases:
   const simpleName = basename(path).replace(/(-[\w-]{8})+\./, '.');
+  // TODO these have all changed drastically
   switch (simpleName) {
     case 'icomoon.svg':    return size >    60_000;
     case 'index.js':       return size > 2_300_000;
     case 'main.js':        return size >   400_000;
     case 'forms.js':       return size > 2_100_000;
+    case 'forms.css':      return size >   500_000;
     case 'web-form.js':    return size > 1_000_000; // Matches web-form-renderer.js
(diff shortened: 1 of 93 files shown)


packages/web-forms/src/components/common/canvas/AsyncCanvas.vue
```diff
@@ -4,11 +4,11 @@
  * This prevents unnecessary bloat in the application bundle, reducing initial load times and improving performance.
  * Use dynamic imports instead (e.g., `await import(importPath)`) for lazy-loading these dependencies when required.
  */
-import { getModeConfig, type CanvasMode } from '@/components/common/canvas/getModeConfig.ts';
-import { loadCanvas, type CanvasBlockComponent } from '@/components/common/canvas/loadCanvas.ts';
-import AsyncLoader from '@/components/common/AsyncLoader.vue';
-import { TRANSLATE } from '@/lib/constants/injection-keys.ts';
-import type { Translate } from '@/lib/locale/useLocale.ts';
+import { getModeConfig, type CanvasMode } from '@getodk/web-forms/components/common/canvas/getModeConfig.ts';
```
garethbowen, line 7: This is unfortunate but this way all packages can have the same alias.

bin/check-bundle-size.js
```diff
@@ -56,18 +56,17 @@ function isTooBig({ path, size }) {
   const simpleName = basename(path).replace(/(-[\w-]{8})+\./, '.');
   switch (simpleName) {
     case 'icomoon.svg':    return size >    60_000;
-    case 'index.js':       return size > 2_300_000;
-    case 'main.js':        return size >   400_000;
-    case 'forms.js':       return size > 2_100_000;
-    case 'web-form.js':    return size > 1_000_000; // Matches web-form-renderer.js
-    case 'MapBlock.js':    return size >   600_000; // A Web Forms' feature bundle
-    case 'geojson-map.js': return size >   500_000;
+    case 'forms.js':       return size > 1_400_000;
```
garethbowen, line 59: All these sizes changed a lot. Some got bigger, some got smaller. I think it's just a function of vite bundling differently, and it's the overall sizes that matter.

packages/web-forms/tsconfig.app.json
```diff
@@ -28,7 +28,8 @@
       "@locales/*": ["./locales/*"],
```
latin-panda: We can remove L27 and L28, right? Since everything will now use `@getodk/web-forms/*`

packages/web-forms/src/index.ts
```diff
@@ -1,8 +1,8 @@
 import { webFormsPlugin } from './web-forms-plugin';
 import OdkWebForm from './components/OdkWebForm.vue';
-import { POST_SUBMIT__NEW_INSTANCE } from '@/lib/constants/control-flow.ts';
+import { POST_SUBMIT__NEW_INSTANCE } from './lib/constants/control-flow.ts';
```
latin-panda: Can these imports use `@getodk/web-forms/` too?

latin-panda (approved):
Nice improvement ✨ I tested locally, and this is what I observed: 

Total dist (with assets):
- Master: 9.55 MB
- This branch: 7.61 MB

js and css wasm raw:
- Master: 9.13 MB
- This branch: 6.53 MB

js and css gzipped: 
- Master: 2.89 MB
- This branch: 1.72 MB

Two minor comments, but approving now to unblock

packages/web-forms/tsconfig.app.json
garethbowen: Done.

packages/web-forms/src/index.ts
garethbowen: Done.

PR #1783, opened 2026-08-19
fix(web-forms#878): calling sum() on empty nodeset should return 0

### LINKED ISSUE #878: Sum of empty nodeset or nodeset with all empty values returns empty value rather than 0
### Expected Behaviour

The sum function returns 0 when called on an empty nodeset or a nodeset with all empty values.

### Observed Behaviour

The sum function returns empty string when called on an empty nodeset or a nodeset with all empty values.

### Reproduction Steps

Form (XML or XLSX file): [sum_diffs.xlsx](https://github.com/user-attachments/files/31198271/sum_diffs.xlsx)

Steps:
1. Launch in Web Forms
2. Notice that the Sum value is blank initially.
3. Select Blackberry and Cherimoya, notice that the Sum value is blank for both
4. Do the same in Collect and Enketo and notice that all three cases lead to a sum of 0

garethbowen:
Closes getodk/web-forms#878

Testing

I couldn't think of another solution.

No bug.

No.

### DIFF
--- a/.changeset/yellow-mugs-act.md
+++ b/.changeset/yellow-mugs-act.md
@@ -4,4 +4,4 @@
 "@getodk/web-forms": patch
 ---
 
-Fixed a bug where the xpath sum() function would return null if no nodes matched the expression
+Fixed a bug where the XPath sum() function returned NaN instead of 0 when no nodes matched the expression.
--- a/.changeset/yellow-mugs-act.md
+++ b/.changeset/yellow-mugs-act.md
@@ -1,0 +1,7 @@
+---
+"@getodk/xforms-engine": patch
+"@getodk/xpath": patch
+"@getodk/web-forms": patch
+---
+
+Fixed a bug where the xpath sum() function would return null if no nodes matched the expression
--- a/packages/xforms-engine/test/integration/repeat.test.ts
+++ b/packages/xforms-engine/test/integration/repeat.test.ts
@@ -2341,12 +2341,9 @@ describe('Tests ported from JavaRosa - repeats', () => {
            *
            * - Fails in direct port due to lack of position predicates in each
            *   reference to `outer_repeat`
-           *
-           * - Fails with parameterized option to make the position explicit, on
-           *   assertion of `0` value: current implementation of XPath `sum`
-           *   produces `NaN` when applied to an empty node-set (which is a bug).
            */
-          it.fails('[is] reevaluated when trigger[dependency?] changes', async () => {
+          const textFn = addExplicitRepeatPositionPredicate ? it : it.fails;
+          textFn('[is] reevaluated when trigger[dependency?] changes', async () => {
             const scenario = await Scenario.init(
               'Predicate trigger',
               html(
@@ -2391,37 +2388,17 @@ describe('Tests ported from JavaRosa - repeats', () => {
 
             if (addExplicitRepeatPositionPredicate) {
               scenario.answer('/data/outer_repeat[1]/cutoff_number', 3);
-            } else {
-              scenario.answer('/data/outer_repeat/cutoff_number', 3);
-            }
-
-            if (addExplicitRepeatPositionPredicate) {
               expect(scenario.answerOf('/data/outer_repeat[1]/sum')).toEqualAnswer(intAnswer(15));
-            } else {
-              expect(scenario.answerOf('/data/outer_repeat/sum')).toEqualAnswer(intAnswer(15));
-            }
-
-            if (addExplicitRepeatPositionPredicate) {
               scenario.answer('/data/outer_repeat[1]/cutoff_number', 7);
-            } else {
-              scenario.answer('/data/outer_repeat/cutoff_number', 7);
-            }
-
-            if (addExplicitRepeatPositionPredicate) {
               expect(scenario.answerOf('/data/outer_repeat[1]/sum')).toEqualAnswer(intAnswer(0));
-            } else {
-              expect(scenario.answerOf('/data/outer_repeat/sum')).toEqualAnswer(intAnswer(0));
-            }
-
-            if (addExplicitRepeatPositionPredicate) {
               scenario.answer('/data/outer_repeat[1]/cutoff_number', -11);
-            } else {
-              scenario.answer('/data/outer_repeat/cutoff_number', -11);
-            }
-
-            if (addExplicitRepeatPositionPredicate) {
               expect(scenario.answerOf('/data/outer_repeat[1]/sum')).toEqualAnswer(intAnswer(21));
             } else {
+              scenario.answer('/data/outer_repeat/cutoff_number', 3);
+              expect(scenario.answerOf('/data/outer_repeat/sum')).toEqualAnswer(intAnswer(15));
+              scenario.answer('/data/outer_repeat/cutoff_number', 7);
+              expect(scenario.answerOf('/data/outer_repeat/sum')).toEqualAnswer(intAnswer(0));
+              scenario.answer('/data/outer_repeat/cutoff_number', -11);
               expect(scenario.answerOf('/data/outer_repeat/sum')).toEqualAnswer(intAnswer(21));
             }
           });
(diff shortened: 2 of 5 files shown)


packages/xpath/src/functions/fn/number.ts
```diff
@@ -19,33 +19,25 @@ export const round = mathAlias('round');
...
-  (context, expressions): number => {
-    if (expressions.length === 0) {
-      return NaN;
-    }
-
-    let result!: number;
+  (context, [expression]): number => {
+    let result = 0;
 
-    for (const expression of expressions) {
```
garethbowen, line 29: It isn't valid in the spec to provide multiple expressions, and we have tests that assert the function throws if multiple are given.

packages/xpath/src/functions/fn/number.ts
```diff
@@ -19,33 +19,25 @@ export const round = mathAlias('round');
...
   'sum',
   [{ arityType: 'required' }],
-  (context, expressions): number => {
-    if (expressions.length === 0) {
-      return NaN;
-    }
-
-    let result!: number;
+  (context, [expression]): number => {
+    let result = 0;
```
garethbowen, line 23: This is the actual fix.

packages/xpath/src/functions/fn/number.ts
```diff
@@ -19,33 +19,25 @@ export const round = mathAlias('round');
...
+    for (const item of resultSet) {
+      const numberValue = item.toNumber();
 
-        if (Number.isNaN(numberValue)) {
-          result = NaN;
-
-          break;
-        }
+      if (Number.isNaN(numberValue)) {
+        return NaN;
```
garethbowen, line 37: The original implementation used break, which only broke out of the inner loop, not the outer loop. Because adding a NaN to a number is NaN, there's no point continuing so just return immediately.

packages/xpath/test/xforms/sum.test.ts
```diff
@@ -65,6 +65,45 @@ describe('#sum()', () => {
...
+        <item>abc</item>
+        <item>99</item>
+      </root>`);
+
+    const contextNode = testContext.document.getElementById('root');
+
+    testContext.assertNumberValue('sum(/root/item)', NaN, { contextNode });
+  });
+
+  it('sum of null value is NaN', () => {
```
lognaturel: That's the "blueberry" case in the form I shared, right? If so, I think the expected result to match Collect and Enketo is 89, it would be good to double check.

garethbowen: No it's not exactly the same... in your form the element isn't returned because the `<weight>` element doesn't exist. In this case it's empty.

When I run your form manually everything works as expected.

I'll double check...

lognaturel: Oh that's fascinating, XLSForm is taking care of that case, really. Would be good to have a test on the no child element case if there isn't one already, I think.

garethbowen: Enketo returns NaN in this case.

garethbowen: Ooof. Collect returns "0".

I've found conflicting information about what is "correct"... how about I standardise with Collect for now?

garethbowen: For example... xpath has a native evaluator test which asserts the opposite. I think Collect is inconsistent with whatwg specs, and now I don't know which to align with  :(

packages/xforms-engine/test/integration/repeat.test.ts
```diff
@@ -2341,12 +2341,9 @@ describe('Tests ported from JavaRosa - repeats', () => {
            *
            * - Fails in direct port due to lack of position predicates in each
            *   reference to `outer_repeat`
-           *
-           * - Fails with parameterized option to make the position explicit, on
-           *   assertion of `0` value: current implementation of XPath `sum`
-           *   produces `NaN` when applied to an empty node-set (which is a bug).
            */
-          it.fails('[is] reevaluated when trigger[dependency?] changes', async () => {
+          const textFn = addExplicitRepeatPositionPredicate ? it : it.fails;
```
latin-panda: typo
```suggestion
          const testFn = addExplicitRepeatPositionPredicate ? it : it.fails;
```

packages/xpath/test/xforms/sum.test.ts
latin-panda: Yes, NaN aligns better with the specs, but I think Collect's "0" is more useful, and many forms are built to work with Collect. 
I'd match Collect and document this decision as intentionally deviating from the spec.

latin-panda (commented):
Just one comment and waiting for the final decision to review the latest code again

packages/xpath/test/xforms/sum.test.ts
lognaturel: The good news is that it looks like XLSForm would never generate this so it's a bit of a theoretical concern.

I think @latin-panda's approach makes sense unless it's onerous.

packages/xforms-engine/test/integration/repeat.test.ts
garethbowen: Done.

packages/xpath/test/xforms/sum.test.ts
garethbowen: Not onerous at all! Done.

garethbowen: Do you think it's worth adding to the [spec docs](https://getodk.github.io/xforms-spec/#fn:sum)?

.changeset/yellow-mugs-act.md
```diff
@@ -0,0 +1,7 @@
+---
+"@getodk/xforms-engine": patch
+"@getodk/xpath": patch
+"@getodk/web-forms": patch
+---
+
+Fixed a bug where the xpath sum() function would return null if no nodes matched the expression
```
latin-panda: IMO, this makes it clearer
```suggestion
Fixed a bug where the XPath sum() function returned NaN instead of 0 when no nodes matched the expression.
```

packages/xpath/src/functions/fn/number.ts
```diff
@@ -19,33 +19,31 @@ export const round = mathAlias('round');
...
+    let result = 0;
 
-    for (const expression of expressions) {
-      const resultSet = expression.evaluate(context);
+    const resultSet = expression!.evaluate(context);
 
-      for (const item of resultSet) {
-        const numberValue = item.toNumber();
+    for (const item of resultSet) {
+      if (item.toString() === '') {
```
latin-panda:
```suggestion
      if (item.toString().trim() === '') {
```

Collect is also trimming 
See this form: [sum-empty-nodeset-test.xml](https://github.com/user-attachments/files/31263352/sum-empty-nodeset-test.xml)

packages/xpath/test/xforms/sum.test.ts
latin-panda: > Do you think it's worth adding to the [spec docs](https://getodk.github.io/xforms-spec/#fn:sum)?

I think it's okay since it's ODK's take on this feature, and the xforms-spec has a disclaimer in the footer: `Documents the XForms and OpenRosa-derived form specification as supported by ODK-compliant tools`

latin-panda (dismissed):
I explored further today and found a case where Collect returns a sum but WF does not.

packages/xpath/src/functions/fn/number.ts
lognaturel: Nice catch! This is strange but not too surprising, there's this kind of trimming in various places in JR/Collect. Agreed that it would be good to match.

.changeset/yellow-mugs-act.md
garethbowen: Done.

packages/xpath/src/functions/fn/number.ts
garethbowen: Done, and unit test added.

packages/xpath/test/xforms/sum.test.ts
garethbowen: PR raised: https://github.com/getodk/xforms-spec/pull/369

lognaturel (approved):
Thanks! 🎉

PR #1784, opened 2026-08-19
Filter app users and links by actor properties

### LINKED ISSUE #1871: Filter the App Users list by user property values
On the App Users tab, allow filtering the list by user property values so project managers can find users assigned to specific properties.

## Scope
- Filter controls above the App Users table
- Filter by one or more user property values
- Works alongside the existing App Users list

## References
- Depends on backend: #1868 
- Related: #1790 (column display and editing)

### LINKED ISSUE #1874: Filter the Public Access Links list by user property values
On the Public Access Links tab, allow filtering the list by user property values so project managers can find links assigned to specific properties.

## Scope
- Filter controls above the Public Access Links table (per Figma)
- Filter by one or more user property values
- Works alongside the existing Public Access Links list/table

## References
- Figma
- Depends on backend: https://github.com/getodk/central/issues/1868
- Mirrors the App Users list filter issue https://github.com/getodk/central/issues/1871

ktuite:
Closes https://github.com/getodk/central/issues/1874 and https://github.com/getodk/central/issues/1871

Screenshot 2026-08-18 at 6 48 11 PM

Screenshot 2026-08-18 at 6 48 21 PM

Tests and trying it out.

Existing actor property and entity filtering docs are probably sufficient.

### DIFF
--- a/.changeset/many-trains-draw.md
+++ b/.changeset/many-trains-draw.md
@@ -1,5 +1,0 @@
----
-"@getodk/central-frontend": patch
----
-
-feat getodk/central#1874 and getodk/central#1871: Filter app users and public access links by actor properties
--- a/.changeset/many-trains-draw.md
+++ b/.changeset/many-trains-draw.md
@@ -1,0 +1,5 @@
+---
+"@getodk/central-frontend": patch
+---
+
+feat getodk/central#1874 and getodk/central#1871: Filter app users and public access links by actor properties
--- a/apps/central/src/components/custom-props-filter.vue
+++ b/apps/central/src/components/custom-props-filter.vue
@@ -21,7 +21,7 @@
               {{ p.name }}
             </option>
           </select>
-          <span class="equals-operator">=</span>
+          <span>=</span>
           <select v-model="pendingValue" class="form-control value-select"
             :disabled="pendingProperty === ''" :aria-label="$t('header.value')">
             <option value="" disabled>{{ $t('header.value') }}</option>
@@ -99,7 +99,7 @@ const availableValues = computed(() => {
 
 // Badge text shown on the trigger button
 const displayValue = computed(() => {
-  if (props.modelValue == null) return t('common.all');
+  if (props.modelValue == null) return t('action.all');
   return `${props.modelValue.property} = ${props.modelValue.value}`;
 });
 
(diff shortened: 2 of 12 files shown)


.changeset/many-trains-draw.md
sadiqkhoja, line 1: maybe delete this?

sadiqkhoja:
when property name is too long then dropdown looks like this:

image

sadiqkhoja:
cosmetic change: I think filter should be left align with the table.

apps/central/src/locales/en.json5
```diff
@@ -423,6 +424,7 @@
     "conflictsCount": "{count} conflict | {count} conflicts"
   },
   "common": {
+    "all": "All",
```
sadiqkhoja: We have i18n for "All" in other filters as well like `creator`, `conflict`, `review state`. @matthew-white is this one of those words which will be translated according to the context?

apps/central/src/components/custom-props-filter.vue
```diff
@@ -0,0 +1,216 @@
...
+      <li class="filter-row">
+        <div class="form-group">
+          <select v-model="pendingProperty" class="form-control property-select"
+            :aria-label="$t('resource.property')" @change="pendingValue = ''">
+            <option value="" disabled>{{ $t('resource.property') }}</option>
+            <option v-for="p of actorProperties" :key="p.name" :value="p.name">
+              {{ p.name }}
+            </option>
+          </select>
+          <span class="equals-operator">=</span>
```
sadiqkhoja: I hope that `=` is a universal symbol and not required to be translated

sadiqkhoja:
Should we reset filter after creation of new record? because if I apply some filtering and create an app user with a different value then new app user doesn't appear on the table.

apps/central/src/locales/en.json5
```diff
@@ -196,6 +196,7 @@
   },
   "action": {
     "add": "Add",
+    "apply": "Apply",
```
sadiqkhoja: We have "Apply" in `multiselect` component so we can use @transifexKey here

sadiqkhoja (approved):

apps/central/src/locales/en.json5
matthew-white: Yes, I think that's needed. E.g., in Spanish I think it's todos vs. todas depending on the noun.

PR #1785, opened 2026-08-19
fix(web-forms#877): do not error trying to parse strings as dates

### LINKED ISSUE #877: Inconsistent handling of malformed dates
### Expected Behaviour

Evaluating a string which is an invalid date (eg: `2026-13-01`) should not throw an error.

### Observed Behaviour

If a string looks like a date the xpath package tries to parse it. Certain invalid date forms are caught and return null, but others throw errors.

### Reproduction Steps

Form (XML or XLSX file):

```xml
<?xml version="1.0"?>
    <h:head>
        <h:title>invalid values</h:title>
            <instance>
                <data id="invalid_values_en" version="1">
                    <bad_date_malformed/>
                    <meta>
                        <instanceID/>
                    </meta>
                </data>
            </instance>
            <bind nodeset="/data/bad_date_malformed" type="string"/>
            <setvalue ref="/data/bad_date_malformed" event="odk-instance-first-load" value="'2026-13-15'"/>
        </model>
    </h:head>
    <h:body>
        <input ref="/data/bad_date_malformed">
            <label>Date defaulted from a malformed date</label>
            <hint>Expected: blank.</hint>
        </input>
    </h:body>
</h:html>
```

Steps:
1. Load the above form

### Screenshots or Videos

Image

### Environment

- Browser version: Any
- OS: Any
- Web Forms version: 1.0.2

### Additional Context

Raised during testing for last-saved: https://github.com/getodk/web-forms/issues/306#issuecomment-5314280890

It's particularly problematic in the case of last-saved because unlike the form above, it doesn't throw on first launch, only if the last-saved date is invalid, so it may happen remotely or to a user without the ability to fix the form. It's also very difficult for the end user to clear the last-saved xml so they can resume their work.

garethbowen:
Closes getodk/web-forms#877

testing

I thought about modifying the `dateTimeFromString` function in coercion.ts so that it would never throw, but it's used elsewhere (such as the format-date function) so it may cause other regressions. Because this is targeting a patch release that felt too risky. I'm going to raise an issue to revisit this for the next minor when we have a little more time to do it right.

No bug.

None.


packages/xforms-engine/test/integration/actions-events/set-value-action.test.ts
```diff
@@ -610,6 +610,25 @@ describe('setvalue action', () => {
...
+              setvalue('odk-instance-first-load', '/data/input', "'2026-13-15'")
+            )
+          ),
+          body(input('/data/input'))
+        )
+      );
+
+      expect(scenario.answerOf('/data/input')).toEqualAnswer(stringAnswer('2026-13-15'));
+    });
+
```
garethbowen, line 631: An integration test feels like overkill but it's caused by the interaction of multiple files so I couldn't find a way to reproduce it in an xpath unit test.

packages/xpath/src/evaluations/StringEvaluation.ts
```diff
@@ -25,17 +25,19 @@ export class StringEvaluation<T extends XPathNode> extends ValueEvaluation<T, 'S
 
     if (isEmpty) {
```
latin-panda, line 26: The NodeEvaluation is also breaking, for example this form: 

[malformed_date_field_value.xml](https://github.com/user-attachments/files/31218236/malformed_date_field_value.xml)

latin-panda: Could you please add these tests using XPath? I'm more interested in the node value cases.  

```
import { createXFormsTestContext } from '../helpers.ts';

describe('malformed date-like strings', () => {
  let testContext: XFormsTestContext;

  describe('string literals', () => {
    beforeEach(() => {
      testContext = createXFormsTestContext();
    });

    [
      { expression: '"2026-13-15"', expected: '2026-13-15' },
      { expression: '"2026-02-30"', expected: '2026-02-30' },
      { expression: '"2026-13-15T00:00:00Z"', expected: '2026-13-15T00:00:00Z' },
    ].forEach(({ expression, expected }) => {
      it(`evaluates ${expression} as a plain string`, () => {
        testContext.assertStringValue(expression, expected);
      });
    });

    it('converts a malformed date literal to NaN', () => {
      testContext.assertNumberValue('number("2026-13-15")', NaN);
    });
  });

  describe('node values', () => {
    beforeEach(() => {
      testContext = createXFormsTestContext('<root><a>2026-13-15</a></root>');
    });

    it('evaluates a node containing a malformed date as a plain string', () => {
      testContext.assertStringValue('concat(/root/a, "!")', '2026-13-15!');
    });

    it('converts a node containing a malformed date to NaN', () => {
      testContext.assertNumberValue('number(/root/a)', NaN);
    });

    it('compares a node containing a malformed date without throwing', () => {
      testContext.assertBooleanValue('/root/a < 1', false);
    });
  });
});
```

latin-panda (changes requested):
The string cases I tested are working as expected; however, I found a case where I got an exception with node values. Could we fix it in this PR as well?

packages/xpath/src/evaluations/StringEvaluation.ts
garethbowen: Thanks for the tests! I added them to the date test.

The NodeEvaluation was creating a StringEvaluation so I now just call `toNumber()` on that which uses the new code.

The `number` function was also breaking so I added try/catch to that too.

latin-panda (approved):
It works well in all the cases I tested! Thanks for the quick fix!

PR #1786, opened 2026-08-19
chore: remove unused hanken grotesk font

garethbowen:
I'm fairly sure this font was only used in the demo app.


latin-panda (approved):
Thank you!

PR #1787, opened 2026-08-19
Increased row height

sadiqkhoja:
Towards https://github.com/getodk/central/issues/1604

image

### DIFF
--- a/apps/central/src/components/public-link/table.vue
+++ b/apps/central/src/components/public-link/table.vue
@@ -71,10 +71,6 @@ const { publicLinks, actorProperties } = useRequestData();
   th.access-link, td.access-link {
     max-width: 350px;
   }
-  td {
-    height: 52px; // Access link cells has scrollbar
-    vertical-align: unset;
-  }
 }
 </style>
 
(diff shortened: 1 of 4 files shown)


apps/central/src/components/public-link/table.vue
```diff
@@ -75,9 +75,6 @@ const { publicLinks, actorProperties } = useRequestData();
     height: 52px; // Access link cells has scrollbar
     vertical-align: unset;
   }
```
sadiqkhoja: this can go away

apps/central/src/components/table/freeze.vue
```diff
@@ -136,6 +136,11 @@ defineExpose({ getRowPair });
 .table-freeze {
   @include clearfix;
   position: relative;
+
+  th, td {
+    height: 56px;
+    vertical-align: middle;
+  }
```
sadiqkhoja: maybe this is not required because table styles are set in app.scss

sadiqkhoja: if required then add mixin

matthew-white (approved):

sadiqkhoja:
remove:

```
.table tbody .custom-props-data-row td {
  height: 37px;
}
```

PR #1790, opened 2026-08-19
chore(#2146): allow failure downloading fonts on redirect

### LINKED ISSUE #2146: [frontend] e2e-tests: intermittent failure: font download cancelled on redirect
Example: https://github.com/getodk/central-frontend/actions/runs/32303355143/job/96245470205

```
1) [firefox] › tests/web-forms.spec.js:136:7 › ODK Web Forms › redirects to login if 401 on load › f prefix 

    Error: Unexpected fatal error(s) logged: 
        0. [firefox|console.error] :0:72173
        message: [JavaScript Error: "downloadable font: download failed (font-family: "Roboto" style:normal weight:400 stretch:100 src index:0): status=2152398850 source: http://central-test.localhost/assets/roboto-latin-400-normal-BqEyEoaF.woff2"]

    expect(received).toHaveLength(expected)

    Expected length: 0
    Received length: 1
    Received array:  ["[firefox|console.error] :0:72173
        message: [JavaScript Error: \"downloadable font: download failed (font-family: \"Roboto\" style:normal weight:400 stretch:100 src index:0): status=2152398850 source: [http://central-test.localhost/assets/roboto-latin-400-normal-BqEyEoaF.woff2\](http://central-test.localhost/assets/roboto-latin-400-normal-BqEyEoaF.woff2/)"]"]
        at     0. [firefox|console.error] :0:72173
        at Object.testBase.extend.browserConsoleToTestStdout.auto (/home/runner/work/central-frontend/central-frontend/central/client/e2e-tests/util.js:88:9)
```

A recent change stopped the font from being bundled into the css so it downloads as a separate request. I think what's happening is the redirect fires before all downloads have completed which cancels the download. This is a special case and we should either ignore cancelled request failures, or disable them for this test specifically.

garethbowen:
Closes getodk/central#2146

This makes the minimal change to get this one test to pass reliably without impacting others.

None.

No.

### DIFF
--- a/e2e-tests/tests/web-forms.spec.js
+++ b/e2e-tests/tests/web-forms.spec.js
@@ -135,6 +135,7 @@ test.describe('ODK Web Forms', () => {
     urls.forEach(t => {
       test(t.name, async ({ allowedLogs, page }) => {
         allowedLogs.push((consoleMsg, normalisedMsg) => {
+          // the redirect causes the browser to cancel outstanding requests which should not fail the test
           if(normalisedMsg.includes('downloadable font: download failed')) return true;
           if(normalisedMsg !== 'Failed to load resource: the server responded with a status of 401 (Unauthorized)') return false;
           const { url } = consoleMsg.location();
--- a/e2e-tests/tests/web-forms.spec.js
+++ b/e2e-tests/tests/web-forms.spec.js
@@ -135,7 +135,8 @@ test.describe('ODK Web Forms', () => {
     urls.forEach(t => {
       test(t.name, async ({ allowedLogs, page }) => {
         allowedLogs.push((consoleMsg, normalisedMsg) => {
-          if(normalisedMsg !== 'Failed to load resource: the server responded with a status of 401 (Unauthorized)') return;
+          if(normalisedMsg.includes('downloadable font: download failed')) return true;
+          if(normalisedMsg !== 'Failed to load resource: the server responded with a status of 401 (Unauthorized)') return false;
           const { url } = consoleMsg.location();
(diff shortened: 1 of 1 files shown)


garethbowen:
I ran this 5 times with 100% success: https://github.com/getodk/central-frontend/actions/runs/32311478868?pr=1790

alxndrsn:
This feels conceptually similar to https://github.com/getodk/central-frontend/pull/1754.

Is there a way to understand if the error occurred after redirect or not?  Perhaps it's impractical.  But it feels like this change would prevent detection if the font loading was actually broken.



> I ran this 5 times with 100% success: https://github.com/getodk/central-frontend/actions/runs/32311478868?pr=1790

Recommended for future:

```diff
 jobs:
   your-job:
+    strategy:
+      fail-fast: false
+      matrix:
+        x: [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 ]
+        y: [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 ]
```

garethbowen:
> Is there a way to understand if the error occurred after redirect or not? Perhaps it's impractical. But it feels like this change would prevent detection if the font loading was actually broken.

Because it's only on the one test where the redirect occurs the font loading should still fail all the other tests.

lognaturel (approved):
I think a comment to say the font failure case is to handle flakes would be helpful! Right now it looks like that's a success case which it isn't really.

lognaturel:
Oh, I see. It is a success case because the font request is stopped *because* of the redirect. Fair enough, The link to this PR and issue is probably enough context.

PR #1791, opened 2026-08-20
fix(web-forms#881): distance function should allow empty input

### LINKED ISSUE #881: Calling distance with empty value throws an error
### Expected Behaviour

When the value of the function is not yet defined the function should return null so I can load the form and enter the value.

### Observed Behaviour

An error is thrown.

### Reproduction Steps

Form (XML or XLSX file):

```xml
<?xml version="1.0"?>
<?xml version="1.0"?>
    <h:head>
        <h:title>distance on load</h:title>
            <instance>
                <data id="distance_on_load" version="1">
                    <distance/>
                    <location1/>
                    <location2/>
                    <meta>
                        <instanceID/>
                    </meta>
                </data>
            </instance>
            <bind nodeset="/data/location1" type="geopoint"/>
            <bind nodeset="/data/location2" type="geopoint"/>
            <bind nodeset="/data/distance" type="string" calculate="round(distance(/data/location1, /data/location2), 2)"/>
        </model>
    </h:head>
    <h:body>
        <input ref="/data/location1" appearance="placement-map">
            <label>First point</label>
        </input>
        <input ref="/data/location2" appearance="placement-map">
            <label>Second point</label>
        </input>
        <input ref="/data/distance">
            <label>Distance</label>
        </input>
    </h:body>
</h:html>
```

Steps:
1. Load the form

### Screenshots or Videos

Image

### Additional Context

Reported on the forum: https://forum.getodk.org/t/webforms-behaviour-notes/58272/35
The equivalent error in javarosa only fires on invalid value (eg: "foo"): https://github.com/getodk/javarosa/blob/1c17b23c81b2e87ec57360bf5a4b48e2b0d15c52/src/test/java/org/javarosa/core/util/GeoDistanceTest.java#L207

garethbowen:
Closes getodk/web-forms#881

Existing test suite.
Compared with Collect and Enketo.

Honestly it feels clunky. Some of the problem is that the distance function and the area function behave differently with invalid geopoints, for example, passing in the letter `a` returns 0 for area, and throws for distance. Perhaps this is wrong but it's enforced by tests...

The other part of the issue is the function can take a single geotrace/geoshape, or an array of geopoints. This makes for messy implementation.

If you have a cleaner solution I'm all ears!

None.

None.


packages/xpath/src/functions/xforms/geo.ts
latin-panda: In this branch, a new exception is thrown, and the form doesn't load when there are invalid points. In master, the form still loads. Maybe a try/catch here.

form: [geofence-bug-repro.xml](https://github.com/user-attachments/files/31260240/geofence-bug-repro.xml)

packages/xpath/src/functions/xforms/geo.ts
```diff
@@ -104,15 +108,34 @@ export const distance = new NumberFunction(
...
+    const lines = geotrace?.lines;
 
     if (lines == null) {
-      throw new JRCompatibleGeoValueError('distance');
+      if (values.length === 1 && values[0]) {
+        // a single valid geopoint given
+        return NaN;
+      }
+      throw new Error(
+        "The function 'distance' received a value that does not represent GPS coordinates"
```
latin-panda: The message is repeated, it can be extracted to a variable

packages/xpath/test/xforms/geo.test.ts
```diff
@@ -30,6 +30,8 @@ describe('geo functions', () => {
     { argument: '0 0;0 1', expected: { area: 0.0, distance: 111318.85 } },
     { argument: '0 0;0 90', expected: { area: 0.0, distance: 10018696.05 } },
     { argument: '90 0;90 1', expected: { area: 0.0, distance: 0.0 } },
+    { argument: '', expected: { area: 0.0, distance: NaN } },
```
latin-panda, line 34: I think we need a new test with empty values, invalid values in multiple points, e.g.`distance('', '0 0', '0 1')`

packages/xpath/src/functions/xforms/geo.ts
```diff
@@ -71,7 +70,12 @@ const evaluateArgumentValues = <T extends XPathNode>(
 
 export const area = new NumberFunction('area', [{ arityType: 'required' }], (context, args) => {
   const values = evaluateArgumentValues(context, args);
-  const geotrace = Geotrace.fromEncodedValues(values);
+  let geotrace;
+  try {
```
latin-panda: It feels weird that the `area` behaves differently, for example, by not skipping empty values like `distance` does. Maybe it's an old JavaRosa discrepancy we are carrying into WF :(

packages/xpath/test/xforms/geo.test.ts
```diff
@@ -40,29 +42,23 @@ describe('geo functions', () => {
     });
   });
 
+  it(`distance(garet) works ()`, () => {
```
latin-panda:
```suggestion
  it(`distance returns NaN for a single geopoint'`, () => {
```

packages/xpath/src/functions/xforms/geo.ts
latin-panda: Another way to make these differences clearer in the code is to create a function that evaluates the result and lets these functions (`area`, `distance`, ...) decide what to do in each case. For example: 

```ts
export const distance = new NumberFunction('distance', ..., (context, args) => {                                                                                                                   
        const result = parsePoints(context, args);                                                                                                            
                                                                                                                                                                                                     
        switch (result.status) {                                                                                                                                                                     
                case 'empty': return NaN;                                                                                                                                                     
                case 'single-point': return NaN;                                                                                                                                                     
                case 'invalid': throw new Error("The function 'distance' received a value...");                                                                                                 
                case 'valid': return toAbsolutePrecision(sum(result.geotrace.lines.map(geodesicDistance)), PRECISION);                                                                        
        }                                                                                                                                                                                            
  }); 

export const area = new NumberFunction('area', ..., (context, args) => {                                                                                                                   
        const result = parsePoints(context, args);                                                                                                            
                                                                                                                                                                                                     
        switch (result.status) {                                                                                                                                                                     
                case 'empty': return ....;                                                                                                                                                     
                case 'single-point': return ..;                                                                                                                                                     
                case 'invalid': throw new Error("The function 'distance' received a value...");                                                                                                 
                case 'valid': return ....;                                                                        
        }                                                                                                                                                                                            
  }); 
```

IMO it's faster to understand something like this

latin-panda (changes requested):
I noticed a minor regression in traces. The rest are small suggestions.

packages/xpath/test/xforms/geo.test.ts
garethbowen: Hah. Sorry. I pulled that out from the array above so I could .only it and gave it a stupid name so you would find it  :D 

It should just be deleted.

garethbowen: Done.

packages/xpath/test/xforms/geo.test.ts
garethbowen: Done.

packages/xpath/src/functions/xforms/geo.ts
garethbowen: I've added a unit test for this.

packages/xpath/src/functions/xforms/geo.ts
garethbowen: I've refactored the code so now it's no longer duplicated.

packages/xpath/src/functions/xforms/geo.ts
garethbowen: Yes you're right. I've made a significant refactor so the common validation just returns a boolean, and the individual functions can handle as they see fit.

100% agree that it's weird that distance throws instead of returning NaN in this case. There's a broader topic about how we handle xpath errors in general. At the moment if an error is thrown the user cannot continue with the form. I think we should treat them the same as validation errors which block form submission but give you a chance to fix them. This is consistent with collect. I'll add this topic to our next team meeting.

latin-panda: > I think we should treat them the same as validation errors which block form submission but give you a chance to fix them. This is consistent with collect.

Yes, I agree with this!

packages/xpath/src/lib/geo/Geotrace.ts
```diff
@@ -36,50 +28,33 @@ const collectLines = (geopoints: GeotracePoints): readonly GeotraceLine[] => {
...
-    if (tail.length === 0) {
-      return this.fromEncodedGeotrace(head);
-    }
+  );
+};
 
-    const geopoints = values.map((value) => {
-      return geopointCodec.decodeValue(value);
-    });
+export const validate = (values: readonly string[]): GeopointValidationResult => {
```
latin-panda, line 42: I tested several edge cases in the latest Collect and compared them with Web Forms. These are the differences: 

- Some display blank
- Some have 2 decimals; we could match the maps component, where we use 5.
- Some have different results in the calculation. 

But here I'm using edge cases for coordinates (0.0, 1.1) 
Do we want to match Collect? 

[geo-distance-area.xml](https://github.com/user-attachments/files/31316079/geo-distance-area.xml)

Image

latin-panda: I forgot to include these:
  - `distance('abc')` and `area('abc')` (invalid string)                                                                                                                                                                                                                                              
  - `distance('0 0;;0 1')` and `area('0 0;;0 1')` (blank segment in the middle)

packages/xpath/src/functions/xforms/geo.ts
```diff
@@ -104,15 +109,25 @@ export const distance = new NumberFunction(
   'distance',
   [{ arityType: 'required' }, { arityType: 'variadic' }],
   (context, args) => {
-    const values = evaluateArgumentValues(context, args);
-    const lines = Geotrace.fromEncodedValues(values)?.lines;
+    const argValues = evaluateArgumentValues(context, args);
+    const values = argValues.filter((value) => value.trim() !== '');
```
latin-panda: The untrimmed values are passed to validate L119
```suggestion
    const values = argValues.map(value => value.trim()).filter((value) => value !== '');
```
The same with area

Maybe this trim and filter could be moved to `validate`

latin-panda (changes requested):
I included some cases where the results differ from Collect. Let me know what you think :)

packages/xpath/src/functions/xforms/geo.ts
garethbowen: Fixed. Thanks.

Area worked anyway, because it takes a single argument as a geotrace/geoshape, which the validate function already cleans up.

I've added more tests.

packages/xpath/src/lib/geo/Geotrace.ts
garethbowen: - "area square": I think those are the same but Collect is using scientific notation, so it should be green or yellow.
- "field area in hectares": That should be yellow I think?

Then there are specific categories;
1. Rounding. IMO it's incorrect for the xpath library to round at all unless instructed by the client. In this case I propose we remove all rounding and let the form designer use the [round](https://getodk.github.io/xforms-spec/#fn:round) function to get the appropriate level of accuracy. If you agree I'll raise an issue to work on this separately.
2. Distance with 0 or 1 point. I think Collect is wrong here as returning 0 is misleading.
3. Distance with a middle blank point. I think Collect is wrong here too. Here WF skips over the blank values and just measures the distance from "10 10" to "20 20". Whereas I think Collect is measuring "10 10" to "0 0" + "0 0" to "20 20". I'm not sure WF is correct, but Collect's answer is almost certainly not what the form designer wanted.

Does that sound right to you?

I hope these are edge cases and form designers are using relevance or similar to only show the distance/area when all required points are given. If not, we should probably discuss further and clarify in the spec.

latin-panda: > Rounding. IMO it's incorrect for the xpath library to round at all unless instructed by the client. In this case I propose we remove all rounding and let the form designer use the [round](https://getodk.github.io/xforms-spec/#fn:round) function to get the appropriate level of accuracy. If you agree I'll raise an issue to work on this separately.

I agree, that makes sense

>Distance with 0 or 1 point. I think Collect is wrong here as returning 0 is misleading.
Distance with a middle blank point. I think Collect is wrong here too. Here WF skips over the blank values and just measures the distance from "10 10" to "20 20". Whereas I think Collect is measuring "10 10" to "0 0" + "0 0" to "20 20". I'm not sure WF is correct, but Collect's answer is almost certainly not what the form designer wanted.

Yes, I prefer to err here rather than try to calculate anything with bad data. What do you think about documenting these differences from Collect and filing issues in Collect repo?

latin-panda:
I ran out of time today but I'll continue this review and test again tomorrow

packages/xpath/src/lib/geo/Geotrace.ts
garethbowen: - https://github.com/getodk/web-forms/issues/885
- https://github.com/getodk/xforms-spec/issues/370

latin-panda (approved):
Looks good, and it works well :)
I didn't find anything new.

packages/xpath/src/functions/xforms/geo.ts
garethbowen: https://github.com/getodk/web-forms/issues/883

PR #1795, opened 2026-08-21
Move placement of EntityUploadDataTemplate and update text

matthew-white:
This PR updates the entity upload modal, making progress on getodk/central#1904:

- Moves the link to download the data template to the first line of the drop zone
- Updates text

Tests continue to pass. I also tried it out locally.

Mostly it's just moving things around.


apps/central/src/components/entity/upload/file-select.vue
```diff
@@ -13,17 +13,21 @@ except according to the terms contained in the LICENSE file.
...
-          </button>
-        </template>
-      </i18n-t>
+      <div>
+        <i18n-t keypath="text.full">
+          <template #chooseFile>
+            <!-- eslint-disable-next-line vuejs-accessibility/form-control-has-label -->
+            <input v-show="false" ref="input" type="file" accept=".csv,.tsv"
+              @change="changeInput">
+            <button type="button" class="btn btn-primary" :aria-disabled="parsing"
```
matthew-white:
```suggestion
            <button type="button" class="btn btn-outlined" :aria-disabled="parsing"
```

matthew-white (commented):

ktuite (approved):

apps/central/src/components/entity/upload/file-select.vue
matthew-white: Done in 7738ba9af6f6c63842a99e3f44cec75635ec7e04

PR #1798, opened 2026-08-22
Make routes case-sensitive

matthew-white:
This PR updates Central's route matching to be case-sensitive.

I added a new test. The test fails without the changes to src/router.js.

This PR uses Vue Router's built-in configuration: https://router.vuejs.org/guide/essentials/route-matching-syntax.html#Sensitive-and-strict-route-options

There's a chance that someone's existing link will break. However, the current, less strict behavior isn't something that we've intentionally supported or documented. I think any user impact should be low.


apps/central/test/router.spec.js
```diff
@@ -12,6 +12,36 @@ import { mockResponse } from './util/axios';
 import { setLanguages } from './util/i18n';
 
 describe('createCentralRouter()', () => {
+  describe('route matching', () => {
+    it('matches case-sensitively', () => {
+      mockLogin();
+      return load('/account/edit')
+        .afterResponses(app => {
+          app.findComponent(NotFound).exists().should.be.false;
```
matthew-white, line 23: This is the easiest way to assert that we're on the "not found" / 404 page.

matthew-white (commented):
Adding a comment to facilitate async code review

matthew-white:
Originally I specified `strict: true` in addition to `sensitive: true`, thinking that that would close out getodk/central#1697. However, I read that issue too quickly and realized that ~`sensitive: true`~ `strict: true` doesn't actually solve it. I've updated this PR to clarify that it is now just about making routes case-sensitive.

**UPDATE:** Actually, maybe `strict: true` would solve getodk/central#1697. 🤔 I've pushed #1800 along those lines. However, I think that can be considered separately from this PR.

matthew-white:
This PR does not close getodk/central#1697. However, I've pushed two separate PRs, either of which would close the issue: #1799 and #1800.

apps/central/test/router.spec.js
alxndrsn: That's very confusing - I *would* expect the `NotFound` component to exist on the "not found" page...

alxndrsn (commented):
(See comment on `router.spec.js`.)

apps/central/test/router.spec.js
matthew-white: > I *would* expect the `NotFound` component to exist on the "not found" page

Sorry, yes, that is how it works. If the `NotFound` component exists, that means you're on the "not found" page. If it doesn't, you're not on that page.

> This is the easiest way to assert that we're on the "not found" / 404 page.

Sorry, I should have said that this is the easiest way to assert *whether* we're on the "not found" page. Initially in the test, we're not on that page. In this test, first we assert that a lowercase path doesn't go to "not found", then we change the path to uppercase and see that now it is "not found".

I've added comments to the test and tried to make it clearer overall. Let me know what you think!

matthew-white:
@alxndrsn I've replied to your comment and tried to make the test clearer. Let me know what you think!

apps/central/test/router.spec.js
alxndrsn: Maybe extracting a function would be helpful here.

E.g. 

```js
assertPageNotFound()
// or
assertPageFound()
```

or maybe more idiomatically

```js
currentPage.should.be(NotFoundPage);
// or
currentPage.should.not.be(NotFoundPage);
```

alxndrsn (approved):

apps/central/test/router.spec.js
matthew-white: I like that idea! I've filed getodk/central#2157 to add something along those lines.

PR #1799, opened 2026-08-22
Remove trailing slash from route path

### LINKED ISSUE #1697: [frontend] broken redirect to `/submissions`
# Repro

2. observe "Page Not Found" Error:

Image

## Additionally

The page header indicates that the user is "Not logged in", even when they are.  Filed at #1698.

# Cause

https://staging.getodk.cloud/projects/203/forms/trees_update/ seems to append `/submissions` to the URL.  This works OK if the original URL had no trailing slash.

matthew-white:
Closes getodk/central#1697.

I added new tests.

A few options were considered in getodk/central#1697. The discussion gravitated toward fixing the issue in Central Frontend (the Vue router) rather than in nginx.

We can always implement something at the nginx level later if we decide that that'd be useful. This change doesn't foreclose that.

This PR fixes the issue by removing the trailing slash. That's the idea that came up in the issue. However, I actually think I prefer the approach in #1800. That PR simply doesn't match any Vue routes if the path has a trailing slash.

This PR adds more redirect logic and a little more complexity to the router. However, I think that aspect of the router is well tested. This PR adds new tests as well.

### DIFF
--- a/apps/central/src/router.js
+++ b/apps/central/src/router.js
@@ -65,16 +65,11 @@ router.afterEach(unlessFailure(to => {
   // user is trying to navigate to a parent route, which isn't expected.
   router.beforeEach(to => (Object.keys(to.meta).length === 0 ? '/' : true));
 
-  // Remove any trailing slash from the path.
+  // Remove trailing slashes from the path.
   router.beforeEach(to => {
-    // Two trailing slashes won't match any route, so return `true` to proceed
-    // to NotFound.
-    if (to.path.endsWith('//')) return true;
-
-    if (to.path.endsWith('/') && to.path !== '/')
-      return { path: to.path.slice(0, -1), query: to.query, hash: to.hash };
-
-    return true;
+    let { path } = to;
+    while (path.endsWith('/') && path !== '/') path = path.slice(0, -1);
+    return path === to.path ? true : { path, query: to.query, hash: to.hash };
   });
 
   // If a route is nested, its relative path is '', and that path is an alias,
--- a/apps/central/src/router.js
+++ b/apps/central/src/router.js
@@ -67,10 +67,13 @@ router.afterEach(unlessFailure(to => {
 
   // Remove any trailing slash from the path.
   router.beforeEach(to => {
-    if (to.path.endsWith('//')) // Not a valid Frontend URL
-      return '/';
+    // Two trailing slashes won't match any route, so return `true` to proceed
+    // to NotFound.
+    if (to.path.endsWith('//')) return true;
+
     if (to.path.endsWith('/') && to.path !== '/')
       return { path: to.path.slice(0, -1), query: to.query, hash: to.hash };
+
     return true;
   });
 
--- a/apps/central/src/router.js
+++ b/apps/central/src/router.js
@@ -70,7 +70,7 @@ router.afterEach(unlessFailure(to => {
     if (to.path.endsWith('//')) // Not a valid Frontend URL
       return '/';
     if (to.path.endsWith('/') && to.path !== '/')
-      return { path: to.path.slice(0, -1), query: to.query };
+      return { path: to.path.slice(0, -1), query: to.query, hash: to.hash };
     return true;
   });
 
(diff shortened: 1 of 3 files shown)


matthew-white:
Sorry @alxndrsn, I thought I'd tagged you on this PR already. It'd be great if you wouldn't mind taking a look. This PR is an alternative to #1800, and you may prefer the approach that this PR takes.

apps/central/test/router.spec.js
```diff
@@ -139,6 +156,19 @@ describe('createCentralRouter()', () => {
         });
     });
 
+    // getodk/central#1697
+    it('redirects to .../submissions despite a trailing slash', () => {
```
matthew-white: This test is very similar to the one directly above. The key difference is this line:

```js
.route('/projects/1/forms/f/')
```

Note that the path has a trailing slash.

matthew-white (commented):
Adding a comment to facilitate async code review.

matthew-white:
@alxndrsn I just pushed a few small changes, but I'm done now. This PR is ready for review.

apps/central/src/router.js
```diff
@@ -65,6 +65,18 @@ router.afterEach(unlessFailure(to => {
...
   router.beforeEach(to => (Object.keys(to.meta).length === 0 ? '/' : true));
 
+  // Remove any trailing slash from the path.
+  router.beforeEach(to => {
+    // Two trailing slashes won't match any route, so return `true` to proceed
+    // to NotFound.
+    if (to.path.endsWith('//')) return true;
+
+    if (to.path.endsWith('/') && to.path !== '/')
+      return { path: to.path.slice(0, -1), query: to.query, hash: to.hash };
```
alxndrsn: Does this return the complete, rewritten `to` object?  Would it be safer to include all of the original props, e.g.

```suggestion
      return { ...to, path: to.path.slice(0, -1) };
```

apps/central/src/router.js
```diff
@@ -65,6 +65,18 @@ router.afterEach(unlessFailure(to => {
   // user is trying to navigate to a parent route, which isn't expected.
   router.beforeEach(to => (Object.keys(to.meta).length === 0 ? '/' : true));
 
+  // Remove any trailing slash from the path.
+  router.beforeEach(to => {
+    // Two trailing slashes won't match any route, so return `true` to proceed
+    // to NotFound.
+    if (to.path.endsWith('//')) return true;
```
alxndrsn: Is there a reason to handle this as a special case, rather than allowing it to fall through, and hit the `.../` stripping code below?

It could also be rewritten as:

```js
let { path } = to;
while (path.endsWith('/') && path !== '/') path = path.slice(0, -1);

return { ...to, path };
```

apps/central/test/router.spec.js
alxndrsn: I only understand this test when it's explicitly linked to the previous one.

If the intent is that the behaviour is identical, maybe the test could be parameterised?  E.g.

```js
[
  '/projects/1/forms/f',
  '/projects/1/forms/f/', // See: https://github.com/getodk/central/issues/1697
].forEach(formRootPath =>
    it('redirects to .../submissions from root path of form', async () => {
      testData.extendedForms.createPast(1);
      return load('/projects/1/forms/f/settings')
        .complete()
        .route(formRootPath)
        .respondForComponent('FormSubmissions')
        .afterResponses(app => {
          app.vm.$route.path.should.equal('/projects/1/forms/f/submissions');
        });
    }));
```

Otherwise, more extensive comments or a different test name might help.

alxndrsn (approved):
This is my preferred approach :+1:

Some thoughts on tests and implementation specifics below, but overall a win as-is.

apps/central/src/router.js
matthew-white: `to` has more stuff on it, e.g., `fullPath` and `redirectedFrom`. Vue Router documents a bunch of different route/location types, but I think it's the difference between these two:

- `to`: https://router.vuejs.org/api/interfaces/RouteLocationNormalizedGeneric.html
- Return value of `beforeEach` function: https://router.vuejs.org/api/interfaces/RouteLocationAsPathGeneric.html

apps/central/src/router.js
matthew-white: That sounds good to me to remove multiple trailing slashes. :+1: Done in c76b084f8cb4005af17e5f48bac411fef50402ac

apps/central/test/router.spec.js
matthew-white: I've parameterized the test in eb83368c6d970be12edf237052eedbcc922a4a68 :+1:

PR #1800, opened 2026-08-22
Routes: do not match path with trailing slash

### LINKED ISSUE #1697: [frontend] broken redirect to `/submissions`

matthew-white:
This PR implements "strict" route matching, in which Vue Router will not match a route if the path has a trailing slash.

This PR closes getodk/central#1697. In that issue, the path had a trailing slash. The user was redirected to a strange URL and saw the 404 page. With this PR, the user will continue to see the 404 page: that is the expected behavior, since no route matches the trailing slash. However, the URL will not change: there will be no double slash in the URL (`//`).

In other words, this PR makes it clear that paths with a trailing slash are not expected to match routes.

I added new tests, including one specifically about getodk/central#1697.

This PR uses Vue Router's built-in `strict` configuration: https://router.vuejs.org/guide/essentials/route-matching-syntax.html#Sensitive-and-strict-route-options. That said, an alternative approach can be seen at #1799, in which the trailing slash is removed before further route matching.

Personally, I think I have a slight preference for this PR, since it involves less code/complexity. However, I'm happy to run with either approach.

~There's a chance that someone's existing link will break. However,~ I think any user impact should be low.

### DIFF
--- a/.changeset/wacky-streets-clean.md
+++ b/.changeset/wacky-streets-clean.md
@@ -1,0 +1,5 @@
+---
+"@getodk/central-frontend": minor
+---
+
+Do not match routes or redirect if the path has a trailing slash (getodk/central#1697)
--- a/apps/central/src/router.js
+++ b/apps/central/src/router.js
@@ -32,7 +32,8 @@ export default (container, {
     history,
     routes,
     scrollBehavior,
-    sensitive: true
+    sensitive: true,
+    strict: true
   });
   const { requestData, toast, redAlert, unsavedChanges, config } = container;
   const { session } = requestData;
--- a/apps/central/test/router.spec.js
+++ b/apps/central/test/router.spec.js
@@ -26,6 +26,20 @@ describe('createCentralRouter()', () => {
           app.findComponent(NotFound).exists().should.be.true;
         });
     });
+
+    it('does not match on a trailing slash', () => {
+      mockLogin();
+      return load('/account/edit')
+        .afterResponses(app => {
+          app.findComponent(NotFound).exists().should.be.false;
+        })
+        .load('/')
+        .complete()
+        .load('/account/edit/')
+        .afterResponses(app => {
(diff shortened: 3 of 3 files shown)


alxndrsn (commented):
The PR title is "Add strict route matching", but the tests just seem to be about trailing slashes.

Would a more appropriate title be "Reject trailing slashes when matching paths"?

Maybe this is a discussion to be had on https://github.com/getodk/central/issues/1697, but I suspect that rejecting paths with trailing slashes will surprise some users and code.  Maybe a more helpful change would be to redirect to the preferred format?

matthew-white:
> The PR title is "Add strict route matching", but the tests just seem to be about trailing slashes.

I've updated the PR title to be more descriptive. I chose the word "strict" because that's the term in the Vue Router configuration. But I agree that it's not the clearest.

> I suspect that rejecting paths with trailing slashes will surprise some users and code.

It may surprise some, but hopefully not too many. When you navigate within Frontend, it will send you to paths without a trailing slash. To get a path with a trailing slash, you'd have to append the slash yourself.

> Maybe a more helpful change would be to redirect to the preferred format?

I'm open to this approach. :+1: #1799 is an alternative to this PR and implements that idea. I have a slight preference for this PR just because the code is a little simpler. (It's just a configuration change.) That said, #1799 isn't particularly complicated either, so I'd be happy to run with that approach if you prefer. I agree that it'd be a nicer experience for the (hopefully rare) case that a slash is appended to a path.

matthew-white:
Closing in favor of #1799.

PR #1802, opened 2026-08-23
Remove unused reference to session resource

matthew-white:
The `SubmissionDownload` component retrieves the `session` resource in its `setup()` function. However, it doesn't actually use it. I've updated the component to remove references to `session`.

I noticed this while working on getodk/central#1315.

Tests continue to pass.

It's just removing an unused variable.

I think our linting will catch this sort of thing automatically for Composition API components. It's harder to spot for Options API components.

No user-facing change; it's just a refactor.

### DIFF
--- a/apps/central/src/components/submission/download.vue
+++ b/apps/central/src/components/submission/download.vue
@@ -141,9 +141,9 @@ export default {
   },
   emits: ['hide'],
   setup() {
-    const { session, keys, fields } = useRequestData();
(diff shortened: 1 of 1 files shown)


matthew-white:
@sadiqkhoja, would you mind taking a look at this PR? It's just a matter of removing references to an unused variable.

sadiqkhoja (approved):

PR #1805, opened 2026-08-23
Restore session for user who ends up on NotFound

### LINKED ISSUE #1698: [frontend] "Page not found" page says user is logged out when they are logged in
Per https://github.com/getodk/central/issues/1697.

# Repro

2. observe "Not logged in" text at the top-right corner of the page:

Image

# Expected

User is logged in, so header should not say "Not logged in".

matthew-white:
This PR closes getodk/central#1698. Currently, we set the `restoreSession` route meta field to `false` for the `NotFound` route. However, as I mentioned at https://github.com/getodk/central/issues/1698#issuecomment-4130465074, I can't think of a reason for why that would be necessary. This PR changes `restoreSession` back to the default value of `true` for the route.

I definitely remember leveraging the fact that `NotFound` doesn't restore the session while working on some early tests. However, I don't think tests rely on that fact anymore.

Beyond tests, I don't see conceptually why it'd be a problem to restore the session when navigating to `NotFound`.

Note that `NotFound` is unusual in another way: it's currently the only route accessible by both anonymous users and logged in users (both `requireLogin` and `requireAnonymity` are `false`). However, I don't think that's a reason not to restore the session.

After this change, the only remaining route for which `restoreSession` is `false` will be `AccountClaim`. I think that's fine, but it's interesting to note. There's related discussion about `AccountClaim` at getodk/central#916.

I updated tests that started failing after I changed `restoreSession` back to the default value of `true`.

It's a pretty small PR. The only thing that needs to change is the route meta field: changing it back to its default value of `true`.

Hopefully users aren't ending up on `NotFound` too often. If there's a regression in this area, hopefully we'll learn about it during regression testing that's starting soon.

### DIFF
--- a/apps/central/test/components/project/show.spec.js
+++ b/apps/central/test/components/project/show.spec.js
@@ -13,6 +13,7 @@ import { mockLogin } from '../../util/session';
 
 describe('ProjectShow', () => {
   it('requires the projectId route param to be integer', async () => {
+    mockLogin();
     const app = await load('/projects/p');
     app.findComponent(NotFound).exists().should.be.true;
   });
--- a/apps/central/test/router.spec.js
+++ b/apps/central/test/router.spec.js
@@ -50,7 +50,6 @@ describe('createCentralRouter()', () => {
     describe('restoreSession is false for the first route', () => {
       const paths = [
         `/account/claim?token=${'a'.repeat(64)}`,
-        '/not-found'
       ];
       for (const path of paths) {
         it(`does not restore session for a user navigating to ${path}`, () =>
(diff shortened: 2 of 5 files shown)


apps/central/test/components/project/show.spec.js
```diff
@@ -13,6 +13,7 @@ import { mockLogin } from '../../util/session';
 
 describe('ProjectShow', () => {
   it('requires the projectId route param to be integer', async () => {
+    mockLogin();
```
matthew-white, line 16: I've added this `mockLogin` because the navigation to /projects/p will end up on `NotFound`, which will now attempt to restore the session. Without this change, that extra request will lead to a `mockHttp()`/`load()` failure: we need either to specify a response to the new request or to `mockLogin` (which prevents the request).

apps/central/test/router.spec.js
```diff
@@ -230,10 +229,12 @@ describe('createCentralRouter()', () => {
   });
 
   describe('requireLogin is false and requireAnonymity is false', () => {
-    it('does not redirect an anonymous user', async () => {
-      const app = await load('/not-found');
-      app.vm.$route.path.should.equal('/not-found');
-    });
+    it('does not redirect an anonymous user', () =>
+      load('/not-found')
+        .restoreSession(false)
```
matthew-white, line 234: A request to restore the session is now being sent. In order to keep the user anonymous as needed for this test, we need to not restore the session. That's why I've added `.restoreSession(false)`

matthew-white (commented):
Adding comments to facilitate async code review

sadiqkhoja (approved):

PR #1806, opened 2026-08-23
Catch session restore promise in router

### LINKED ISSUE #2072: Unhandled 401 error on session restore
## Description

[Sentry](https://getodk.sentry.io/issues/7604341520/events/2d16cee8d9dd4a83a0de887b0b1fa957/) is currently logging expected 401 Unauthorized responses as unhandled AxiosError exceptions. This happens when the frontend attempts to restore a user session via GET `/v1/sessions/restore` and the session has expired or is invalid. 

While this is expected behavior, it creates unnecessary noise in the error tracker. We can wrap the call in `try/catch` or handle it in an Axios response interceptor, so it doesn't crash/propagate unhandled.

matthew-white:
Closes getodk/central#2072.

Nothing in particular. Tests continue to pass. We can monitor Sentry for whether reports continue to come in.

I think the previous approach of using `Promise.allSettled()` was valid. However, now I'm explicitly `.catch()`-ing `sessionPromise`. Now that it's explicitly caught, we can use `Promise.all()` instead of `Promise.allSettled()`.

Instead of adding the `.catch(noop)` in the router, I could have added it to the `restoreSession()` function in src/util/session.js. However, I think it makes the code a little clearer to catch the promise where it's used (in the router).

I don't think the issue has to do with the specific type of error. The `restoreSession()` function already has logic to handle 401 errors specifically. We just need to `.catch()` the promise.

### DIFF
--- a/apps/central/src/router.js
+++ b/apps/central/src/router.js
@@ -118,7 +118,7 @@ router.afterEach(unlessFailure(to => {
     // navigation.
     const needsLogin = to.meta.restoreSession && !session.dataExists;
     const sessionPromise = needsLogin
-      ? restoreSession(session)
+      ? restoreSession(session).catch(noop)
       : Promise.resolve();
 
     // A test can skip this request by setting `config` before the initial
@@ -139,12 +139,12 @@ router.afterEach(unlessFailure(to => {
 
     const locale = userLocale();
     const localePromise = locale != null
-      ? loadLocale(container, locale)
+      ? loadLocale(container, locale).catch(noop)
       : Promise.resolve();
 
     // Once the session and the config have been received, we can complete
     // login.
-    await Promise.allSettled([sessionPromise, configPromise]);
+    await Promise.all([sessionPromise, configPromise]);
     if (needsLogin && session.dataExists) {
       // If this is the first time that the session has been restored since the
       // most recent OIDC login, set sessionExpires in local storage. If
@@ -155,7 +155,7 @@ router.afterEach(unlessFailure(to => {
       await logIn(container, newSession).catch(noop);
     }
 
-    await localePromise.catch(noop);
+    await localePromise;
     return config.loadError != null ? '/load-error' : true;
   });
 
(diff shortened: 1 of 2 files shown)


apps/central/src/router.js
```diff
@@ -139,12 +139,12 @@ router.afterEach(unlessFailure(to => {
...
     const locale = userLocale();
     const localePromise = locale != null
-      ? loadLocale(container, locale)
+      ? loadLocale(container, locale).catch(noop)
       : Promise.resolve();
 
     // Once the session and the config have been received, we can complete
     // login.
-    await Promise.allSettled([sessionPromise, configPromise]);
+    await Promise.all([sessionPromise, configPromise]);
```
matthew-white, line 147: - `configPromise` already had a `.catch()`
- Now `sessionPromise` has a `.catch()`
- So we can use `Promise.all()` instead of `allSettled()`

apps/central/src/router.js
```diff
@@ -139,12 +139,12 @@ router.afterEach(unlessFailure(to => {
 
     const locale = userLocale();
     const localePromise = locale != null
-      ? loadLocale(container, locale)
+      ? loadLocale(container, locale).catch(noop)
```
matthew-white, line 142: This change isn't strictly necessary. We were already `.catch()`-ing this promise below (which you can see in the diff). However, I like the consistency of this change:

- This code involves three promises: `sessionPromise`, `configPromise`, `localePromise`
- `configPromise`: `.catch()` was already part of the variable's definition.
- `sessionPromise`: `.catch()` is now part of the variable's definition.
- `localePromise`: I think it's nice to make `.catch()` part of the variable's definition. The pattern is: every promise is responsible for its own `.catch()`-ing. We `.catch()` immediately rather than later on in the function.

matthew-white (commented):
Adding comments to facilitate async code review

latin-panda (approved):
Looks good and functions well, thank you!
I like your comments in the PR; they make the review process nice!

PR #1807, opened 2026-08-24
Reload page after vite:preloadError

### LINKED ISSUE #2073: CSS preload failures after deploy bubble up as Vue errors
## Description

After a new version of Central is deployed, users still running the previous version trigger [Sentry errors](https://getodk.sentry.io/issues/7601136677/events/89c15366b43d4b2ebf83ac6b1fe3f96f/) like:

```
  Unable to preload CSS for /assets/outdated-version-BscxB6kc.css
  Unable to preload CSS for /assets/dl-data-FXEGUtHv.css
```

The app already polls `/version.txt` every X seconds and shows a "Refresh Page" toast when a version change is detected. These errors occur in the window before that check runs.

An option could be to listen for Vite's `vite:preloadError` event in main.js and reload the page. This is the pattern [Vite documents](https://vite.dev/guide/build#load-error-handling) for this case.

```js
  window.addEventListener('vite:preloadError', () => { window.location.reload(); });
```

This could also replace the "Refresh Page" toast for the deploy-transition case, since the reload happens automatically.

matthew-white:
Closes getodk/central#2073.

Mostly I just wrote a new test.

I didn't try this, but one thing I could do is try to trigger a real `vite:preloadError`, e.g., by building files locally, changing some code, then taking an action that attempts to load the changed asset. That wouldn't work with HMR, but we could use `npm run dev:build` to try to see that.

Another thing we'll do is continue to watch whether this error appears in Sentry.

The main thing I want to make note of is that I put this code in a new composable, `useVersionMonitor()`. The /version.txt check lives in the `App` component, but I feel like there's already too much going on in that component; I didn't want to add to it. Instead, I plan to move the /version.txt check from the `App` component into the new `useVersionMonitor()` composable. The /version.txt check and the `vite:preloadError` event listener are two different ways to monitor for a change to the version of Central deployed.

I think the risk of regression is low, as it's entirely new code (just additions). In the issue, we discussed the potential user impact.


apps/central/src/composables/version-monitor.js
matthew-white, line 1: As mentioned in the PR description, I plan to move the /version.txt check into this file soon. I like the idea of centralizing that logic here and simplifying the `App` component.

apps/central/src/composables/version-monitor.js
```diff
@@ -0,0 +1,13 @@
+import { inject } from 'vue';
+
+import useEventListener from './event-listener';
+
+// useVersionMonitor() watches for indications that a new version of Central has
+// been deployed. When it detects a likely version change, it either reloads the
+// page automatically or prompts the user to do so.
+export default () => {
+  const location = inject('location');
```
matthew-white, line 9: In production, `location` is the same as `window.location`. In testing, it can be mocked. It is a property of the `container` object and is `provide`d via `container`: see src/container.js.

matthew-white (commented):
Adding comments to facilitate async code review.

matthew-white:
Unfortunately, it looks like e2e tests are consistently failing with this change. You can see that in CI above and also in getodk/web-forms#884, which I filed before closing.

I can think of two options for how to proceed:

1. Go back to the idea of showing a toast instead of automatically reloading the page.
2. Somehow detect whether it's a CI/test/e2e environment and not reload if so.
    - Maybe we could check for `localhost` in the URL?
    - Or maybe we add a new property to src/config.js or client-config.json to indicate whether we're in CI? Not exactly sure what that plumbing would look like, but basically adding some sort of configuration to drive whether the reload actually happens.

Why is `vite:preloadError` being triggered in e2e tests? I feel like I've seen related discussion about that, but I don't think I understand the cause of it.

matthew-white:
Another idea: we could wait a few seconds before adding the event listener. Definitely feels like a workaround, but it'd fix the e2e tests without having to add configuration.

matthew-white:
> we could wait a few seconds before adding the event listener.

I've implemented logic around those lines, and e2e tests now pass. I don't wait to add the event listener itself, but the event listener will skip the automatic reload within the first seconds of app startup.

@latin-panda, I think this PR is fully ready for review now!

latin-panda:
Thanks @matthew-white. It's the first item on my list to review tomorrow!

apps/central/src/composables/version-monitor.js
```diff
@@ -0,0 +1,17 @@
...
+
+  // getodk/central#2073
+  const start = Date.now();
+  useEventListener(window, 'vite:preloadError', () => {
+    // Don't reload right after app startup, as that can break e2e tests.
+    if (Date.now() - start >= 30000) location.reload();
+  });
```
latin-panda, line 16: Can we add `event.preventDefault()` here? Without it, the error still throws even though the page reloads. 





```suggestion
  useEventListener(window, 'vite:preloadError', (event) => {
    // Don't reload right after app startup, as that can break e2e tests.
    if (Date.now() - start >= 30000) {
      event.preventDefault(); 
      location.reload();
    }
  });
```

latin-panda (approved):
Thank you! Just one suggestion below. Approving to unblock

apps/central/src/composables/version-monitor.js
matthew-white: Done in e497d5ef12bffa780d1f09920bc8f4dc24fc57c3

matthew-white:
> I plan to move the /version.txt check from the `App` component into the new `useVersionMonitor()` composable.

I've filed a follow-up issue about this: getodk/central#2022.

PR #1808, opened 2026-08-24
fix(#2032): request project using verbs query parameter

### LINKED ISSUE #2032: Loading a webform does a request for extended project metadata which is slow
We use a user's verbs on the form to see what they're allowed to do. We can't get that information just from fetching the XML because a user who can preview a form may not be allowed to fill it.

Currently getting the verbs requires doing an extended project metadata request which is slow. Could we introduce an endpoint to get just the verbs?

Related: https://github.com/getodk/central/issues/1318

CC @garethbowen

garethbowen:
Closes: getodk/central#2032

NB the e2e tests are failing because this depends on: https://github.com/getodk/central-backend/pull/1917 

Manual testing - verified that verbs is included but other extended properties are not.

...

Should lead to much better form load performance.

No.

### DIFF
--- a/.changeset/perky-tools-pump.md
+++ b/.changeset/perky-tools-pump.md
@@ -2,4 +2,4 @@
 "@getodk/forms": minor
 ---
 
-Improved performance by fetching the project verbs from a more efficient server endpoint.
+Improved form loading times by fetching the project verbs from a more efficient server endpoint.
--- a/.changeset/perky-tools-pump.md
+++ b/.changeset/perky-tools-pump.md
@@ -1,0 +1,5 @@
+---
+"@getodk/forms": minor
+---
+
+Improved performance by fetching the project verbs from a more efficient server endpoint.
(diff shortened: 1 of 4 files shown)


garethbowen:
@alxndrsn Is backwards compatibility something we need to worry about? ie: running a new version of frontend with an old version of backend? I assume not, and it makes the code a lot simpler, so I decided to leave it out unless required.

matthew-white:
FWIW it's not something we worry about it in Central Frontend. Instead, we prompt the user to reload the page if we detect that a new version of Central has been deployed.

.changeset/perky-tools-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+"@getodk/forms": minor
+---
+
+Improved performance by fetching the project verbs from a more efficient server endpoint.
```
alxndrsn: "Improved performance" seems vague; would more details be helpful?  E.g. "Improved page load times on ___ pages"

alxndrsn (commented):
LGTM, but e2e tests are failing.

garethbowen:
> LGTM, but e2e tests are failing.

Yes, because they need the backend change. Once that's merged I'll rerun before merging this.

.changeset/perky-tools-pump.md
garethbowen: Done.

alxndrsn (approved):

PR #1811, opened 2026-08-24
feat(web-forms): navigate to questions

### LINKED ISSUE #104: After adding a repeat instance, give focus to first item in new repeat instance
#### Web Forms version and package (e.g. are you using `ui-vue` or only the engine?)
ui-vue

#### Browser and version
Firefox

#### Device used
macOS

#### Problem description
Keyboard navigation works really well! But when I add a repeat using the keyboard, focus stays on the `Add` button. Ideally, it would go to the first item in the new repeat instance.

Try a form with repeats like https://dev.getodk.cloud/#/projects/2/forms/groups_and_repeats/
Navigate with the keyboard
Add a new instance
Notice that focus is still on Add

#### Expected behavior
Focus is on the first question in the new repeat instance

#### Other information 

Things you tried, stack traces, related issues, suggestions on how to fix it...

### LINKED ISSUE #189: Navigate to questions with errors
User story:
As an enumerator, I should have a way to navigate to questions with errors so that I could easily fix them. 

Background:
If there are any errors on the Form when user presses the "send" button they see a banner at the top of the Form that says "N  questions with error". We need to add "issue error" button that would take user to the first invalid question.

Some ideas about implementations: https://github.com/getodk/web-forms/pull/164#discussion_r1714415457

Figma: https://www.figma.com/design/6v0NnvayfDlaQhYpjv37Bn/ODK-Web-Forms?node-id=2217-23619&t=GstUtShabbnWPrPI-4

### LINKED ISSUE #867: "Fill in form again" scrolls to bottom instead of top (regression)
### Expected Behaviour

After submitting a Web Form and choosing "Fill in form again," the reloaded form should be scrolled to the top with focus on the first question. This was previously fixed in Web Forms (see getodk/central-frontend#1554), but it's no longer working.

### Observed Behaviour

On Send, the form scrolls to the top and the "fill out again" prompt appears. But clicking the prompt lands the user back at the bottom of the form (at the Send button) instead of the top. On a long form this forces a manual scroll back up to the first question. 

### Reproduction Steps

Steps:
1. Open a long Web Form (enough content that the top and Send button are not visible at the same time).
2. Fill it in and click Send.
3. When the "Fill in form again" prompt appears, click it.
4. Observe that the reloaded form is scrolled to the bottom (Send button) rather than the top / first question.

### Environment

- Browser version: chrome
- OS: Macos
- Web Forms version: bundled with ODK Central v2026.2.0 

### Additional Context

Reported on the ODK forum. Previously fixed in Web Forms via getodk/central-frontend#1554, so this looks like a regression.

latin-panda:
Closes https://github.com/getodk/web-forms/issues/867
Closes https://github.com/getodk/web-forms/issues/189
Closes https://github.com/getodk/web-forms/issues/104

Manual test - [see test plan covered](https://github.com/getodk/central-frontend/pull/1811#issuecomment-5398687641).
New unit tests.

<details><summary>VIDEO: pageless - First load, fill again, repeats</summary>

https://github.com/user-attachments/assets/3d2bbfd5-dd00-4297-803d-805433f6e635

</details>

<details><summary>VIDEO: pageless - Error Nav</summary>

https://github.com/user-attachments/assets/518a8215-4fa7-4ed3-94b6-b09e3ef946cb

</details>

<details><summary>VIDEO: paged - Error Nav</summary>

https://github.com/user-attachments/assets/950513f3-0c17-4c71-ab44-9237ecf1daeb

</details>

<details><summary>VIDEO: paged - First load, fill again, repeats</summary>

https://github.com/user-attachments/assets/82c0696e-c13e-4aec-afdd-841b56204fed

</details>

<details><summary>VIDEO: paged - repeats with long scroll</summary>

https://github.com/user-attachments/assets/25353d40-9622-463c-8392-79b3641df9f3

</details>

The engine identifies the question requiring attention, while the client's library manages scrolling and focus. So, the library handles all scenarios, including load, page changes, repeats, and errors. 

The form now directs user focus to the most relevant element, like:
- the first question on open
- after clicking "Fill out again"
- the first field of a new repeat instance
-  the first error message on form submit or clicking the error's "View" button.

If the question is already in view, don't scroll to avoid weird jumps.

Date and time fields retain their behavior because focusing opens the calendar popup at an incorrect position (inlined at the end of the page); this requires more research and testing, so it is left for another delivery. The focus is set on input text and select dropdowns, where regular DOM focus works well in PrimeVue.

No


latin-panda:
## Manual test plan of navigate to questions

### Opening the form

- [x] When a form opens, the page is at the top and the cursor is already in the first question (inputs and select dropdowns)
- [x] On a form with pages, the first page shows and its first question has the focus (inputs and select dropdowns)

### Submitting and filling again

Fixes: https://github.com/getodk/web-forms/issues/867

- [x] After submitting, choosing "Fill out again" shows a fresh empty form back at the top (no jump to the bottom)
   - Known low-impact gap - after the dialog closes, the cursor is not in the first input or select dropdown due to timing with the PrimeVue dialog, but the form position is correct. Fixing the cursor was complicating the code. 

### Errors when submitting

Fixes: https://github.com/getodk/web-forms/issues/189

- [x] Submitting with missing required answers shows the error message and brings the first problem question to the middle of the screen.
- [x] The "View" button in the error banner always moves you to the first unresolved problem.
- [x] On a form with pages, the error jump also switches to the right page.

### Moving between pages

- [x] "Next" and "Back" put the focus in the question of the page you land on (inputs and select dropdowns)
- [x] When a page has scroll, landing on a new page shows the top and first question (never the middle or bottom)

### Repeats

Fixes: https://github.com/getodk/web-forms/issues/104

- [x] Adding an instance and if it is already visible, nothing scrolls. The focus just moves into its first field. 
- [x] Adding an instance, that would appear off screen, the page scrolls so the new instance is at the top, the focus in its first field.
- [x] Removing an instance on a one-long-page form: nothing moves or scrolls (the list just gets shorter). 
- [x] Removing an instance on a form with one instance per page, you land on the entry that took its place; removing the last one lands you on the previous one; removing the only one shows the Add button.
- [x] A repeat with a fixed count (set by another answer) grows and shows new empty instances when the count goes up; it never deletes entries when the count goes down (that would lose data).

packages/web-forms/src/assets/styles/reset.scss
```diff
@@ -104,6 +104,6 @@ summary {
 }
 
 :focus-visible {
```
latin-panda, line 106: We rarely want to add an outline, except for select options; those are already explicit in the code.

packages/web-forms/src/components/common/SearchableDropdown.vue
```diff
@@ -53,7 +53,7 @@ const selectValue = (value: string) => {
 
 <template>
 	<Select
-		:id="question.nodeId"
+		:input-id="question.nodeId"
```
latin-panda, line 56: it matches multiselect dropdown and allows set focus

packages/web-forms/src/components/form-layout/FormQuestion.vue
```diff
@@ -48,7 +49,8 @@ provide(QUESTION_HAS_ERROR, questionHasError);
 
 <template>
 	<div
-		:id="question.nodeId + '_container'"
+		:id="containerId(question.nodeId)"
+		tabindex="-1"
```
latin-panda, line 53: Helps better scroll to the question container so the label and anything inside are in the viewport.

latin-panda:
@garethbowen This is ready for review. I still need to test in Safari. I will do it tomorrow.

packages/web-forms/src/lib/useNavigationTarget.ts
```diff
@@ -0,0 +1,61 @@
...
+const navigateTo = (nodeId: string, align: ScrollAlignment = 'start') => {
+  // Controls without a container are their own scroll target
+  const container = document.getElementById(containerId(nodeId)) ?? document.getElementById(nodeId);
+
+  if (!container) {
+    return;
+  }
+
+  scrollToContainer(container, align);
+  const focusTarget = document.getElementById(nodeId) ?? container;
```
garethbowen: Don't get the node twice...

```suggestion
  const node = document.getElementById(nodeId);
  const container = document.getElementById(containerId(nodeId));
  const scrollTarget = container ?? node;

  if (!scrollTarget) {
    return;
  }

  scrollToContainer(scrollTarget, align);
  const focusTarget = node ?? container;
```

packages/web-forms/src/lib/useNavigationTarget.ts
```diff
@@ -0,0 +1,61 @@
...
+const scrollToContainer = (container: HTMLElement, align: ScrollAlignment) => {
+  const rect = container.getBoundingClientRect();
+  const visibleHeight = Math.min(rect.bottom, window.innerHeight) - Math.max(rect.top, 0);
+  const isMostlyVisible = visibleHeight >= rect.height * 0.5;
+
+  if (align === 'start' && isMostlyVisible) {
+    return;
+  }
+
+  container.scrollIntoView({ behavior: 'smooth', block: align });
```
garethbowen: I think it would be better to use `block: "nearest"`, rather than 'start'/'center'. This has the advantage of not doing anything if the element is already in view, so we don't need to do the `isMostlyVisible` check. And secondly it means that if the element is only a little out of view it will only scroll a little bit.

packages/web-forms/src/components/common/SearchableDropdown.vue
garethbowen, line 56: I think there are more changes like this required. The form I tested used an InputInt which calls InputNumeric which uses PVs InputNumber, except it's setting `id` not `input-id` so the focus doesn't work. I haven't tested all types but we should go through and make sure they all focus as expected.

There's a similar but more complicated problem for radio buttons - ideally we'd focus on the first radio so keyboard navigation would work but we'd have to either calculate the ID or have special handling in the navigateTo that finds the first input of the container?

packages/web-forms/src/lib/useNavigationTarget.ts
```diff
@@ -0,0 +1,61 @@
...
+  const focusTarget = document.getElementById(nodeId) ?? container;
+  focusTarget.focus({ preventScroll: true });
+};
+
+const navigateToFirstViolation = (root: RootNode | null) => {
+  if (!root) {
+    return;
+  }
+
+  root.navigateToFirstViolation();
```
garethbowen, line 43: In root, this updates the navigation target, which triggers the watch on line 48, and sometimes ends up calling navigateTo twice with the same node id. It's not expensive but it's not ideal.

I assume this is happening because if someone clicks `view` or submit without actually fixing the violation, then the nodeId hasn't changed and the signal doesn't fire so you also need to explicitly call to make sure the field scrolls back into view?

It might be better to have the root's nodeNavigation signal always fire regardless of whether the nodeId has changed or not. This can be done by passing `{ equals: false }` to the `createSignal` call in `createNodeNavigation`. If you do this then you no longer need the `navigateToFirstViolation` function, you just set the navigation target on the root and the reactivity will do it for you.

It feels cleaner, but also we'd need to be careful about when we set the navigation target...

garethbowen (changes requested):
I'm really enjoying the keyboard navigation! Thanks.

A few comments inline...

Also, not strictly part of this change, but now that tab navigation is so good, I think we should add `tabindex="-1"` to the back button, because you usually want to fill in the fields, and tab on to the next button.

packages/web-forms/src/lib/useNavigationTarget.ts
latin-panda: Nice, thanks! I used the `nearest`.

packages/web-forms/src/lib/useNavigationTarget.ts
latin-panda: Done!

packages/web-forms/src/lib/useNavigationTarget.ts
latin-panda: Good catch! I tried `{ equals: false }`, but it didn't work. Vue skips updates when the value hasn't changed, so the watcher never fired for a repeated ID.
Instead, I check whether the target changed after calling `navigateToFirstViolation`. If it changed, the watcher scrolls. If not, I call `navigateTo` directly. So it's one call now.

packages/web-forms/src/components/common/SearchableDropdown.vue
latin-panda: I fixed the number input with `input-id`. 
For the rest, when the question ID doesn't point to a focusable element, it focuses the first focusable element inside the question. That covers radios, checkboxes, upload, geopoint, slider, rating, and rank. 

The Date/time pickers are a bit more complex to control, so I left them out for now. For example, the Time cannot be controlled properly with the keyboard yet.

packages/web-forms/src/components/form-layout/FormFooter.vue
```diff
@@ -41,6 +32,16 @@ const t: Translate = inject(TRANSLATE)!;
 			<span>{{ t('odk_web_forms.submit.label') }}</span>
 			<IconSVG name="mdiSendVariantOutline" variant="inverted" />
 		</Button>
+		<Button
```
latin-panda, line 35: > Also, not strictly part of this change, but now that tab navigation is so good, I think we should add tabindex="-1" to the back button, because you usually want to fill in the fields, and tab on to the next button.

Done, but differently. The `tabindex="-1"` makes the element unreachable with the tab key. So I inverted the order of buttons in the DOM and fixed the display with CSS.

The display is the same, and tab focus lands first in Next, then in Back.

latin-panda:
@garethbowen, this is ready for another review.

packages/web-forms/src/lib/useNavigationTarget.ts
```diff
@@ -0,0 +1,67 @@
+import type { RootNode } from '@getodk/xforms-engine';
+import { nextTick, watch } from 'vue';
+import { containerId } from '@getodk/web-forms/lib/format/ids.ts';
+
+const findFocusTarget = (
+  node: HTMLElement | null,
+  container: HTMLElement | null
+): HTMLElement | null => {
+  const FOCUSABLE =
+    'button, input:not(.p-datepicker-input), textarea, [tabindex]:not([tabindex="-1"])';
```
garethbowen: This is surprisingly challenging to do... It might be worthwhile checking for disabled as well, but I'll leave that up to you.

The only problem if we get it wrong is the focus will fail silently so not a big risk.

garethbowen (changes requested):
One minor comment inline.

The only thing I found in my testing is when I submit and choose to fill out again, the first field doesn't get focus. Looking at the code I'd expect it to just work, but maybe it's missing something in this specific case, or there's some race condition when resetting the form?

packages/web-forms/src/components/OdkWebForm.vue
```diff
@@ -254,8 +253,16 @@ const init = async () => {
...
+const releaseFocus = () => {
+	const active = document.activeElement;
+	if (active instanceof HTMLElement) {
+		active.blur();
+	}
+};
+
 const handleSubmit = (currentState: FormStateSuccessResult) => {
 	const { root } = currentState;
+	releaseFocus(); // so follow-up dialogs don't restore focus here and scroll back to it
```
latin-panda, line 265: > The only thing I found in my testing is when I submit and choose to fill out again, the first field doesn't get focus. Looking at the code I'd expect it to just work, but maybe it's missing something in this specific case, or there's some race condition when resetting the form?

@garethbowen Yes, I documented that in the [test plan](https://github.com/getodk/central-frontend/pull/1811#issuecomment-5398687641).



It's an issue in the apps/forms dialog; it has a scrolling issue, but I fixed it in this line. 

I can file a ticket to investigate the focus problem further.

garethbowen: Ahh right. Thanks for writing it up!

garethbowen (approved):

packages/web-forms/src/components/OdkWebForm.vue
latin-panda: It'd be nice if the PrimeVue dialog in apps/forms didn't scroll to the last focused element (the submit button). 
I also considered moving the submission handler that clears the form after the dialog closes (today it's linked to the submit button), but that would be a more significant change

garethbowen: I've been tempted to move the submission handler too... it feels strange that the form clears on submission and not when you click "fill in again".

packages/web-forms/src/lib/useNavigationTarget.ts
latin-panda: Fixed and tested! Thanks

PR #1817, opened 2026-08-24
EntityUpload: show both header warnings and data error

matthew-white:
This PR makes progress on getodk/central#1904. For that issue, we want users to be able to see both errors and warnings at the same time. The idea is to show the user as much information as possible. Whatever issues with the CSV Central knows about, it can show to the user.

There are four types of errors/warnings:

1. Errors with the column header
2. Warnings about the column header
3. Errors in the data below the column header
4. Warnings about the data below the column header

Not all combinations of these four are possible. The following table summarizes the combinations:

&nbsp; | Header errors | Header warnings | Data error | Data warnings
-- | -- | -- | -- | --
Header errors | ✅ | Future PR | ❌ | ❌
Header warnings | Future PR | ✅ | This PR | ✅
Data error | ❌ | This PR | ✅ | ❌
Data warnings | ❌ | ✅ | ❌ | ✅

✅ = possible today
❌ = not possible and also not currently planned to become possible

Some of these combinations we're not currently planning to make happen: they would require larger changes to `EntityUpload`. However, some of these combinations are straightforward to implement.

This PR implements one of those straightforward combinations: showing header warnings alongside a data error.

I updated tests.

I considered using `.finally()`, i.e., setting `warnings` and/or `fileMetadata` in `.finally()`. However, I was worried that there would be a tick in which not all expected refs were set. I thought Vue might would try to re-render during that tick and that there could end up being a required component prop that was missing.

### DIFF
--- a/apps/central/src/components/entity/upload.vue
+++ b/apps/central/src/components/entity/upload.vue
@@ -300,6 +300,7 @@ const parseEntities = async (file, headerResults, signal) => {
 };
 const selectFile = (file) => {
   redAlert.hide();
+  fileMetadata.value = null;
   headerErrors.value = null;
   dataError.value = null;
 
@@ -320,15 +321,17 @@ const selectFile = (file) => {
         return Promise.resolve();
       }
 
+      const metadata = { name: file.name, size: file.size };
       return parseEntities(file, headerResults, signal)
         .then(results => {
           csvEntities.value = results.data;
-          fileMetadata.value = { name: file.name, size: file.size };
+          fileMetadata.value = metadata;
           if (validation.warnings != null || results.warnings.count !== 0)
             warnings.value = { ...validation.warnings, ...results.warnings.details };
         })
         .catch(error => {
           if (!signal.aborted) {
+            fileMetadata.value = metadata;
             dataError.value = error.message;
             warnings.value = validation.warnings;
           }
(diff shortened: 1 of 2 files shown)


apps/central/test/components/entity/upload.spec.js
```diff
@@ -235,12 +235,36 @@ describe('EntityUpload', () => {
       warnings.largeCell.should.equal(3);
     });
 
-    it('does not show warnings if there is a data error', async () => {
+    it('only shows header warnings if there is a data error', async () => {
```
matthew-white:
```suggestion
    it('does not show data warnings if there is a data error', async () => {
```

ktuite (approved):

matthew-white (commented):

apps/central/src/components/entity/upload.vue
matthew-white, line 1: Let's probably set `fileMetadata` as well. As of #1828, that's needed for `EntityUploadWarnings`.

matthew-white:
I'm going to rebase now that #1815 has been merged.

apps/central/test/components/entity/upload.spec.js
matthew-white: Done in d867e8e01fe38896f822aaf7c21807bd8e9364f4

apps/central/src/components/entity/upload.vue
matthew-white: Done in efc174a9a690669afa2191650b95198cd8b2ce0f

PR #1818, opened 2026-08-25
fix(web-forms#885): removes rounding from geo functions

### LINKED ISSUE #885: The xpath geo functions should not round the result
### Expected Behaviour

When calling [distance](https://getodk.github.io/xforms-spec/#fn:distance) or [area](https://getodk.github.io/xforms-spec/#fn:area) the result should have as much precision as possible.

### Observed Behaviour

Both functions round the result to 2 decimal places. This is inconsistent with JavaRosa, and not dictated by the spec. It's better to leave the rounding to the client to implement in the UI or the form designer to explicitly request with the [round](https://getodk.github.io/xforms-spec/#fn:round) function.

garethbowen:
Closes getodk/web-forms#885

Testing

The challenging part was getting the tests to play nicely. Because each js engine has it's own rules about floating point numbers we have to round the results to get them to pass in each browser. This obviously has implications for form designers who may be relying on a high level of precision, however I still believe it's correct to leave it to the designers and the UX to round as appropriate.

Forms that used to get a rounded area or distance are now getting more precise results. This is consistent with Collect so I don't expect anyone is relying on it.

No.

### DIFF
--- a/.changeset/small-sites-crash.md
+++ b/.changeset/small-sites-crash.md
@@ -1,0 +1,7 @@
+---
+"@getodk/xpath": patch
+"@getodk/web-forms": patch
+"@getodk/xforms-engine": patch
+---
+
+Removed rounding from area() and distance() xpath functions.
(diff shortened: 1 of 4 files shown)


latin-panda (approved):
Looks good! Thanks!

PR #1819, opened 2026-08-25
EntityUpload: ignore columns that start with __

### LINKED ISSUE #1785: When importing Entities from CSV, ignore columns that start with __
It can be helpful to download an Entity List from one project and upload it to another. Currently, doing so without modification gives an error like:

Image

In this case, the best action to take is to delete the __ columns. Central knows that those are system-reserved so it should ignore them.

I think it would be ok to do so silently. Alternatively, we could show a message like 

> Skipped 4 system properties with __ prefix: __id, __created_at...

matthew-white:
Closes getodk/central#1785.

Mostly new tests, some local testing.

Mostly it follows the same pattern as the `missingProperties` warning.

The one main difference is that we need to ignore these system properties when constructing the entity JSON. I updated `rowToEntity()` to do so. I'll almost certainly need to return to that function when working on getodk/central#1786.

We should test entity upload fully once all the related issues are complete.

### DIFF
--- a/apps/central/src/components/entity/upload.vue
+++ b/apps/central/src/components/entity/upload.vue
@@ -253,10 +253,13 @@ const rowToEntity = (values, columns) => {
   let hasProperty = false;
   for (const [i, value] of values.entries()) {
     if (value === '') continue; // eslint-disable-line no-continue
+
     const column = columns[i];
-    if (column === 'label' || dataset.propertyMap.has(column)) {
+    if (column === 'label') {
+      obj.label = value;
+    } else if (dataset.propertyMap.has(column)) {
       obj[column] = value;
-      if (column !== 'label') hasProperty = true;
+      hasProperty = true;
     }
   }
   const { label } = obj;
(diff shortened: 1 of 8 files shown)


apps/central/src/components/entity/upload.vue
```diff
@@ -208,17 +208,27 @@ const validateHeader = ({ columns, errors, meta }, file) => {
...
+      }
+    }
+
+    errorDetails.unknownProperty = columnSet.size !==
+      dataset.properties.length -
+      warningDetails.missingProperties.length +
+      warningDetails.systemProperties.length +
       (hasLabel ? 1 : 0);
+
+    for (const [name, value] of Object.entries(warningDetails)) {
```
matthew-white, line 230: Add a comment about this line, something like:

> We won't show warnings about empty arrays; remove them.

Make it clear that we need to be able to count the number of keys to determine the number of warnings.

ktuite (approved):

matthew-white (commented):

matthew-white:
I'm going to rebase now that #1814 is merged.

apps/central/src/components/entity/upload.vue
matthew-white: Done in 233ad12f2db9976e4f9faf3cd7595669e32bad35

PR #1820, opened 2026-08-26
fix(web-forms#879): makes date error handling consistent

### LINKED ISSUE #879: Make xpath date parsing consistent for error cases
### Expected Behaviour

When calling an xpath function that takes a date, there are multiple ways the input can be invalid, but the value returned should be consistent regardless of exactly how the string is invalid.

### Observed Behaviour

Calling format-date sometimes throws an error and sometimes returns null. I think the correct approach is to throw in all cases, but we need to make sure this isn't going to do anything unexpected. This also impacts all other functions that take a date string.

### Reproduction Steps

Form (XML or XLSX file):

```xml
<?xml version="1.0"?>
    <h:head>
        <h:title>date formatting</h:title>
            <instance>
                <data id="date_formatting" version="1">
                    <date/>
                    <meta>
                        <instanceID/>
                    </meta>
                </data>
            </instance>
            <bind nodeset="/data/date" type="string"/>
            <setvalue ref="/data/date" event="odk-instance-first-load" value="format-date('2026-12-12', '%Y-%b-%e')"/>
        </model>
    </h:head>
    <h:body>
        <input ref="/data/date">
            <label>Date</label>
        </input>
    </h:body>
</h:html>
```

Steps:
1. The form above works fine - notice the field is set to "2026-Dec-12"
2. Change the first parameter to the format-date function to `'2026-13-12'` which has an invalid month - notice an error is thrown on load
3. Change the first parameter to the format-date function to `'26-12-12'` which isn't a recognised format - notice the value is set to blank

### Additional Context

Found during patching of #877

garethbowen:
Closes getodk/web-forms#879

Testing.

The other option was to throw errors every time, but so much calling code expects null for the other cases that this made more sense.

Hopefully none.

No.


packages/xpath/test/xforms/format-date-time.test.ts
garethbowen, line 1: Removed because it had like 1 test that is being sufficiently covered by the format-date test

packages/xpath/src/lib/datetime/coercion.ts
```diff
@@ -6,22 +6,26 @@ import {
...
   }
 
   return null;
 };
 
+const hasTimeZone = (value: string) => {
+  return value.endsWith('Z') || TIMEZONE_OFFSET_PATTERN.test(value) || !/^\d{4}/.test(value);
+};
+
 export const dateTimeFromString = (
```
latin-panda, line 9: I compared it with Collect. The ones with a green check I consider good output, more correct.

1: format-date('2026-12-12', '%Y-%m-%d') = 2026-12-12
WF: :white_check_mark:
Collect: :white_check_mark:

2: format-date('2012-08-20T00:00:00.00+00:00', '%Y-%m-%d') = 2012-08-20
WF: :white_check_mark:
Collect: :white_check_mark:

3: format-date('26-12-12', '%Y-%m-%d') = [blank] 
WF: [blank] :white_check_mark:
Collect: 0026-12-12 :x:

8: (Apr 31 + offset): format-date('2026-04-31T10:00:00+05:00', '%Y-%m-%d') 
WF: 2026-05-01 :x:
Collect: Error dialog `The value XXX cant be converted to date` :white_check_mark:

10: (Midnight next day) format-date('2026-12-01T24:00:00Z', '%Y-%m-%d') 
WF: 2026-12-02 :x:
Collect: Error dialog `The value XXX cant be converted to date` :white_check_mark:

13: format-date('2026-12-01T10:30Z', '%Y-%m-%d') = 2026-12-01
WF: 2026-12-01 :white_check_mark: 
Collect: Error dialog `The value XXX cant be converted to date`  :white_check_mark:

15: format-date-time('2026-02-29T10:00:00Z', '%Y-%m-%d %H:%M') = 
WF: 2026-03-01 17:00 :x:
Collect: Error dialog `The value XXX cant be converted to date` :white_check_mark:

latin-panda: These other cases involve invalid values, and we need to decide whether to print blank or fail with an error message: 
- WF: [blank]
- Collect: Error dialog `The value XXX cant be converted to date`

Cases: 
4: format-date('2026-13-01', '%Y-%m-%d') 
5: format-date('invalid', '%Y-%m-%d') 
6: format-date(number('invalid'), '%Y-%m-%d') 
7: (Feb 29 in non-leap year): format-date('2026-02-29', '%Y-%m-%d') or same with timezone (2026-02-29T10:00:00Z)
9: (hour 25 + Z): format-date('2026-12-01T25:00:00.000Z', '%Y-%m-%d') 
11: format-date('2026-12-01T24:00:00', '%Y-%m-%d') 
12: (hour only + Z): format-date('2026-12-01T10Z', '%Y-%m-%d')

latin-panda (changes requested):
I finished testing, and comparing with Collect, I left my findings below

packages/xpath/src/lib/datetime/coercion.ts
garethbowen: Ok! All the weird ones (like 29 days in February) were caused by JS Date() function being rubbish. So I deleted a whole lot of code and relied more on Temporal and it seemed to work.

I also added all your tests as unit tests to prove it.

garethbowen: > we need to decide whether to print blank or fail with an error message

The problem now is we have no way to show a message except for an unrecoverable one. Do you feel comfortable using blank for now and we can go back to returning a message when we fix this: https://github.com/getodk/web-forms/issues/883

latin-panda: Yes, that works. Thanks

packages/xpath/test/xforms/format-date.test.ts
```diff
@@ -66,29 +66,44 @@ describe('#format-date()', () => {
...
+  describe('invalid dates', () => {
+    [
+      "format-date('invalid', '%e | %a' )",
+      "format-date('2026-12-01T25:00:00.000Z', '%e | %a' )", // 25 hours
+      "format-date('2026-12-01T25:00:00.000+10:00', '%e | %a' )", // 25 hours
+      "format-date('2026-13-01', '%e | %a' )", // 13 months
+      "format-date(number('invalid'), '%Y-%m-%d')", // not a number
+      "format-date('2026-04-31T23:00:00+12:00', '%Y-%m-%d')", // there aren't 31 days in April
+      "format-date-time('2026-02-29T10:00:00Z', '%Y-%m-%d %H:%M')", // there aren't 29 days in February non leap year
+      "format-date('2026-12-01T24:00:00Z', '%Y-%m-%d')", // there aren't 24 hours in a day
```
latin-panda, line 101: A date with only hour is an invalid format and defaults to an empty string, correct? 
`2026-12-01T10Z`

latin-panda (approved):
Nice! Approving now; just one comment to add the case below to the unit test

packages/xpath/test/xforms/format-date.test.ts
garethbowen: Temporal parses that example to 10am, which seems reasonable to me. What do you think?

latin-panda: I was concerned that xforms-engine doesn't support it (the format was a big issue before), but then I saw this [regex for xpath](https://github.com/getodk/central-frontend/blob/a291d38388d0b91755eea0b48296f389f50d1e09/packages/common/src/constants/datetime.ts#L34), and it seems xpath is more flexible with these formats.

Okay leaving it like this

latin-panda (approved):

PR #1821, opened 2026-08-26
Fixes: don't close new/upload entity modal on outside click or ESC

### LINKED ISSUE #1961: Clicking outside the New Entity modal discards all entered values
1. Go to an Entity List
2. Click "New Entity"
3. Fill in a bunch of important values
4. Click outside the modal, see that it dismisses, and that your entered values have disappeared

sadiqkhoja:
Closes https://github.com/getodk/central/issues/1961

Added a new prop `persistent` which disables hiding on outside click and ESC. 

No

No

### DIFF
--- a/apps/central/test/components/modal.spec.js
+++ b/apps/central/test/components/modal.spec.js
@@ -194,7 +194,21 @@ describe('Modal', () => {
     });
   });
 
-  describe('persistent prop', () => {
+  describe('hiding the modal', () => {
+    it('emits hide on ESC key by default', async () => {
+      const modal = mountComponent();
+      await modal.get('.modal').trigger('keydown.esc');
+      should.exist(modal.emitted().hide);
+    });
+
+    it('emits hide on click outside by default', async () => {
+      const modal = mountComponent();
+      const modalEl = modal.get('.modal');
+      await modalEl.trigger('mousedown');
+      await modalEl.trigger('click');
+      should.exist(modal.emitted().hide);
+    });
+
     it('does not emit hide on ESC key when persistent is true', async () => {
       const modal = mountComponent({
         props: { persistent: true }
--- a/apps/central/test/components/modal.spec.js
+++ b/apps/central/test/components/modal.spec.js
@@ -193,4 +193,24 @@ describe('Modal', () => {
       });
     });
   });
+
+  describe('persistent prop', () => {
+    it('does not emit hide on ESC key when persistent is true', async () => {
+      const modal = mountComponent({
+        props: { persistent: true }
+      });
+      await modal.get('.modal').trigger('keydown.esc');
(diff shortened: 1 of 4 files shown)


apps/central/test/components/modal.spec.js
```diff
@@ -193,4 +193,24 @@ describe('Modal', () => {
       });
     });
   });
+
+  describe('persistent prop', () => {
+    it('does not emit hide on ESC key when persistent is true', async () => {
+      const modal = mountComponent({
+        props: { persistent: true }
+      });
+      await modal.get('.modal').trigger('keydown.esc');
```
matthew-white, line 216:
```suggestion
      await modal.get('.modal').trigger('keydown', { key: 'Escape' });
```

From a quick look at the Vue Test Utils docs, it's not clear to me that it supports the same `.esc` syntactic sugar as Vue itself.

Although, it's also not throwing an error, so maybe that means that it does?

apps/central/test/components/modal.spec.js
matthew-white, line 1: This might be outside the scope of this PR, but I see that we don't have parallel tests for the default case, when `persistent` is `false`. I feel like it'd build confidence for us to have tests of the default hide behavior along with tests of the `persistent` behavior.

matthew-white (approved):

apps/central/test/components/modal.spec.js
matthew-white: OK I see that #1823 also uses that syntactic sugar, so I guess Vue Test Utils does support it! #1823 makes a positive assertion about the presence of an event, not the absence, so I think it's clear that `.esc` works.

apps/central/test/components/modal.spec.js
```diff
@@ -193,4 +193,24 @@ describe('Modal', () => {
       });
     });
   });
+
+  describe('persistent prop', () => {
+    it('does not emit hide on ESC key when persistent is true', async () => {
```
matthew-white, line 212: The test in #1823 has a `nextTick()`. Does this test need one?

matthew-white: And maybe this test needs `state: true` among its props? I'll leave it to you, I'm just noticing those differences between the two PRs' tests.

apps/central/test/components/modal.spec.js
sadiqkhoja: yeah it's not documented but I was find a test https://github.com/vuejs/vue-test-utils/blob/9627a809a177ef9798ca66a669891c73fc740d8d/test/specs/wrapper/trigger.spec.js#L95

apps/central/test/components/modal.spec.js
sadiqkhoja: `state` is already set to true in `mountComponent`.

PR #1823, opened 2026-08-26
Fixes/focus modal on shown

### LINKED ISSUE #2163: “Connect data” modal doesn’t close on escape key
Tested on staging.

versions:
040ad8696bde2d8d3cecfd992aad724d49d9d88f (v2026.2.4-10-g040ad86)
 22757ca26c55ab08a0002d69b8b5565bb19b49fc client (@getodk/web-forms@1.0.0-100-g22757ca2)
 5d88c1ebfe707b013e0a64da3f353e3576824af9 server (v2026.2.0-54-g5d88c1eb)

PS: works correctly on test server, so it's a regression.

sadiqkhoja:
Closes https://github.com/getodk/central/issues/2163

Brings back the function that was there with the bootstrap.

- Onshow, modal or it's element should have the focus
- any time focus is back to the document (after focus shift to browser control), it should come back to the modal

no

no

### DIFF
--- a/apps/central/src/components/modal.vue
+++ b/apps/central/src/components/modal.vue
@@ -17,7 +17,7 @@ checkScroll() method may add the `has-scroll` class. -->
     <div v-show="state" ref="el" v-bind="$attrs" class="modal" tabindex="-1"
       :style="backdrop ? { backgroundColor: 'rgba(0, 0, 0, 0.5)' } : null"
       role="dialog" :aria-labelledby="titleId" @mousedown="modalMousedown"
-      @click="modalClick" @keydown.esc="hideOnEsc" @focusout="refocus">
+      @click="modalClick" @keydown.esc="hideOnEsc">
       <div class="modal-dialog" :class="sizeClass" role="document">
         <div class="modal-content">
           <div class="modal-top-actions">
@@ -166,6 +166,16 @@ const observer = new MutationObserver(() => {
   }
 });
 
+// Redirects focus back to the modal if it moves outside. Similar to bootstrap's internal
+// implementation
+const enforceFocus = (event) => {
+  if (!props.state || el.value == null) return;
+  // Do not focus the .modal element if it is already focused or if it
+  // contains the focused element.
+  if (event.target === el.value || event.target.closest('.modal') === el.value) return;
+  el.value.focus();
+};
+
 const show = () => {
   addModalOpenClass();
   checkScroll();
@@ -176,8 +186,16 @@ const show = () => {
     characterData: true
   });
   window.addEventListener('resize', handleWindowResize);
+  document.addEventListener('focusin', enforceFocus);
+
   // Emit shown after nextTick to ensure DOM is updated (v-show has made element visible)
-  nextTick(() => { emit('shown'); });
+  nextTick(() => {
+    // Focus the modal so that pressing Escape will hide it.
+    if (document.activeElement != null && document.activeElement.closest('.modal') !== el.value) {
+      el.value.focus();
+    }
+    emit('shown');
+  });
   openModal.shown(el.value);
 };
 const removeSelection = () => {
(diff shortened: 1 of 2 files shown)


apps/central/src/components/modal.vue
```diff
@@ -164,6 +164,16 @@ const observer = new MutationObserver(() => {
...
 
+// Redirects focus back to the modal if it moves outside. Similar to bootstrap's internal
+// implementation
+const enforceFocus = (event) => {
+  if (!props.state || el.value == null) return;
+  // Do not focus the .modal element if it is already focused or if it
+  // contains the focused element.
+  if (event.target === el.value || event.target.closest('.modal') === el.value) return;
+  el.value.focus();
+};
```
sadiqkhoja, line 177: this ensures the when focus comes back from browser controls to the document, it goes to the modal instead of "ODK Central" heading in the navbar.

apps/central/src/components/modal.vue
```diff
@@ -174,8 +184,16 @@ const show = () => {
...
   window.addEventListener('resize', handleWindowResize);
+  document.addEventListener('focusin', enforceFocus);
+
   // Emit shown after nextTick to ensure DOM is updated (v-show has made element visible)
-  nextTick(() => { emit('shown'); });
+  nextTick(() => {
+    // Focus the modal so that pressing Escape will hide it.
+    if (document.activeElement != null && document.activeElement.closest('.modal') !== el.value) {
+      el.value.focus();
+    }
```
sadiqkhoja: this focuses on the modal after modal is added to the dom and it is focusable. The `if` condition ensures the we don't move the focus back to the modal if child component focuses a first control element.

apps/central/src/components/modal.vue
matthew-white, line 1: Would you mind checking whether this changes fixes getodk/central#1294? It'd be fun to close out that long-standing issue.

apps/central/test/components/modal.spec.js
```diff
@@ -138,6 +138,27 @@ describe('Modal', () => {
...
+      document.activeElement.should.equal(modal.get('.modal').element);
+    });
+
+    it('emits hide event after Escape key is pressed', async () => {
+      const modal = mountComponent({
+        props: { state: true, hideable: true },
+        attachTo: document.body
+      });
+      await nextTick();
+      await modal.get('.modal').trigger('keydown.esc');
```
matthew-white: I know in a real browser, the modal needs to be focused in order for the Escape key to work. But since we're programmatically triggering the Escape on a specific element, I feel like the focus probably isn't necessary. The test would pass without it. In other words, I think this is a good test, but ultimately it's not extremely related to focus. So maybe it doesn't need to go in this `describe()` block about focus. And maybe things like `attachTo` and `nextTick()` aren't needed.

apps/central/src/components/modal.vue
```diff
@@ -174,8 +184,16 @@ const show = () => {
     characterData: true
   });
   window.addEventListener('resize', handleWindowResize);
+  document.addEventListener('focusin', enforceFocus);
+
   // Emit shown after nextTick to ensure DOM is updated (v-show has made element visible)
-  nextTick(() => { emit('shown'); });
+  nextTick(() => {
+    // Focus the modal so that pressing Escape will hide it.
+    if (document.activeElement != null && document.activeElement.closest('.modal') !== el.value) {
```
matthew-white: Is it possible for `document.activeElement` to be `null`? Probably not, right? If it is possible, maybe this logic should be slightly different:

```suggestion
    if (document.activeElement == null || document.activeElement.closest('.modal') !== el.value) {
```

matthew-white (approved):

apps/central/test/components/modal.spec.js
sadiqkhoja: removed this test. It is same as `emits hide on ESC key by default` added in the previous PR

apps/central/src/components/modal.vue
sadiqkhoja: no it doesn't fixes https://github.com/getodk/central/issues/1294 :(

what I've found with quick observation is that firefox doesn't triggers `focusin` event when an active element is removed from the DOM and focus is switched to the `body`.

PR #1824, opened 2026-08-26
fix(web-forms#850): show thank you message when window can’t be closed

### LINKED ISSUE #850: Remove window.close
On "New Submission" page for the logged in user, we show a success modal with close button which tries to close the window by calling `window.close`. It currently works because we always open new submissions page in a new tab but if you copy the link and open it in a new tab/window/incognito then close button doesn't work, browser throws: `Scripts may only close windows that were opened by a script.`

Since we intend to getodk/central#933, we should remove the window.close functionality. My pitch is to replace it with "Back to Submissions" button.

latin-panda:
Closes https://github.com/getodk/web-forms/issues/850

Reusing existing modals, minimal changes. 

Users are instructed to manually close modal

No


latin-panda:
@garethbowen, could you please review when you get a chance?

garethbowen (changes requested):
I'm not sure about the UX on this one.

1. I'm always surprised by the "close" action - in my head "close" means close the dialog, not the window.
2. Even more surprising is clicking the "close" button and getting a second dialog saying "you can close this now".
3. What if I change my mind and now I want to fill it in again?

I prefer either of @lognaturel suggestions here: https://github.com/getodk/central/issues/1077#issuecomment-2887352114

Either (a) check window.opener and decide whether to show the "submissionModal" or "thankYouModal" based on that, so you only ever get 1 modal, or (b) remove the close button altogether (people know how to close a tab, right?).

garethbowen (approved):
Sweet!

PR #1826, opened 2026-08-26
fix(web-forms#891): reject forms with computation cycles at load

### LINKED ISSUE #891: Forms with cycles break while filling questions
### Expected Behavior

<!-- A clear and concise description of what you expected to happen. -->

 The form should not load. It should show a clear error naming the problem nodes. 
Collect does this. JavaRosa rejects the form at load with:                                                                                                                                         
                                                                                                                                                                                                     
```
Cycle detected in form's relevant and calculation logic!                                                                                                                                         
The following nodes are likely involved in the loop:                                                                                                                                             
 /data/details 
```

### Observed Behavior

<!-- A clear and concise description of what actually happened, including any error messages. -->

The form loads fine. It breaks later, when the user answers a question, which can be hard for a form designer to detect in a big form.
`An error occurred while loading this form. Potential Infinite Loop Detected`.
It doesn't say which field caused it. And it says "while loading" even though the user was filling the form already.  

### Reproduction Steps

<!-- Provide the form and a clear, step-by-step guide to reproduce the bug. -->
Form (XML or XLSX file):

[nav-test-d.xml](https://github.com/user-attachments/files/31482566/nav-test-d.xml)

Steps:
1. Open the form, and it loads (Collect rejects it here)
2. The group's relevant reads a question inside that same group:                                                                                                                                   
  `<bind nodeset="/data/details" relevant="/data/details/keep != 'no'"/>  `
3. Type `no` in "Keep this page?". 
4. The error modal appears.

### Screenshots or Videos

<!-- Add any relevant screenshots, videos, or links to help illustrate the issue. -->

### Environment

- Browser version: All
- OS: macos
- Web Forms version: 1.0.3

### Additional Context

<!-- Any other details, such as device info, network conditions, or related forum discussions? -->

latin-panda:
Closes https://github.com/getodk/web-forms/issues/891

Manual test

Message matching Collect, and shows the problematic expression:

I checked how JavaRosa detects cycles and used it as inspiration.

Quicker feedback for the form designer when a form fails, at load time, with more information about the problem.

No


packages/xforms-engine/test/integration/form-definition-validity.test.ts
```diff
@@ -17,387 +17,172 @@ import { describe, expect, it } from 'vitest';
 import { intAnswer } from '../scenario/answer/ExpectedIntAnswer.ts';
 import { Scenario } from '../scenario/jr/Scenario.ts';
 
-/**
```
latin-panda, line 20: We had tests for these, but off. I just enabled them, improved the file, and reduced noise.

latin-panda:
@garethbowen what do you think? I found this while testing pagination

packages/xforms-engine/src/parse/model/rejectComputationCycles.ts
```diff
@@ -0,0 +1,83 @@
...
+  const remaining = unresolved.filter((reader: Computation) => {
+    return unresolved.some((writer) => dependsOn(reader, writer, modelNodesets));
+  });
+
+  if (remaining.length === unresolved.length) {
+    return remaining;
+  }
+
+  return dropResolvable(remaining, modelNodesets);
+};
```
garethbowen, line 72: This can be more efficient but I'm not expecting there will be that many computations in a form, and most will be removed on the first pass, so it's probably not worth the effort. What do you think?

packages/xforms-engine/test/integration/form-definition-validity.test.ts
```diff
@@ -17,387 +17,172 @@ import { describe, expect, it } from 'vitest';
...
+      "supports self references in constraints [and subsequent value assignments, where permitted by the field's `constraint` expression]",
+      async () => {
+        const scenario = await initScenario(selfReferencingConstraintForm);
 
         scenario.next('/data/count');
         scenario.answer(5);
 
-        // assertThat(scenario.answerOf("/data/count"), is(nullValue()));
         expect(scenario.answerOf('/data/count').getValue()).toBe('');
+        expect(scenario.answerOf('/data/count')).toEqualAnswer(intAnswer(5));
```
garethbowen: This doesn't make sense to me... isn't one of these guaranteed to fail?

garethbowen (changes requested):
Nice work!

Thank you for cleaning up that test.

packages/xforms-engine/src/parse/model/rejectComputationCycles.ts
latin-panda: I checked and agree it could be more efficient, but it's not worth it. The code would be 2-3x bigger, and the ms saved wouldn't be perceptible.

packages/xforms-engine/test/integration/form-definition-validity.test.ts
latin-panda: You're right, this test came from the original port. Without the bad line, it's a copy of the test above it, so I deleted the whole test.

garethbowen (approved):
Nice!

PR #1829, opened 2026-08-26
fix(web-forms#889): fixed styling for form level errors

### LINKED ISSUE #889: Error message that appears when trying to send form with errors disappear after scrolling page on phone
### Expected Behaviour

Error message should be always visible when scrolling

### Observed Behaviour

Error message disappear after scrolling page

### Reproduction Steps

Steps:
1. Go to https://staging.getodk.cloud/projects/139/forms/audit%20repeat%20required%20const%20track%20changes/submissions
2. Tap on New submission
3. Try to send form without filling questions so error message appears
4. Scroll down the page

### Screenshots or Videos

https://github.com/user-attachments/assets/7288a490-f2dc-4312-a49f-73269b91c86c

### Environment

Browser version: Chrome 149.0.7827.137, Safari
OS: iOS 26.6
Web Forms version: v1.0.3

garethbowen:
Closes getodk/web-forms#889

Testing in desktop and mobile on firefox and chrome
Needed: testing on safari

"sticky" is the native way to do this.

None.

No.

### DIFF
--- a/packages/web-forms/src/components/OdkWebForm.vue
+++ b/packages/web-forms/src/components/OdkWebForm.vue
@@ -422,7 +422,6 @@ onUnmounted(() => {
 		}
 
 		.form-error-message.p-message.p-message-error {
-			display: inline-block;
 			position: sticky;
 			z-index: var(--odk-z-index-error-banner);
 			border-radius: var(--odk-radius);
--- a/packages/web-forms/src/components/OdkWebForm.vue
+++ b/packages/web-forms/src/components/OdkWebForm.vue
@@ -333,7 +333,6 @@ onUnmounted(() => {
 		:class="{ 'submit-pressed': submitPressed }"
 	>
 		<div class="form-wrapper">
-			<div v-if="showValidationError" class="error-banner-placeholder" />
 			<!-- Closable error message to clear the view and avoid overlap with other elements -->
 			<Message
 				v-if="showValidationError"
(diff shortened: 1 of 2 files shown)


garethbowen:
@latin-panda Would you mind having a look at this? Unfortunately I couldn't test on iOS so I couldn't reproduce the original bug, but I'm hoping that the different approach will resolve the bug, as well as improve the UX in the other ways mentioned in [this comment](https://github.com/getodk/web-forms/issues/889#issuecomment-5431818823).

packages/web-forms/src/components/OdkWebForm.vue
```diff
@@ -422,20 +421,17 @@ onUnmounted(() => {
 			padding: 2rem;
 		}
 
-		.error-banner-placeholder {
-			height: 4rem;
-		}
-
 		.form-error-message.p-message.p-message-error {
-			position: fixed;
+			display: inline-block;
```
latin-panda: I removed this line and everything seems okay, I think it can be removed.

latin-panda (approved):
I don't have an iPhone so I couldn't check the original bug, but on macOS Safari it works fine.                                             
One thing I noticed while testing but _I think_ is preexisting, so it could be fix here or file an issue.                                                                             

If a form has `<odk:setgeopoint event="odk-instance-first-load" ref="/data/start_location"/>` and you reject the location permission in the browser, the banner shows up with the location error (good), but it also says "3 questions with errors". At that point, I haven't tapped Send yet, so nothing is highlighted and the message is confusing. I'd expect the banner to only mention the location error until Send is tapped, and only then show the count and highlight the questions in red.

[error-banner-test.xml](https://github.com/user-attachments/files/31522825/error-banner-test.xml)

garethbowen:
> I don't have an iPhone so I couldn't check the original bug,

Oh sorry! I thought you did. I'll send it back to Wiktor to test. As you've approved I'll make that one suggestion and merge and then he can test on staging.

Regarding the geolocation error I think it's preexisting and I think could use some UX input, so I created a new issue: https://github.com/getodk/web-forms/issues/897

PR #1830, opened 2026-08-27
fix(web-forms#893, #888): fix focus behavior when navigating to a question 

### LINKED ISSUE #888: Note displayed with blue outline
#### Web Forms version
1.0.3

#### Browser
Chrome, Firefox
 
#### Problem description
 A note is displayed with a blue outline. The outline disappears after clicking something.
Is it expected?
 #### Screenshots
https://jam.dev/c/fbb8e033-948e-4fed-8398-8fc9fd33c347

1. Open a new submission in 
https://staging.getodk.cloud/projects/235/forms/selects_images_pages/submissions
or
https://staging.getodk.cloud/projects/235/forms/all-widgets/submissions
2. Check the blue outline in the first question.

### LINKED ISSUE #893: Calendar from a date question displayed below the form page
### Observed Behaviour
The calendar from a date question is displayed below the form page.
I think it occurs when there are some questions on a field-list which go beyond the screen view (I tried to reproduce it with 1,2,3 question on a field-list but the issue doesn't occur then). The issue occurs with and without pages style.

### Reproduction Steps

1. Click "new submission" in https://staging.getodk.cloud/projects/235/forms/datepages/submissions 
2. Scroll to the bottom of the page

### Screenshots or Videos

Image

### Environment

- Browser version: Chrome, Firefox
- Web Forms version: 1.0.3

latin-panda:
Closes https://github.com/getodk/web-forms/issues/888
Closes https://github.com/getodk/web-forms/issues/893

Manual tests with forms from the issues

Adjusted selectors

Better UI

No


latin-panda:
@garethbowen, this is ready for review. I fixed them in one PR because the question container for the datepicker also needed the outline fix.

garethbowen (commented):
The outline:none works for me.

For the date picker, it's a shame to lose this functionality. The root cause appears to be a race condition where we're rendering the picker control before the page is ready, which may cause other issues for other question types too. If I wrap the Root.navigateToFirstControl in a 1s setTimeout then everything works as expected.

Is there a way to trigger the navigation from the client on content loaded? Or set in the Root immediately but only register the watcher once the page is loaded?

latin-panda:
@garethbowen I tried lifecycle hooks and watching DOM elements, but at the end, what fixed it for pageless and paged forms was nextTick.

I'm concerned about the user experience because the PrimeVue datepicker opens the popup on load, covering other questions. The video shows form load and page navigation, and it would be the same for new repeat instances and error questions (same code). @lognaturel, I would appreciate your feedback as well on the datepicker with focus.


https://github.com/user-attachments/assets/1814c885-4998-4a17-a40e-9f5272dfed81

garethbowen (commented):
Yes I see what you mean if the date is the first field when the page loads (either first time, or after clicking next). I'm ambivalent about the two options...

I think it's useful when the user explicitly clicks the "view" button. How easy would it be to separate the code so dates are focused when the view button is clicked, but not when it's a page navigation?

latin-panda:
> I think it's useful when the user explicitly clicks the "view" button. How easy would it be to separate the code so dates are focused when the view button is clicked, but not when it's a page navigation?

I've proceeded with this idea and will monitor user feedback. When an error occurs in a question using the PrimeVue date picker, the focus is triggered and the calendar opens; in the other cases (page load, etc...), the focus isn't set. Separating the code specifically for dates would require some refactoring, so I decided not to do it and keep the code simple. I added a simple special case by checking the style classes. 

It's ready for another review

lognaturel:
> I added a simple special case by checking the style classes

Thanks for thinking through this, seems like a good solution! If you haven't already done so, making sure there's a unit test on this would be good. The goal would be to have something that fails if any of the involved classes change.

garethbowen (approved):
Thanks! Looks great.

PR #1831, opened 2026-08-27
fix(web-forms#887): remove map feature when value is cleared  

### LINKED ISSUE #887: Map pin not removed when geopoint value is cleared
### Expected Behaviour

When a geopoint value is cleared programatically (form logic), the map should go blank. No pin, and the default message should show. 

### Observed Behaviour

The old pin stays on the map. The message says "tap to move point". But the status bar says "no point saved". The pin can only be removed by tapping the map to save it, then using "remove point".
     

### Reproduction Steps

Form (XML or XLSX file): [geopoint-cleared-by-setvalue.xml](https://github.com/user-attachments/files/31431561/geopoint-cleared-by-setvalue.xml)

Steps:
 1. Open the form in Web Forms.                                                                                                                                                                     
  2. Select "Item 1 (has geometry)". A pin appears on the map.                                                                                                                                       
  3. Change the selection to "Item 2 (no geometry)".                                                                                                                                                 
  4. The pin from Item 1 is still on the map, but the status bar says "no point saved".  

### Screenshots or Videos

https://github.com/user-attachments/assets/7854535f-496f-4b2a-87aa-528555bcc8dc

### Environment

- Browser version: All
- OS: All
- Web Forms version: 1.0.3

### Additional Context

[forum post](https://forum.getodk.org/t/webforms-behaviour-notes/58272/39?u=jenniferq)

latin-panda:
Closes https://github.com/getodk/web-forms/issues/887

<!-- 

-->

CI pass and manual tests using form from the issue and `All questions` form:

https://github.com/user-attachments/assets/986607c9-f0b3-439d-b7d7-c2d40373a88f

The map never noticed when the form cleared the value (e.g. setValue). Now, we add a couple of simple validations to catch that scenario. 

Users see the pin (or line/polygon) disappear when the form clears the value. Before, it stayed on the map by mistake.                                                                             
                                                                                                                                                                                                     
Nothing else changed. Placing, dragging, deleting, undo and select-from-map all work as before.

No

### DIFF
--- a/.changeset/puny-wings-cry.md
+++ b/.changeset/puny-wings-cry.md
@@ -1,0 +1,5 @@
+---
+"@getodk/web-forms": patch
+---
+
+Remove point, trace or shape from the map when its value is cleared by form logic.
--- a/packages/web-forms/src/components/common/map/MapBlock.vue
+++ b/packages/web-forms/src/components/common/map/MapBlock.vue
@@ -127,6 +127,7 @@ watch(
 	() => props.savedFeatureValue,
 	(newValue) => {
 		mapHandler.applySavedFeatureValue(newValue);
+		pointPlaced.value = !!mapHandler.getSavedFeatureValue()?.length;
 	}
 );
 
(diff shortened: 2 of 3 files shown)


latin-panda:
@garethbowen this is ready for review!

garethbowen (approved):
Very nice.

PR #1836, opened 2026-08-27
EntityUpload: update titles above tables

matthew-white:
This PR makes progress on getodk/central#1904. It adds titles above the tables in the `EntityUpload` modal:

- Above the first table: "Your current Entities"
  - Remove the previous `.panel-title` of "[Entity List name] server data"
  - With the `.panel-title` gone, we don't need to wrap the table in a `.panel` anymore.
- Above the second table: "New Entities"
- Also adding a class `entity-upload-table-container` to set the margin between the table and the title above it, as well as the margin below each table (including the margin between the first table and the title above the second table).
  - With the new, fixed `margin-bottom` on the `.entity-upload-table-container`, we no longer need `margin-top` on `EntityUploadErrors`.
  - Move the `margin-top` on `EntityUploadWarnings` to `margin-bottom` on `EntityUploadErrors`. `margin-bottom` is more common in the codebase than `margin-top`.
  - `EntityUploadWarnings` doesn't need any margin anymore. That's in part because `EntityFileSelect` still sets `margin-top`. It's possible that I'll change that in a later PR.

I tried it out locally. Tests continue to pass.

I wasn't exactly sure how to structure some things. For example, I debated whether to place the `.entity-upload-section-title` inside the new `.entity-upload-table-container` or above/before it. I think the approach I took makes sense, but it should also be easy to adjust if we want to later.

### DIFF
--- a/apps/central/src/components/entity/upload.vue
+++ b/apps/central/src/components/entity/upload.vue
@@ -446,6 +446,7 @@ watch(() => props.state, (state) => {
   font-weight: bold;
   margin-bottom: 4px;
 
+  // Explanatory/descriptive text that immediately follows the title
   + p, + p + p {
     margin-bottom: 10px;
     &:last-of-type { margin-bottom: 20px; }
--- a/apps/central/src/components/entity/upload.vue
+++ b/apps/central/src/components/entity/upload.vue
@@ -417,14 +417,12 @@ watch(() => props.state, (state) => {
 
   .panel-simple {
     .panel-heading {
-      @include text-overflow-ellipsis;
       background-color: $color-action-background;
       border-bottom: none;
       color: #fff;
     }
 
     .panel-body { padding: 0; }
-
     thead { background-color: #c5dfe7; }
   }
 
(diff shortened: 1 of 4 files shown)


apps/central/src/components/entity/upload.vue
```diff
@@ -452,21 +441,14 @@ watch(() => props.state, (state) => {
...
-  }
 }
 
 .entity-upload-section-title {
   font-size: 16px;
   font-weight: bold;
   margin-bottom: 4px;
 
-  ~ p {
+  + p, + p + p {
```
matthew-white, line 450: Add a comment that this is selecting introductory/explanatory text.

ktuite (approved):

matthew-white (commented):

apps/central/src/components/entity/upload.vue
matthew-white: Done in 8a058768e3c5b5f51f7673b06244b3ec0590002e

PR #1842, opened 2026-08-28
fix(web-forms#892): better handling of dynamically updating attachment names

### LINKED ISSUE #892: Dynamic media question label filenames can fail when dynamic value typed too fast
### Description
See https://github.com/getodk/web-forms/issues/892#issuecomment-5434442308

Forum test form: https://forum.getodk.org/t/dynamic-images-loaded-to-an-image-widget-from-a-choice-csv-selection/41921

Simplest case: [dynamic_media.zip](https://github.com/user-attachments/files/31483698/dynamic_media.zip)

---

### Checklist
  - [x] No
  - [x] No
  - [ ] No
  - [ ] No
---

### User Stories
As a form designer, I want to specify a filename that is dynamic so that the displayed media adapts to previous answers the user has given.

---

### Additional Notes

-

---

### Related Issues

-

garethbowen:
Closes getodk/web-forms#892

Manual testing
- adding a random delay in responding to requests to test the abort
- testing of filenames with non-url safe characters
- testing entities and other attachments

The first issue is about the requests completing in a random order, and the last request to finish winning. Aborting previous requests means the last one initiated wins.

The second issue was about issuing one request to the server per character typed. By checking the known attachments we can only make a request when it matches.

It shouldn't do.

No.

### DIFF
--- a/apps/forms/src/components/web-form-renderer.vue
+++ b/apps/forms/src/components/web-form-renderer.vue
@@ -55,6 +55,11 @@ const withToken = (url) => `${url}${queryString({ st: props.st })}`;
 
 const getAttachment = (requestUrl: URL) => {
   const encodedName = encodeURIComponent(requestUrl.pathname.split('/').pop()!);
+  // TODO for testing purposes, put a random delay here to exacerbate the race condition
+  if (encodedName !== 'a.jpg') {
+
+    return new Response("Not Found", { status: 404 });
+  }
   const url = withToken(`/v1/projects/${props.form.projectId}/forms/${props.form.xmlFormId}${draftPath.value}/attachments/${encodedName}`);
   return fetch(url);
 };
--- a/apps/forms/src/components/web-form-renderer.vue
+++ b/apps/forms/src/components/web-form-renderer.vue
@@ -22,7 +22,7 @@ export interface WebFormsRendererProps {
   instanceId?: string | null;
   submissionAttachments?: string[] | null;
   defaultParameters?: Record<string, string>;
-  st?: string | null
+  st?: string | null;
 }
 
 const props = defineProps<WebFormsRendererProps>();
@@ -54,13 +54,12 @@ const visibleModal = ref();
 const withToken = (url) => `${url}${queryString({ st: props.st })}`;
 
 const getAttachment = (requestUrl: URL) => {
-  const encodedName = encodeURIComponent(requestUrl.pathname.split('/').pop()!);
-  // TODO for testing purposes, put a random delay here to exacerbate the race condition
-  if (encodedName !== 'a.jpg') {
-
-    return new Response("Not Found", { status: 404 });
+  const fileName = requestUrl.pathname.split('/').pop()!;
+  const decoded = decodeURIComponent(fileName);
+  if (!props.form.attachments.some(a => a.name === decoded)) {
+    return new Response('Not Found', { status: 404 });
   }
-  const url = withToken(`/v1/projects/${props.form.projectId}/forms/${props.form.xmlFormId}${draftPath.value}/attachments/${encodedName}`);
+  const url = withToken(`/v1/projects/${props.form.projectId}/forms/${props.form.xmlFormId}${draftPath.value}/attachments/${fileName}`);
   return fetch(url);
 };
 
--- a/apps/forms/src/utils/api.ts
+++ b/apps/forms/src/utils/api.ts
@@ -1,3 +1,7 @@
+export interface Attachment {
+  name: string
+}
+
 export interface Form {
   name: string;
   xmlFormId: string;
@@ -7,6 +11,7 @@ export interface Form {
   enketoOnceId?: string;
   draft: boolean;
   webformsEnabled: boolean;
+  attachments: Attachment[];
 }
 
 export interface Project {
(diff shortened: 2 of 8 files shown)


apps/forms/src/components/web-form-renderer.vue
```diff
@@ -54,8 +54,12 @@ const visibleModal = ref();
 const withToken = (url) => `${url}${queryString({ st: props.st })}`;
 
 const getAttachment = (requestUrl: URL) => {
-  const encodedName = encodeURIComponent(requestUrl.pathname.split('/').pop()!);
```
garethbowen, line 57: The filename was already encoded so this was encoding it a second time. I doubt it worked...

apps/forms/src/utils/api.ts
```diff
@@ -1,3 +1,7 @@
+export interface Attachment {
+  name: string
```
garethbowen, line 2: I think about having an array of string, instead of an actual object, but I think it's likely we'll want more properties later.

apps/forms/src/utils/api.ts
```diff
@@ -92,13 +97,24 @@ export const getFormXml = async (projectId: number, formId: string, draft: boole
   return await response.text();
 };
 
-const getForm = async (url: string): Promise<Form> => {
+const getAttachments = async (projectId: number, xmlFormId: string, st?: string | null): Promise<Attachment[]> => {
+  const qs = queryString({ st });
+  const url = `/v1/projects/${projectId}/forms/${xmlFormId}/attachments${qs}`;
```
latin-panda: The media is fetched from `.../draft/attachments/` when the form is a `draft` (see `getAttachment` in web-form-renderer) but this list is always from the published form, right? Then this function should receive the `props.draft` flag and add `/draft` here too.

apps/forms/src/utils/api.ts
```diff
@@ -92,13 +97,24 @@ export const getFormXml = async (projectId: number, formId: string, draft: boole
...
+  const response = await fetch(url);
+  const attachments = await response.json() as BackendAttachmentResponseBody[];
```
latin-panda, line 109: It needs to check for `response.ok` (like [getSubmissionAttachmentNames](https://github.com/garethbowen/central-frontend/blob/8a966f78dcc8e06672ee5294cda345cdd4d7ee47/apps/forms/src/utils/api.ts#L78) does) because if this request fails the `.json()` returns a object and `.filter` won't work.

packages/web-forms/src/components/common/media/MediaBlockBase.vue
```diff
@@ -73,14 +96,22 @@ const handleError = (error: string) => {
 };
 
 watchEffect(() => {
+	loading.value = true;
```
latin-panda, line 99: Is it simple to add a short debounce to this component?                                                                                                                                          
When the user starts typing, set loading to true to show the skeleton and start the debounce. So while typing it won't switch back and forth between skeleton and 404 error image.

apps/forms/test/utils/api.spec.ts
```diff
@@ -120,10 +122,39 @@ describe('Test api utility', () => {
...
+          type: 'image',
+          exists: false,
+          blobExists: true,
+          datasetExists: false,
+          hash: 'd41d8cd98f00b204e9800998ecf8427e',
+          updatedAt: '2018-03-21T12:45:02.312Z'
+        },
+      ];
+
+      vi.stubGlobal('fetch', vi.fn().mockImplementation((url: string) => {
```
latin-panda, line 146: Could you please add a case to cover the draft case?

apps/forms/test/components/web-form-renderer.spec.ts
```diff
@@ -131,7 +140,17 @@ describe('WebFormRenderer', () => {
...
+      name: 'simple',
+      xmlFormId: 'simple',
+      projectId: 1,
+      enketoId: '',
+      state: 'open',
+      draft: false,
+      webformsEnabled: true,
+      attachments: [ { name: 'cities.csv' } ]
+    };
+    const component = await mountComponent({ form: given, xform: formWithAttachmentXml });
```
latin-panda, line 154: Could you add a test where a media name is not in the attachments list? It should return 404 and not call fetch.

latin-panda (changes requested):
Very nice! Some comments below

apps/forms/src/utils/api.ts
garethbowen: Done.

apps/forms/test/utils/api.spec.ts
garethbowen: Done.

apps/forms/src/utils/api.ts
garethbowen: Done. Nice catch.

apps/forms/test/components/web-form-renderer.spec.ts
garethbowen: Done.

packages/web-forms/src/components/common/media/MediaBlockBase.vue
garethbowen: I thought about this for a while...

I think this is an edge case - while it makes sense to calculate the image, it's more likely it'll be calculated from a select than a text input, because expecting users to type in the name of one of the attached images would be strange. Other input methods aren't likely to be changed as quickly.

The debouncing code could be fragile with race conditions on when loading is set.

The ideal solution would be for this component to have access to the map of attachments so it could do the synchronous check and only set loading if a request to the backend was being sent, but that would require an API change.

Are you ok to leave this without the debounce, unless this turns out to be a real world use case?

latin-panda: yes, it's okay

latin-panda (approved):
Nice! Ready to merge 🚀 Thank you!

PR #1843, opened 2026-08-28
EntityUpload: list columns that can become new properties

matthew-white:
This PR makes progress on getodk/central#1786. It's the first PR for the issue, but it doesn't fully complete it: there will be follow-up PRs. This PR does the following:

- Removes the error when there is an unknown property.
- Instead, it shows a warning.
- The warning lists the specific unknown properties (i.e., it doesn't just say, "there is some unknown property").
- It allows the user to select one or more unknown properties.
- It allows the user to select all unknown properties at once via a "Select all" checkbox.
- Selecting a property shows it in the second table (the table for the CSV data).
- However, the property is not yet actually included in the request. Frontend doesn't create the new property, and it doesn't send the property's entity data to Backend as part of the upload request. That'll come in a follow-up PR.
- In a follow-up PR, I'll also make the modal remember the property selection even as the user moves from one file selection to the next (up until the modal is closed entirely). In this PR, the property selection is reset whenever a file is selected.

New tests. I also tried it out locally.

One change I made was to `rowToEntity()`, which converts a row/array of CSV values to an entity object. Previously, the resulting entity object was optimized for transport. For example, if its `data` object was empty, it would get a `toJSON` property that would remove it from the payload sent to Backend. However, things get more complicated now that extra properties can be selected/added after the CSV is parsed. That means that the entity objects we parse may differ from the ones we ultimately send to Backend. For that reason, it makes less sense to try to optimize them for transport at parse time.

In this PR, I've gone ahead and removed some of those optimizations, with the thought that if there is a degradation in performance, that's likely something that we'll notice during testing. However, now I'm leaning toward adding back modified versions of those optimizations in a follow-up PR. Those modified optimizations would take place at the time of the request rather than at parse time.

### DIFF
--- a/apps/central/src/components/entity/upload.vue
+++ b/apps/central/src/components/entity/upload.vue
@@ -283,7 +283,7 @@ const { i18n: globalI18n, redAlert } = inject('container');
 const parseEntities = async (file, headerResults, extraProperties, signal) => {
   const results = await parseCSV(globalI18n, file, headerResults.columns, {
     delimiter: headerResults.meta.delimiter,
-    transformRow: rowToEntity(new Set(extraProperties)),
+    transformRow: rowToEntity(new Set(extraProperties ?? [])),
     signal
   });
   if (results.data.length === 0) throw new Error(t('alert.noData'));
--- a/apps/central/src/components/entity/upload.vue
+++ b/apps/central/src/components/entity/upload.vue
@@ -258,26 +258,32 @@ const validateHeader = ({ columns, errors: papaErrors }) => {
   return result;
 };
 const { t } = useI18n();
+// noPropertyData is used to minimize the JSON sent to Backend: the JSON won't
+// specify a `data` property for an entity without property data.
+const noPropertyData = { toJSON: () => undefined };
 const rowToEntity = (extraProperties) => (values, columns) => {
   let label;
   const data = Object.create(null);
+  let hasProperty = false;
   const extraData = extraProperties.size !== 0 ? Object.create(null) : undefined;
   for (const [i, value] of values.entries()) {
     if (value === '') continue; // eslint-disable-line no-continue
 
     const column = columns[i];
-    if (column === 'label')
+    if (column === 'label') {
       label = value;
-    else if (dataset.propertyMap.has(column))
+    } else if (dataset.propertyMap.has(column)) {
       data[column] = value;
-    else if (extraProperties.has(column))
+      hasProperty = true;
+    } else if (extraProperties.has(column)) {
       extraData[column] = value;
+    }
   }
 
   if (label == null || /^\s+$/.test(label))
     throw new Error(t('alert.blankLabel'));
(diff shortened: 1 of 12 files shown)


apps/central/src/components/entity/upload.vue
```diff
@@ -384,15 +379,27 @@ const showWarningRows = (range) => {
...
+  else
+    selectedProperties.delete(name);
+};
+const propertiesToCreate = computed(() =>
+  warnings.value?.extraProperties?.filter(name => selectedProperties.has(name)));
+
 const { request, awaitingResponse: uploading } = useRequest();
 const uploadProgress = ref(0);
 const upload = () => {
+  const entitiesToSend = csvEntities.value.map(entity =>
```
matthew-white: Remove `extraProperties` from these. Add a test to verify this case.

apps/central/src/components/entity/upload.vue
```diff
@@ -329,7 +321,8 @@ const selectFile = (file) => {
         return Promise.resolve();
       }
 
-      return parseEntities(file, headerResults, signal)
+      const extraProperties = validation.warnings?.extraProperties;
```
matthew-white, line 332:
```suggestion
      const extraProperties = validation.warnings?.extraProperties ?? [];
```

apps/central/src/components/entity/upload.vue
```diff
@@ -356,15 +349,17 @@ const selectFile = (file) => {
...
 watch(() => props.state, (state) => {
   if (state) return;
   abortParse();
   csvEntities.value = null;
   fileMetadata.value = null;
   errors.value = null;
   warnings.value = null;
+  selectedProperties.clear();
 });
-onBeforeUnmount(() => { abortParse(); });
```
matthew-white, line 367: Maybe revert

matthew-white (commented):

ktuite (approved):

matthew-white:
I've rebased now that #1834 has been squash-merged.

apps/central/src/components/entity/upload.vue
matthew-white: I ended up deciding not to revert. It's nice to move this lifecycle hook closer to where the parsing happens.

apps/central/src/components/entity/upload.vue
matthew-white: > Remove `extraProperties` from these.

Done in 4743963d187c66ec28d724fbdee5154cc3425ba7

> Add a test to verify this case.

Done in 04f286707d158f2bded90ae98116dbd64ed6a724

apps/central/src/components/entity/upload.vue
matthew-white: Solved the problem in a slightly different way in a4fb1f8ad129dc1b2ea3cd6c1f51b33c4ea0d5f0

PR #1844, opened 2026-08-31
fix(web-forms#886): allow commas as decimal places

### LINKED ISSUE #240: String to number conversion: Differences between JavaRosa and Webforms
While working on #49, I stumbled upon some implementation differences between webforms and JavaRosa, that came about while thinking about "how and why does JR allow a randomization seed from a node with text, is that intentional or accidental, and should Webforms do the same" (so far I only have answers to #1 and maybe #2). Anyway:

The equivalent of a JS `Number` as used in Webforms is the Java `Double`. Same thing, same precision, so in theory this should allow identical outcomes on both platforms.

- [JavaRosa implementation](https://github.com/getodk/javarosa/blob/3efbb35b86cf22359eef85d426ca9b39af005714/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L702-L746) (think of it as a `toDouble`, as that's what it does)

- [The correspondent code site in Webforms might be here](https://github.com/getodk/web-forms/blob/1f868e80801e37e9d849337b1f2d9ffedd554adf/packages/xpath/src/evaluations/NodeEvaluation.ts#L60)

Evidently the JR implementation is much more extensive; and there's talk about "xpath type conversion rules" — which I haven't looked up, but those sound as if they could be relevant for webforms.

Poignantly, JR allows for `"123,123"` -> `123.123`, while Webforms makes this a `NaN`.
Conversely, Webforms allows `"0x12"` -> `18`, while JR would make this a `NaN`.

The question is thus: which one (if any) of the two is right? And if JR is right, what'd be the appropriate site for the Webforms implementation of same?

### LINKED ISSUE #886: Support European comma decimal places in xpath number functions
### Expected Behaviour

Calling `number("3,5")` should evaluate to `3.5`. While this is not valid in the xpath spec it's a common format and supported by JavaRosa.

### Observed Behaviour

Currently we return NaN.

### Additional Context

- Raised in this issue: https://github.com/getodk/web-forms/issues/878#issuecomment-5373170779
- JavaRosa code: https://github.com/getodk/javarosa/blob/f8262bbd3f09b4ff4bef673d22c9df0876fb38d7/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java#L751-L753

garethbowen:
Closes getodk/web-forms#886
Closes getodk/web-forms#240

Unit test.
Manual test.
Comparison with Collect for...

```
number(1)
number('1')
number(1.2)
number('1.2')
number('1,23')
number('0x12')
number('abc')
number('   13   ')
number('.56')
```

The two implementations match in all cases except the hexadecimal `0x12` which WF parses to 18 but JavaRosa returns NaN. I'm comfortable that this is enough of an edge case to ignore.

This mostly matches the implementation in Collect.

Allows entering numbers with commas for decimals.

No.

### DIFF
--- a/packages/xpath/src/evaluations/StringEvaluation.ts
+++ b/packages/xpath/src/evaluations/StringEvaluation.ts
@@ -2,8 +2,6 @@ import type { XPathNode } from '../adapter/interface/XPathNode.ts';
 import type { LocationPathEvaluation } from './LocationPathEvaluation.ts';
 import { ValueEvaluation } from './ValueEvaluation.ts';
 
-const parseNumber = (value: string) => Number(value.replace(',', '.'));
-
 export class StringEvaluation<T extends XPathNode> extends ValueEvaluation<T, 'STRING'> {
   readonly type = 'STRING';
   readonly nodes = null;
@@ -20,7 +18,7 @@ export class StringEvaluation<T extends XPathNode> extends ValueEvaluation<T, 'S
     super();
 
     this.booleanValue = !isEmpty;
-    this.numberValue = isEmpty ? NaN : parseNumber(value);
+    this.numberValue = isEmpty ? NaN : Number(value);
     this.stringValue = value;
 
     const numberFunction = context.functions.getDefaultImplementation('number');
@@ -28,7 +26,7 @@ export class StringEvaluation<T extends XPathNode> extends ValueEvaluation<T, 'S
     if (isEmpty) {
       this.numberValue = NaN;
     } else if (numberFunction === null) {
-      this.numberValue = parseNumber(value);
+      this.numberValue = Number(value);
     } else {
       try {
         this.numberValue = numberFunction
@@ -39,7 +37,7 @@ export class StringEvaluation<T extends XPathNode> extends ValueEvaluation<T, 'S
           ])
           .toNumber();
       } catch {
-        this.numberValue = parseNumber(value);
+        this.numberValue = Number(value);
       }
     }
   }
--- a/packages/xpath/src/evaluations/StringEvaluation.ts
+++ b/packages/xpath/src/evaluations/StringEvaluation.ts
@@ -2,6 +2,8 @@ import type { XPathNode } from '../adapter/interface/XPathNode.ts';
 import type { LocationPathEvaluation } from './LocationPathEvaluation.ts';
 import { ValueEvaluation } from './ValueEvaluation.ts';
 
+const parseNumber = (value: string) => Number(value.replace(',', '.'));
+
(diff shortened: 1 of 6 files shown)


packages/xpath/src/evaluations/StringEvaluation.ts
```diff
@@ -2,6 +2,8 @@ import type { XPathNode } from '../adapter/interface/XPathNode.ts';
 import type { LocationPathEvaluation } from './LocationPathEvaluation.ts';
 import { ValueEvaluation } from './ValueEvaluation.ts';
 
+const parseNumber = (value: string) => Number(value.replace(',', '.'));
```
latin-panda: IMO, this is better in `functions/xforms/number.ts` because it's more like an ODK thing than pure XPath (`1,2` parses to `NaN`) and `StringEvaluation` is shared with the plain evaluator

latin-panda (approved):
Works as expected! 

I have a small comment for your consideration

packages/xpath/src/evaluations/StringEvaluation.ts
garethbowen: That's where I started too but it has a couple of problems... (a) we want to accept commas in other places (like sum) so you'd end up having to try and call this everywhere, and (b) the number function doesn't actually get the string representation, it just ends up with an evaluation and calls `toNumber` which ends up here. Instead of calling `toNumber` we could call `toString` and do this parsing but that would have to be done everywhere (problem (a)).

Do you have any ideas for how to get around these limitations?

latin-panda: I tried it the other way in my fork [886-alternative](https://github.com/latin-panda/test-central-frontend/commit/725844aa5e27bcc4d95254b161b80d489fc829cd) where I moved the parsing into functions/xforms/number.ts.  

> we want to accept commas in other places (like sum) so you'd end up having to try and call this everywhere
         
We don't need to call it everywhere. StringEvaluation already delegates the conversion to the default number function (the `getDefaultImplementation('number')` call in its constructor; same hook the date handling uses), so sum, comparisons, etc. already get it. I've copied your sum tests to try, and they passed.  

> the number function doesn't actually get the string representation, it just ends up with an evaluation and calls toNumber which ends up here

The number function gets the string. It looks at the argument's stringValue and handles the comma before it becomes a number.

What do you think?

garethbowen: I see what you mean! Thanks for the example. I was trying to "fix" it earlier in the function, but delaying it to be beside the date formatting makes sense.

PR #1845, opened 2026-08-31
feat(web-forms#461): enforce single submission on public access links

### LINKED ISSUE #461: Enforce single submission from browser state
### Description
To implement ODK Central's Single Public Link feature for Web Forms to enable one-time form submissions via a shareable URL. We need to implement "local persistence (single)" and offline capabilities in Web Forms.

---

### Checklist
  - [x] No
  - [ ] No
  - [x] No
  - [ ] No
---

### User Stories

- **As a** [user type],
  **I want** [functionality],
  **so that** [benefit].

---

### Design

- **Figma**: [Insert URL]

---

### References
- [ODK Central Docs](https://docs.getodk.org/central-submissions)
- [ODK Web Forms Feature Matrix](https://github.com/getodk/web-forms)

### Related Issues

- https://github.com/getodk/central/issues/1187

garethbowen:
Closes getodk/web-forms#461

Manual testing
e2e test added

The implementation is quite simple. The only bit I really debated was whether to use cookies (like Enketo does) or local storage. I chose local storage because of the better API.

None.

No.

### DIFF
--- a/apps/forms/src/components/web-form-renderer.vue
+++ b/apps/forms/src/components/web-form-renderer.vue
@@ -11,6 +11,7 @@ import Location from '../utils/location';
 import { getDeviceId } from '../utils/device-id';
 import { hideSpinner } from '../utils/spinner';
 import { deleteLastSaved, getLastSaved, setLastSaved } from '../utils/last-saved';
+import { hasSubmitted, setSubmitted } from '../utils/once-store';
 defineOptions({
   name: 'WebFormRenderer'
 });
@@ -37,7 +38,7 @@ interface PostPrimaryInstanceParams {
   deviceID?: string | undefined;
 }
 
-let clearForm:Function;
+let clearForm: Function;
 let submissionData: SubmissionData;
 
 const submissionResult:any = {};
@@ -97,7 +98,7 @@ const postPrimaryInstance = async (file:File) => {
   }
 };
 
-const isProblem = (data:any) => {
+const isProblem = (data: any) => {
   return data != null &&
     typeof data === 'object' &&
     typeof data.code === 'number' &&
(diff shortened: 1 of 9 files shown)


garethbowen:
@latin-panda I haven't given this enough time to settle in, but thought it was worth you having a look to see if you think it's worth including in 2026.3.0.

latin-panda:
I wasn't able to review this today, but I'll get it done tomorrow!

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -135,6 +136,10 @@ const handleResult = () => {
   // Success handler
   if (submissionResult.primaryInstanceResult.success && attachmentResultArr.every(r => r.success)) {
 
+    if (props.form.once && props.form.enketoOnceId) {
+      setSubmitted(props.form.enketoOnceId);
```
latin-panda: If the form data is sent but an attachment upload fails, and the user closes the tab instead of retrying. The form data is saved at this point, and the flag is never set. They can submit the whole form again later. Could we call setSubmitted right after the primary instance succeeds, instead of after everything succeeds?

latin-panda (approved):
The solution looks good (local storage over cookies) and works well. I didn't explore it in depth, but reloading the same link, navigating the browser history, and deleting local storage all work fine. 

Just one comment on timing when uploading attachments, but approving now to unblock since it seems like a one-line change.

apps/forms/src/components/web-form-renderer.vue
garethbowen: Yes that works for me. It'll have to move again if/when we support offline submission...

PR #1848, opened 2026-08-31
feat(web-forms#515): add quick appearance

### LINKED ISSUE #515: "quick" appearance for map and select questions to automatically flip page
Blocked by https://github.com/getodk/web-forms/issues/254

#### Description
Add support for `quick` appearance functionality when form is displaying one question at the time, to allow users to rapidly get to the next question in the form.

---

#### Checklist
  - [x] No
  - [x] No
  - [x] No
  - [x] No

---

#### User Stories
- **As a** map user,
  **I want** to quickly navigate to the next question
  **so that** I can fill out the form in agile way

---

#### Design
- **Figma**: N/A
- **Design Docs**: N/A

---

#### Dependencies
- https://github.com/getodk/web-forms/issues/254

#### Related Issues
- https://github.com/getodk/web-forms/issues/305

latin-panda:
Closes https://github.com/getodk/web-forms/issues/515

Compared with Collect. All cases with select and pagination were manually tested. Unit tests were added too.

Appearances are a client concern, so the implementation lives in the client, reusing the existing pagination API (nextPage)

In paginated forms, answering a `quick` select1 now auto-advances to the next page. No advance for `likert`/`label` variants or when the question has invalid answers. 

No

### DIFF
--- a/packages/web-forms/src/lib/pagination/pagination.ts
+++ b/packages/web-forms/src/lib/pagination/pagination.ts
@@ -1,4 +1,8 @@
-import type { AnyControlNode, RepeatRangeUncontrolledNode, SelectNode } from '@getodk/xforms-engine';
+import type {
+  AnyControlNode,
+  RepeatRangeUncontrolledNode,
+  SelectNode,
+} from '@getodk/xforms-engine';
 
 export const isOnCurrentPage = (node: AnyControlNode | RepeatRangeUncontrolledNode): boolean => {
   if (!node.root.isPaginated) {
@@ -10,7 +14,10 @@ export const isOnCurrentPage = (node: AnyControlNode | RepeatRangeUncontrolledNo
 
 export const advanceIfQuick = (question: SelectNode) => {
   const { appearances } = question;
-  if ((appearances.quick || appearances.quickcompact) && !appearances.label && !appearances.likert) {
+  const isQuick = appearances.quick || appearances.quickcompact;
+  const supportsAutoAdvance = !appearances.label && !appearances.likert;
+
+  if (isQuick && supportsAutoAdvance) {
     question.root.nextPage();
   }
 };
--- a/packages/web-forms/src/lib/pagination/pagination.ts
+++ b/packages/web-forms/src/lib/pagination/pagination.ts
@@ -1,4 +1,4 @@
-import type { AnyControlNode, RepeatRangeUncontrolledNode } from '@getodk/xforms-engine';
+import type { AnyControlNode, RepeatRangeUncontrolledNode, SelectNode } from '@getodk/xforms-engine';
 
 export const isOnCurrentPage = (node: AnyControlNode | RepeatRangeUncontrolledNode): boolean => {
   if (!node.root.isPaginated) {
@@ -6,4 +6,11 @@ export const isOnCurrentPage = (node: AnyControlNode | RepeatRangeUncontrolledNo
   }
   const currentPage = node.root.currentState.currentPage;
   return currentPage != null && node.currentState.pageBoundary === currentPage;
+};
+
+export const advanceIfQuick = (question: SelectNode) => {
+  const { appearances } = question;
+  if ((appearances.quick || appearances.quickcompact) && !appearances.label && !appearances.likert) {
+    question.root.nextPage();
+  }
 };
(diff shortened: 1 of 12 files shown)


latin-panda:
@garethbowen this is ready for review :)

packages/web-forms/src/lib/pagination/pagination.ts
```diff
@@ -7,3 +11,14 @@ export const isOnCurrentPage = (node: AnyControlNode | RepeatRangeUncontrolledNo
...
+export const advanceIfQuick = (question: SelectNode) => {
+  const { appearances } = question;
+  const isQuick = appearances.quick || appearances.quickcompact;
+  const supportsAutoAdvance = !appearances.label && !appearances.likert;
+  const isValid = question.validationState.violation == null;
+
+  if (isQuick && supportsAutoAdvance && isValid) {
+    question.root.nextPage();
+  }
+};
```
garethbowen: It's difficult to know when this is going to advance, because the logic about which types it applies to is split - some of it is here (likert and label), and some is in the controls which either call this or don't. So it's hard to be confident that it's happening only when it's supposed to.

It would be cleaner, I think, if every time a value was set this function was called, and then all the logic about whether to advance or not was contained inside this function.

I can think of three ways to do this...
1. Change all the controls so we pass in a setValue fn which actually sets the value on the control and calls this function. This will change a lot of code, so I don't like it much.
2. Move this code to the engine and change the setValue code to call it, or have a reactive update call it. This goes against your comment about quick being a client side concern, which is a downside, but in some ways it does make it well contained because engine already looks after the current page.
3. Have something in the client react to the value change and call this function. The problem is it would fire all the time and very seldom change the page, so I don't like this option either.

Personally my favourite option is (2) but what you've done is a close second so if you feel strongly about it then I'm ok with it. I just worry I'm going to change the code in the future and quick by forgetting to explicitly call advanceIfQuick...

garethbowen (commented):
Very cool.

I've left one comment inline.

I didn't get around to testing this yet, but do we have to guard against multiple questions on the page, for example, if the form isn't configured to use pages or the question is in a field-list? Or is this not possible? Or do we just expect the form designer to do the right thing?

packages/web-forms/src/lib/pagination/pagination.ts
latin-panda: I've moved the auto-advance logic into the select component and use events to emit the value change (we already do this in some component like AsyncMap), so it remains the client's concern and is in one place.

latin-panda:
> I didn't get around to testing this yet, but do we have to guard against multiple questions on the page, for example, if the form isn't configured to use pages or the question is in a field-list? Or is this not possible? Or do we just expect the form designer to do the right thing? 

Nothing happens if the form doesn't have pages; the `nextPage()` is a no-op there, so `quick` is ignored.                                                                                   
When the form has pages and other questions are located after the `quick` question, answering it will skip the other questions. I verified this is exactly what Collect does, and it may be intentional, and it's the form designer's responsibility.

Manually tested cases (compared against Collect):                                                                                                                                                  
                                                                                                                                                                                                     
  - [x] `quick` radios advance on tap                                                                                                                                                                
  - [x] `quick minimal` (dropdown) advances on pick                                                                                                                                                  
  - [x] `quick autocomplete` advances on pick                                                                                                                                                        
  - [x] `quick columns-n` / `columns-pack` advance on tap                                                                                                                                            
  - [x] `quick no-buttons` advances on tap                                                                                                                                                           
  - [x] `quickcompact` (deprecated name) advances on tap                                                                                                                                             
  - [x] `quick list` / `list-nolabel` advance on tap                                                                                                                                                 
  - [x] `quick likert` saves but does not advance                                                                                                                                                    
  - [x] `quick label` taps do nothing, no advance                                                                                                                                                    
  - [x] `quick` inside a field-list jumps past the whole screen                                                                                                                                      
  - [x] `quick` inside a repeat advances to the next page                                                                                                                                            
  - [x] select_multiple with `quick` never advances                                                                                                                                                  
  - [x] readonly `quick` does nothing                                                                                                                                                                
  - [x] a constraint violation blocks the advance and shows the error                                                                                                                                                                                                                                                                    
  - [x] `quick map` advances after saving a selection                                                                                                                                                                                                                                                                                              
  - [x] form without pages: `quick` does nothing

garethbowen (approved):
Yes! Great. Thanks.

PR #1849, opened 2026-08-31
EntityUpload: remember selection of extra properties

matthew-white:
This PR makes progress on getodk/central#1786. Right now, the selection of new properties in `EntityUploadExtraProperties` is not persisted when the user selects a new file. This PR makes it so that the selection is remembered up until the user closes the modal. Doing so makes it easier for the user to iterate on the upload file if there are errors or warnings.

New tests.

We already store the user's `selectedProperties`. Now we won't clear them on file select.

We'll pass `selectedProperties` to `EntityUploadExtraProperties` via its new `selected` prop. In some ways, we're actually able to simplify `EntityUploadExtraProperties` by adding this prop.

However, it's not convenient that in order for `selected` to reach `EntityUploadExtraProperties`, it has to go through the intermediary `EntityUploadWarnings` component first. This PR makes that easier by having the `EntityUpload` modal component render `EntityUploadExtraProperties` directly. That way, the modal can pass it props directly and handle events directly. There's no need for `EntityUploadWarnings` to get involved in the communication between the two components. `EntityUploadWarnings` still needs to contain `EntityUploadExtraProperties`, but it can be passed it as a slot rather than handling its rendering itself.

### DIFF
--- a/apps/central/test/components/entity/upload.spec.js
+++ b/apps/central/test/components/entity/upload.spec.js
@@ -648,6 +648,11 @@ describe('EntityUpload', () => {
     });
 
     const extraCSV = createCSV('label,height,circumference,species\ndogwood,1,2,dogwood\nelm');
+    const toggleExtra = (modal, name, checked = true) => {
+      const input = modal.get(`#entity-upload-extra-properties input[value="${name}"]`);
+      input.element.checked.should.equal(!checked);
+      return input.setChecked(checked);
+    };
 
     it('shows a warning if there are extra properties', async () => {
       const modal = await showModal();
@@ -677,12 +682,6 @@ describe('EntityUpload', () => {
 
     it('remembers the property selection until the modal is hidden', async () => {
       const modal = await showModal();
-      // Checks or unchecks the first property.
-      const toggleFirst = (checked = true) => {
-        const input = modal.get('#entity-upload-extra-properties .checkbox:nth-child(2) input');
-        input.element.checked.should.equal(!checked);
-        return input.setChecked(checked);
-      };
       const getSelected = () => {
         const { selected } = modal.getComponent(EntityUploadExtraProperties).props();
         return [...selected];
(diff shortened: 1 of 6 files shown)


apps/central/test/components/entity/upload.spec.js
```diff
@@ -672,6 +675,64 @@ describe('EntityUpload', () => {
...
+      const getTableExtra = () => {
+        const tables = modal.findAllComponents(EntityUploadTable);
+        tables.length.should.equal(2);
+        return tables[1].props().extraProperties ?? [];
+      };
+
+      await selectFile(modal, extraCSV);
+      await toggleFirst();
+
+      const secondCSV = createCSV('label,height,foo,bar\ndogwood,1,x,y');
```
matthew-white:
```suggestion
      const foobarCSV = createCSV('label,height,foo,bar\ndogwood,1,x,y');
```

apps/central/test/components/entity/upload.spec.js
```diff
@@ -672,6 +675,64 @@ describe('EntityUpload', () => {
...
+      getTableExtra().should.eql(['circumference']);
+
+      await selectFile(modal, secondCSV);
+      getSelected().should.eql(['circumference', 'foo']);
+      getChecked().should.eql(['foo']);
+      getTableExtra().should.eql(['foo']);
+      await toggleFirst(false);
+
+      await selectFile(modal, extraCSV);
+      await selectFile(modal, secondCSV);
```
matthew-white: Add a code comment explaining the intention here.

matthew-white (commented):

ktuite (approved):

apps/central/test/components/entity/upload.spec.js
matthew-white: Done in 7ed958933a63e66548743d16b6951cb18091e052

apps/central/test/components/entity/upload.spec.js
matthew-white: Done in 7ed958933a63e66548743d16b6951cb18091e052

PR #1850, opened 2026-08-31
Disable EntityUploadExtraProperties if CSV file has error

matthew-white:
This PR makes progress on getodk/central#1786. The Figma design shows that if the CSV file has an error, the checkboxes in `EntityUploadExtraProperties` should be disabled. Further, the text above the checkboxes should be a little different.

New tests.

### DIFF
--- a/apps/central/src/components/entity/upload/warnings.vue
+++ b/apps/central/src/components/entity/upload/warnings.vue
@@ -202,7 +202,7 @@ defineEmits(['rows']);
         "multiple": "Select which ones to create, otherwise they will be ignored."
       },
       "error": {
-        "one": "Once you’ve fixed all errors, you’ll be able to choose whether to create a new property.",
+        "one": "Once you’ve fixed all errors, you’ll be able to select the column.",
         // "Ones" refers to "columns".
         "multiple": "Once you’ve fixed all errors, you’ll be able to select which ones to create."
       }
--- a/apps/central/src/components/entity/upload/warnings.vue
+++ b/apps/central/src/components/entity/upload/warnings.vue
@@ -88,11 +88,21 @@ except according to the terms contained in the LICENSE file.
       </template>
       <template #body>
         <p>
-          <template v-if="extraProperties.length === 1">
-            {{ $t('extraProperties.description.one') }}
+          <template v-if="errors === 0">
+            <template v-if="extraProperties.length === 1">
+              {{ $t('extraProperties.description.one') }}
+            </template>
+            <template v-else>
+              {{ $t('extraProperties.description.multiple') }}
+            </template>
           </template>
           <template v-else>
-            {{ $t('extraProperties.description.multiple') }}
+            <template v-if="extraProperties.length === 1">
+              {{ $tc('extraProperties.error.one', errors) }}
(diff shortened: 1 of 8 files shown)


apps/central/src/components/entity/upload/warnings.vue
```diff
@@ -188,6 +204,11 @@ defineEmits(['rows']);
         "one": "Select the column to create a new property, otherwise it will be ignored.",
         // "Ones" refers to "columns".
         "multiple": "Select which ones to create, otherwise they will be ignored."
+      },
+      "error": {
+        "one": "Once you’ve fixed the error, you’ll be able to choose whether to create a new property. | Once you’ve fixed the errors, you’ll be able to choose whether to create a new property.",
```
matthew-white: Let's reword this message so that we don't need to pluralize it based on the number of errors:

> Once you've fixed ***all*** errors

Then maybe we just pass a `hasError` prop rather than an `errors` count?

apps/central/test/components/entity/upload/extra-properties.spec.js
```diff
@@ -36,6 +36,17 @@ describe('EntityUploadExtraProperties', () => {
     });
   });
 
+  it('disables checkboxes if the disabled prop is true', () => {
+    const component = mountComponent({
+      props: { properties: ['foo', 'bar'], disabled: true }
+    });
+    const checkboxes = component.findAll('.checkbox');
```
matthew-white, line 43: Assert that `.length.should.equal(3)`.

ktuite (approved):

apps/central/test/components/entity/upload.spec.js
```diff
@@ -733,6 +733,15 @@ describe('EntityUpload', () => {
...
     });
 
+    it('disables selection if there is an error', async () => {
+      const modal = await showModal();
+      await selectFile(modal, createCSV('height,circumference\n1,2'));
+      modal.findComponent(EntityUploadErrors).exists().should.be.true;
+      const extraComponent = modal.getComponent(EntityUploadExtraProperties);
+      expect(extraComponent.props().properties).to.eql(['circumference']);
+      extraComponent.props().disabled.should.be.true;
+    });
```
matthew-white, line 768: Add an assertion that the `errors` prop is passed to `EntityUploadWarnings`.

matthew-white (commented):

apps/central/test/components/entity/upload/extra-properties.spec.js
matthew-white: Done in 28e11a8875cedede1e33ab364433b4d1eef09f5e

apps/central/src/components/entity/upload/warnings.vue
matthew-white: Done in cf910cd8dd426bbbd998632ff369455fc1e24c74

apps/central/test/components/entity/upload.spec.js
matthew-white: Done in cf910cd8dd426bbbd998632ff369455fc1e24c74

PR #1851, opened 2026-08-31
EntityUpload: create new properties and include their data in upload

matthew-white:
This PR makes progress on getodk/central#1786. Up until now, the user could see unknown/extra properties in their CSV file. They could select which ones to create. However, Frontend wouldn't actually send the requests to create the properties, and it wouldn't include the extra properties' data in the upload request to create new entities. This PR changes that, implementing the last main piece of getodk/central#1786.

New tests. There were a couple of places in particular that I tried to make sure we had test coverage:

1. Verifying what data is sent to Backend; including the correct subset of extra properties and merging their data with the data for the existing entity properties
    - In particular, `mergeDataWithExtra()` has a lot of branches. I tried to cover them all in at least one test though.
2. Error cases
    - What if one property is created, but not another
    - What if properties are created, but then the subsequent upload fails

**Performance**

I'm a little concerned about the memory footprint associated with the definition `entitiesToSend`: the fact that we're creating a whole new array for the entity data to send to Backend (which, depending on the property selection, may differ from the entity data we've parsed from the CSV). For that reason, I've tried to make `mergeDataWithExtra()` as efficient as possible:

- Don't create the new array or call `mergeDataWithExtra()` if there aren't any extra properties
- Within `mergeDataWithExtra()`, don't create new objects or iterate over properties unless necessary. If there's a well-defined case that will be reasonably common where we can return early with a simpler result, do that instead of doing the full merge of the `data` and `extra` objects.
  - Example 1: `propertiesToCreate.value == null`. The user uploaded a file with extra properties, but then didn't select any. Just use the `data` object rather than creating a new object to merge `data` and `extra`.
  - Example 2: `propertiesToCreate.value.length === warnings.value.extraProperties.length &&
    entity.data === noPropertyData`. The user uploaded a file with extra properties, then selected all the properties. The file had one or more entities that didn't have any data for existing entity properties (perhaps because the entity list doesn't have any properties yet). I think this will be common if the user uploads their entity data right after creating the entity list (before explicitly creating any entity properties).

**Error cases**

I also thought about error cases. Hopefully they aren't too common, but I think they will definitely happen, so it's important to cover them. For example, getodk/central#691 is still a live issue. You could imagine a user attempting an upload, successfully creating their new properties, but then seeing the actual upload request fail. The end state is that the properties have been created, but their entities haven't been created yet (or at least so it seems).

What I've implemented is that

### DIFF
--- a/apps/central/src/components/entity/upload.vue
+++ b/apps/central/src/components/entity/upload.vue
@@ -484,7 +484,6 @@ watch([errors, warnings, csvEntities], () => {
 });
 
 const hide = () => { emit('hide', createdProperties.size !== 0); };
-
 watch(() => props.state, (state) => {
   if (state) return;
   abortParse();
--- a/apps/central/src/components/entity/upload.vue
+++ b/apps/central/src/components/entity/upload.vue
@@ -11,7 +11,7 @@ except according to the terms contained in the LICENSE file.
 -->
 <template>
   <modal id="entity-upload" :state="state" :hideable="!uploading" :persistent="true" size="full"
-    backdrop @hide="$emit('hide')" @mutate="resizeColumnIfShown">
+    backdrop @hide="hide" @mutate="resizeColumnIfShown">
     <template #title>{{ $t('title') }}</template>
     <template #body>
       <div :class="{ backdrop: uploading }">
@@ -53,8 +53,8 @@ except according to the terms contained in the LICENSE file.
           @rows="showWarningRows">
           <template #extra-properties>
             <entity-upload-extra-properties :properties="warnings.extraProperties"
-              :selected="selectedProperties" :disabled="errors != null"
-              @toggle="toggleExtraProperty"/>
+              :selected="selectedProperties" :created="createdProperties"
+              :disabled="errors != null" @toggle="toggleExtraProperty"/>
           </template>
         </entity-upload-warnings>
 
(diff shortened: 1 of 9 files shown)


apps/central/src/components/entity/upload.vue
```diff
@@ -443,10 +483,18 @@ watch([errors, warnings, csvEntities], () => {
...
-    for (const table of tables) table.resetScroll();
-  }
+  if (state) return;
+  abortParse();
+  csvEntities.value = null;
+  fileMetadata.value = null;
+  errors.value = null;
+  warnings.value = null;
+  selectedProperties.clear();
+  createdProperties.clear();
```
matthew-white, line 495: Add a test that this is cleared out.

apps/central/src/components/entity/upload/extra-properties.vue
```diff
@@ -1,18 +1,29 @@
...
+      :class="{ disabled: disabled || createdAll }">
+      <label v-tooltip.sr-only>
+        <input type="checkbox" :checked="selectedAll"
+          :disabled="disabled || createdAll"
+          :aria-describedby="createdAll ? disabledMessageId('all') : null"
           data-select-all="true">
         <span>{{ $t('action.selectAll') }}</span>
       </label>
+      <p v-if="createdAll" :id="disabledMessageId('all')" class="sr-only">
+        {{ $t('created') }}
```
matthew-white: This message doesn't really make sense for the "Select all" checkbox. Let's either:

1. Use a separate message for "Select all"; or
2. Don't add a tooltip at all and just let the individual properties speak for themselves

matthew-white (commented):
Notes from interactive review

ktuite (approved):

apps/central/src/components/entity/upload/extra-properties.vue
matthew-white: I went with (1). Done in 0e71831919d308edbebeb2a25a6852f7ce6f6d29

apps/central/src/components/entity/upload.vue
matthew-white: Done in 40618edff5efb2e6a12fdce5e6d283a15e408e46

PR #1855, opened 2026-09-02
fix(web-forms#899): likert connector line and focus highlight 

### LINKED ISSUE #899: Likert broken connector line and blue focus box
### Expected Behaviour

The focus is only on the radio button, and the radio buttons are connected with a line. 

### Observed Behaviour

When it has few options, the likert connector line doesn't reach the radios, and when a likert radio receives focus the invisible option box shows a large blue background.  

### Reproduction Steps

Form (XML or XLSX file): 

[likert-variations.xml](https://github.com/user-attachments/files/31748732/likert-variations.xml)

Steps:
1. Load form
2. Use Tab key to set focus; see the blue box.
3. Observer those with few radios and how the line doesn't connect them. 

### Environment

- Browser version: all
- OS: all
- Web Forms version: 1.0.3

latin-panda:
Closes https://github.com/getodk/web-forms/issues/899

[likert-variations.xml](https://github.com/user-attachments/files/31748861/likert-variations.xml)

The line was drawn with calculations that only worked when the options were narrow. Now it's based on the radios, so it works no matter how many options there are or how wide they get. For the blue box, I only changed the likert styles, the highlight is correct for regular selects, so the shared styles stay untouched.

Purely visual fixes. Only likert questions are affected.

No


latin-panda:
@garethbowen this is ready for a review when you have a chance please :)

packages/web-forms/e2e/test-cases/visual/image-labels.test.ts-snapshots/likert-chromium.png
latin-panda, line 1: Tiny diff in placement after the css fix
Image

packages/web-forms/src/components/common/LikertWidget.vue
```diff
@@ -35,30 +35,31 @@ defineEmits(['change']);
...
-			right: max(20px, calc(50% - 20px));
+		&::before {
+			content: '';
+			position: absolute;
+			background-color: var(--odk-border-color);
+			height: 3px;
+			top: calc(var(--odk-spacing-m) + (var(--p-radiobutton-height) / 2));
+			transform: translateY(-50%);
+			inset-inline-start: calc(-1 * var(--odk-spacing-m));
+			inset-inline-end: calc(-1 * var(--odk-spacing-m));
```
latin-panda, line 48: The negative values extend each option's line segment into the margin between options, so each covers half the gap, and the segments join into one continuous line.

garethbowen (approved):
It works in my testing.

Thanks!

PR #1866, opened 2026-09-02
fix(web-forms#898): blocks users resubmitting single submission forms

### LINKED ISSUE #898: It is possible to navigate through form and send/edit it using keyboard after filling single submission public link
### Expected Behaviour

It should not be possible to naviagte through form using keyboard after filling single submission public link

### Observed Behaviour

It is possible to navigate through form and send/edit it using keyboard after filling single submission public link

### Reproduction Steps

Steps:
1. Go to any form e.g. All question types and create single submission public link
2. Open previously created public link
3. Fill form and tap "Send" button
4. When "Thanks for participation" dialog is open press "Tab" key few times until it will be focused on "String" question
5. Type anything in question
6. Use "Tab" key again until you move to "Send" button
7. When there is focus on "Send" button press enter

### Screenshots or Videos

https://jam.dev/c/b0f48c85-aad2-45db-b8d6-8edbf3055ff3

### Environment

- Browser version: Chrome, Firefox
- Web Forms version: 1.0.3

garethbowen:
Closes getodk/web-forms#898

Testing

I also considered removing the form from the page altogether, but that felt disruptive when this was likely to be unintentional, and the feature isn't designed to be bullet proof.

Another option would be to introduce another constant like `POST_SUBMIT__NEW_INSTANCE`, for example, `POST_SUBMIT__DISABLE` which could disable the submit button in WF, and also potentially disable all the input fields. In this case I preferred to keep the changes on the app side.

### DIFF
--- a/apps/forms/src/components/web-form-renderer.vue
+++ b/apps/forms/src/components/web-form-renderer.vue
@@ -143,7 +143,7 @@ const handleResult = () => {
   // Success handler
   if (submissionResult.primaryInstanceResult.success && attachmentResultArr.every(r => r.success)) {
 
-    clearForm();
+    clearForm(props.form.once ? {} : { next: POST_SUBMIT__NEW_INSTANCE });
 
     if (isPublicLink.value) {
       visibleModal.value = { type: 'thankYouModal', hideable: false };
@@ -226,7 +226,7 @@ const submitData = async () => {
   handleResult();
 };
 
-const initializeSubmissionState = (data:SubmissionData, clearFormCallback:Function) => {
+const initializeSubmissionState = (data: SubmissionData) => {
   submissionData = data;
 
   submissionResult.primaryInstanceResult = {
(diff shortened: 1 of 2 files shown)


apps/forms/src/components/web-form-renderer.vue
```diff
@@ -143,7 +143,7 @@ const handleResult = () => {
   // Success handler
   if (submissionResult.primaryInstanceResult.success && attachmentResultArr.every(r => r.success)) {
 
-    clearForm();
+    clearForm(props.form.once ? {} : { next: POST_SUBMIT__NEW_INSTANCE });
```
garethbowen, line 146: The form no longer clears when it's a single submission. I think this is an improvement because the only reason to clear the form is to get it ready for the next submission. This also means that the instanceId doesn't get reset, so it provides some defence against submitting a second time.

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -279,7 +275,13 @@ const handleSubmit = async (
     // hence this branch should never execute.
     return;
   }
-  initializeSubmissionState(data as unknown as SubmissionData, clearFormCallback);
+  if (alreadySubmittedOnce()) {
```
garethbowen, line 278: However if the user _does_ submit again then this makes it a no-op so they don't get an ugly 409 conflict response.

apps/forms/src/components/web-form-renderer.vue
lognaturel: Wonderful, I came to ask whether we could do this as well. I was demoing this earlier today and it was very confusing to see the form refresh behind the thank you message.

latin-panda (approved):
The solution look good. I couldn't submit twice as described in the issue. I can still navigate back in paginated forms, but no harm since they cannot submit.

PR #1867, opened 2026-09-02
fix(web-forms#901): makes active value options more subtle

### LINKED ISSUE #901: Web Forms focus styling on selects looks more selected than when it's actually selected
Web Forms puts focus on selects, which is a nice accommodation, but the current styling makes it look like an item is selected. I'm worried users will not know to click and assume the first item is selected.

Image

garethbowen:
Closes getodk/web-forms#901

### DIFF
--- a/.changeset/jolly-clocks-shop.md
+++ b/.changeset/jolly-clocks-shop.md
@@ -1,0 +1,5 @@
+---
+"@getodk/web-forms": patch
+---
+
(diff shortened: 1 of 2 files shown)


garethbowen:
@latin-panda I've asked Nicole for a UX review, but let me know if I've missed something else.

latin-panda (approved):
Thank you! I think it's okay. The selected vs focused are notoriously different.

PR #1869, opened 2026-09-03
Show toast in more cases after submitting iframe form

### LINKED ISSUE #2188: Show error toast in more cases after submitting iframe form for submission download
Filing an issue based on https://github.com/getodk/central/issues/2080#issuecomment-5055168590

The `checkForProblem()` method of the `SubmissionDownload` component in Frontend continually checks an iframe for the receipt of a Backend Problem response after the submission of a form. It checks the following:

- If the form is still in the DOM: do nothing; keep checking.
- If the page is blank: evidence of form submission success; do nothing; stop checking.
- If the text content of the page is valid Problem JSON: show the Problem message in a toast.
- If the text content of the page is not JSON at all: show an error in a toast.

However, there are a couple of other cases we could check for. In both cases, we should show a toast, stop checking, and `console.log()` more details.

1. Unable to access the iframe DOM
    - Currently it just stops checking for a Problem response. It doesn't show a toast.
2. The text of the content page is valid JSON, but not valid Backend Problem JSON
    - Currently it just stops checking for a Problem response. It doesn't show a toast.

matthew-white:
This PR is a work in progress that I started during development on v2026.3, when I was looking at getodk/central#2080. This PR closes getodk/central#2188.

The main left to do is to add more tests. I feel pretty good about the changes to src/ itself.

I've added one new test so far, but there's still more to test.

### DIFF
--- a/apps/central/src/components/submission/download.vue
+++ b/apps/central/src/components/submission/download.vue
@@ -265,21 +265,37 @@ export default {
       form.reset();
     },
     checkForProblem() {
-      const doc = this.$refs.iframe.contentWindow.document;
+      let doc;
+      try {
+        doc = this.$refs.iframe.contentWindow.document;
+      } catch (error) {
+        this.logger.error('[SubmissionDownload] Could not access iframe document');
+        this.logger.error(error);
+        this.redAlert.show(this.$t('alert.somethingWentWrong'));
+        return true;
+      }
+
       // If Backend returns a Problem, the iframe changes pages. However, if the
       // form submission is successful, it seems that the iframe does not change
       // pages, and the form remains on the page.
       if (doc.querySelector('form') != null || doc.body == null) return false;
+
       let problem;
+      const { textContent } = doc.body;
       try {
         // Note that the Problem may be wrapped in another element, for example,
         // a <pre> element.
-        problem = JSON.parse(doc.body.textContent);
-      } catch (e) {
-        this.logger.log(doc.body.textContent);
-        this.redAlert.show(this.$t('alert.parseError'));
+        problem = JSON.parse(textContent);
+      } catch (error) {}
+      if (isProblem(problem)) {
+        this.redAlert.show(problem.message);
+      } else {
+        this.logger.error('[SubmissionDownload]', problem == null
+          ? 'Response is not valid JSON'
+          : 'Response is not a Backend Problem');
+        this.logger.log(textContent);
+        this.redAlert.show(this.$t('alert.somethingWentWrong'));
       }
-      if (isProblem(problem)) this.redAlert.show(problem.message);
       return true;
     },
     decrypt(action) {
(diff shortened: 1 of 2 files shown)


apps/central/src/components/submission/download.vue
```diff
@@ -265,21 +265,37 @@ export default {
...
       // If Backend returns a Problem, the iframe changes pages. However, if the
       // form submission is successful, it seems that the iframe does not change
       // pages, and the form remains on the page.
       if (doc.querySelector('form') != null || doc.body == null) return false;
+
       let problem;
+      const { textContent } = doc.body;
       try {
         // Note that the Problem may be wrapped in another element, for example,
         // a <pre> element.
```
matthew-white, line 287: Maybe this comment should be moved above, above where I define `textContent`?

apps/central/src/components/submission/download.vue
```diff
@@ -417,7 +433,8 @@ $actions-padding-left: $label-icon-max-width + $margin-right-icon;
     "alert": {
       "unavailable": "The data download is not yet available. Please try again in a moment.",
       "submit": "Data download should begin soon. Once it begins, you can close this message. If it hasn’t started in 20 seconds, please try again.",
-      "parseError": "Something went wrong while requesting your data."
+      "parseError": "Something went wrong while requesting your data.",
+      "somethingWentWrong": "@:alert.parseError"
```
matthew-white, line 437: I want to rename `parseError` to `somethingWentWrong`. That's because I want to use the same message for more cases (the cases that this PR addresses). In other words, I just want to rekey the message. `@:` is Vue I18n syntax. We could also use `@transifexKey` instead. Usually we use `@transifexKey`, but `@:` is possible in this case because it's a rekey within the same file.

apps/central/test/components/submission/download.spec.js
```diff
@@ -555,6 +560,19 @@ describe('SubmissionDownload', () => {
     });
   });
 
+  // getodk/central#2080
+  it.skip('shows a red alert if iframe cannot be accessed after form submission', async () => {
+    const clock = sinon.useFakeTimers(Date.now());
+    const modal = await setup(event => {
+      event.preventDefault();
+      // TODO. Mock the iframe document so that it throws.
```
matthew-white, line 568: This is where I got stuck. I want to make it so that accessing `window.document` throws. Can I use `sinon.replaceGetter()` for that (specifying a getter that throws)? Or will that not work, since `window.document` isn't originally a getter?

apps/central/test/components/submission/download.spec.js
```diff
@@ -512,17 +512,22 @@ describe('SubmissionDownload', () => {
...
+          clock.tick(1000);
+          modal.should.alert('danger', 'Something went wrong while requesting your data.');
         });
-        modal.get('a').trigger('click');
-        await modal.setProps({ state: false });
-        clock.tick(1000);
-        modal.should.alert('danger', 'Something went wrong while requesting your data.');
       });
 
       it('logs the response', async () => {
```
matthew-white, line 533: I should probably update and/or copy this test as well. Both because there are new cases that do logging and because the existing case does more logging than before.

matthew-white (commented):
Notes about decisions I made and questions I have

PR #1870, opened 2026-09-03
Add single-select mode to multi-select dropdown for entities viewAs filter

### LINKED ISSUE #2171: Add styling for viewAs entity filter
This viewAs filter (from this [issue](https://github.com/getodk/central/issues/1792)) is functionally done but the style doesn't match the figma yet.

ktuite:
Closes https://github.com/getodk/central/issues/2171

Screenshot 2026-09-02 at 8 12 20 PM

### DIFF
--- a/apps/central/src/components/multiselect.vue
+++ b/apps/central/src/components/multiselect.vue
@@ -465,6 +465,13 @@ const emptyMessage = computed(() => (searchValue.value === ''
     }
   }
 
+  .display-value {
+    max-width: 200px;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
   .icon-angle-down {
     font-size: 16px;
     color: #555555;
@@ -539,8 +546,7 @@ const emptyMessage = computed(() => (searchValue.value === ''
     font-size: 14px;
     list-style: none;
     max-height: 250px;
-    overflow-x: hidden;
-    overflow-y: auto;
+    overflow: visible auto;
     padding-bottom: 3px;
     padding-left: 0;
     padding-top: $vpadding;
@@ -577,10 +583,6 @@ const emptyMessage = computed(() => (searchValue.value === ''
       &:hover { background-color: darken($color-action-light, 7%); }
     }
 
-    input:focus-visible + span {
-      box-shadow: $btn-focus-box-shadow;
-    }
-
     input[type="checkbox"] {
       margin-top: 0;
       margin-right: 5px;
@@ -603,6 +605,8 @@ const emptyMessage = computed(() => (searchValue.value === ''
 
   &.single .option-list {
     font-size: 12px;
+    overflow-x: hidden;
+    overflow-y: auto;
 
     li {
       box-sizing: border-box;
@@ -626,6 +630,10 @@ const emptyMessage = computed(() => (searchValue.value === ''
     }
 
     label { line-height: 16px; }
+
+    input:focus-visible + span {
+      box-shadow: $btn-focus-box-shadow;
+    }
   }
 
   &.single {
--- a/apps/central/src/components/multiselect.vue
+++ b/apps/central/src/components/multiselect.vue
@@ -62,9 +62,9 @@ except according to the terms contained in the LICENSE file.
                 'search-match': searchMatches.has(value),
                 selected: single && selected.has(value)
               }">
-              <div :class="single ? 'radio' : 'checkbox'">
+              <div :class="single ? 'single-select-option' : 'checkbox'">
                 <label>
-                  <input :type="single ? 'radio' : 'checkbox'" :name="single ? radioName : null"
+                  <input :type="single ? 'radio' : 'checkbox'" :name="single ? singleSelectName : null"
                     :class="{ 'sr-only': single }" :data-index="i"
                     :aria-describedby="description != null ? descriptionId(i) : null">
                   <span v-if="description == null" v-tooltip.text>{{ text }}</span>
@@ -188,7 +188,7 @@ const { i18n, buildMode } = inject('container');
 const idPrefix = `multiselect${id}`;
 id += 1;
 const descriptionId = (i) => `${idPrefix}-description${i}`;
-const radioName = `${idPrefix}-radio`;
+const singleSelectName = `${idPrefix}-single`;
 
 const optionList = ref(null);
 
@@ -564,11 +564,13 @@ const emptyMessage = computed(() => (searchValue.value === ''
       label { @include text-overflow-ellipsis; }
     }
 
-    .radio {
+    .single-select-option {
       display: block;
 
       label {
         @include text-overflow-ellipsis;
+        font-weight: normal;
+        margin-bottom: 0;
         width: 100%;
       }
     }
--- a/apps/central/src/components/multiselect.vue
+++ b/apps/central/src/components/multiselect.vue
@@ -49,7 +49,7 @@ except according to the terms contained in the LICENSE file.
       </li>
       <li v-else class="change-all single">
         <button type="button" class="btn btn-outlined" @click.prevent="clearSelection">
-          {{ clear }}
+          {{ none }}
         </button>
       </li>
       <li>
@@ -160,11 +160,7 @@ const props = defineProps({
   },
   none: {
     type: String,
-      required: false
-  },
-  clear: {
-    type: String,
-    required: false
+    required: true
   },
   search: {
     type: String,
--- a/apps/central/src/components/multiselect.vue
+++ b/apps/central/src/components/multiselect.vue
@@ -60,7 +60,7 @@ except according to the terms contained in the LICENSE file.
             <li v-for="({ value, key = value, text = value, description }, i) in options"
               :key="key" :class="{
                 'search-match': searchMatches.has(value),
-                selected: single && selected.has(value)
+                selected: single && selected.has(value),
               }">
               <div :class="single ? 'single-select-option' : 'checkbox'">
                 <label>
--- a/apps/central/src/components/multiselect.vue
+++ b/apps/central/src/components/multiselect.vue
@@ -10,7 +10,7 @@ including this file, may be copied, modified, propagated, or distributed
 except according to the terms contained in the LICENSE file.
 -->
 <template>
-  <dropdown ref="dropdownEl" tag="div" class="multiselect form-group"
+  <dropdown ref="dropdownEl" tag="div" class="multiselect form-group" :class="{ single }"
     placement="bottom-start" :close-on-menu-click="false" @show="onShow" @hide="onHide">
     <template #toggle="{ toggle, attrs }">
       <button v-bind="attrs" type="button" class="dropdown-trigger"
@@ -37,7 +37,7 @@ except according to the terms contained in the LICENSE file.
           </button>
         </div>
       </li>
-      <li class="change-all">
+      <li v-if="!single" class="change-all">
         <button type="button"
           class="btn btn-outlined select-all" @click.prevent="changeAll(true)">
           {{ all }}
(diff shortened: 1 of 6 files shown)


apps/central/src/components/multiselect.vue
```diff
@@ -47,16 +47,25 @@ except according to the terms contained in the LICENSE file.
...
         <ul ref="optionList" class="option-list"
           :class="{ 'shows-all': searchValue === '' }" @change="changeCheckbox">
           <template v-if="options != null">
             <!-- eslint-disable-next-line vue/object-curly-newline -->
             <li v-for="({ value, key = value, text = value, description }, i) in options"
-              :key="key" :class="{ 'search-match': searchMatches.has(value) }">
-              <div class="checkbox">
+              :key="key" :class="{
+                'search-match': searchMatches.has(value),
+                selected: single && selected.has(value),
```
ktuite, line 63: Make a computed property

apps/central/src/components/multiselect.vue
```diff
@@ -395,6 +412,13 @@ const changeAll = (selectAll) => {
   }
 };
 
+const clearSelection = () => {
```
ktuite, line 415: Is there a way we could reuse changeall instead of introducing a new function

apps/central/src/components/multiselect.vue
```diff
@@ -405,10 +429,17 @@ const selectOption = computed(() => {
   if (props.loading) return i18n.t('common.loading');
   if (props.options == null) return i18n.t('common.error');
   const { placeholder } = props;
-  return placeholder({
+  const payload = {
```
ktuite, line 432: rename payload to data

apps/central/src/components/multiselect.vue
```diff
@@ -405,10 +429,17 @@ const selectOption = computed(() => {
...
     selected: i18n.n(props.modelValue.length, 'default'),
     total: i18n.n(props.options.length, 'default')
-  });
+  };
+  if (props.single) {
+    const selectedOption = props.modelValue.length !== 0
+      ? props.options.find(option => option.value === props.modelValue[0])
+      : null;
+    payload.selectedText = selectedOption?.text ?? selectedOption?.value ?? null;
```
ktuite, line 440: possibly investigate

apps/central/src/components/multiselect.vue
```diff
@@ -434,6 +465,13 @@ const emptyMessage = computed(() => (searchValue.value === ''
     }
   }
 
+  .display-value {
```
ktuite, line 468: use text overflow ellipsis mixin here

apps/central/src/components/multiselect.vue
```diff
@@ -524,6 +564,27 @@ const emptyMessage = computed(() => (searchValue.value === ''
...
+    .single-select-option {
+      display: block;
+
+      label {
+        @include text-overflow-ellipsis;
+        font-weight: normal;
+        margin-bottom: 0;
+        width: 100%;
+      }
+    }
```
ktuite, line 576: strange to have it here too, also shouldn't be limited to just the single select case. 

matt says we are doing it but only with the .checkbox class.
could the checkbox class be reused?

apps/central/src/components/multiselect.vue
```diff
@@ -524,6 +564,27 @@ const emptyMessage = computed(() => (searchValue.value === ''
...
+
+    li.selected {
+      background-color: $color-action-light;
+      margin-left: -$hpadding;
+      margin-right: -$hpadding;
+      padding-left: 2 * $hpadding;
+      padding-right: 2 * $hpadding;
+
+      &:hover { background-color: darken($color-action-light, 7%); }
```
ktuite, line 585: Not necessary

apps/central/src/components/multiselect.vue
```diff
@@ -544,6 +605,68 @@ const emptyMessage = computed(() => (searchValue.value === ''
...
+    overflow-y: auto;
+
+    li {
+      box-sizing: border-box;
+      // Override the max-width set above: in single-select mode, rows (and
+      // especially the selected row's highlight) should track the menu's
+      // actual width rather than being capped, or the highlight can fall
+      // short of the row's edge when the menu's width changes.
+      max-width: none;
+      height: 30px;
```
ktuite, line 620: we don't usually hardcode heights

apps/central/src/components/multiselect.vue
```diff
@@ -544,6 +605,68 @@ const emptyMessage = computed(() => (searchValue.value === ''
...
+      &:not(.empty-message):hover { background-color: #e6e6e6; }
+    }
+
+    li.selected {
+      margin-left: -8px;
+      margin-right: -8px;
+      padding-left: 16px;
+      padding-right: 16px;
+      &:hover { background-color: $color-action-light; }
+    }
```
ktuite, line 632: is this duplicate styling with above styling?

apps/central/src/components/multiselect.vue
```diff
@@ -544,6 +605,68 @@ const emptyMessage = computed(() => (searchValue.value === ''
...
+
+    label { line-height: 16px; }
```
ktuite, line 634: rm

apps/central/src/components/multiselect.vue
```diff
@@ -544,6 +605,68 @@ const emptyMessage = computed(() => (searchValue.value === ''
...
+
+    input:focus-visible + span {
+      box-shadow: $btn-focus-box-shadow;
+    }
+  }
+
+  &.single {
```
ktuite, line 641: double check more of these styles to see which ones are really needed

ktuite (commented):
Notes from interactive review

apps/central/src/components/multiselect.vue
```diff
@@ -137,15 +142,14 @@ const props = defineProps({
   // Text, including for form controls and actions
   label: {
     type: String,
-    required: true
```
ktuite: not sure what this one is about

apps/central/src/components/multiselect.vue
```diff
@@ -237,6 +241,9 @@ const syncWithModelValue = () => {
 };
 
 const changeCheckbox = ({ target }) => {
+  if (props.single) {
+    for (const value of [...selected]) change(value);
+  }
```
ktuite: Add a comment about turning off the one previously-selected value before changing the new one.

apps/central/src/components/multiselect.vue
```diff
@@ -405,9 +416,13 @@ const selectOption = computed(() => {
   if (props.loading) return i18n.t('common.loading');
   if (props.options == null) return i18n.t('common.error');
   const { placeholder } = props;
+  const singleSelectText = props.single
+    ? props.options.find(({ value }) => value === props.modelValue[0])?.text ?? null
```
ktuite: Code is too defensive in a weird way... the only important defensive case is when nothing is selected.
Change to check props.modelValue length first. If 0, don't search options at all. Otherwise, expect to find the matching option and just use its text.

apps/central/src/components/multiselect.vue
ktuite: Could a tooltip be added to this for really long selected text?

apps/central/src/components/multiselect.vue
```diff
@@ -499,6 +519,8 @@ const emptyMessage = computed(() => (searchValue.value === ''
       flex: 1 1 50%;
       line-height: 15px;
     }
+
+    &.single button { width: 100%; }
```
ktuite: Remove. It takes up the full width anyway.

apps/central/src/components/multiselect.vue
ktuite: Maybe move it back to it's own and leave a comment about how checkbox already has the font-weight: normal style.

apps/central/src/components/multiselect.vue
```diff
@@ -544,6 +571,49 @@ const emptyMessage = computed(() => (searchValue.value === ''
     }
   }
 
+  &.single .option-list {
+    overflow-x: hidden;
+    overflow-y: auto;
+
+    // Keep the radio in the DOM so it remains focusable for keyboard users,
+    // while hiding it visually.
```
ktuite: comment applies to input[type="radio"] section below

ktuite: Just use sr-only class on input instead

apps/central/src/components/multiselect.vue
```diff
@@ -53,10 +53,11 @@ except according to the terms contained in the LICENSE file.
           <template v-if="options != null">
             <!-- eslint-disable-next-line vue/object-curly-newline -->
             <li v-for="({ value, key = value, text = value, description }, i) in options"
-              :key="key" :class="{ 'search-match': searchMatches.has(value) }">
-              <div class="checkbox">
+              :key="key" :class="searchSelectionClasses(value)">
```
ktuite: CSS now uses `li:has(input:checked)` so the selected class doesn't seem to be used anymore.

apps/central/test/components/multiselect.spec.js
```diff
@@ -23,6 +23,10 @@ const assertChecked = (component, checked) => {
   const inputs = component.findAll('input[type="checkbox"]');
```
ktuite:
```suggestion
  const inputs = component.findAll('input[type="checkbox"],input[type="radio"]');
```

add this and then assertChecked can be used in place of assertSingleChecked

apps/central/test/components/entity/filters.spec.js
```diff
@@ -378,9 +378,11 @@ describe('EntityFilters', () => {
...
         app.vm.$container.requestData.dataset.entities.should.equal(2);
       })
-      .request(component => {
-        const { id } = testData.extendedFieldKeys.last();
-        component.get('#entity-filters-view-as select').setValue(String(id));
+      .request(async component => {
+        const filter = component.get('#entity-filters-view-as');
+        await filter.get('.dropdown-trigger').trigger('click');
+        await filter.find('input[type="radio"]').setValue(true);
+        await filter.get('.action-bar button').trigger('click');
```
ktuite: Could use `changeMultiselect` instead? Would need to update that code to select on radios too but would simplify these tests.

apps/central/test/components/entity/filters/view-as.spec.js
```diff
@@ -120,4 +116,22 @@ describe('EntityFiltersViewAs', () => {
...
+  });
+
+  it('does not open when disabled', async () => {
+    const component = mountComponent(createFieldKeys(1), {
+      props: { disabled: true },
+      attachTo: document.body
+    });
+    await toggle(component);
+    assertDisabled(component);
+  });
```
ktuite: component.findComponent('Multiselect').props.disabled should exist and have the right value/message
We know that Multiselect, once it gets that prop, it'll do the right thing.

apps/central/test/components/entity/filters/view-as.spec.js
```diff
@@ -120,4 +116,22 @@ describe('EntityFiltersViewAs', () => {
...
   });
+
+  it('filters app users by search text', async () => {
+    const [fieldKey1, fieldKey2] = createFieldKeys(2);
+    const component = mountComponent([fieldKey1, fieldKey2], { attachTo: document.body });
+    await toggle(component);
+    await component.get('.search input').setValue('0');
+    component.findAll('.search-match label').map(label => label.text())
+      .should.eql([fieldKey1.displayName]);
+  });
```
ktuite: Remove the test, trust the multiselect to test its own search.

apps/central/test/components/entity/filters/view-as.spec.js
```diff
@@ -22,77 +22,73 @@ const mountComponent = (fieldKeys, options) => {
...
-    component.get('select').element.value.should.equal(String(fieldKey.id));
+  it('renders a single-select input for each app user', async () => {
+    const fieldKeys = createFieldKeys(2);
+    const component = mountComponent(fieldKeys, { attachTo: document.body });
+    await toggle(component);
+    component.findAll('input[type="radio"]').length.should.equal(2);
   });
 
-  it('passes a new value for modelValue prop to the select', async () => {
+  it('shows and checks the app user specified by modelValue', async () => {
```
ktuite: Ok to remove this test, and with the disabled test.

apps/central/test/components/entity/filters/view-as.spec.js
ktuite: Or just remove this test because there's no extra logic to test.

ktuite (commented):
Interactive code review with Matt

matthew-white (approved):

PR #1871, opened 2026-09-03
Show viewAs only when useful and change text from "Viewing as" to "View as"

ktuite:
Request from @nicoleorlowski to change the text, also decided to only show the filter when it was going to be useful (e.g. when a dataset access filter is actually set).

Tests.


ktuite:
Requesting a review from copilot was accidental 😭 and now I can't stop it.

lognaturel (approved):

PR #1873, opened 2026-09-03
fix(web-forms#902): correct the paths to the locales directory

### LINKED ISSUE #902: Translation of component chrome no longer works
### Expected Behaviour

When loading a form with your browser language set to French the send button should be translated to Envoyer

### Observed Behaviour

The send button is in English

### Reproduction Steps

Form (XML or XLSX file):

Steps:
1. Set your browser language to French
2. Load any form

### Screenshots or Videos

Image

garethbowen:
Closes getodk/web-forms#902

This started failing when I made the change to build from src, not dist. Unfortunately this wasn't covered by tests so it didn't get picked up.

Wrote an integration test
Manual testing

### DIFF
--- a/packages/web-forms/src/lib/locale/useLocale.ts
+++ b/packages/web-forms/src/lib/locale/useLocale.ts
@@ -20,8 +20,9 @@ const availableTranslations = import.meta.glob<{ default: TransifexTranslation }
 );
 
 const getAvailableTranslationsKey = (locale: string) => {
-  // must match the import path above
-  return `../../../locales/strings_${locale}.json`;
+  return Object.keys(availableTranslations).find((path) => {
+    return path.endsWith(`/strings_${locale}.json`);
+  });
 };
 
 /**
@@ -49,8 +50,13 @@ const loadMessages = async (locale: string): Promise<ICUMessage> => {
     return enMessages;
   }
 
+  const key = getAvailableTranslationsKey(locale);
+  if (!key) {
+    return enMessages;
+  }
+
   try {
-    const raw = await availableTranslations[getAvailableTranslationsKey(locale)]!();
+    const raw = await availableTranslations[key]!();
     return { ...enMessages, ...normalizeMessages(raw.default) };
   } catch (error) {
     // eslint-disable-next-line no-console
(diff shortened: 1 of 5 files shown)


packages/web-forms/src/lib/locale/useLocale.ts
```diff
@@ -19,6 +19,11 @@ const availableTranslations = import.meta.glob<{ default: TransifexTranslation }
   '../../../locales/strings_*.json'
 );
 
+const getAvailableTranslationsKey = (locale: string) => {
+  // must match the import path above
+  return `../../../locales/strings_${locale}.json`;
+};
```
latin-panda, line 26: It can use availableTranslations, so there's no need to duplicate the path
```suggestion
 const getAvailableTranslationsKey = (locale: string) => {                                                                                                                                          
    return Object.keys(availableTranslations).find((path) => path.endsWith(`/strings_${locale}.json`));                                                                                                                                                                                              
  };  
```

packages/web-forms/src/lib/locale/useLocale.ts
```diff
@@ -45,7 +50,7 @@ const loadMessages = async (locale: string): Promise<ICUMessage> => {
   }
 
   try {
-    const raw = await availableTranslations[`/locales/strings_${locale}.json`]!();
+    const raw = await availableTranslations[getAvailableTranslationsKey(locale)]!();
```
latin-panda: If applying the suggestion from before, then this is

```suggestion
    const raw = getAvailableTranslationsKey(locale);
```

latin-panda (approved):
I didn't have any issues testing forms with languages. It's worth sending a note to the QA Team to explore more later. One suggestion below but approving to unblock

packages/web-forms/src/lib/locale/useLocale.ts
garethbowen: Done.

packages/web-forms/src/lib/locale/useLocale.ts
garethbowen: This suggestion didn't work, because the `getAvailableTranslationsKey` still just returns the key, not the raw translations, so I left this as is.

PR #1874, opened 2026-09-03
chore: updates translations for web forms

latin-panda:
verified translations

rutine task

no

no


garethbowen (approved):
Thanks!

PR #1875, opened 2026-09-03
fix(web-forms#900): block Next on pages with invalid questions

### LINKED ISSUE #900: Required questions in Web Forms do not block forward progress when using pages
When a question is required when using pages, you can still proceed.  I find this behavior surprising. 

@lognaturel says it was done on purpose. I think we should revisit this.

Relatedly, if in a required text question, you type a space and then delete it, you get the `This field is required.` message, but you can still hit next.

latin-panda:
Closes https://github.com/getodk/web-forms/issues/900

Executed test plan manually:

**Forms:** 

[geolocation-banner-manual.xml](https://github.com/user-attachments/files/31809617/geolocation-banner-manual.xml)
[pagination-21-constraint-behind.xml](https://github.com/user-attachments/files/31809618/pagination-21-constraint-behind.xml)
[pagination-19-required.xml](https://github.com/user-attachments/files/31809624/pagination-19-required.xml)
[2-simple-required.xml](https://github.com/user-attachments/files/31809638/2-simple-required.xml)

##### Paged form (`pagination-19-required.xml`)

- [x] First load: no banner, no highlights
- [x] Next on blank required question: stays on page, red highlight, banner says "1 question with error", question is focused
- [x] Fill the question: highlight and banner disappear immediately
- [x] Next: advances with no banner, no highlight
- [x] Next on the blank age: blocked again
- [x] Back while blocked: always works
- [x] Back with the error still behind you: banner stays visible
- [x] Type then clear age, navigate away and back: highlight persists
- [x] Close the banner: stays hidden until the next blocked Next or failed Send
- [x] Send with an invalid question on a previous page (use `pagination-21-constraint-behind.xml`): blocked, navigates back to it, shows its constraint message
- [x] Fill everything, Send: submits; form is clean afterwards

##### Non-paged form (`2-simple-required.xml`)

- [x] First load: no banner, no highlight
- [x] Send with the question blank: banner and highlight appear
- [x] Fill the question: both clear; empty it again: both return
- [x] Type then clear before ever pressing Send: banner and highlight appear together 

##### Geolocation (form with `odk:setgeopoint`)

- [x] Deny the location permission: banner shows the geolocation error
- [x] Allow the permission, re-trigger the fetch: banner clears

##### Regression

- [x] Paged form with no required/constraint questions: navigation unchanged
- [x] Likert selects never auto-advance
- [x] Non-relevant required questions never block Next
- [ ] Repeats: add/remove instances on pages **(pending team decision on add-instance gating)**

The form already keeps a live list of invalid questions, so we just filter it for the current page, no extra work. The banner and red highlights now follow the same rule (show an error once the user has seen the question), so they can never diverge. 
I considered blocking inside the engine, checking the DOM, and walking the whole form; all discarded as slower or fragile solution, this current solution has the best balance. 

Next now stops on a page with errors, shows them, and Back always works. And error highlights no longer disappear when you navigate away and back

No


latin-panda:
@garethbowen I have an open question about repeats (asked in WF Slack) that could be addressed in a follow-up PR if this one is good to send to QA. Let me know what you think!

garethbowen (approved):
Approved to get it in before code freeze. There's an e2e test fail but it looks like flake so I've kicked it off again.

You mentioned you considered blocking in the engine and that would be my preferred approach. In fact I think all of the web-forms packages pagination.ts could go in the engine?

It feels more intuitive to me because that's where the majority of the page handling lives now, but also it means that the "next" button and the "quick" appearance will both call the same validation code rather than being in two places.

Was the engine approach slow or fragile in some way I'm not getting?

If we _do_ decide to switch to the engine I'm comfortable with this being a subsequent change in the next release if necessary...

latin-panda:
Thanks @garethbowen for the review and flexibility! 

> Was the engine approach slow or fragile in some way I'm not getting?

There's no performance gain either way, and the solution gets a bit bigger if moved to the engine, so it's more about placement. I don't think everything can move, some things are better handled in the client, like which questions get error styling, showing/hiding the error banner, and focusing a specific question. So the engine would block and report (return the violations from `nextPage` or expose them reactively), and the client decides what to show. A nice thing about moving it is that we can revert the node getter I added to `DescendantNodeViolationReference`, keeping that interface a bit slimmer.                                                                                                                                                                    

In this PR I aimed to keep it contained, but when fixing the repeats "add" button (to block nav) I'll move some of this into the engine and we can see how it feels together. 

Merging now, and I'll do that in a follow-up PR.

PR #1878, opened 2026-09-04
EntityUpload: hide table for CSV data until there is data

matthew-white:
This PR addresses https://github.com/getodk/central/issues/1904#issuecomment-5520855236, specifically:

> It looks like the second table should be hidden until the upload file has been selected.

More precisely: the second table (the table for the CSV data) should be hidden until data has been parsed without error.

You can see this in action here:

https://github.com/user-attachments/assets/6a432d55-ef7f-458d-b638-fbe1f52917d2

I tried it out locally. I also updated an existing test.

Mostly it's just a matter of wrapping the second table in a `v-show`.

The margin between elements in the modal is simplified enough by now that nothing in that area needs to change. In general, I don't think styles need to change for this: everything looks OK to me visually.

### DIFF
--- a/apps/central/src/components/entity/upload.vue
+++ b/apps/central/src/components/entity/upload.vue
@@ -31,18 +31,21 @@ except according to the terms contained in the LICENSE file.
             :count="serverPage.count" :size-options="pageSizeOptions"
             :spinner="serverEntities.awaitingResponse"/>
         </div>
-        <p class="entity-upload-section-title">{{ $t('newEntities') }}</p>
-        <div class="entity-upload-table-container panel panel-simple">
-          <div class="panel-heading">
-            <h1 class="panel-title">{{ $t('table.file') }}</h1>
-          </div>
-          <div class="panel-body">
-            <entity-upload-table :ref="setTable(1)" :entities="csvSlice"
-              :extra-properties="propertiesToCreate" :row-index="csvRow"
-              :page-size="csvPage.size" :highlighted="warningRows"/>
-            <pagination v-if="csvEntities != null" v-model:page="csvPage.page"
-              v-model:size="csvPage.size" :count="csvEntities.length"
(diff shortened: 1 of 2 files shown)


apps/central/src/components/entity/upload.vue
matthew-white, line 1: I recommend hiding whitespace during code review.

apps/central/test/components/entity/upload.spec.js
```diff
@@ -324,33 +324,61 @@ describe('EntityUpload', () => {
     });
   });
 
-  it('shows the data template link until there is CSV data', async () => {
+  it('renders correctly as the user moves through the flow', async () => {
```
matthew-white, line 327: It's the same logic for whether to show the table as it is for whether to show the link to download a data template. Given that, I think we can repurpose this test to be more general. In this test, we'll assert that the correct things are shown initially and as the user selects different types of files.

apps/central/test/components/entity/upload.spec.js
```diff
@@ -324,33 +324,61 @@ describe('EntityUpload', () => {
...
-      errors: modal.findComponent(EntityUploadErrors).exists(),
-      warnings: modal.findComponent(EntityUploadWarnings).exists()
-    });
+    const getState = () => {
+      let tableVisible = true;
+      try {
+        table.should.be.visible();
+      } catch (_) {
+        tableVisible = false;
+      }
```
matthew-white, line 338: `showModal()` doesn't `attachTo: document.body`, so we can't use Vue Test Utils' `isVisible()`. I didn't want to modify `showModal()` for this one test. If this sort of thing comes up again, maybe we could move this pattern into our own utility function.

matthew-white (commented):
Notes to facilitate async code review

ktuite (approved):

PR #1881, opened 2026-09-05
Check for receipt of project response before checking verbs

### LINKED ISSUE #2189: Deleting an Entity List property can cause an unhandled frontend error
#### Problem description

https://getodk.sentry.io/issues/7672147231/?project=4511508552220672
https://getodk.sentry.io/issues/7710275015/?project=4511508552220672

`TypeError: Cannot read properties of undefined (reading 'has')`

Sentry points to the dataset properties/overview bundle and shows the error occurring through Vue’s error handler around the property deletion flow.

#### URL of the page
https://sucden.getodk.cloud/projects/6/entity-lists/gh_parcel_creation/properties

Delete a property and then ... ?

#### Expected behavior

No crash logged.

#### Other notes (if any)

The dataset properties component contains this template expression:

`project.verbs.has('dataset.update')`

which would produce the reported error if project.verbs is undefined.

The deletion flow, `useRequest()` and `isProblem()` don't call `has`.

Deleting a property changes causes the component to re-render and it's possible that `project.verbs` is undefined during a rerender? Is there a general state-management issue here?

matthew-white:
This PR closes getodk/central#2189. It checks `project.dataExists` before trying to access `project.verbs`.

I wrote a test that fails without the code change. In the future, I think something like getodk/central#2198 would be a good way to catch this type of bug.

Once this PR is merged, we can see whether the error continues to appear in Sentry.

I could have added the check of `project.dataExists` on an ancestor element, but I think it makes sense to check it at its usage site. Also, maybe it's good to try to render as much of the page as possible once the response for the entity list has been received, even before the response for the project has been received.

I'll leave a few comments about changes I made to test/.


apps/central/test/util/request-data.js
```diff
@@ -21,11 +21,17 @@ the response or response data. For example:
 */
 export const setRequestData = (requestData, responsesOrData) => {
   for (const [name, responseOrData] of Object.entries(responsesOrData)) {
+    // Adding this to make it easy for an individual test to prevent setting
+    // requestData resources that are set by default.
+    if (responseOrData === undefined) continue; // eslint-disable-line no-continue
```
matthew-white: This change supports the following line in the test above:

```js
requestData: { project: undefined }
```

apps/central/test/util/axios.js
```diff
@@ -59,7 +59,7 @@ export const mockResponse = {
   // returned, with the object as the response data. (In that case, the response
   // will not have a `config` property.)
   of: (responseOrData) => {
-    if (typeof responseOrData === 'object' &&
+    if (typeof responseOrData === 'object' && responseOrData != null &&
```
matthew-white, line 62: This change isn't strictly necessary, but I was looking at this function while updating `setRequestData()`, which calls this function. `setRequestData()` can now receive values of `null` or `undefined`, and I wondered what would happen if they were passed to this function. Currently, this function tries to access properties on `responseOrData` after checking `typeof responseOrData === 'object'`, but I don't think that totally works, since `null` would also pass that `typeof` check.

apps/central/test/components/dataset/overview/dataset-properties.spec.js
```diff
@@ -113,6 +117,34 @@ describe('DatasetProperties', () => {
       component.findAll('.delete-button').length.should.be.eql(2);
     });
 
+    it('does not show a button for a project viewer', () => {
```
matthew-white, line 120: This test verifies existing behavior. I just noticed that there was no test for it while I was looking at these tests. It tests the same button that I modify in this PR.

apps/central/test/components/dataset/overview/dataset-properties.spec.js
```diff
@@ -113,6 +117,34 @@ describe('DatasetProperties', () => {
...
+        properties: [{ name: 'height' }]
+      });
+      const component = mountComponent({
+        container: {
+          requestData: { project: undefined }
+        }
+      });
+      component.find('.col-actions').should.be.true;
+      component.find('.delete-button').exists().should.be.false;
+      setRequestData(component.vm.$container.requestData, {
```
matthew-white, line 141: It's not typical for a test to call `setRequestData()` directly. Usually `requestData` is set automatically by `mount()` or `load()`. However, it's needed in this case. As I mentioned in the PR description, ultimately I think something like getodk/central#2198 would be a better way to catch bugs like this one. But since I'm making a code change, I thought it was best practice to add a test that fails without the change.

apps/central/test/components/dataset/overview/dataset-properties.spec.js
```diff
@@ -1,23 +1,27 @@
...
-const mountComponent = () => mount(DatasetProperties, {
-  container: {
-    router: mockRouter('/'),
-    requestData: {
-      project: testData.extendedProjects.last(),
-      dataset: testData.extendedDatasets.last()
+import { setRequestData } from '../../../util/request-data';
+
+const mountComponent = (options = undefined) =>
+  mount(DatasetProperties, mergeMountOptions(options, {
```
matthew-white, line 16: Adding `mergeMountOptions` so that we can prevent `project` from being set in a test below.

matthew-white (commented):
Adding notes to facilitate async code review

matthew-white:
This PR is ready for code review. I've tagged multiple people, but I'm only looking for a single checkmark. I think it'd be safe to merge this PR during regression testing: it doesn't need to happen today to make it into the release.

apps/central/test/util/request-data.js
matthew-white: Maybe move this line to another layer, maybe `mergeMountOptions()` or `createTestContainer()` or `testRequestData()`.

matthew-white (commented):

ktuite (approved):

apps/central/test/util/request-data.js
matthew-white: Done in 4a97e626b79793c5a74119aa956741bceafecffd

PR #1882, opened 2026-09-06
Do not upload entity CSV data for deleted property

### LINKED ISSUE #2200: Entity upload attempts to upload data for deleted property
#### Problem description

Even after being deleted, an entity property is still partially remembered (in `dataset.propertyMap`) until the page is refreshed or the `dataset` resource is otherwise refreshed.

1. Create an entity list.
2. Add new properties: `height` and `circumference`
3. Delete `circumference`
4. Without refreshing the page, navigate to the Entities page
5. Click "Upload Entities"
6. Select a CSV file that contains a column for `circumference`
7. Observe that a warning is shown for `circumference`: it is correctly identified in the warning as a nonexistent property.
8. Don't check the checkbox for `circumference`. The intention is to not recreate the property.
9. Click "Append Entities"
10. Observe that an error message is shown: "The entity is invalid. You specified the dataset property [circumference] which does not exist."

#### Expected behavior

Frontend should not attempt to upload data for a property after it has been deleted.

matthew-white:
This PR closes getodk/central#2200. I noticed that issue when looking at the `DatasetProperties` component while working on getodk/central#2189.

I wrote a new test that fails without the code change.

Here's the problematic line that I noticed in `DatasetProperties`:

https://github.com/getodk/central-frontend/blob/f1e85b00d2d1db4c9a4e7f2f013126d2a0cdfe2b/apps/central/src/components/dataset/overview/dataset-properties.vue#L135

It's an issue because `EntityUpload` assumes that any change to `properties` happens via a call to `dataset.replaceData()`. That method was introduced in #1685 as a way to normalize `dataset.properties` whenever it changes. `EntityUpload` relies on `dataset.propertyMap`, which is recalculated whenever `dataset.replaceData()` is called:

https://github.com/getodk/central-frontend/blob/f1e85b00d2d1db4c9a4e7f2f013126d2a0cdfe2b/apps/central/src/request-data/resources.js#L109-L113

However, the line in `DatasetProperties` replaces `dataset.properties` without calling `dataset.replaceData()`.

I thought about changing the line in `DatasetProperties` to call `dataset.replaceData()`, but honestly that line likely predates even #1685. It's filtering the existing list of properties, not adding a new property that needs to be normalized, so it doesn't seem unreasonable to me for it to replace `dataset.properties` directly. It's common for code to modify `requestData` resources like that; it's less common to need to call a method like `dataset.replaceData()`.

Instead of using `dataset.replaceData()`, I changed how `dataset.propertyMap` is calculated. Instead of doing so imperatively in `dataset.replaceData()`, we can use a Vue computed ref. `dataset.properties` is reactive, so we can define a computed ref for `dataset.propertyMap` that should just work.

### DIFF
--- a/apps/central/src/request-data/resources.js
+++ b/apps/central/src/request-data/resources.js
@@ -106,12 +106,6 @@ export default (container, createResource) => {
       // it here.
       if (!('accessFilter' in data)) Object.assign(data, { accessFilter: null });
 
-      if (data.properties != null) {
-        data.propertyMap = new Map();
-        for (const property of data.properties)
-          data.propertyMap.set(property.name, property);
-      }
-
       return data;
     };
     /* eslint-enable no-param-reassign */
@@ -125,6 +119,10 @@ export default (container, createResource) => {
       replaceData: (data) => {
         Object.assign(dataset.data, transformData(data));
       },
+      propertyMap: computeIfExists(() => dataset.properties.reduce(
+        (map, property) => map.set(property.name, property),
+        new Map()
+      )),
       hasGeometry: computeIfExists(() =>
         dataset.properties.some(({ name }) => name === 'geometry'))
     };
--- a/apps/central/test/components/entity/upload.spec.js
+++ b/apps/central/test/components/entity/upload.spec.js
@@ -1136,4 +1136,42 @@ describe('EntityUpload', () => {
           }));
     });
   });
+
+  it('does not upload data for a deleted property', () => {
+    mockLogin();
+    testData.extendedDatasets.createPast(1, {
+      properties: [{ name: 'height' }, { name: 'circumference' }]
+    });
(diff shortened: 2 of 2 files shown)


matthew-white:
I tagged both @sadiqkhoja and @lognaturel for code review, but I'm only looking for a single checkmark. @sadiqkhoja is familiar with `dataset.replaceData()` from #1685, so he'd be a good person to review. I'm only tagging @lognaturel in case she has time to review it today before I update the QA server. I feel like the PR description should give enough context about what's going on. I don't feel like this change needs to be merged before regression testing: it's a minimal change that should be safe to include in the release either way.

lognaturel:
Looks reasonable to me but I agree that it's ok to merge as a bug fix so I'll let @sadiqkhoja take a real look.

sadiqkhoja (approved):
looks good to me

PR #1883, opened 2026-09-06
Remove unused displayNameLabel messages

### LINKED ISSUE #1350: Remove unused i18n messages
I've noticed a few unused i18n messages in the codebase. It'd be great to remove those messages, as unused messages lead to unnecessary work for translators.

At some point, it'd be ideal to automate the detection of unused messages: #829. However, even before that, we should remove any unused messages that we know about. Let's use this issue to track such messages.

- [x] `FieldKeyEdit`, `PublicLinkEdit`: `displayNameLabel` message
  - getodk/central-frontend#1883
- [x] `DatasetShow` component: `back` message
  - getodk/central-frontend#1804
- [x] `OdataLoadingMessage`: `middle` and `last` messages, maybe others

matthew-white:
This PR closes getodk/central#1350. It removes two unused i18n messages. I noticed these while working on getodk/central#2165.

I searched the codebase for the message key (`displayNameLabel`), but I didn't find any uses.

At some point, it'd be great to try to detect unused messages automatically. That's filed at getodk/central#829.

Should be no impact for users. It'll be one less thing for translators to translate.

### DIFF
--- a/apps/central/src/components/field-key/edit.vue
+++ b/apps/central/src/components/field-key/edit.vue
@@ -83,9 +83,7 @@ watch(() => props.state, (state) => {
   "en": {
     // This is the title at the top of a pop-up for editing an App User. {displayName} is the
     // the name of the App User
-    "title": "Edit App User “{displayName}”",
-    // Shown as label before the display name of the App User in the edit pop-up.
-    "displayNameLabel": "App User"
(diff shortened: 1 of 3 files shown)


matthew-white:
I've tagged multiple people for code review, but I'm just looking for a single checkmark to merge.

ktuite (approved):

PR #1876, opened 2026-09-03
Create new actor properties at same time as new actor

matthew-white:
Closes getodk/central#2165.

PR #1884, opened 2026-09-06
DatasetPropertyNew: show duplicate Problem inline

matthew-white:
This PR is a follow-up to #1876 (for getodk/central#2165). It is related to getodk/central#2173.

As of #1876, the `DatasetPropertyNew` modal will show an error inline if the user enters a property that is a case-insensitive duplicate of a property that is known to Frontend. However, if such a property is *unknown* to Frontend, having been created in a different tab or by a different user, Frontend will allow the request to go ahead, resulting in a Problem response from Backend, which Frontend will show in an ordinary alert at the top of the modal. This PR changes that so that if Frontend receives a Problem response about a duplicate property (either a 409.3 or a 409.24), it will indicate that in the same inline error. That way, the user will see the same error in the two cases: when the error is detected by Frontend and when it is detected by Backend.

New/updated tests. I also tried it out locally.

If Backend returns a Problem response, Frontend could refresh the entity list once the modal is hidden in order to retrieve the latest list of entity properties. I'd be happy to do that in a follow-up PR, but it also feels like maybe too much / unnecessary.

This shouldn't be a common case, so I don't feel like it's important for this PR to be merged to be released with v2026.3, which is about to start regression testing. Then again, I also don't feel like this PR is particularly risky. I'll discuss with @ktuite.

### DIFF
--- a/apps/central/src/components/dataset/property/new.vue
+++ b/apps/central/src/components/dataset/property/new.vue
@@ -18,8 +18,7 @@ except according to the terms contained in the LICENSE file.
         <p>{{ $t('introduction[0]') }}</p>
       </div>
       <form @submit.prevent="submit">
-        <property-input ref="input" v-model="name"
-          :properties="dataset.properties"/>
+        <property-input ref="input" v-model="name" :properties="propertyNames"/>
         <p>{{ $t('introduction[1]') }}</p>
         <div class="modal-actions">
           <button type="button" class="btn btn-link"
@@ -37,16 +36,15 @@ except according to the terms contained in the LICENSE file.
 </template>
 
 <script setup>
+import { computed, reactive, ref, watch } from 'vue';
 import { equals } from 'ramda';
-import { ref, watch } from 'vue';
-import { useI18n } from 'vue-i18n';
 
 import Modal from '../../modal.vue';
 import PropertyInput from '../../property-input.vue';
 import Spinner from '../../spinner.vue';
 
 import useRequest from '../../../composables/request';
-import { apiPaths } from '../../../util/request';
+import { apiPaths, isProblem } from '../../../util/request';
 import { useRequestData } from '../../../request-data';
 import { noop } from '../../../util/util';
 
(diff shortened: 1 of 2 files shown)


apps/central/src/components/dataset/property/new.vue
```diff
@@ -66,24 +64,30 @@ const emit = defineEmits(['hide', 'success']);
 
 const input = ref(null);
 const name = ref('');
+const duplicateNames = reactive([]);
```
matthew-white: Add a comment about what this is. It collects any duplicate names detected by Backend (from a 409.3 or 409.24).

apps/central/src/components/dataset/property/new.vue
```diff
@@ -66,24 +64,30 @@ const emit = defineEmits(['hide', 'success']);
...
+    data,
+    fulfillProblem: ({ code, details }) =>
+      (code === 409.3 && equals(details.fields, ['name', 'datasetId'])) ||
+      code === 409.24
   })
-    .then(() => {
-      emit('success');
+    .then(response => {
+      if (isProblem(response.data))
+        duplicateNames.push(data.name);
```
matthew-white, line 88: In the case of a 409.24, let's add the correct existing duplicate name to the list rather than the input name.

ktuite (approved):

matthew-white (commented):

matthew-white:
@ktuite and I discussed that we can wait until the release is out before merging this PR. Given that, I think I should also go ahead and add a changeset.

apps/central/src/components/dataset/property/new.vue
matthew-white: Done in 3ced444006549b63412fee90e1ef0b56e10f6545

apps/central/src/components/dataset/property/new.vue
matthew-white: Done in 2cf5dbab87b0c89380422ed557f6451abb43904b

matthew-white:
> Given that, I think I should also go ahead and add a changeset.

Done in bc6dc4d16334b13fa9bd82b3b1619b5f8b57fe2a

PR #1886, opened 2026-09-07
Add transifex:pull to Central

matthew-white:
This PR adds a new command to Central: `npm run transifex:pull`. This should make it easier to pull the latest translations for Central. It runs the `tx` CLI with the correct arguments and from the correct directory. Web Forms already has a `transifex:pull` command, so this PR is to add the command to Central as well.

Mostly I just copied the command from Web Forms to Central. However, I did make one change. Web Forms specifies `-a` to `tx`, but we don't want that for Central. We only want to download translations for existing locales, not new ones. We generally don't add new locales until they're sufficiently translated. E.g., if someone requests a new language in Transifex, but doesn't actually translate it, we don't want to download that translation file during a Transifex pull; we would just have to delete it. Certain steps are required when a new locale is added to Central, which are documented in CONTRIBUTING.md.

I tried running the new command, and it appeared to be successful. 🎉 I saw that the Transifex files were updated.

### DIFF
--- a/apps/central/docs/CONTRIBUTING.md
+++ b/apps/central/docs/CONTRIBUTING.md
@@ -177,11 +177,7 @@ Also note about comments:
 - `restructure.js` will automatically generate comments for any message whose path ends with `.full`, because such messages are used for component interpolation.
 - Use JSON with comments, but do not use other features of JSON5, which our workflow might not support.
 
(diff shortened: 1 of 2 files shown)


garethbowen (approved):
Nice one.

Let's also add `transifex:pull` in the top level package.json to do them all... something like...

`"transifex:pull": "npm run transifex:pull --workspaces --if-present",`

matthew-white:
> Let's also add `transifex:pull` in the top level package.json to do them all

Sounds good, I'll push a follow-up PR for that soon. :+1:

PR #1888, opened 2026-09-07
chore: run e2e tests in Playwright container

latin-panda:

Run CI multiple times with different commits

- Reproducible locally with Docker. Before, we had differences between local macOS and Linux vs GHA's runner.  
  - Downside: we must bump the Playwright image tag when upgrading @playwright/test. It fails loudly if missed, so we'll find out fast and fix it.  I consider it worth it because of the improvement made in this branch.                                                                                             
- Before, 6 jobs each downloaded browsers from a CDN, and a job was sometimes canceled when the download was slow. Now jobs use the Playwright container image, so there's no per-browser install step and less exposure to network flakiness.                                                                                                                                                                
- One shared cache action, cache keys computed once.     
- Removed the node-version matrix, bc it had a single entry and only fed the cache action. Node now comes from package.json (setup-node) and the Playwright image. This is what renames the check names (see below).                                                                                                                                                                                        
- It renames the e2e jobs (drops the node version from the name). The three old names (`Test web-forms e2e / Functional e2e (24.16.0, <browser>`) are required status checks in branch protection, so after this merges, rebased and newly created branches will show them stuck at "Expected" and can't merge until our GH Admin updates the required check names. The same would happen anyway on any future node-version bump, so it's worth fixing the naming now. 
   - To update branch protection, I think it's in Settings -> Branches -> required status checks. I'll also give the team a heads-up.

N/A

No


latin-panda:
@garethbowen Last week, I had an issue with checks timing out because the browser download took too long, even though the tests weren't broken. [I found in Playwright docs that they have containers](https://playwright.dev/docs/ci#via-containers). So far, it looks good to me. What do you think?

.github/actions/node-cache/action.yml
```diff
@@ -0,0 +1,82 @@
...
+      path: |
+        node_modules
+        apps/*/node_modules
+        packages/*/node_modules
+      key: ${{ steps.keys.outputs.install }}
+
+  - name: 'Install packages (when cache missed)'
+    shell: sh
+    if: ${{ inputs.mode == 'restore' && steps.cache-install.outputs.cache-hit != 'true' }}
+    run: npm ci --no-audit
```
garethbowen, line 82: Is this just a rename or are there other changes here too?

.github/workflows/wf-ci-e2e.yml
```diff
@@ -4,25 +4,53 @@ on:
...
+
+      - uses: 'actions/setup-node@v6'
+        with:
+          node-version-file: package.json
+          cache: 'npm'
+
+      - run: 'npm ci --no-audit'
+
+      # Builds only what the e2e tests need.
+      - run: 'npx turbo run build --filter=@getodk/web-forms'
```
garethbowen, line 27: The cache already has everything pre-built - why build again?

.github/workflows/wf-ci-e2e.yml
```diff
@@ -4,25 +4,53 @@ on:
...
+
+      - uses: ./.github/actions/node-cache
+        with:
+          mode: save
```
garethbowen, line 31: If I understand correctly this will save the dist to the node-cache, but you've only built the packages not the apps. Why save this state?

.github/workflows/wf-ci-e2e.yml
```diff
@@ -4,25 +4,53 @@ on:
...
+
   web-forms-e2e-functional:
     name: 'Functional e2e'
+    needs: web-forms-e2e-build
     runs-on: 'ubuntu-latest'
     timeout-minutes: 4
+    container:
+      # The image version must match the @playwright/test version in the root package-lock.json.
+      image: mcr.microsoft.com/playwright:v1.60.0-noble
+      options: --user 1001
```
garethbowen, line 41: Are we confident that downloading this image will be more reliable than downloading browsers directly?

.github/workflows/wf-ci.yml
```diff
@@ -25,24 +25,20 @@ jobs:
       fail-fast: false
       matrix:
         target: ['Node']
-        node-version: ['24.16.0']
```
garethbowen, line 28: The reason the matrix was left here was because Node 26 is going to LTS in October and we wanted to add it while still testing Node 24. In fact we should probably be testing against it now so we find anything before release.

garethbowen (changes requested):
Nice find!

To be confident of reliability changes like this we usually run the build multiple times (maybe 100). Have you tried that?

It looks like there are multiple threads here, not just the container change.

Does this also resolve https://github.com/getodk/central/issues/2060 ? @alxndrsn

alxndrsn:
> Does this also resolve [getodk/central#2060](https://github.com/getodk/central/issues/2060) ? @alxndrsn

It might very well side-step it, but it doesn't seem to right now:

https://github.com/getodk/central-frontend/actions/runs/34334779372/job/102523218655?pr=1888

```
2026-09-09T15:13:22.7806563Z [e2e-tester] Playwright: Version 1.60.0
2026-09-09T15:13:22.7807069Z [e2e-tester] Installing playwright deps...
2026-09-09T15:13:23.5450341Z Installing dependencies...
2026-09-09T15:13:23.5541478Z Switching to root user to install dependencies...
2026-09-09T15:13:23.6514198Z Get:1 file:/etc/apt/apt-mirrors.txt Mirrorlist [144 B]
2026-09-09T15:13:23.6958403Z Hit:2 http://azure.archive.ubuntu.com/ubuntu noble InRelease
2026-09-09T15:13:23.6978958Z Get:3 http://azure.archive.ubuntu.com/ubuntu noble-updates InRelease [126 kB]
2026-09-09T15:13:23.7040767Z Get:4 http://azure.archive.ubuntu.com/ubuntu noble-backports InRelease [126 kB]
2026-09-09T15:13:23.7079522Z Get:5 http://azure.archive.ubuntu.com/ubuntu noble-security InRelease [126 kB]
```

.github/actions/node-cache/action.yml
latin-panda: It adds a save mode in this file, so save and restore share the same keys. 
Node always comes from package.json. 
The cache paths are `apps/*/node_modules` and `packages/*/node_modules` instead of `**/node_modules`, which also matched nested folders and made the cache bigger. It also falls back to `npm ci` if the cache is missing, instead of failing and blocking people.

.github/workflows/wf-ci-e2e.yml
latin-panda: I tried removing it and the e2e jobs failed with a cache miss. I hit this before but forgot why, so I've added a comment (L7). The cache from `check-and-build` can't be restored inside the container because the runner compresses with zstd and the Playwright image doesn't have it.                                                                                                                    
So this job builds and saves the cache from inside the container. I looked into installing zstd instead, but it needs root in the container and an apt-get step in every e2e job, and I'm afraid that extra installation brings back the slowness this PR is trying to avoid.

.github/workflows/wf-ci-e2e.yml
latin-panda: Same reason [as above](https://github.com/getodk/central-frontend/pull/1888#discussion_r4018829371). It only builds web forms because that's all the e2e tests need, and it doesn't replace the full cache since the runner and the container caches are stored separately.

.github/workflows/wf-ci-e2e.yml
latin-panda: so far is good, no failures, I'm still running more times

.github/workflows/wf-ci.yml
latin-panda: During the last team call, we decided to remove the matrix

.github/actions/node-cache/action.yml
garethbowen: I'm not a fan of falling back if the cache is missing. My guess is that if we broke something and the cache never matched then this would be missed and the build would silently get slower. Is there a legitimate reason why it would be ok for the cache to miss in restore mode?

.github/workflows/wf-ci-e2e.yml
garethbowen: That's a pain... Ideally the entire CI would test the same build of the app for correctness as well as performance.

It looks like it is possible but messy to convince the cache to use gzip instead of zstd but this is fine. Another approach would be to use artifacts instead of cache for actual application build in the same way we do for releases, which would mean we can install branch code just as easily as we do final code. But let's leave that for another day.

packages/web-forms/tests/fixtures/3-notes-with-markdown.xml
```diff
@@ -178,7 +178,7 @@ Phone number: [<output value="/data/group/phone" />](tel:<output value="/data/gr
 
 Escaping: \*hello\*
 
-&lt;span style="color:red;font-family:cursive"&gt;signature&lt;/span&gt;
+&lt;span style="color:red;font-family:Liberation Serif,serif"&gt;signature&lt;/span&gt;
```
garethbowen, line 181: Why did this change? Was there no cursive font at all on the containers?

garethbowen (changes requested):
Almost there!

packages/web-forms/tests/fixtures/3-notes-with-markdown.xml
latin-panda: correct, it's not available

.github/actions/node-cache/action.yml
latin-panda: I believe I experienced a cache miss some time ago, but I can't recall the specifics or how it occurred. I removed the fallback. If it happens again, I will investigate further.

.github/workflows/wf-ci-e2e.yml
latin-panda: sounds good!

.github/workflows/wf-ci.yml
latin-panda: The xform-engine tests are still installing browsers and recently failed to install Chrome (`Run npx playwright install chromium --with-deps`). This branch fix could also be applied to the WF's packages, but I think it would be better as a follow-up PR.

It'd be nice if GHA had a way to download everything the CI needs (tools, images, browsers, etc...) without merging jobs. We have so many jobs downloading the same things.

latin-panda: I've been running at different times of day, and there have been no failures with this approach.

garethbowen (approved):
Thanks!

PR #1890, opened 2026-09-08
Remove skipAutoLogout route meta field

### LINKED ISSUE #2154: Remove skipAutoLogout mechanism
Frontend has a route meta field called `skipAutoLogout`. It is described as follows [here](https://github.com/getodk/central-frontend/blob/efedd8fd620d2ac992b961e1936bd5144f38463b/apps/central/src/routes.js#L82-L84):

> If `true`, no alert will be displayed when session is about to expire. Also user will be not be redirected to login page when session has expired.

This was useful when submissions were filled out within Central Frontend, i.e., when Central rendered the Web Forms component, before the addition of the separate Web Forms app. Given the new Web Forms app, I don't think this `skipAutoLogout` mechanism is used anymore on any of the remaining Central pages.

Given that the mechanism isn't used anymore, I think we can remove it from Frontend. Doing so would reduce code complexity related to login/logout.

matthew-white:
Closes getodk/central#2154.

Tests continue to pass.

There was discussion on the issue that this mechanism is no longer needed.

I searched the codebase for `skipAutoLogout` and removed the associated code.

If we ever need this mechanism again in the future, we can always add it back and revert this PR.

I think the risk of regression is probably low given (1) the minimal code changes and (2) the fact that `skipAutoLogout` is no longer used. However, given the importance of the code in src/util/session.js, I'm thinking I probably won't merge this PR until after regression testing is complete. In other words, this PR is intended for v2026.4, not v2026.3.

### DIFF
--- a/apps/central/src/routes.js
+++ b/apps/central/src/routes.js
@@ -79,10 +79,6 @@ The following meta fields are supported for bottom-level routes:
     However, NotFound requires neither: a user can navigate to NotFound whether
     they are logged in or anonymous.
 
-  - skipAutoLogout (default: false): If `true`, no alert will be displayed when
-    session is about to expire. Also user will be not be redirected to login
-    page when session has expired.
-
   requestData
   -----------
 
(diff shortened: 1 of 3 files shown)


sadiqkhoja (commented):
looks good to me

sadiqkhoja (approved):

PR #1891, opened 2026-09-08
fix(web-forms#910): resolve potential race condition when redirecting

### LINKED ISSUE #910: TypeError: undefined is not an object (evaluating 't.form.once')
Sentry Issue: [WEBFORMS-CLOUD-2H](https://getodk.sentry.io/issues/7719525547/?referrer=github_integration)

```
TypeError: undefined is not an object (evaluating 't.form.once')
  at re (/assets/web-form-renderer-AuocIkEM.js:2345:101709)
  at None (/assets/web-form-renderer-AuocIkEM.js:2345:101786)
  at _s (/assets/browserTracingIntegration-C93H1ad4.js:2:30024)
  at kt (/assets/browserTracingIntegration-C93H1ad4.js:2:30096)
  at None (/assets/browserTracingIntegration-C93H1ad4.js:4:7733)
...
(2 additional frame(s) were not displayed)
```

garethbowen:
Closes getodk/web-forms#910

Testing.

The redirect is async so would execute _after_ the synchronous code has run, including trying to render the form which may be null. By returning immediately we leave the loading state set to true so web-forms isn't rendered.

This probably wasn't a big deal because the redirect will still occur and the user probably won't even notice, but it's a waste of cycles and reports to sentry.

None.

None.

### DIFF
--- a/apps/forms/src/components/submission.vue
+++ b/apps/forms/src/components/submission.vue
@@ -135,7 +135,7 @@ const fetchFormConfig = async (): Promise<Form> => {
   throw new RequestError('Form not found', 404);
 };
 
-const fetchForm = async (): Promise<Form | undefined> => {
+const fetchForm = async (): Promise<Form> => {
   const formConfig = await fetchFormConfig();
 
   redirectEnketoUrls(formConfig);
@@ -158,11 +158,7 @@ const fetchForm = async (): Promise<Form | undefined> => {
       // editing in Enketo isn't supported
       throw new RequestError('Form not found', 404);
     }
-    if (offline.value) {
-      // TODO: Update once Web Forms has support for offline
-      window.location.replace(`/-/x/${formConfig.enketoId}${queryString(route.query)}`);
-      return;
-    }
+
     webFormsEnabled.value = false;
   }
   return formConfig;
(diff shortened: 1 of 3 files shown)


apps/forms/src/components/submission.vue
```diff
@@ -135,7 +135,7 @@ const fetchFormConfig = async (): Promise<Form> => {
   throw new RequestError('Form not found', 404);
 };
 
-const fetchForm = async (): Promise<Form | undefined> => {
+const fetchForm = async (): Promise<Form> => {
```
garethbowen, line 138: Removing the `undefined` option makes it clearer what's going on.

apps/forms/src/components/submission.vue
latin-panda, line 156: Before, a form with Web Forms enabled that was opened through an offline link was still rendering Web Forms. But after moving the `if(offline.value){...` code to L197, it seems to redirect to Enketo, correct?                                                                                                                                          
That feels surprising. There's nothing that explains why it suddenly renders Enketo, despite the Web Forms selection in the form settings.                                                           
If that's the intention, I think it needs a warning message somewhere here:

latin-panda (changes requested):
It looks good, just one comment below

apps/forms/src/components/submission.vue
garethbowen: Hmm... I'm not sure what the intended path is now. The comment implies to me that offline should always go to Enketo but I agree that's not what the code was doing before. I've updated the code so the behaviour is consistent 2026.2.

latin-panda (approved):
Looks good! Thanks!

PR #1892, opened 2026-09-09
Upgraded node to 24.20

### LINKED ISSUE #2212: Update Node in dev/CI
Now that #2120 is complete, Node has been updated for prod (to 24.20.0). However, we still need to finish updating it in dev/CI:

- `central-backend`
- `central-frontend`

sadiqkhoja:
Partially closes https://github.com/getodk/central/issues/2212

Sibling PR: https://github.com/getodk/central-backend/pull/1936


packages/web-forms/bin/package.json
matthew-white, line 1: Do we definitely want to bump the required Node version for Web Forms with each release? I don't remember what the policy is. @garethbowen, would you mind taking a look?

garethbowen: For the npm packages, if anyone else is using them, it would be very annoying to bump the version every time. And if we leave the "engines" version at 24.16 then I think we need to leave the CI at 24.16 to ensure we don't break anything.

The intention was to use the node matrix when 26 goes LTS next month so we would test 24.16 and 26.x.

It's fine to bump the volta version to be consistent with frontend so dev uses 20.20.

garethbowen (changes requested):
Thanks. I don't know if the WF policy is documented anywhere (will discuss in the next team meeting) but added a comment with my understanding.

packages/web-forms/bin/package.json
matthew-white: Is there a way to bump the top-level `engines` without bumping `engines` for the individual Web Forms packages? I.e., can we bump `engines` for Central Frontend and the `forms` app without bumping `engines` for the individual packages?

I feel like it's a valuable verification step to check that Central Frontend CI runs successfully under the version of Node that we're actually deploying in production.

Or will that level of verification be achieved once we move to the node matrix?

garethbowen: Never mind! I just met with the brains trust and we decided WF will align with the central process for node versions, until someone who is actually using the npm packages starts complaining.

garethbowen (approved):

matthew-white:
I'm going to go ahead and merge this just to make sure there's no conflict with #1899.

matthew-white:
I feel like @garethbowen's checkmark is sufficient to merge in terms of code review.

matthew-white:
I see that there was some issue with required checks not starting. Is it because the checks reference the old Node version `24.16.0`? E.g., "Test web-forms e2e / Functional e2e (24.16.0, chromium)". Maybe we need to update the list of required checks. 🤔

sadiqkhoja:
yup that's the reason. I thought it was github's glitch. For now I have removed `Test web-forms e2e / Functional e2e (24.16.0 *` from the required checks.

I wonder if we should remove the node version number from the action name.

PR #1894, opened 2026-09-10
fix(web-forms#909): stop recomputing form logic that cycles

### LINKED ISSUE #909: Crash after clicking on a checkbox
### Expected Behaviour

There should not be a crash after clicking on a checkbox

### Observed Behaviour

There is a crash after clicking on a checkbox

### Reproduction Steps

Steps:
1. Go to https://test.getodk.cloud/projects/655/forms/indefinite_repeat_and_remove_already_selected_from_species_list/submissions
2. Tap on New submission
3. Scroll down and click on "pas d'autre taxon" checkbox

### Environment
- Browser version: Chrome, Firefox, Safari
- OS: Linux, IOS
- Web Forms version: 1.0.03

latin-panda:
Closes https://github.com/getodk/web-forms/issues/909

<!-- 

-->

Manual tests:

- [x] Single select whose own answer filters its choice list (circular reference)
- [x] Same as before but with multi-select
- [x] Same, but with ranking question
- [x] Select filtered by an answer from another question
- [x] Cascading selects, country → city (regression)
- [x] Very large choice list, ~thousands of options (regression)
- [x] Multilingual form, switching language mid-answer (regression)

The form in the issue is incorrect; however, Web Forms should not break. As with Collect, the form remains usable and can be used to submit data. A proper way to implement that form's intention is this: 
[exclude_previous_repeats_test (1).xml](https://github.com/user-attachments/files/32074660/exclude_previous_repeats_test.1.xml)

Each select question's option list now has its own counter. It counts how many times that list changes during a single update (one answer edit), then resets when the update finishes. That is an arbitrary number; I didn't find a case that made sense where it could exceed this number. Open to suggestions.
When it reaches 100, the memo returns the list it already has, preferring the result with more options, so the person's answer stays visible and saved. If the same list is returned, so nothing changes, nothing re-runs, and the loop ends.                                                                                                                      
                                                                                                                                                                                                     
Other approaches considered (and failed at some point during testing):                                                                                                                                                        
                                                                                                                                                                                                     
- Restrict cycles in every expression globally (my first version). The form froze on values mid-update and caused incorrect states.
- Detect the cycle at parse time and refuse the form or show an error. Much bigger change, and it rejects any form with a self-reference that Collect accepts today. So their forms will stop working in Web Forms. 
- Remember visited states (history cache) to detect repeats. Too complicated to store every past result per expression and compare each new result for values of any type,  
  on a hot path shared by all fields. 
- Time-based (stop after X milliseconds) or form-size-scaled limits (limit = number of expressions in the form). Hacky and fragile.                                        

- A valid form doing 100+ real option-list changes in one single update 

No


latin-panda:
@garethbowen This isn't my favorite solution, but after exploring and failing with some of my solutions, it's the most contained one that works. Very open to better ideas!

packages/xforms-engine/src/lib/reactivity/createItemCollection.ts
```diff
@@ -103,14 +103,49 @@ interface ItemsetItem {
...
+    }
+    if (nodeListsEqual(result, previous)) {
+      return previous;
+    }
+    changeCount += 1;
+    if (changeCount <= MAX_CHANGES_PER_UPDATE) {
+      return result;
+    }
+    // Settle on the phase showing more options, so a filtered-out answer stays visible.
+    return result.length > previous.length ? result : previous;
```
latin-panda, line 136: I'd like to surface this to the user when the cycle is detected but doesn't block the form, show a warning on top of the form or under the field (when we extend the violations mechanism), for example: "This question has a circular computation. Please notify the person who sent you this form".

garethbowen: Yes that would be great. I've raised https://github.com/getodk/web-forms/issues/913 to revisit this.

garethbowen (approved):
Like you, I don't love this solution, but let's go with it and revisit when we look at getodk/web-forms#913

PR #1897, opened 2026-09-11
fix(web-forms#914): improves form definition error messages

### LINKED ISSUE #863: Form parsing failures are being reported to sentry
### Description

If a form has a typo in it then executing the function throws and a message is shown to the user, but it's also being reported to sentry, eg: `Function 'subsr' is not defined.`

We should filter out Errors that can be identified as user error to reduce spam in sentry.

### Reproduction

```xml
    <h:head>
        <h:title>invalid substring function</h:title>
        <model>
            <instance>
                <data id="unknownfn">
                    <meta>
                        <instanceID/>
                    </meta>
                    <name/>
                    <first/>
                </data>
            </instance>
            <bind nodeset="/data/meta/instanceID" type="string" readonly="true()" calculate="concat('uuid:', uuid())"/>
            <bind nodeset="/data/name" type="string"/>
            <bind nodeset="/data/first" type="string"/>
        </model>
    </h:head>
    <h:body>
        <input ref="/data/name">
            <label>Name</label>
            <setvalue event="xforms-value-changed" ref="/data/first" value="subs(/data/name, 0, 1)"/>
        </input>
        <input ref="/data/first">
            <label>First letter</label>
        </input>
    </h:body>
</h:html>
```

1. Load the form
2. Type something in the name field to trigger the function to execute

Image

### LINKED ISSUE #914: Show a helpful error message on invalid forms
### Expected Behaviour

When previewing a broken form the error message should be helpful enough to guide you to a solution.

### Observed Behaviour

The error message is next to useless, eg: `Argument 2 not allowed`

### Reproduction Steps

Form (XML or XLSX file):

```xml
    <h:head>
        <h:title>invalid coalesce</h:title>
        <model>
            <instance>
                <data id="invalidcoalesce">
                    <meta>
                        <instanceID/>
                    </meta>
                    <name/>
                </data>
            </instance>
            <bind nodeset="/data/meta/instanceID" type="string" readonly="true()" calculate="concat('uuid:', uuid())"/>
            <bind nodeset="/data/name" type="string" calculate="coalesce()"/>
        </model>
    </h:head>
    <h:body>
        <input ref="/data/name">
            <label>Name</label>
        </input>
    </h:body>
</h:html>
```

Steps:
1. Load the form

### Screenshots or Videos

Image

garethbowen:
Closes getodk/web-forms#914
Closes getodk/web-forms#863

Manual and CI testing
It was difficult to verify that error messages weren't propagating to sentry, but I've stopped them being logged to console so I'm 99% sure this will work.

It would be cleaner to use try/catch but it could be thrown in many places and this works as a catch all.

Hopefully just shows more useful error messages with less sentry spam.

No.

### DIFF
--- a/packages/xpath/src/evaluator/functions/FunctionImplementation.ts
+++ b/packages/xpath/src/evaluator/functions/FunctionImplementation.ts
@@ -13,31 +13,8 @@ import type { EvaluationContext } from '../../context/EvaluationContext.ts';
 import type { Evaluation } from '../../evaluations/Evaluation.ts';
 import type { EvaluationType, PrimitiveEvaluationType } from '../../evaluations/EvaluationType.ts';
 import { LocationPathEvaluation } from '../../evaluations/LocationPathEvaluation.ts';
-
-export class UnknownFunctionError extends Error {
-  constructor(functionName: string) {
-    super(`Unknown function ${functionName}`);
-  }
-}
-
-export class InvalidArgumentError extends Error {
-  constructor(argumentIndex: number, parameter: Parameter | null) {
-    if (parameter == null) {
-      super(`Argument ${argumentIndex} not allowed`);
-    } else {
-      const { typeHint } = parameter;
-
-      const causeMessage =
-        typeHint == null
-          ? `Expected argument at index ${argumentIndex}`
-          : `Expected argument compatible with type ${typeHint} at index ${argumentIndex}`;
-
-      super(`Invalid argument at index: ${argumentIndex}`, {
-        cause: new Error(causeMessage),
-      });
-    }
-  }
-}
+import { TooFewArgumentsError } from '../../error/TooFewArgumentsError.ts';
+import { TooManyArgumentsError } from '../../error/TooManyArgumentsError.ts';
 
 export type ParameterArityType = 'optional' | 'required' | 'variadic';
 
@@ -46,11 +23,7 @@ export type ParameterTypeHint =
   // as lazy? This would allow call-site optimizations that would
   // be more challenging or fussy to implement in the individual
   // `FunctionImplementation`.
-  | 'any' // TODO: naming? Could be 'unknown' or 'result' or...?
-  | 'boolean'
-  | 'node'
-  | 'number'
-  | 'string';
+  'any' | 'boolean' | 'node' | 'number' | 'string';
 
 export interface Parameter {
   readonly arityType: ParameterArityType;
@@ -58,8 +31,6 @@ export interface Parameter {
 }
 
 // TODO: this is the parameter signature, what about return? (partly addressed by `TypedFunction`)
-// TODO: is it possible to enforce order? I.e.:
-// [...RequiredParameter, ...OptionalParameter, ...([] | [VariadicParameter])]
 export type FunctionSignature = readonly Parameter[];
 
 // prettier-ignore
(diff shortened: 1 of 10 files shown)


packages/xpath/src/evaluator/functions/FunctionImplementation.ts
```diff
@@ -46,20 +23,14 @@ export type ParameterTypeHint =
   // as lazy? This would allow call-site optimizations that would
   // be more challenging or fussy to implement in the individual
   // `FunctionImplementation`.
-  | 'any' // TODO: naming? Could be 'unknown' or 'result' or...?
```
garethbowen, line 49: I removed this todo... the names are well established now.

packages/xpath/src/evaluator/functions/FunctionImplementation.ts
```diff
@@ -46,20 +23,14 @@ export type ParameterTypeHint =
...
+  'any' | 'boolean' | 'node' | 'number' | 'string';
 
 export interface Parameter {
   readonly arityType: ParameterArityType;
   readonly typeHint?: ParameterTypeHint;
 }
 
 // TODO: this is the parameter signature, what about return? (partly addressed by `TypedFunction`)
-// TODO: is it possible to enforce order? I.e.:
-// [...RequiredParameter, ...OptionalParameter, ...([] | [VariadicParameter])]
```
garethbowen, line 62: I don't think it is possible with a straight array.

packages/xpath/src/evaluator/functions/FunctionImplementation.ts
```diff
@@ -101,7 +72,6 @@ export class FunctionImplementation {
     readonly signature: FunctionSignature,
     runtimeImplementation: FunctionCallable | FunctionImplementation
   ) {
-    // TODO: *validate signature order!*
```
garethbowen, line 104: This is only useful when writing new functions which we hardly do any more so decided it wasn't worth writing this code, or executing it at runtime.

If we think it is worth it it'd probably better to do it as a unit test.

packages/xpath/src/evaluator/functions/FunctionImplementation.ts
```diff
@@ -154,27 +124,15 @@ export class FunctionImplementation {
...
-    }
-
-    for (const [index, parameter] of signature.entries()) {
-      // TODO: `typeHint` checking...
-      if (parameter.arityType !== 'required') {
-        break;
-      }
-
-      if (args[index] == null) {
-        throw new InvalidArgumentError(index, parameter);
```
garethbowen, line 176: Removed because it should never happen. We've already checked max and min which is set by during the init.

packages/xpath/src/evaluator/functions/FunctionImplementation.ts
```diff
@@ -154,27 +124,15 @@ export class FunctionImplementation {
...
-      throw new InvalidArgumentError(min, null);
+      throw new TooFewArgumentsError(this.localName, argumentCount, min);
     }
 
     if (argumentCount > max) {
-      throw new InvalidArgumentError(max, null);
-    }
-
-    for (const [index, parameter] of signature.entries()) {
-      // TODO: `typeHint` checking...
```
garethbowen, line 170: I don't think this is viable.

garethbowen:
@latin-panda This is ready for review. If you think it's safe we could sneak it in to 1.1.0 but otherwise let's hold off until after the release.

latin-panda (approved):
Looks good to me! Thank you!

PR #1898, opened 2026-09-11
feat(web-forms#855): emit an event when language is set

### LINKED ISSUE #855: Vue component should emit an event when the language is changed
### Description

The forms app attempts to pick the right language but it's limited because, a) it doesn't know anything about the languages configured in the xform, and b) it doesn't know when the user manually picks a language using the components language selector. This means the language of the app and component may get out of sync, particularly for first load, when the user hasn't set the locale in local storage yet.

Extend the API with an event that fires when the locale is set, both when the web-forms i18n code updates it, and when the user picks a language from the dropdown. Update the forms app to listen to this event and update its locale.

### LINKED ISSUE #912: Language set in a submission of a form is applied in another form
#### Problem description
A language set in a submission of a form is applied in another form  in the dialog after clicking "Send".
In the proper form the set language appears in "Send" button but not in the dialog after clicking "Send".
I reproduced it in French.
I've tried to reproduce it in Spanish and Polish but the issue doesn't occur.
I'm not sure if it's a Central or WF issue. If it's WF let me know and I'll file the issue in WF.

#### URL of the page
https://test.getodk.cloud/projects/746/forms/all-widgets/submissions
1. Open a new submission in https://test.getodk.cloud/projects/746/forms/all-widgets/submissions 
2. Change the language - set French.
3. Send the form (observe the language in Send button and the confirmation dialog after sending).
4. In the first tab open a new submission of any other form.
5. Send the submission.
6. Observe the language in Send button and the confirmation dialog after sending.
#### Screenshot
https://jam.dev/c/e21b91b7-5c1a-45eb-86ad-84e434bfe8ce

#### Expected behavior
The language should be applied in the correct instance.
#### Central version shown in version.txt
https://staging.getodk.cloud/
versions:
bc46bbcb80bcb7a0cc04eb57aaeecd3590414671 (v2026.2.4-24-gbc46bbc)
 7f40e6f565785da4a433e27f4f7902fa61b5560e client (v2026.3.0-beta.0-6-g7f40e6f5)
 710329246f3e472335b3346c72f340ab8c973913 server (v2026.2.0-68-g71032924)
https://test.getodk.cloud/
versions:
bc46bbcb80bcb7a0cc04eb57aaeecd3590414671 (v2026.2.4-24-gbc46bbc)
 0000000000000000000000000000000000000000 client (v2026.3.0-beta.0)
 a61b0acbc7d5d8e3bc381523e78e108e60fddfdb server (v2026.2.0-67-ga61b0acb)
#### Browser
Chrome, Firefox

garethbowen:
Closes getodk/web-forms#855
Closes getodk/web-forms#912

Manual testing

It's the simplest way I could think of.

It shouldn't do anything except it keeps both app and component in sync.

No.

### DIFF
--- a/.changeset/blue-parks-stick.md
+++ b/.changeset/blue-parks-stick.md
@@ -1,0 +1,6 @@
+---
+"@getodk/web-forms": minor
+"@getodk/forms": minor
+---
+
+Added component emit for language selection so app and component can stay in the same language
--- a/apps/forms/src/i18n.ts
+++ b/apps/forms/src/i18n.ts
@@ -75,6 +75,11 @@ addLocale('pt', 'Português');
 addLocale('zh', '汉语', '');
 addLocale('zh-Hant', '漢語', '');
 
+export const setLocale = (locale: string) => {
+  i18n.global.locale = locale;
+  document.documentElement.setAttribute('lang', locale);
+};
+
 const loadUsersLocale = () => {
   const locale = userLocale() ?? fallbackLocale;
   document.documentElement.setAttribute('lang', locale);
--- a/packages/web-forms/README.md
+++ b/packages/web-forms/README.md
@@ -385,6 +385,8 @@ The `<OdkWebForm>` component accepts the following props:
 The component emits the following events:
 
 - `loaded`: Emitted when the form is loaded and displayed to the user
+- `languageSelected`: Emitted when the form language is set, either when the form is loaded, or when the user selects an option from the language selector.
+  - Payload: ([language: string])
 - `submit`: Emitted when the user presses the "Send" button on a valid form
   - Payload: ([submissionPayload: MonolithicInstancePayload, callback: HostSubmissionResultCallback])
 - `submitChunked`: Emitted for chunked submissions when the form is valid
--- a/packages/web-forms/src/components/OdkWebForm.vue
+++ b/packages/web-forms/src/components/OdkWebForm.vue
@@ -111,6 +111,7 @@ const hostSubmissionResultCallbackFactory = (
 // eslint-disable-next-line @typescript-eslint/consistent-type-definitions -- evidently a type must be used for this to be assigned to a name (which we use!); as an interface, it won't satisfy the `Record` constraint of `defineEmits`.
 type OdkWebFormEmits = {
 	loaded: [],
+	languageSelected: [language: string],
 	submit: [submissionPayload: MonolithicInstancePayload, callback: HostSubmissionResultCallback];
 	submitChunked: [
 		submissionPayload: ChunkedInstancePayload,
@@ -223,9 +224,20 @@ const errorBannerDismissed = ref(false);
 const geolocationErrorMessage = ref<string | null>(null);
 const isFormEditMode = ref(false);
 provide(IS_FORM_EDIT_MODE, readonly(isFormEditMode));
-const { setLanguage, t } = useLocale(computed(() => state.value.root));
+const { setLanguage, getLanguage, t } = useLocale(computed(() => state.value.root));
 provide(TRANSLATE, t);
 const { navigateToFirstViolation, navigateToNode } = useNavigationTarget(() => state.value.root);
+
+if (isEmitSubscribed('onLanguageSelected')) {
+	const language = computed(() => getLanguage());
+	watch(
+		() => language.value,
+		(selected) => {
+			emit('languageSelected', selected);
+		},
+		{ immediate: true }
+	);
+}
 
 onErrorCaptured(err => {
 	runtimeError.value = FormInitializationError.from(err);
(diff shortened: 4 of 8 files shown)


packages/web-forms/src/lib/locale/useLocale.ts
```diff
@@ -195,7 +195,6 @@ export const useLocale = (formRef: Ref<RootNode | null>) => {
 
   const applyLocale = (candidates: string[], formBaseLocale?: string) => {
     const newContentLocale = formBaseLocale ?? FALLBACK;
-    document.documentElement.lang = newContentLocale;
```
garethbowen, line 198: I felt uncomfortable with the component setting the attribute on the app document. Now that we emit I think we can leave the app to do it.

garethbowen:
@latin-panda Absolutely no hurry on this but when you get around to it I think this will save a lot of confusion!

packages/web-forms/README.md
```diff
@@ -385,6 +385,8 @@ The `<OdkWebForm>` component accepts the following props:
 The component emits the following events:
 
 - `loaded`: Emitted when the form is loaded and displayed to the user
+- `languageSelected`: Emitted when the form language is set, either when the form is loaded, or when the user selects an option from the language selector.
```
latin-panda: It's better to explain a bit how it works, for example

> The language used by the component for its buttons and messages (e.g., "fr") may differ from the form's language. If no translation is provided, it defaults to "en".

packages/web-forms/src/components/OdkWebForm.vue
```diff
@@ -223,10 +224,21 @@ const errorBannerDismissed = ref(false);
...
 provide(IS_FORM_EDIT_MODE, readonly(isFormEditMode));
-const { setLanguage, t } = useLocale(computed(() => state.value.root));
+const { setLanguage, getLanguage, t } = useLocale(computed(() => state.value.root));
 provide(TRANSLATE, t);
 const { navigateToFirstViolation, navigateToNode } = useNavigationTarget(() => state.value.root);
 
+if (isEmitSubscribed('onLanguageSelected')) {
+	const language = computed(() => getLanguage());
+	watch(
+		() => language.value,
```
latin-panda: I think this should work: 
```suggestion
	watch(
		() => getLanguage()
```

apps/forms/src/i18n.ts
```diff
@@ -75,6 +75,11 @@ addLocale('pt', 'Português');
 addLocale('zh', '汉语', '');
 addLocale('zh-Hant', '漢語', '');
 
+export const setLocale = (locale: string) => {
```
latin-panda, line 78: The component has a map that we can use to validate: 
```suggestion
export const setLocale = (locale: string) => {
   if (!locales.has(locale)) { 
     return;
  }  
```

packages/web-forms/tests/lib/locale/useLocale.test.ts
```diff
@@ -35,99 +33,105 @@ describe('useLocale', () => {
...
-      expect(document.documentElement.lang).toBe('fr-CA');
+      expect(setLanguage).toHaveBeenCalledTimes(1);
+      expect(setLanguage).toHaveBeenCalledWith(canadian);
+      expect(localStorage.getItem(STORAGE_KEY)).toBe('fr-CA');
     });
   });
 
   describe('resolveLocale translation fallback', () => {
     it('loads base locale translations when exact locale has no translation file', async () => {
       document.documentElement.lang = 'fr';
```
latin-panda: this line can be removed too

latin-panda (approved):
Nice fix! Just minor comments below, approving now to unblock

packages/web-forms/README.md
garethbowen: Updated.

packages/web-forms/src/components/OdkWebForm.vue
garethbowen: Done.

apps/forms/src/i18n.ts
garethbowen: Thanks!

packages/web-forms/tests/lib/locale/useLocale.test.ts
garethbowen: Thanks!

.changeset/blue-parks-stick.md
```diff
@@ -0,0 +1,6 @@
+---
+"@getodk/web-forms": minor
+"@getodk/forms": minor
```
garethbowen, line 3: @latin-panda Technically this is a minor because it's adding a feature, but it's _also_ resolving a bug that we want to fix in 1.1.1. If I leave this as "minor" then the version will be bumped to 1.2.0. I suggest calling this a "patch" instead and sneaking a feature in to the 1.1.1 release. What do you think?

latin-panda: Yes, I agree, we can leave it as a "patch"

latin-panda (approved):
Looks good! Thanks

PR #1900, opened 2026-09-13
Move /version.txt check to useVersionMonitor()

### LINKED ISSUE #2202: Move /version.txt check to useVersionMonitor()
Right now, the regular check of /version.txt happens in the `App` component. However, I'd like to reduce the complexity of `App`. In my opinion, the component should be as simple as possible and mostly just tie together logic from other parts of the code. In getodk/central-frontend#1807 (the PR for #2073), I added a new `useVersionMonitor()` composable. My plan is to move the /version.check to the new composable in order to simplify `App`.

matthew-white:
This PR is a follow-up to #1807. In that PR, I added a new composable named `useVersionMonitor()` to listen for `vite:preloadError` events that indicate that Central has been updated. However, as I noted in that PR, we have another, older mechanism for identifying a Central update: the check of /version.txt in the `App` component. Now that we have the new composable though, I've been wanting to move the check of /version.txt from `App` to the composable:

- I think it makes sense to have all code related to checking for a Central update in the same place
- It's nice to simplify the `App` component
  - The `App` component is quite a bit simplified with this PR. After this, I think it will be ready to become a Composition API component.

This PR closes getodk/central#2202.

For the most part, I just copy/pasted code from one place to the other. I did make some small changes and added some comments. I'll leave comments about the changes I made.


apps/central/test/composables/version-monitor.spec.js
```diff
@@ -1,20 +1,212 @@
...
 import useVersionMonitor from '../../src/composables/version-monitor';
+import { logOut } from '../../src/util/session';
 
 import createTestContainer from '../util/container';
+import { mockHttp, load } from '../util/http';
+import { mockLogin } from '../util/session';
 import { withSetup } from '../util/lifecycle';
 
 describe('useVersionMonitor()', () => {
+  describe('change in /version.txt', () => {
```
matthew-white, line 12: I moved these tests as-is from app.spec.js. Other than the title of the top `describe()`, I didn't change anything in these tests.

apps/central/src/composables/version-monitor.js
```diff
@@ -1,12 +1,51 @@
...
 // been deployed. When it detects a likely version change, it either reloads the
 // page automatically or prompts the user to do so.
 export default () => {
-  const location = inject('location');
+  const { i18n, toast, location } = inject('container');
+  const { centralVersion } = useRequestData();
+  const { callWait } = useCallWait();
+
+  // Continually request /version.txt, checking for a change.
+  const checkVersion = async () => {
```
matthew-white, line 16: I changed this function to use `async`/`await`. In `App`, it was a promise chain. I think it's clearer this way.

apps/central/src/composables/version-monitor.js
```diff
@@ -1,12 +1,51 @@
...
+    if (previousVersion == null || centralVersion.versionText === previousVersion)
+      return false;
+
+    // Alert the user about the version change, then keep alerting them. One
+    // benefit of this approach is that the user should see the toast even if
+    // there is another toast (say, about session expiration).
+    callWait(
+      'versionChange',
+      () => {
+        toast.show(i18n.t('alert.versionChange'), { autoHide: false })
```
matthew-white, line 41: `App` used `alert`, but here I'm using `toast`. In general, I'm trying to use `toast` in more places.

apps/central/src/components/app.vue
```diff
@@ -76,41 +72,6 @@ export default {
...
-            () => {
-              this.alert.info(this.$t('alert.versionChange'))
-                .cta(this.$t('action.refreshPage'), () => { this.location.reload(); });
-            },
-            (count) => (count === 0 ? 0 : 60000)
-          );
-          return true;
-        })
-        // This error could be the result of logout, which will cancel all
-        // requests.
```
matthew-white, line 110: I didn't copy this comment to `useVersionMonitor()`, because I'm pretty sure it's out-of-date. `logOut()` shouldn't call `reset()` or `cancelRequest()` for the `centralVersion` resource:

https://github.com/getodk/central-frontend/blob/45281cb346ef888651f459f304b5f76621d12134/apps/central/src/util/session.js#L107-L114

matthew-white (commented):
Adding notes to facilitate async code review.

apps/central/test/composables/version-monitor.spec.js
```diff
@@ -1,20 +1,212 @@
...
+          })
+          .respondWithData(() => '(v2024.1.2-sha)')
+          .complete()
+          .request(() => {
+            clock.tick(60000);
+          })
+          .respondWithData(() => '(v2024.1.3-sha)')
+          .afterResponse(async (app) => {
+            clock.tick(0);
+            await app.get('.red-alert .close').trigger('click');
```
sadiqkhoja, line 112: I think this should be 
```suggestion
            await app.get('.toast .close').trigger('click');
```

sadiqkhoja (approved):

PR #1901, opened 2026-09-13
Reduce the maximum number of warnings allowed in test/run.sh

matthew-white:
test/run.sh checks for the number of warnings when running Frontend tests. It exits with 1 if the number is greater than 2. However, for a while, I've noticed that the actual number of warnings is 1. Given that, I think we can reduce the number of warnings allowed from 2 to 1. Doing so will help prevent us from accidentally adding a new warning without realizing it.

I think it's good for the number to be as low as possible so that we catch new warnings as they're introduced. Whenever there's a new warning, we can decide whether to fix it or to allow it by increasing the max count.


apps/central/test/run.sh
```diff
@@ -19,10 +19,10 @@ awk '
   /Module Warning/       { ++warnings; print "WARNING: " $0 }
   /WARN \[web-server\]:/ { ++warnings; print "WARNING: " $0 }
   END {
-    if(warnings > 2) {
+    if(warnings > 1) {
       print "All tests passed, but there were " warnings " warnings: see above."
       exit 1
     }
-    print "There were " warnings " warnings, which is within the accepted threshold."
+    print "There were " warnings " warning(s), which is within the accepted threshold."
```
matthew-white, line 26: I guess I could have put this line in a separate PR, but it feels at least somewhat related. Now that the warning count is 1, the message displayed has been "There were 1 warnings". The "1 warnings" has been grating on me, so I thought I'd go ahead and make it "warning(s)" now that 1 is the expected number of warnings.

matthew-white (commented):
Adding a comment to facilitate async code review.

alxndrsn (approved):

apps/central/test/run.sh
alxndrsn: Seems fine here - directly relevant to the aim of the PR.

PR #1902, opened 2026-09-14
fix(web-forms#911): show full error details on submission failure

### LINKED ISSUE #911: Show Problem message returned by Central Backend
### Expected Behaviour

If Central Backend returns an error response in its standard "Problem" format, I'd expect Web Forms to show the message contained within the Problem object.

### Observed Behaviour

In getodk/central#2219, Backend returns a 404 error. The response data is a standard Problem object. However, Web Forms indicates that there was no response:

> Something went wrong: there was no response to your request.

### Reproduction Steps

1. Navigate to https://test.getodk.cloud/projects/749/forms/styletest/submissions/new
2. Click Send

### Screenshots or Videos

Image

### Environment

See getodk/central#2219.

### Additional Context

getodk/central#2219 describes a bug in Central Backend. However, it also shows that Web Forms isn't surfacing the error from Backend.

garethbowen:
Closes getodk/web-forms#911

Fixed unit tests which revealed the issue

It's a bug...

None.

No.


apps/forms/test/components/web-form-renderer.spec.ts
```diff
@@ -1,25 +1,19 @@
...
 import formWithImageAttachmentXml from '../../../central/test/data/xml/image-attachment/form.xml?raw';
 import imageUploaderXml from '../../../central/test/data/xml/image-uploader/form.xml?raw';
 import simpleSubmission from '../../../central/test/data/xml/simple/submission.xml?raw';
 import imageUploaderSubmission from '../../../central/test/data/xml/image-uploader/submission.xml?raw';
-
-/* eslint-disable @typescript-eslint/no-unsafe-member-access */
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
-/* eslint-disable @typescript-eslint/no-unsafe-return */
-/* eslint-disable @typescript-eslint/no-unsafe-call */
-/* eslint-disable @typescript-eslint/no-unsafe-argument */
```
garethbowen, line 22: I took the time to fix these link disables too.

apps/forms/test/components/web-form-renderer.spec.ts
```diff
@@ -32,37 +26,6 @@ vi.mock('../../src/utils/last-saved.ts', () => ({
...
-      // 'retryModal.body': () => 'not fully submitted, try again',
-      'action.tryAgain': () => 'try again',
-      'editSubmissionModal.title': () => 'edit successful',
-      'editSubmissionModal.body': () => 'redirecting now...',
-      'thankYouModal.title': () => 'cheers!',
-      'thankYouModal.body': () => 'successfully submitted through this public link',
-    },
-  },
-});
-
```
garethbowen, line 65: This was an attempt to mock the i18n, but to get the templates to work it's easiest just to use the i18n the app uses.

apps/forms/test/components/web-form-renderer.spec.ts
```diff
@@ -72,37 +35,55 @@ describe('WebFormRenderer', () => {
...
-          enketoId: '',
-          state: 'open',
-          draft: false,
-          webformsEnabled: true,
-          attachments: testProps.form?.attachments ?? []
-        },
-        actionType: testProps.actionType ?? 'new',
-        instanceId: testProps.instanceId ?? null,
-        submissionAttachments: testProps.submissionAttachments ?? null,
-        st: testProps.st ?? null
```
garethbowen, line 98: Most of this was about fixing those pesky lint warnings.

apps/forms/src/components/web-form-renderer.vue
```diff
@@ -115,9 +115,7 @@ const submissionPath = () => {
 };
 
 const isSessionTimeout = (error) => {
-  return error?.response &&
-    isProblem(error.response.data) &&
-    error.response.data.code === 401.2;
+  return isProblem(error) && error.code === 401.2;
```
latin-panda, line 123: I spent a bit more time looking at these error code and found that backend sends [valueTooLong: 400.13](https://github.com/getodk/central-backend/blob/94f3b82e034d2aa3a07b56573e994c7479490f0a/lib/util/problem.js#L86), which won't [match L413](https://github.com/getodk/central-frontend/blob/3d948827961b66a558e967131a5d8f88db99f09b/apps/forms/src/components/web-form-renderer.vue#L131), I was wondering if you could fix that error code in this branch. Otherwise it can be in another ticket.

latin-panda (approved):
Looks good! I added a tiny finding below

apps/forms/src/components/web-form-renderer.vue
garethbowen: It looks like 400.13 is used when some write exceeds the varchar max length of the pg column. Did you actually hit this limit?

However the line you linked is talking about 413 which is talking about the payload size. This is returned from nginx, not central. Your comment reminded me that this error will not be in the central format so won't have JSON payload to check. The original change in this PR meant it was failing to parse the JSON and throwing an error which was showing the generic failure modal. I have updated the code and added tests to handle the 413 error.

garethbowen:
@latin-panda The additional commit warrants another check if you don't mind.

latin-panda (approved):
Looks good!

PR #1903, opened 2026-09-14
fix(web-forms#906): block leaving a page with invalid questions

### LINKED ISSUE #906: Block "Add" on invalid questions in paginated repeats
### Description

Web Forms now blocks "next" navigation when the current page has invalid questions, but in a paginated repeat, the "Add" button still advances past them. Block it the same way. 
This doesn't apply to repeats that display all their instances on one page. Also, consider move the shared blocking logic into the engine so both paths use one validation path ([reference](https://github.com/getodk/central-frontend/pull/1875#issuecomment-5536269846)).

### Additional Notes
It's a follow up work from this PR: https://github.com/getodk/central-frontend/pull/1875

### Related Issues

- https://github.com/getodk/web-forms/issues/900

latin-panda:
Closes https://github.com/getodk/web-forms/issues/906

[Executed the test plan](https://github.com/getodk/web-forms/issues/906#issuecomment-5667129023)

Moves page navigation logic to the engine, leaving purely UIUX stuff to the client

It now blocks the repeat add feature when questions have violations; no other side effects are expected. 

No


latin-panda:
@garethbowen this is ready for review. I'm thinking we can include it in the patch release.

packages/xforms-engine/src/client/RootNode.ts
```diff
@@ -92,8 +93,11 @@ export interface RootNode extends BaseNode {
    */
   setCurrentPage(page: PageBoundary): PageBoundary | null;
 
-  // Moves to the next page. Does nothing when {@link RootNodeState.hasNextPage} is `false`.
-  nextPage(): void;
+  // Whether the node's page is the current page. Always `true` on a non-paginated form.
+  isNodeInPage(node: AnyControlNode | RepeatRangeUncontrolledNode): boolean;
+
+  // Moves to the next page, unless the current page has violations, then it stays and returns them.
+  nextPage(): BlockingViolations;
```
garethbowen, line 100: I spent a while thinking about this api specifically. It seems like the "next page" and "submit" APIs should be more similar - ie: I want to do an action, which may be blocked by violations.

Currently the `handleSubmit` in OdkWebForm.vue explicitly checks `root.validationState.violations.length === 0` before preparing the instance payload, but with this PR `handleNext` just calls `nextPage()` and checks what is returned.

Do you think these should be consistent? If so, should the `prepareInstancePayload` function return violations somehow, or should the client be expected to check for violations before calling `nextPage()`?

I'm leaning towards the latter. Firstly because the former would be a breaking change in the engine, but more importantly because it feels like a cleaner api, and means other clients can have different opinions about whether navigation is allowed or not.

I know we're constantly revising what goes in the client vs the engine so I'm sorry if this is annoying!

garethbowen (commented):
Very nice. Worked well in my manual testing, and good test coverage.

One API question inline, but otherwise all good!

packages/xforms-engine/src/client/RootNode.ts
latin-panda: Not annoying at all!

When I looked at this last week, I was thinking about who holds the knowledge and who does the action. The engine holds the knowledge. It's the source of truth and decides what happens to the form, and clients stay "simple".

For "next" and "repeat add" actions, the engine is the one that knows what a page is, does the actual page move, and knows which violations belong to the current page (field-lists and repeats make that tricky in the client). So it decides if navigation is blocked, and tells the client why.

For "submit", the engine doesn't do the action, it just builds the payload. The client owns the send to Central, so the client checks the violations and decides. That check is simple, any violation blocks, and the client already has that in validationState.

The actor is different in each case, but both follow the same principle.

If we move the logic to the client, we need to expose more of the engine internals, something like `getBlockingViolations()`, and have the client check before every call (next page, add repeat, and later the table of contents). It's a bit more API, and every client needs to remember to check.                                                  

I'm fine going that way if you prefer it, but I'd keep it as is. It keeps the form rules in the engine.

garethbowen: Yes that's a fair point - the client owns submission and the engine owns the page navigation.

The prepare xml call is potentially another example where it should return a result with either the xml or the violations so it's a single call, but that's outside the scope of this issue.

Thanks for the explanation!

garethbowen (approved):

PR #1906, opened 2026-09-16
fix(web-forms#858): calculate and setvalue on upload nodes  

### LINKED ISSUE #858: Calculations and actions on upload nodes are not supported
### Expected Behaviour

Using setvalue to update an upload control should work. While it doesn't make sense to compute a binary file, if you compute a valid `jr:` url then the resource should be loaded. This also applies to calculate on the `bind` element.

This can be useful particularly for annotating one of a set of images.

### Observed Behaviour

Both setvalue and calculate silently no-op.

### Reproduction Steps

Form (XML or XLSX file): Form and attachments on the forum: https://forum.getodk.org/t/collect-2026-3-0-default-image-inline-preview-missing-take-choose-disabled-crash-on-null-annotation/58380/4

Steps:
1. Load the form
2. Select a radio button in the first field
3. You should see the image load on the annotate field below

For the calculate version, the simplest form has a bind something like this...

```xml
<bind nodeset="/data/group_image/imagecrash" type="binary" calculate="concat('jr://images/','koala.jpg')"/>
```

### Additional Context

Nodes like InputControl extend ValueNode, but UploadControl does not. This means that `createInstanceValueState` is never called so calculates are never registered. Once the calculate is registered we'll need a way to set the _string_ value of the UploadControl, and then that can be used for the setvalue actions too.

latin-panda:
Closes https://github.com/getodk/web-forms/issues/858

<!-- 

-->

Added test coverage and ran a manual [test plan](https://github.com/getodk/web-forms/issues/858#issuecomment-5702987704)

The upload node's value is now the source of truth, using the same `createInstanceValueState` as other value nodes, so `calculate` and `setvalue` work through the existing engine flows. The file is looked up from that value, fetching each `jr://` form attachment once per instance.

The other option was to special-case actions inside the attachment state, but that duplicated the value logic and would not have covered readonly or relevance the same way as other nodes. 

Annotate uploads now reload the canvas whenever the engine sets the value (edit, default, calculate, setvalue). When the user picks an image, the canvas already has it, so nothing is reloaded. In both cases the canvas saves a flattened copy as the answer right away, the same as it already did for a default image.                                                                             
                                                                                                                         
Upload questions now respond to calculate and setvalue, showing the referenced form attachment

No


packages/xforms-engine/src/lib/reactivity/createInstanceAttachment.ts
```diff
@@ -79,193 +70,185 @@ export interface BaseInstanceAttachmentState {
...
 ): InstanceAttachment => {
   return context.scope.runTask(() => {
-    const { rootDocument } = context;
-    const { attachments } = rootDocument;
+    const [getInstanceValue, setInstanceValue] = instanceValueState;
+    const [getState, setState] = createSignal<BaseInstanceAttachmentState>(BLANK_STATE);
+    const [getClientWrite, setClientWrite] = createSignal<ClientWrite | null>(null);
+    const resolver: ReferenceResolver = { context, getInstanceValue, getClientWrite, setState };
+
+    createComputed(() => {
```
latin-panda, line 213: Before, the attachment resolved once on init from the static instance node. Now it reacts to the instance value, so `calculate` and `setvalue` work like on any other node

packages/xforms-engine/src/instance/UploadControl.ts
```diff
@@ -146,13 +150,22 @@ export class UploadControl
     this.nodeOptions = definition.bodyElement.options;
     this.maxPixels = definition.bind.maxPixels;
 
-    const instanceAttachment = createInstanceAttachment(this);
+    this.decodeInstanceValue = (value) => value;
 
-    this.instanceAttachment = instanceAttachment;
+    const { valueState: instanceValueState, setValueFromAction } = createInstanceValueState(this);
```
latin-panda, line 155: The upload node now owns a normal instance value like other value nodes, and the attachment layer reads from it. The other changes in this file just connect the new value state to the node.

packages/xforms-engine/test/integration/instance-attachments/calculate-setvalue.test.ts
```diff
@@ -0,0 +1,445 @@
...
+  const getUploadNode = (scenario: ReactiveScenario, reference: string): UploadNode => {
+    const node = scenario.getInstanceNode(reference);
+    assert(node.nodeType === 'upload');
+    return node;
+  };
+
+  const settle = (scenario: ReactiveScenario, node: UploadNode): Promise<void> => {
+    return new Promise((resolve) => {
+      scenario.createEffect(() => {
+        if (!node.currentState.attachmentState.loading) {
```
latin-panda, line 72: Resolves when the engine reports the attachment is done loading, so the test never guesses with a timeout. The two ReactiveScenario tweaks are just to make this work

latin-panda:
@garethbowen this is ready for review.

packages/xforms-engine/src/instance/attachments/InstanceAttachmentsState.ts
```diff
@@ -2,22 +2,24 @@ import {
...
+import { getBlobData } from '@getodk/common/lib/web-compat/blob.ts';
 import type { FetchFormAttachment } from '../../client/resources.ts';
 import { AttachmentNotFoundError } from '../../error/AttachmentNotFoundError.ts';
-import type { StaticLeafElement } from '../../integration/xpath/static-dom/StaticElement.ts';
 import type { InstanceAttachmentMap } from '../input/InstanceAttachmentMap.ts';
 import type { InstanceAttachmentContext } from '../internal-api/InstanceAttachmentContext.ts';
 import type { InstanceAttachment } from './InstanceAttachment.ts';
 
 export class InstanceAttachmentsState extends Map<InstanceAttachmentContext, InstanceAttachment> {
+  private readonly formAttachmentFiles = new Map<JRResourceURLString, Promise<File>>();
```
garethbowen: Is this a cache for performance or to debounce concurrent requests for the same URL (or both)?

If it's for performance, won't the browser request cache do the same thing?

If it's to debounce, then you don't need to cache all files requested, just the latest one, right?

My main concern is storing multiple potentially very large files in memory even after the user has changed the url value.

packages/xforms-engine/src/instance/UploadControl.ts
garethbowen, line 1: I was hoping these changes would mean UploadControl could now extend ValueNode instead of DescendantNode to make it more similar to other Control types. Do you think that's possible/desirable?

garethbowen (changes requested):
Thanks!

I've added a couple of high level questions.

packages/xforms-engine/src/instance/attachments/InstanceAttachmentsState.ts
latin-panda: Both. It avoided re-downloading a file already fetched and coalesced duplicate in-flight requests (switching a selection back and forth or two questions calculating the same file). The client's media cache covers only label and hint media, not engine fetches.

However, after spending more time digging through the forum and GH issues, I found no reports of problems like this. So I removed it. It's easy to add back in the future if necessary.

packages/xforms-engine/src/instance/UploadControl.ts
latin-panda: Yes. I initially left it out because it needs changes in the base classes (ValueNode and ValueCodec), which I wanted to keep out of a bug fix. I have implemented the change in its own commit in case we need to revert. The engine unit and integration tests pass, which gives me good confidence :) 
What do you think?

packages/xforms-engine/src/instance/abstract/ValueNode.ts
```diff
@@ -120,12 +120,12 @@ export abstract class ValueNode<
     this.instanceState = createValueNodeInstanceState(this);
 
     this.setEncodedValue = (value: string, bypassReadonly = false) => {
```
latin-panda, line 122: It now writes the instance string directly instead of going through the runtime value. In the end, it's the same result for every codec, but the upload codec can't decode synchronously (the file is fetched async), so this is the only path that works for it. It also removes the second createRuntimeValueState call (L106), which existed only for setActionValue.

packages/xforms-engine/src/instance/UploadControl.ts
latin-panda: I also went through the test plan again; all good.

garethbowen (changes requested):
The hierarchy change is a huge win! Thanks for taking the time.

I ran a test using the `setvalue` form provided in your test plan and noticed something not quite working as expected. If you fill it in and submit it then the xml has values like `<photo_load>jr://images/koala.jpg</photo_load>`, but no attachment is posted. What this means is Central thinks there's a problem with the submission because it's expecting 4 attachments but only got 2.

Image

If you inspect the submission and try to download the image you get a 404. The URL looks something like this: `/v1/projects/11/forms/858-setvalue/submissions/uuid%3Aa5a22091-422f-4c24-9110-b34f3adf7b1b/attachments/jr%3A%2F%2Fimages%2Fkoala.jpg`

It's worth checking what Collect does in this situation. I suspect what we should do is rename the image with the generated name just like other uploads, and then send it to the server as an attachment, even if it's a duplicate of a form attachment. This feels wasteful, but I suspect a full solution which doesn't duplicate files will required changes to Central which is out of scope for this fix.

latin-panda:
@garethbowen,
Yes, I saw that. In Collect, the behavior is the same (see the video below, submissions from Collect). I thought it was intentional to save disk space. I'm going to ask in the channel [thread](https://getodk.slack.com/archives/C051L6AKG4F/p1790076288319879?thread_ts=1789666314.661139&cid=C051L6AKG4F) to be sure. If we decide to fix it in both WF and Collect, I think it's better to do it in a separate PR, more targeted. 

https://github.com/user-attachments/assets/1c98fe0f-e59e-4671-8882-b04bef1026b9

latin-panda:
Coming back from the channel's thread, we'll keep it as is for now and make a note of this behavior ([docs PR](https://github.com/getodk/docs/pull/2110)). And see if someone asks for it later.

garethbowen (approved):
Thank you! The code feels much cleaner now.

PR #1908, opened 2026-09-18
fix(web-forms#883): allow users to recover from xpath errors

### LINKED ISSUE #883: Allow users to recover from xpath errors
When the xpath throws a validation error the web-forms UI displays the error in a dialog that is impossible to recover from without refreshing the form.

Image

We can't necessarily show an error beside the field because it may be that it's not shown to the user, or the xpath expression may reference several fields. Instead show an error at the top of the form, and disable submit. That way the user is not blocked from fixing the error, and not interrupted if they're in the process of providing valid data.

Also consider if throwing errors is the correct approach or if the API should instead return error objects.

There may be a common approach that can be used for [similar errors in the xforms-engine](https://github.com/getodk/web-forms/issues/297).

garethbowen:
Closes getodk/web-forms#883

Manual testing, CI.

Discussed on the issue.

Error messages are now shown without interrupting flow, and inline if possible. 

No.


packages/web-forms/src/components/OdkWebForm.vue
```diff
@@ -389,6 +397,13 @@ onUnmounted(() => {
 			>
 				<IconSVG name="mdiAlertCircleOutline" variant="error" />
 				<ul class="form-error-text-wrap">
+					<li v-for="[message, references] in errorViolations" :key="message">
+						{{ t('odk_web_forms.evaluation.error', {
+							message,
+							fields: references.join(', '),
+							count: references.length
+						}) }}
+					</li>
```
garethbowen, line 406: The most common case is a single error violation, potentially impacting multiple fields, so I've optimised the UX for that case.

packages/xpath/test/xforms/format-date.test.ts
```diff
@@ -91,14 +91,23 @@ describe('#format-date()', () => {
 
   describe('invalid dates', () => {
```
garethbowen, line 92: Now some of these throw and some return null... It may still not be right but it's closer.

garethbowen:
@latin-panda This is now ready for review, but given the size the change ended up being I'm not sure we want to merge it in 1.1.1. I'm happy to leave this for 1.2.0 if that feels safer to you.

packages/xforms-engine/src/instance/abstract/DescendantNode.ts
```diff
@@ -296,4 +297,19 @@ export abstract class DescendantNode<
 
     this.scope.dispose();
   }
+
+  private compute(expression: DependentExpression<'boolean'>, defaultValue: boolean) {
+    return createMemo(() => {
+      const r: Result<'boolean'> = createComputedExpression(this, expression, {
+        defaultValue,
+      })();
```
latin-panda: The `createComputedExpression` already returns a memo, and if it's wrapped in another memo Solid does not dispose it, then a fresh one is created each time

```suggestion
   const computed = createComputedExpression(this, expression, { defaultValue });
   return createMemo(() => {
     const result = computed();
```

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
```diff
@@ -99,6 +108,7 @@ const createTextChunks = <Role extends TextRole>(
   const chunks: TextChunk[] = [];
   const mediaSources: MediaSources = {};
   const chunkExpressions = getChunkExpressions(context, definition);
+  context.setError(null);
```
latin-panda: Let me see if I'm getting this correctly:

Each node/field has one error "slot". And many things write to that slot (the field's calculate, relevant, label, constraint). Each of them sets the slot to null when it succeeds.
So the label succeeds and sets it to null. But if the calculate failed before then, the error is gone. Does it make sense? 🤔
Maybe it's better to have a map of actors (calculate, constraint, etc.) and the corresponding error.

packages/xforms-engine/test/lib/reactivity/createAggregatedViolations.test.ts
```diff
@@ -80,7 +80,7 @@ describe('createAggregatedViolations - reactive aggregated `constraint` and `req
   interface SimplifiedViolation {
     readonly condition: ValidationCondition;
     readonly valid: false;
-    readonly message: ViolationMessage<ValidationCondition> | null;
+    readonly message: ViolationMessage<ValidationCondition> | string | null;
```
latin-panda, line 84: Maybe not super common, but group and repeat range nodes can have failing expressions (relevant, readonly, jr:count). So their error is stored on the node but not shown and not blocking the submit (because the submit checks the violation list). 
Could you please check that group and repeat errors are added to the violation list and add a test here?

packages/web-forms/src/components/OdkWebForm.vue
```diff
@@ -310,22 +300,36 @@ const revealViolations: RevealViolations = (violations) => {
...
+	const grouped = new Map<string, string[]>();
+	violations.forEach(error => {
+		const key = error.violation.condition === 'error' && error.violation.message;
+		if (!key) {
+			return;
+		}
+		let fieldReference = error.reference;
+		if (fieldReference.startsWith('/data/')) {
+			fieldReference = fieldReference.replace('/data/', '');
+		}
```
latin-panda: Is it always named 'data'? Is there any way the engine can clean the error before sending it to the client? Maybe it has more context about the field name so we don't need to hardcode it here

packages/web-forms/src/components/OdkWebForm.vue
```diff
@@ -339,7 +343,11 @@ const showValidationError = computed(() => {
 	if (errorBannerDismissed.value) {
```
latin-panda, line 357: I haven't tested it, but I _think_ `errorBannerDismissed` should be set to `false` here so that new evaluations with errors display the banner.

latin-panda (changes requested):
I haven't finished reviewing or testing, but I wanted to share my feedback so far. I'll continue tomorrow

packages/xforms-engine/src/instance/abstract/DescendantNode.ts
garethbowen: Ahh good catch. Thanks.

packages/web-forms/src/components/OdkWebForm.vue
garethbowen: We don't currently do that for other violations until you hit the next page or submit button, right? I feel like it would just be annoying if you have explicitly dismissed it for it to pop up again, even if now the message is different.

packages/web-forms/src/components/OdkWebForm.vue
garethbowen: No it could be named anything. "data" is what pyxform calls it so it's overwhelmingly more popular than anything else.

I've modified this to look up the root node name rather than assume it's data. I preferred this to having engine do it because it feels client specific - other clients may want the full path for some reason.

packages/xforms-engine/test/lib/reactivity/createAggregatedViolations.test.ts
garethbowen: Done, thanks!

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
garethbowen: Done, thanks for spotting this!

My solution feels clunky so if you have a better way to do it I'd love to hear it, but this does work.

latin-panda: Thank you! The map is better, but each place still has to set and clear its own key. I haven't checked all cases but I found that `createTextRange` L111 sets the label error and clears it on the line below, so a broken itext shows nothing, and label, hint and messages share the same key, so one can hide the other's error.

I was having a deeper look, and each expression is already a memo that returns a `Result`, and `result.error.message` has the text we need. So we could have one memo per node that reads all the results and returns the ones that failed. Since it's reactive, it re-runs when something changes and then a fixed expression returns success, the error is removed automatically. 
The `createErrorValidation` can read that memo instead of `getError()`, and the violations list stays the same. 

The only thing I see so far with that approach is the calculate, setvalue and the text chunks don't return a `Result` yet, so they'd need that.

packages/xforms-engine/src/lib/reactivity/validation/createAggregatedViolations.ts
```diff
@@ -37,20 +48,14 @@ const collectViolationReferences = (
...
-        if (reference == null) {
-          return [];
-        }
-
-        return [reference];
-      }
-
+      case 'upload':
+        // leaf node
+        return violationReference(child);
```
latin-panda: In L38 collects attributes for the parent, right? but a leaf question never becomes context, so its attributes are skipped (the submit is not blocked on error)
```suggestion
        return [...violationReference(child), ...child.getAttributes().flatMap(violationReference)];
```

packages/xforms-engine/src/lib/reactivity/createItemCollection.ts
```diff
@@ -68,6 +73,8 @@ class ItemsetItemEvaluationContext implements EvaluationContext {
   readonly evaluator: EngineXPathEvaluator;
   readonly contextReference: Accessor<string>;
   readonly getActiveLanguage: Accessor<ActiveLanguage>;
+  readonly errorState: Accessor<ErrorState>;
```
latin-panda: I couldn't find anything that reads this error state. When an item label is broken (`<label ref="badFn()"/>`) the error seems lost?. I wonder if it could bubble up to the control's state.

packages/xforms-engine/src/lib/reactivity/validation/createValidation.ts
```diff
@@ -102,13 +110,24 @@ const createRequiredValidation = (
...
+  return createMemo(() => {
+    const error = context.getError();
+    if (error) {
+      return {
+        condition: 'error',
+        valid: false,
+        message: error,
+      } as const;
+    }
+    return null;
```
latin-panda: We don't block the form if the question is relevant=false, right?
I think it should return null in that case 
```if !context.isRelevant() { returns null```

latin-panda (changes requested):
I'm still digging into the details :) Meanwhile, I wanted to share some feedback, and I replied to one of the [threads](https://github.com/getodk/central-frontend/pull/1908#discussion_r4086640207) with an idea for catching errors from `Result` more reactively. Let me know what you think!

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
garethbowen: :thinking: 

It's definitely a more reactive approach, and solves all the issues about keeping the error state correct. If I understand correctly, the InstanceNode would have an abstract errors accessor that child nodes would override with a memo which reactively updates based on the computation Results that it knows about. For example, the ValueNode would check for errors on the value Result, and the RepeatRangeControlled could check for errors on the computeCount Result. Each Node's accessor would probably also call the accessor on their parent node to get a combined list of errors. Is that what you were thinking?

It's another significant change so I wanted to confirm I understood what you meant before spending too much time building it out.

packages/xforms-engine/src/lib/reactivity/validation/createAggregatedViolations.ts
garethbowen: True! I've fixed this by moving the attributes violations into the `violationReference` function, so when it's getting violations for the node it also checks the attributes.

packages/xforms-engine/src/lib/reactivity/validation/createValidation.ts
garethbowen: Hmm.. In this case I'm not sure. One example is what if the expression for `relevant` is the thing with the error? Ultimately if this field is in an error state we can't be confident if it is relevant or not. To be on the safe side I think it's fair to block any form with an unresolved error in it.

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
latin-panda: Yes, correct! Just one thing about `call the accessor on their parent`, if you mean the parent class (super.errors() + the node's own results), yes. But if you mean the parent node, the `createAggregatedViolations` already walks the children up to the root, so pulling from the parent would list the same error once per descendant, right?

Thinking about `setvalue` from my last comment, the actions run on events, so there is no `Result` to read. They would need to store the error (or null) of their last run on the destination node, and the node's `errors()` reads that 🤔

packages/xforms-engine/src/lib/reactivity/validation/createValidation.ts
latin-panda: Sounds good! If the question isn't visible and the relevant value is `false`, could this be used to improve the error message? 
"the hidden field has invalid  ....." 
Or the question name/ref.

packages/xforms-engine/src/lib/reactivity/text/createTextRange.ts
garethbowen: > if you mean the parent class (super.errors() + the node's own results), yes

Yes this is what I meant.

> Thinking about setvalue from my last comment, the actions run on events, so there is no Result to read.

Yes I think there will need to be some calculations that store errors outside of the Result. I'll see what I can come up with today.

PR #1909, opened 2026-09-18
fix(web-forms#919): min and max should ignore empty items

### LINKED ISSUE #919: The min() and max() return blank when a repeat has an empty value
### Expected Behaviour

The `min()` and `max()` over a repeat ignore empty rows, like Collect.  With values 2, 3.5 and one empty row: `min()` is 2, `max()` is 3.5, and a lookup for the smallest row returns that row.

### Observed Behaviour

`min()` and `max()` return blank as soon as one row is empty. A comparison against that blank matches the empty row, so the lookup returns the wrong row.  

### Reproduction Steps

Form (XML or XLSX file): [decimal-join-min-repeat.xml](https://github.com/user-attachments/files/32305314/decimal-join-min-repeat.xml)

Steps:
1. Upload the form to Central and open it in Web Forms.                                                                                                                                                
2. Add three repeats: `A` = 2, `B` = 3.5, `C` = empty.                                                                                                                                                 
3. Compare results of questions: 2, 5a and 8 with Collect.

### Environment

- Browser version: chrome
- OS: macos
- Web Forms version: 1.1.0

### Additional Context

reported in the forum https://forum.getodk.org/t/webforms-behaviour-notes/58272/53

garethbowen:
Closes getodk/web-forms#919


latin-panda (approved):
It works well! Thank you for the quick fix!

PR #1910, opened 2026-09-20
fix(web-forms#920): numbers in different formats should match

### LINKED ISSUE #920: Decimal fields with whole numbers don't match calculated values
### Expected Behaviour

A decimal field holding 2 reads as `2` in calculations (Collect does it). The `join()` prints `2`, and comparing the field with a calculated number matches. 

### Observed Behaviour

The field reads as `2.0`. `join()` prints `2.0`, and `v = min(...)` never matches for whole numbers because `"2.0"` is not `"2"`. Lookups only work when the value has a fraction, like 2.5.

### Reproduction Steps

Form (XML or XLSX file):

[decimal-join-min-repeat.xml](https://github.com/user-attachments/files/32305751/decimal-join-min-repeat.xml)

Steps:
1. Upload the form to Central and open it in Web Forms.                                                                                                                                                
2. Add three repeats: `A` = 2, `B` = 3.5, `C` = empty.                                                                                                                                                 
3. Compare results 1 and 6 with Collect. Result 7 shows the workaround with `number()`.   

### Environment

- Browser version: chrome
- OS: macos
- Web Forms version: 1.1.0

### Additional Context

reported in the forum https://forum.getodk.org/t/webforms-behaviour-notes/58272/53

garethbowen:
Closes getodk/web-forms#920

Manual and unit testing.

This solution works for other number values, such as dates with different timezones (test added). 

The other approach I considered was to revert the change that added a ".0" to the string format, and then find a different solution to that problem, possibly only adding the .0 at the serialization stage. We don't do any other formatting at that stage so it didn't feel correct.

This is a more liberal definition of equality which may give some false positives but I can't think of any where this would not match a human's definition of equality.

No.


garethbowen:
@latin-panda This solution solves test number 6 in the issue. It _does not_ solve the join issue, which is lower priority (IMO) and not even necessarily incorrect. If you are happy with this solution I'll create a new issue for the join problem.

packages/xpath/test/xforms/predicates.test.ts
```diff
@@ -244,4 +228,42 @@ describe('predicates with function calls', () => {
     // assertTrue('selected("a b", "a")');
     testContext.assertNumberValue('count(/data/a[selected("a b", "a")])', 3);
   });
+
+  it('can compare nodes containing equal numbers in different forms', () => {
+    testContext = createXFormsTestContext(`
```
latin-panda, line 233: For a moment, I got confused because I added a case where "1,000" equals 1, and it was `true`, because I was thinking of it as 1 thousand. Then I remembered that commas are replaced with dots for decimals. 
Just noting for the record, it might come up again 😅

latin-panda (approved):
It looks good to me. I didn't do deep testing; I just ran a few cases.

PR #1911, opened 2026-09-21
fix(web-forms#918): fields in panels should not have rounded corners

### LINKED ISSUE #918: Error style in repeats and groups shouldn't have rounded corners
Minor stylistic improvement...

When a field constraint is violated the field background is coloured red. When not in a repeat or group this is shown with subtly rounded corners which looks good. However when it _is_ in a repeat or group the question runs right up to the edge of the border of the group and the rounded corners don't look good.

Image

Remove the rounded corners on error blocks when a border exists, so it looks like this:

Image

NB: on the smallest screens the repeats and groups don't have a border so the error should be rounded again...

garethbowen:
Closes getodk/web-forms#918


latin-panda (approved):
Looks good! I also tested with maps and geopoint (no-map). They keep their border radius, but it looks fine because of the surrounding padding.

PR #1912, opened 2026-09-21
fix(web-forms#866): gap around markdown elements in select items

### LINKED ISSUE #866: Whitespace ignored between markdown elements in select options list
### Expected Behaviour

Select options which have markdown elements should render the same in the select as they would on the page.

### Observed Behaviour

The whitespace is ignored. If you select the option then the markdown is rendered as expected in the select value.

### Reproduction Steps

Form (XML or XLSX file):

Steps:
1. Load a form with a select with a label like `Kipchunu **Dispensary**`
2. Open the select

### Screenshots or Videos

Image

### Additional info

It's to do with the `display: flex` of the list item.

garethbowen:
Closes getodk/web-forms#866

I thought about a css solution but we actually want the gap for other reasons, so wrapping _our_ elements in a span so the gap doesn't apply is a better solution.


packages/web-forms/src/components/common/SearchableDropdown.vue
```diff
@@ -63,7 +63,9 @@ const selectedLabel = computed(() => {
 		@change="$emit('change', $event.value)"
 	>
 		<template #option="slotProps">
-			<MarkdownBlock v-for="elem in slotProps.option.label" :key="elem.id" :elem="elem" />
+			<span>
```
latin-panda: Could you also include Multiselect?  The wrapper is in 2 places:
- https://github.com/garethbowen/central-frontend/blob/296a9415b842e4c2faa2abf2c7bde01014100076/packages/web-forms/src/components/common/MultiselectDropdown.vue#L82 
- https://github.com/garethbowen/central-frontend/blob/296a9415b842e4c2faa2abf2c7bde01014100076/packages/web-forms/src/components/common/MultiselectDropdown.vue#L91

The last one has different style from this select L74

latin-panda (approved):
Simple and it works for select and multiselect.

